PR #6, opened 2020-04-10
Bugfix: delete key would rename if set

theHamsta:
I just copied the default settings from the README and was very confused that `d` would rename a file...


kyazdani42:
Hi, thanks for this i didn't realize as i use the default keybindings.

PR #13, opened 2020-05-18
Add config option to follow buffer's directory

Akin909:
@kyazdani42 I tried the most recent change to master and realised it wasn't working. The last change you made was the correct logic but it was in a function that only ran on `VimEnter` and had `auto_open` set to 1. The line is [here](https://github.com/Akin909/nvim-tree.lua/blob/5fcbc21edb08c77b3afd938958ad70178ee74d1f/plugin/tree.vim#L19).

I moved the logic you added into a separate function and call it on `BufEnter` that way when you change buffer the Lua Tree updates to the new root dir.

nvim-tree-follow-buffer

Lemme know what you think. I added it as an option rather than the default, although I think maybe it should be default behaviour

### DIFF
--- a/lua/tree.lua
+++ b/lua/tree.lua
@@ -153,6 +153,15 @@ function M.check_windows_and_close()
   end
 end
 
+function M.navigate_to_buffer_dir(bufname)
+  local new_path = get_cwd()
+  if new_path ~= '/' then
+    new_path = new_path .. '/'
+  end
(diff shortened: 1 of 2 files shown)


kyazdani42:
Ah now i know what you meant, well looks like it should be default behavior indeed, but providing an option to disable it might be useful.
Thanks !

PR #14, opened 2020-05-19
Rewrite the whole api

### LINKED ISSUE #12: Refactoring
Doing:
- use vim nightly features
- when updating file system, check for open buffers, if some name matches, change the buffer names to avoid conflicts when saving a file
- fix various bugs
- use fs_scandir instead of ls which makes ignoring also available on macos and makes the tree much much much faster

not doing:
- bufferize project in the background to open directories faster -> too slow
- whole tree async with luv fs operations -> codebase becomes messy

### LINKED ISSUE #15: Default icon
Would it be possible to add a default icon so that files that don't match any filetype/filename icon are aligned at the same level as the rest?

Screenshot from 2020-05-22 00-14-37

### LINKED ISSUE #17: 'winhl' is being overwritten
In my config files, I set the `winhl` option for LuaTree when it starts, but it's being overwritten. LuaTree should use `set winhl+=` to preserver existing options.

Btw, as soon as you merge the refactor PR I plan to help.

Below is a screenshot of the reason why I wanna modify it. Had to monkey-patch some files to make it.
Screenshot from 2020-05-25 04-50-42

### LINKED ISSUE #18: Add :LuaTreeOpen/Close

### LINKED ISSUE #22: Nvim Tree prevents gx mapping working
It seems like `g:loaded_netrw` is being set by `nvim-tree` at some point, which I probably wouldn't have noticed if not for the fact that I use `gx` often to follow links in nvim. I know Netrw interferes with other alternative tree plugins as this has been something I've seen across several plugins. But is it possible to avoid disabling it altogether or at least maintaining the `gx` behaviour. Maybe a replacement mapping to `gx` that calls loads netrw only for that purpose.

Not entirely sure how that would work tbh

kyazdani42:
Big rewrite of the tree:

Goals:
- [x] The tree is created with `libuv` functions, which makes it blazingly fast.
- [x] The tree may now be faster than any other vim trees, it can handle directories with thousands of files without any latency at all (tested on 40K files, works flawlessly).
- [x] More solid logic for opening and closing the tree.
- [x] tree state is saved (closing / opening a folder should keep opened subdiretories open)
- [x] detection of multiple git projects in the tree (so git infos are properly added to different files)
- [x] user icons (default for files without icons and git icons)
- [x] smart updates (do not reload the whole tree when doing an update)
- [x] implement `gx` function to spawn file with xdg-open/open(macos)
- [x] fix: use noautocmd on wincmd when editing a new buffer
- [x] filesystem operations based on libuv instead of spawning unix processes

closes #12 
closes #22 
closes #15 
closes #18 
closes #17


Akin909:
@kyazdani42 regarding multiple workspaces, in `vscode` this is done based on the language server integration which I don't think is necessary for this feature.

If the objective is simply to have two trees, which I think is what the feature essentially is (without the vscode lsp magic). Btw this PR sounds great without that feature anyway (I've never really used vscode so no idea how useful multi root workspaces is or not). If you did decide to do it I wonder if a good approach would be just having a mode where you can open two parallel `LuaTree` windows :man_shrugging: . Anyway just butting my nose in this looks great :raised_hands: . I work in a giant mono repo so this will be awesome to have

```
|-------------|
| Tree 1      |
| ~/Project 1 |
|             |
|             |
|-------------|
| Tree 2      |
| ~/project 2 |
|             |
|             |
```

kyazdani42:
thanks for the tips.
Well i don't really know how useful it could be, and how it would impact performance, or introduce new bugs so im still wondering about adding this feature. I'll add that at the end and see if it behaves properly.
And i wonder if i could handle icons differently because the vim devicon package takes way too much time and i could make it much faster with lua. Still some work to do !

kyazdani42:
thanks for the tips.
Well i don't really know how useful it could be, and how it would impact performance, or introduce new bugs so im still wondering about adding this feature. I'll add that at the end and see if it behaves properly.
And i wonder if i could handle icons differently because the vim devicon package takes way too much time and i could make it much faster with lua. Still some work to do !

Akin909:
Regarding dev icons being slow one thing I saw in someone else's dotfiles was [this](https://github.com/haorenW1025/dotfiles/blob/master/nvim/lua/devicon.lua) he maintains a list of icons in lua, the issue with this approach is you have to keep the list of icons updated whenever a new language or extension gets added. Although I guess that's not that often.

kyazdani42:
that's what i just did :) it super fast now. But still a lot of colors and file extensions to manage...

Akin909:
I wonder if a separate `lua` library for devicons will be a useful project. I was thinking of doing something similar with [a tiny project of mine](https://github.com/Akin909/nvim-bufferline.lua), not sure how dependencies like that would work in lua but if it was a community thing and it caught on it wouldn't have to be just one person maintaining the list hopefully :shrug:

kyazdani42:
your project looks cool !
Well it would be quite easy, we would have to expose a function that takes an extension, a filename and that would return an icon. We could even return the color of the icon. and in a dev code, it would look like:
```lua
local icon, color = require'devicon-lua'.get_icon(fname, ext)
```

Akin909:
I guess that means a user would have to install an extra plugin though :thinking: , that might be a blocker for some people but I guess you have to do the same with `vim-devicons` anyway.

kyazdani42:
well just use git submodules

romgrk:
hey :) a few question/comments:
 - why do you keep force-pushing your whole branch when there are changes? isn't it easier to simply `git commit/git push`?
 - if you split your PR in smaller chunks (and multiple commits) it's easier to review, so people lurking around can give feedback, spot typos, etc. Also people who wanna help can better gage where they can help.

Feel free to ignore, I'm just eager to use help, I like the approach of this plugin.

kyazdani42:
well, i just kept an habit when doing PR's to make only one commit because i'm always worried to blow everything up with git rebase ^^' 
That's something i've got to work on, i agree it's not easy for people to review a big commit like this :)

PR #43, opened 2020-06-19
add indent markers

### LINKED ISSUE #42: Feature request: indent markers
This is a completely stylistic addition which I found quite nice in other file explorers e.g. `coc-explorer` below.

image

maybe this could be added as a config option so anyone who does or doesn't like it can toggle it.

I had a quick look at the code to see if it could be added easily and looks like in the `renderer.lua` where the padding is added you could insert a special character one for most indents `|` and one at the end of that indent level (can't find the exact unicode symbol atm). Naively just inserting the symbol breaks the position based highlighting. I'm happy to dig into it when I have some free time if you haven't already had a look (if you think it's fine to implement)

kyazdani42:
@Akin909 can you review this works as expected for you ? Or tell me if you'd prefer something else.
(add `let g:lua_tree_indent_markers = 1` to your init.vim)

fixes #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ let g:lua_tree_ignore = [ '.git', 'node_modules', '.cache' ] "empty by default
 let g:lua_tree_auto_open = 1 "0 by default, opens the tree when typing `vim $DIR` or `vim`
 let g:lua_tree_auto_close = 1 "0 by default, closes the tree when it's the last window
 let g:lua_tree_follow = 1 "0 by default, this option allows the cursor to be updated when entering a buffer
+let g:lua_tree_indent_markers = 1 "0 by default, this option shows indent markers when folders are open
 let g:lua_tree_show_icons = {
     \ 'git': 1,
     \ 'folders': 0,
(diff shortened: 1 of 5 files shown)


Akin909:
Wow so quick 😍 I'm not by my laptop atm but will have a look in a bit

Akin909:
@kyazdani42 just tried it out and it looks really cool :+1:. One thing I noticed which is pretty minor is that the the indent mark for the parent of a directory changes depending on which sub folder is open rather than maintaining it's indent level. Not sure if that makes sense
side-by-side
Lua tree is on the right. The indent marker for the `.config` ends once the alacritty folder is open and if I close it then extends to the end again. Also I wonder if it can be highlighted more subtly. Anyway this is already pretty great as it is thanks for looking into it so quickly

kyazdani42:
ah you are right ! i didn't look at the picture from your issue properly

kyazdani42:
This should be ok ! 
You can also change the color by overriding the hl group `LuaTreeIndentMarker`. By default it will be the same color as the folders.

Akin909:
Awesome that works :rocket: thanks so much for adding this :+1:

PR #44, opened 2020-06-19
Add option for disabling bindings

rickysaurav:
* Option for disabling keybindings by plugin. When disabled, all bindings need to be created by the users on their own accord.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,12 @@ let g:lua_tree_bindings = {
     \ 'edit_tab':    '<C-t>',
     \ 'preview':     '<Tab>',
     \ 'cd':          '<C-]>',
+    }
+
+" Disable default mappings by plugin
+" Bindings are enable by default, disabled on any non-zero value
+" let lua_tree_disable_keybindings=1
+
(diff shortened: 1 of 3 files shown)


kyazdani42:
I'm not too sure this is an useful thing to add. Could you tell me why you would need this ?

rickysaurav:
I find the current keybinding method slightly unintuitive .  I would prefer to be able to allocate multiple binds to the same action like "edit"  . For example : i prefer tab and enter to have the same behaviour as "edit" . That's why I would like to roll my own set of keybindings and not have the plugin interfere with them.

kyazdani42:
Can you not disable the keybindings with the configuration by passing empty strings ? 
I'm not sure it would work but if it does, this PR would lose its interest :) otherwise i'll merge.

rickysaurav:
It'd work for most of the stuff. But IIRC, that won't disable mouse bindings, xdg open bindings.

kyazdani42:
well i'll merge but this is a special case, i will make something better another time :) thanks!

PR #52, opened 2020-07-16
Add cut,copy and paste functionality.

kristijanhusak:
This PR adds cut, copy and paste functionality that I mentioned in issue https://github.com/kyazdani42/nvim-tree.lua/issues/51

Note that I don't have much experience with Lua, so do as detailed review as possible.

### DIFF
--- a/lua/lib/fs.lua
+++ b/lua/lib/fs.lua
@@ -142,6 +142,7 @@ local function do_copy(source, destination)
 end
 
 local function do_paste(node, action_type, action_fn)
+  if node.name == '..' then return end
   local clip = clipboard[action_type]
   if #clip == 0 then return end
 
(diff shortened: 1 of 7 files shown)


kyazdani42:
thanks :) I'll review this PR in depth later but it looks already pretty good imo.

kristijanhusak:
I think handling of existing files on copy/move needs to be handled. Now it will just throw an error, but I guess it should prompt for override/cancel. Will try to add it.

lua/lib/fs.lua
```diff
@@ -109,6 +113,92 @@ local function remove_dir(cwd)
...
+      api.nvim_err_writeln('Could not '..action_type..' '..entry.absolute_path..' - '..msg)
+    end
+    ::continue::
+  end
+  clipboard[action_type] = {}
+  return refresh_tree()
+end
+
+local function add_to_clipboard(node, clip)
+  if node.name == '..' then return end
```
kyazdani42, line 190: this should be also checked when pasting or it will error out.

lua/lib/fs.lua
```diff
@@ -109,6 +113,92 @@ local function remove_dir(cwd)
...
+      local ans = vim.fn.input(dest..' already exists, overwrite ? y/n: ')
+      clear_prompt()
+      if not ans:match('^y') then goto continue end
+    end
+
+    local success, msg = action_fn(entry.absolute_path, dest)
+    if not success then
+      api.nvim_err_writeln('Could not '..action_type..' '..entry.absolute_path..' - '..msg)
+    end
+    ::continue::
```
kyazdani42, line 183: goto is not available for some distributions of neovim (as neovim uses lua 5.1 where goto is not available). You could leave it for now as i still have some piece of code with `goto`s but this issue will need to be addressed in another PR.

lua/lib/fs.lua
kristijanhusak: Added it to `do_paste` function.

lua/lib/fs.lua
kristijanhusak: I changed the logic to not use continue.

kristijanhusak:
Yeah I can add something. Maybe echo will be sufficient?
Do you want to have it like a mapping or command?

kyazdani42:
a command would be more appropriate. Echo should be enough indeed

kristijanhusak:
I added `:LuaTreeClipboard` command. See if the format is ok.

PR #53, opened 2020-07-17
Add mapping for toggling ignored folders visibility.

kristijanhusak:
This is another feature that I mentioned in https://github.com/kyazdani42/nvim-tree.lua/issues/51


kyazdani42:
LGTM, thanks for editing the docs too :)

Issue #51, opened 2020-07-14
Few feature requests


kristijanhusak:
Hi,

very nice plugin, works really fast!

I have few feature requests that you might consider:

* Allow selecting files and operating on them (For example, multiple rename)
* DONE in #52: Introduce cut/copy/paste
* DONE IN #56: Add mappings for jumping to previous/next git changed file, something like `vim-gitgutter` does with `]c` and `[c`
* DONE in #53: Allow toggling visibility of ignored files

I'm currently using these features in [defx.nvim](https://github.com/Shougo/defx.nvim), and I'm also author of git and icons plugins for defx.

kyazdani42:
think i can close this

MKrbm:
Is the feature for allowing multiple copy & paste still in progress?

alex-courtis:
> Is the feature for allowing multiple copy & paste still in progress?

That's done. You should be able to see them marked e.g.

2 copied
![20231031_093017](https://github.com/nvim-tree/nvim-tree.lua/assets/1505378/9532d0a3-1395-4a19-bacd-b31a4d02c3e4)

2 more cut
![20231031_093038](https://github.com/nvim-tree/nvim-tree.lua/assets/1505378/8a987239-6317-4573-bcf7-a22c934c6663)

PR #56, opened 2020-07-22
Add mappings for jumping to previous or next git item.

kristijanhusak:
This is one more feature request that i mentioned in #51 .

It's just a simple search for any git icon from current line. This is how I implemented it for defx, and nerdtree git plugin does the same. No one complained so far :)

### DIFF
--- a/lua/tree.lua
+++ b/lua/tree.lua
@@ -1,5 +1,6 @@
 local luv = vim.loop
 local lib = require'lib.lib'
+local config = require'lib.config'
 local colors = require'lib.colors'
 local renderer = require'lib.renderer'
 local fs = require'lib.fs'
@@ -28,6 +29,21 @@ function M.open()
   end
 end
 
+
+local function go_to_git_item(mode)
+  local icon_state = config.get_icon_state()
+  if not icon_state.show_git_icon then return end
+
+  local icons = {}
+  for _, icon in pairs(icon_state.icons.git_icons) do
+    table.insert(icons, icon)
+  end
+
+  icons = table.concat(icons, '\\|')
+  local flags = mode == 'prev_git_item' and 'b' or ''
+  return vim.fn.search(icons, flags)
+end
+
 function M.on_keypress(mode)
   local node = lib.get_node_at_cursor()
   if not node then return end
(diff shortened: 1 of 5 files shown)


lua/tree.lua
```diff
@@ -28,6 +29,21 @@ function M.open()
...
 
+
+local function go_to_git_item(mode)
+  local icon_state = config.get_icon_state()
+  if not icon_state.show_git_icon then return end
+
+  local icons = {}
+  for _, icon in pairs(icon_state.icons.git_icons) do
+    table.insert(icons, icon)
+  end
```
kyazdani42, line 40: this should be cached to improve performance. Also checkout `vim.tbl_values` function.

lua/tree.lua
```diff
@@ -28,6 +29,21 @@ function M.open()
...
+
+  icons = table.concat(icons, '\\|')
+  local flags = mode == 'prev_git_item' and 'b' or ''
```
kyazdani42, line 43: you could make a closure over this variable and load specific functions for each action.
```lua
function gen_go_to(mode)
   local flags = mode == 'prev_git_item' and 'b' or ''
   return function()
    -- uses mode
   end
end

local go_to_prev = gen_go_to('prev_git_item')
local go_to_next = go_to_to()
```

lua/tree.lua
```diff
@@ -55,6 +71,10 @@ function M.on_keypress(mode)
     return lib.toggle_ignored()
   end
 
+  if mode == 'next_git_item' or mode == 'prev_git_item' then
+    return go_to_git_item(mode)
+  end
```
kyazdani42, line 76: and then you could load specific function if it's next or prev.

lua/tree.lua
kristijanhusak: Done, thanks.

lua/tree.lua
kristijanhusak: Done, thanks.

lua/tree.lua
kristijanhusak: I'm using separate if blocks now to call cached functions.

PR #68, opened 2020-08-01
refacto display and add LuaTreeFolderDirty group

### LINKED ISSUE #65: Git status changes
Hi,

is there any specific reason why folders have git status icon after, and files have it before? :)

I would like to make that consistent, and also to introduce an option to highlight the whole item when it is "dirty". What do you think about that?

kyazdani42:
fixes #65 
You can now set `g:lua_tree_git_hl = 1` to color file from their git attributes.

TODO: update the pictures

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ let g:lua_tree_auto_open = 1 "0 by default, opens the tree when typing `vim $DIR
 let g:lua_tree_auto_close = 1 "0 by default, closes the tree when it's the last window
 let g:lua_tree_follow = 1 "0 by default, this option allows the cursor to be updated when entering a buffer
 let g:lua_tree_indent_markers = 1 "0 by default, this option shows indent markers when folders are open
+let g:lua_tree_git_hl = 1 "0 by default, will enable file highlight for git attributes (can be used without the icons).
 let g:lua_tree_show_icons = {
     \ 'git': 1,
     \ 'folders': 0,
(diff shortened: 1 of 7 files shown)


Akin909:
Just tried this out, the highlight group works well :+1: . The position of the git icons seems a bit odd to me. I think it's much nicer that files and folders have icons in the same place but I *personally* think it would be even better if they came after the filename or before the icon rather than in between the file/folder icon and the filename. That's just my opinion though not sure if this is what @kristijanhusak had in mind

nvim-tree-git-icons

kristijanhusak:
I agree with @Akin909 that having icon before like for files look a bit odd. Maybe we are just used to old way :)

Why introducing hl group only for dirty folders? What about files?

kyazdani42:
I'll try the icon after and come back to you. Adding more groups wouldnt be too hard, but it could take a bit more time.

kristijanhusak:
Ok, but icon before seems fine also, just to remove that one space before the icon to be aligned. For example, on screenshot above for folder nvim, removing a space before + sign would make it perfectly aligned with the next level.

kyazdani42:
I'll update the code so users can enable git icons or git highlight groups then (or both). Shouldn't be too hard.

kyazdani42:
I changed the default folder icons (thanks @Akin909) because the one i used took to much space and thus causing all kinds of weird visual effects. It should work fine now, could you try this out one more time to make sure everything works as expected ?

kristijanhusak:
@kyazdani42 looks good to me :)

Akin909:
Gave it a try also and it all works nicely :rocket: :100:

PR #69, opened 2020-08-02
Feat: Allow hiding dotfiles

kosayoda:
\o
This PR allows the user to toggle showing hidden files (files starting with a `.` character).
There is also a setting to hide them by default.

The toggling is not too performant the first time it runs, but I can't find a better way to do so other than checking the first character of every path.

Let me know what you think!


Akin909:
@kosayoda I believe this feature was already added recently by @kristijanhusak. The default mapping for it is `I` I think

kristijanhusak:
@kosayoda not sure if it will improve performance, but maybe setting some `hidden = true` flag on `node` when it is generated and comparing that flag instead of doing substring will work better?

kyazdani42:
yeah well `refresh_tree` needs to be optimized anyway, but the logic is a little bit complicated, this is why it might be a little bit slow.
I don't think this would affect performance too much overall. I'm testing this, if it doesn't make everything slow, i'll merge it :)

kyazdani42:
Also i think the plugin should not filter hidden files by default, as we developer work a lot with them.

lua/lib/populate.lua
```diff
@@ -6,7 +6,8 @@ local api = vim.api
 local luv = vim.loop
 
 local M = {
-  show_ignored = false
+  show_ignored = false,
+  show_hidden = vim.g.lua_tree_show_hidden == 1,
```
kyazdani42, line 10: i think this should be called `show_dotfiles` for clarity though.
the option should be `lua_tree_hide_dotfiles` because it makes more sense to show them by default.

kosayoda: sure

kyazdani42:
resolve the conflicts and i'll merge it

kosayoda:
Conflicts fixed, thank you for the quick review.

PR #79, opened 2020-08-03
Improvements

kristijanhusak:
These are the improvements I mentioned in https://github.com/kyazdani42/nvim-tree.lua/issues/76.

For multiple keys for single action, I added `o` as part of default for open action. I believe a lot of people are used to that mapping, especially if they come from nerdtree. If you disagree we can remove it.

For root folder name, I wasn't sure what is the best way to introduce it, so again, I copied nerdtree :)

I didn't provide an option to disable it, because that would introduce a lot more checks. If you think we need to have it, i'll introduce it.

Here's the screenshot how it looks for me:

screenshot

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -192,7 +192,7 @@ you can change default keybindings by defining this variable.
 default keybindings will be applied to undefined keys.
 >
     let g:lua_tree_bindings = {
-        \ edit:          '<cr>',
+        \ edit:          ['<cr>', 'o'], // Multiple keys provided via list
         \ edit_vsplit:   '<c-v>',
         \ edit_split:    '<c-x>',
         \ edit_tab:      '<c-t>',
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -146,6 +146,12 @@ Can be `0` or `1`. When `1`, will hide dotfiles, files or folders which start
 with the `.` character.
 Default is 0
 
+|g:lua_tree_root_folder_modifier|			*g:lua_tree_hide_dotfiles*
+
+In what format to show root folder. See `:help filename-modifiers` for
+available options.
+Default is `:~`
+
 ==============================================================================
 INFORMATIONS				        *nvim-tree-info*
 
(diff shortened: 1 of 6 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -146,6 +146,12 @@ Can be `0` or `1`. When `1`, will hide dotfiles, files or folders which start
 with the `.` character.
 Default is 0
 
+|g:lua_tree_root_folder_modifier|			*g:lua_tree_hide_dotfiles*
```
kyazdani42, line 149: `*g:lua_tree_root_folder_modifier*`

lua/lib/renderer.lua
```diff
@@ -160,13 +160,22 @@ local special = {
...
 }
 
+local root_folder_modifier = vim.g.lua_tree_root_folder_modifier or ':~'
+
 local function update_draw_data(tree, depth, markers)
   if tree.cwd and tree.cwd ~= '/' then
-    table.insert(lines, "..")
-    table.insert(hl, {'LuaTreeFolderName', index, 0, 2})
+    table.insert(lines, ".. (up a dir)")
+    table.insert(hl, {'LuaTreeFolderName', index, 0, 13})
```
kyazdani42, line 168: should we add this `up a dir` thing ? I guess people using unix already know about that ^^

kyazdani42: yep this `up a dir` takes too much space on small screens, my font for directories is bold and it does not look very good ^^

kyazdani42:
could you maybe switch the cwd and the `..` part, and add an empty line between the path and the rest ? i think it would look better.

lua/lib/renderer.lua
kyazdani42: could you also switch the `..` and the path so the path is at the top ?

kristijanhusak: @kyazdani42 so you mean to merge first two lines into one?
In my example, it would look like this?
```
~/github/nvim-tree/..
doc/
lua/
...
```

kristijanhusak: I just did that. It is indeed better:

screenshot

kyazdani42:
thanks :)

PR #87, opened 2020-08-05
Handle git deleted state.

kristijanhusak:
Git deleted state (` D`) wasn't handled, so it threw an error, because fallback wasn't set properly.
This PR introduces handling the delete state, and properly falling back to default dirty state if not found in the table. Feel free to adjust the icon, I wasn't sure what to put.

### DIFF
--- a/lua/lib/renderer.lua
+++ b/lua/lib/renderer.lua
@@ -1,5 +1,6 @@
 local colors = require'lib.colors'
 local config = require'lib.config'
+local utils = require'lib.utils'
 
 local api = vim.api
 
@@ -81,6 +82,7 @@ if vim.g.lua_tree_git_hl == 1 then
     local icons = git_hl[git_status]
 
     if icons == nil then
+      utils.echo_warning('Unrecognized git state "'..git_status..'". Please open up an issue on https://github.com/kyazdani42/nvim-tree.lua/issues with this message.')
       icons = git_hl.dirty
     end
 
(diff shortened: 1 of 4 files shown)


lua/lib/renderer.lua
```diff
@@ -71,13 +71,19 @@ if vim.g.lua_tree_git_hl == 1 then
...
   get_git_hl = function(node)
     local git_status = node.git_status
     if not git_status then return end
 
     local icons = git_hl[git_status]
+
+    if icons == nil then
+      icons = git_hl.dirty
+    end
+
```
kyazdani42, line 86: i think i prefer to get an error whenever we encounter a git state we don't have :)

lua/lib/renderer.lua
```diff
@@ -112,7 +119,7 @@ if icon_state.show_git_icon then
     if not git_status then return "" end
 
     local icon = ""
-    local icons = git_icon_state[git_status]
+    local icons = git_icon_state[git_status] or git_icon_state.dirty
```
kyazdani42, line 122: same here :)

lua/lib/renderer.lua
kristijanhusak: What do you think about a warning message alongside the fallback ? If it errors out it completely breaks the tree, it's a bit annoying when it happens.

kyazdani42: I think it might be a better solution indeed. Just print the git state that wasn't recognized (and maybe a suggestion to the user to open an issue with the warning message).

PR #91, opened 2020-08-14
Delete current buffer if tree is last window

akinsho:
Currently if `LuaTree` is the last window and `g:autoclose_tree` is set to `1` `:qa!` is executed. This command will **quit vim** which I'm _guessing_ is not the desired functionality because what this means is if you have the tree open and a buffer in the other split and press `q` or `bd` with the tree open your vim session will be closed which is very frustrating 😅 .

This PR switch the command to use to `:bd`, which works by unloading the tree buffer and closing **it's** window but not vim which is what I'd expect this command to do.

### DIFF
--- a/lua/tree.lua
+++ b/lua/tree.lua
@@ -131,7 +131,7 @@ end
 function M.on_leave()
   vim.defer_fn(function()
     if #api.nvim_list_wins() == 1 and lib.win_open() then
-      api.nvim_command(':qa!')
+      api.nvim_command(':bd!')
     end
   end, 50)
 end


kyazdani42:
Hum actually no, this option is for closing vim whenever the tree is the last window.
I have recently stopped using it because i now work on a large project and i need to keep my editor open.
When i was switching between multiple small project, i needed to: see the project structure, make some small changes, quit the buffer when im done which would also quit vim, and then use my terminal to go to some other place and repeat the process multiple times.
You could create a new option if you wished to use the `:bd` command in that fashion :)

akinsho:
Ah that makes sense @kyazdani42 I hadn't even thought of that type of work flow. If you're open to it then I will change this PR to introduce a new option like `g:lua_tree_switch_buffer_if_last` which will use `bd` instead

kyazdani42:
no worries, just come back here when you have decided :)

akinsho:
Turns out this isn't really necessary tbh 🤷🏿‍♂️ or at least I haven't felt the need in almost 2 months

helmecke:
I came across this because nvim was quitting if i changed worktrees with `<cmd>Telescope git_worktree git_worktrees<cr>`.

With the following change I got the desired behavior.
```diff
diff --git a/nvim/.config/nvim/lua/plugins/nvim-tree.lua b/nvim/.config/nvim/lua/plugins/nvim-tree.lua
index 1aaa1f6..a21fed0 100644
--- a/nvim/.config/nvim/lua/plugins/nvim-tree.lua
+++ b/nvim/.config/nvim/lua/plugins/nvim-tree.lua
@@ -8,6 +8,7 @@ vim.g.nvim_tree_icons = {
   symlink = '',
 }
 vim.g.nvim_tree_disable_window_picker = 1
+vim.g.nvim_tree_quit_on_open = 1
 
 local tree_cb = require('nvim-tree.config').nvim_tree_callback
 
@@ -15,7 +16,6 @@ require('nvim-tree').setup {
   disable_netrw = false,
   hijack_netrw = false,
   update_cwd = true,
-  auto_close = true,
   view = {
     mappings = {
       list = {
```

kyazdani42:
auto_close is conflicting with a lot of thing actually.

PR #92, opened 2020-08-18
feat: add ability for the tree to open in new tab when changing tab

### LINKED ISSUE #54: Open LuaTree on multiple/all tabs
I would love to have the ability to open LuaTree on multiple tabs. 

I use NERDTree with the [vim-nerdtree-tabs](https://github.com/jistr/vim-nerdtree-tabs) plugin, which basically makes the tree behave like a panel as found in apps like VSCode or Atom. 

This seems to be somewhat controversial (see [NERDTree#faq](https://github.com/preservim/nerdtree#can-i-have-the-nerdtree-on-every-tab-automatically)) but it best fits the way I use vim, and judging by the popularity of the vim-nerdtree-tabs plugin, I am not alone.

I would be happy to work on a PR if this is something amicable to the goals of the project but which you don't want to implement yourself.

P.s. wow, this is an excellent plugin. It is considerably faster than NERDTree. Great job Kiyan!

### LINKED ISSUE #63: [bug] tab pages broken when opening the tree in a new tab
The tree window is breaking neovim tab pages when opening the tree in a new tab.
- open neovim
- `:LuaTreeOpen`
- `:tabnew`
- `:LuaTreeClose`
- `:LuaTreeOpen`

kyazdani42:
`let g:lua_tree_auto_open=1`
fixes #63 (not really, it was api.nvim_win_set_option that made nvim segfault while changing tab and it was in my statusline...)
fixes #54


kyazdani42:
for information: this is hacky, i just close/open the tree on tabenter, but it works fine so i'll keep it this way :)

PR #100, opened 2020-09-04
Add option to rename files when copying/moving. Closes #95.

### LINKED ISSUE #95: Add more options to move/rename conflicts
When renaming or moving files to different folders, and file with same name already exists in the destination folder, it asks for overwrite, which has either yes or no.
It would be neat if it would allow choosing an option to either overwrite or rename. If rename is selected, it would show a pre-populated prompt where you can edit the filename before it is moved.

kristijanhusak:
This adds an option to rename a file/folder when doing copy or move.

### DIFF
--- a/lua/lib/fs.lua
+++ b/lua/lib/fs.lua
@@ -141,6 +141,30 @@ local function do_copy(source, destination)
   return true
 end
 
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
Nice ! Thanks again :)

PR #116, opened 2020-10-20
Fix setup snippet in README.md

geoboom:
Hey, I've just discovered the plugin and while it's pretty awesome so far, I found a little bug in the README vim setup snippet which results in syntax error after being loaded by neovim. I think you forgot to put a `\` before the last `}` to close the `g:lua_tree_bindings` dictionary.

Hope this helps others!


kyazdani42:
indeed ! thanks :)

Issue #119, opened 2020-10-21
Render error?


glepnir:
Run command `LuaTreeToggle`, Nothing to show on LuaTree buffer. but when I type `o` Show the folder.. and save file I got
```
E5108: Error executing lua ...github.com/kyazdani42/nvim-tree.lua/lua/lib/populate.lua:78: bad argument #1 to 'fs_scandir' (string expected, got nil)
```

glepnir:
OK find problem..If use `on_cmd`  with nvim-tree that will caused this error. if use `on_event` works fine.

PR #121, opened 2020-10-21
call on_enter in open and toggle function

### LINKED ISSUE #120: Remove VimEnter event?
Hey check here. First It will take some startuptime time. Second If we lazyload nvim-tree by dein `on_cmd`, dein will add the  nvim-tree.lua to `rtp` we trigger command `LuaTreeToggle`, So it will not trigger `tree.on_enter` function in `augroup LuaTree`. so will cause #119 ,  If use `on_event` with `BufEnter`, nvim-tree works fine. but check startuptime. vimenter events  will take some startuptime.
https://github.com/kyazdani42/nvim-tree.lua/blob/87a4d60bbcdf1161cb70d54a700106fd84b07f0f/plugin/tree.vim#L17

I write a function that works fine with dein `on_cmd  = LuaTreeToggle`, Just need to manually add nvim-tree to rtp before calling the command. This doesn't take startuptime and works fine.
```lua
function M.load_nvim_tree()
  if not vim.o.runtimepath:find("nvim-tree.lua") then
    vim.o.runtimepath = vim.o.runtimepath ..','.. global.cache_dir ..'dein/repos/github.com/kyazdani42/nvim-tree.lua'
  end
  local has_tree,tree = pcall(require,'tree')
  if has_tree then
    tree.on_enter()
  end
  pcall(vim.cmd,'LuaTreeToggle')
end

```
So is it more correct to add the on_enter function to `open` and `toggle`?

```log
110.599  033.494: VimEnter autocommands
110.614  000.015: UIEnter autocommands
110.620  000.006: before starting main loop
112.923  002.303: first screen update
112.926  000.003: --- NVIM STARTED ---
```

After: 
```log
075.940  007.822: VimEnter autocommands
075.943  000.003: UIEnter autocommands
075.947  000.003: before starting main loop
079.055  003.108: first screen update
079.058  000.003: --- NVIM STARTED ---
```

glepnir:
fix #120 and #119

### DIFF
--- a/lua/tree.lua
+++ b/lua/tree.lua
@@ -9,10 +9,11 @@ local api = vim.api
 
 local M = {}
 
-function M.toggle()
+function M.toggle()  
   if lib.win_open() then
     lib.close()
   else
+    M.on_enter()
     if vim.g.lua_tree_follow == 1 then
       M.find_file(true)
     else
@@ -30,6 +31,7 @@ end
 
 function M.open()
   if not lib.win_open() then
+    M.on_enter()
     lib.open()
   end
 end
--- a/plugin/tree.vim
+++ b/plugin/tree.vim
@@ -14,7 +14,6 @@ augroup LuaTree
   if get(g:, 'lua_tree_auto_close') == 1
     au WinClosed * lua require'tree'.on_leave()
   endif
-  au VimEnter * lua require'tree'.on_enter()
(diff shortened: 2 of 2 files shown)


glepnir:
Seems like write augroup  into a lua function will be better?

glepnir:
this will cause render multiple folder node ..

PR #123, opened 2020-10-24
Refactoring (rewrite almost everything)

kyazdani42:
Goals:
- provide a setup function where everything is loaded from so lazy loading works fine and the whole config is made in lua
- better code organization and proper naming
- better separation of modules so we can extend functionnalities more easily
- make the code a bit faster (even though it was quite fast already)
- maybe add workspaces
- better fs support (creating qfix buffers or editable popups over the tree)
- windows support ? (would need someone to test though)
- libgit2 instead of system commands (that would be nice)
- fs watcher:
  - a pain because fs_event_t doesn't allow recursive file watch on linux... only macos and windows are supported (who knows why...).
  - choices:
     - add a watcher on current cwd and sub cwds (1 depth), and add watchers for each explore call => will miss some fs changes
     - recursively add a watcher on every directories under the current directory => potentially will be very slow if running outside of a project
     - use an external tool and communicate over stdin handle to dispatch actions (such as inotify or watchexec) => that would require lot more code to write, might not be cross platform and users would need to install 3rd party binary

Todos:
- refresh logic
- find file
- git integration (libgit2 seems quite complicated for this use case, but we'll see. Also it would require user installation)
- faster code execution with function caching and some reviews
- fs edition support + copy/cut functionnalities (maybe add batch operations too)
- find a way to make setup less efficient... it's too fast and conflicts with colorscheme when setup on start

To think:
- git: LibGit2 seems not to be ready yet for use in neovim. The whole thing is complicated to setup for lua (yet, we have to wait for bfredl work to land in neogit). Even with this it might be too complicated for most neovim users to install this library, it would probably need to wait for neovim integrating libgit2 in core, which is not for tomorrow imo.
- and discuss:
  - watcher: find a proper solution
  - workspaces: think about how we would handle workspace cwds against vim cwd. Also might imply side effects on existing functionnalities. Also it should only work on projects (directories with .git, cargo.toml, makefile and so one).


Shatur95:
Any chance to have Windows support after it? I use two operating systems and it would be nice to have the same experience on both.

kyazdani42:
i have started to refactor with windows support in mind. Using the libuv and vim functions, it should be windows compatible (although i'm not sure about the C function from the unix standard lib like strrchr and stuff like this). If you can test on windows i'll ping you when it's good enough on my system. Maybe commands like copy cut etc etc would be harder for me to reproduce but if you know some windows programming i'll be happy to receive some help :)

Shatur95:
Yeah, ping me please, I will try to help with Windows :)

kyazdani42:
since this might take too long to complete, i went in favor of a step by step refactoring that i can merge progressively. I will list features from this PR as github issues that i'll resolve over time, based on what people want to implement on top of the existing apis.

PR #128, opened 2020-10-29
no wrap in luatree buffer

glepnir:
If file has long name and buffer has no width enough to show complete name, it will wrap in luatree buffer that will looks like ugly.
image

### DIFF
--- a/lua/lib/renderer.lua
+++ b/lua/lib/renderer.lua
@@ -284,6 +284,7 @@ function M.draw(tree, reload)
     api.nvim_win_set_cursor(tree.winnr, cursor)
   end
   api.nvim_buf_set_option(tree.bufnr, 'modifiable', false)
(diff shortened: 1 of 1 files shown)


kyazdani42:
please be less rude and make some research.
`au FileType LuaTree setlocal nowrap`
`:help nowrap`

glepnir:
> please be less rude and make some research.

Sorry, I want to say that I know the usage and api of neovim well. and I am maintaining neovim's built-in lsp and other lua plugins, I also know that autocmd can be used to do this, but referring to defx,Certain options should be built-in.
https://github.com/Shougo/defx.nvim/blob/2dd64ef7592de02d69153f0a2a58a2c693b12b59/rplugin/python3/defx/view.py#L465

Not all vim users know how to configure these options in some special buffers, if you want more people to use your plugin.

PR #130, opened 2020-10-31
fix: `Previewing a file with g:lua_tree_quit_on_open closes the file browser`

### LINKED ISSUE #129: Previewing a file with g:lua_tree_quit_on_open closes the file browser
### Issue
When trying to preview a file with the option `g:lua_tree_quit_on_open = 1` it will close the file browser instead of keeping it open as one would expect when only previewing the file.

I believe the problem is related to the function `M.open_file` in `lib.lua` where we have the statement

```lua
if vim.g.lua_tree_quit_on_open then
  M.close()
end
```

which does not consider what mode we are in.

### Expected behaviour
With `g:lua_tree_quit_on_open = 1`
When previewing a file (`<Tab>` with default keybindings) I expect the file browser to stay open
When editing a file (`o` or `<Cr>` with default keybindings) I expect the file browser to close

### Actual behaviour
When previewing a file (`<Tab>` with default keybindings) the file browser is closed
When editing a file (`o` or `<Cr>` with default keybindings) the file browser is closed

### Possible solution

Add another condition to the if-statement checking what mode we are in

```lua
function M.open_file(mode, filename)
  ...
  if vim.g.lua_tree_quit_on_open and mode ~= 'preview' then
    M.close()
  end
end
```

mortepau:
Fixes #129

### DIFF
--- a/lua/lib/lib.lua
+++ b/lua/lib/lib.lua
@@ -202,7 +202,7 @@ function M.open_file(mode, filename)
     api.nvim_command('vertical resize '..M.Tree.win_width)
     M.win_focus(cur_win)
   end
-  if vim.g.lua_tree_quit_on_open == 1 then
+  if vim.g.lua_tree_quit_on_open == 1 and mode ~= 'preview' then
     M.close()
   end
 end


kyazdani42 (approved):
thanks :)

PR #139, opened 2020-11-25
Fixes non-git project slow saving

willfish:
## What

The configured BufWritePost autocmd refreshes the current Tree
and all of its descendant nodes.

This is fine in actual git projects but makes file writes really slow
when in my home directory (it recurses through all of the nodes - quite a
few of them if you have node_modules installed).

In this PR I'm updating the behaviour to only refresh git in git projects

## Why

To speed up saving files in not git projects

## AC

- [x] I've tested that it's fast on non-git project trees
- [x] I've tested that this still updates the icons for git project trees

### DIFF
--- a/lua/lib/lib.lua
+++ b/lua/lib/lib.lua
@@ -120,12 +120,17 @@ end
 function M.refresh_tree()
   -- local stat = luv.fs_stat(M.Tree.cwd)
   -- if stat.mtime.sec ~= M.Tree.last_modified then
-    refresh_nodes(M.Tree)
+  refresh_nodes(M.Tree)
   -- end
+  local _, status = git.get_git_root(M.Tree.cwd)
+
   if config.get_icon_state().show_git_icon or vim.g.lua_tree_git_hl then
-    git.reload_roots()
-    refresh_git(M.Tree)
+    if status ~= git.not_git then
+      git.reload_roots()
+      refresh_git(M.Tree)
+    end
   end
+
   if M.win_open() then
     renderer.draw(M.Tree, true)
   else
--- a/lua/lib/git.lua
+++ b/lua/lib/git.lua
@@ -27,7 +27,19 @@ function M.reload_roots()
   end
 end
 
-local function get_git_root(path)
+local function create_root(cwd)
+  local git_root = vim.fn.system('cd "'..cwd..'" && git rev-parse --show-toplevel')
+
+  if not git_root or #git_root == 0 or git_root:match('fatal') then
+    roots[cwd] = not_git
+    return false
+  end
+
+  update_root_status(git_root:sub(0, -2))
+  return true
+end
+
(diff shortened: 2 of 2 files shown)


willfish:
Just seen this PR https://github.com/kyazdani42/nvim-tree.lua/pull/123 @kyazdani42. From what I could understand from your significant and massively cleaner refactor you're no longer relying on the BufWritePost autocmd to generate keep the git icons up-to-date?

Would you prefer that I close this and reduce the need for you to resolve conflicts in your branch and just roll this as a patch locally until you've made your change?

kyazdani42:
Well thanks for this PR :)
the refactor might take some time because i need to figure out some things before going forward (like how to implement a watcher). Basically there is the concept of being in a project or moving around in the filesystem that causes me many headaches so i'm still wondering how to do things.
Also, the libuv that i want to use to watch files has a watcher that works recursively on windows and macos, but not on linux so implementing a performant recursive watcher is kind of hard. I might need to add testing to make sure performance is not degraded.
Well, as for this PR, i will test it briefly when i can to make sure it doesn't break anything and i'll merge it :) i do not worry about conflicts because i've changed the file hierarchy and naming.

willfish:
So I'm super not familiar with lua but could you use https://github.com/EvandroLG/lua-watcher? Or is that not compatible with LuaJIT?

willfish:
Thanks @kyazdani42

kyazdani42:
i would like to avoid deps :) libuv is already good enough imo

lua/lib/lib.lua
```diff
@@ -120,12 +120,17 @@ end
...
+
   if config.get_icon_state().show_git_icon or vim.g.lua_tree_git_hl then
-    git.reload_roots()
-    refresh_git(M.Tree)
+    if status ~= git.not_git then
+      git.reload_roots()
+      refresh_git(M.Tree)
+    end
   end
+
```
kyazdani42, line 133: did you test the case where:
your root is not a git repository, but some folders inside are git repositories :) ?

willfish: So I've simplified quite a bit by just saying if the root of the tree isn't a git directory then don't recurse through the branches

willfish: I'm not actually sure how we would keep the aucmd fast and still recurse through all of the branches? I'm basically assuming that the majority case use of vim is in a project directory that's initialised for git or for making configuration changes in potentially vast trees. The truth is that that assumption won't cover some use cases and therefore people will be losing out functionality for speed.

kyazdani42:
Ok so i might have found an easier solution.
could you try #142 and tell me if this fixes your performance issues ? It just delays the refresh on the event loop so the ui does not block.

willfish:
Fixed https://github.com/kyazdani42/nvim-tree.lua/pull/142

PR #142, opened 2020-11-27
wrap refresh logic in vim.schedule callback

kyazdani42:
fix refresh textlock issue


willfish:
I just ran this locally and giving it a go. What is happening is the first save of a config file in my home directory takes the usual amount of time but subsequent runs are slower.

Running it in a project directory it's as fast as ever

kyazdani42:
allright, we can close your PR then ?

PR #144, opened 2020-11-30
fix tree winnr

smartding:
This PR should fix https://github.com/kyazdani42/nvim-tree.lua/issues/134
tree.winnr is a function that returns winnr of luatree.
So we should use tree.winnr() instead of tree.winnr.


kyazdani42 (approved):
woups thanks

PR #147, opened 2020-12-02
fix issue with data race when opening with luatreefindfile

### LINKED ISSUE #146: LuaTreeFindFile positions cursor at the top
Cursor should be positioned to the current file opened and not the top

Exists on this commit ba4dac1e5904ddf91f3993bb9119d0487cb16a1d

kyazdani42:
fixes #146

### DIFF
--- a/lua/tree.lua
+++ b/lua/tree.lua
@@ -28,9 +28,12 @@ function M.close()
   end
 end
 
-function M.open()
+function M.open(cb)
   if not lib.win_open() then
-    vim.schedule(lib.open)
+    vim.schedule(function()
+        lib.open()
+        cb()
+      end)
   end
 end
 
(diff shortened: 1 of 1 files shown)


kristijanhusak:
After this PR, when tree is opened, and you try to do `:LuaTreeFindFile`, it doesn't do anything. Before it focused the tree window on the given file.

kyazdani42:
@kristijanhusak can you provide more information ? Works for me.

kristijanhusak:
Here the gif:

[
luatree
](url)

I executed `LuaTreeFindFile`, but tree didn't got the focus.

kyazdani42:
i pushed a small fix, the tree cursor properly updates but it doesn't focus the tree though. Tell me if it's all ok for you

kristijanhusak:
I hoped to get it to focus, at least that was the old functionality.

kyazdani42:
Fixed in master, please tell me if there are other things that went wrong !

kristijanhusak:
@kyazdani42 works great now, thanks!

PR #153, opened 2020-12-14
Add nil check for link_to

s-zeng:
I've been getting some errors on startup due to link_to being nil, so I figured I should add a nil check. Let me know if you have any suggestions.


kyazdani42 (approved):
oh yes indeed i forgot realpath sometimes fails silently even with a base C code for no reason. I even added a comment for that. thanks !

PR #157, opened 2020-12-15
feat: g:lua_tree_ignore support *.extension

cqroot:
For example, I don't want to display `*.pyc` files in python project. Now I can achieve it by adding `*.pyc` to the `g:lua_tree_ignore`.

This is a simple but small change implementation.


kyazdani42 (approved):
well tested and this works without affecting performance too much.
Don't know if toggle ignored would work on this though.

PR #162, opened 2020-12-20
fix file icon length

### LINKED ISSUE #160: incorrect render of file icons (folder icons are ok)
I'm using suckless terminal st, and GoMono Nerd Font.
The folder icons are properly rendered, but only the left half of file icons are displayed. 
As shown in the following figure, I can only see the left half of file icons for lua, vim and git files. 
pic-selected-201218-1517-52

smartding:
This PR fixes the inconsistency in how we calculate the length of folder icon and file icon. 
Length of folder icon was calculated after a space is appended, and length of file icon was calculated without any space appended.
It is fixed by adding one after `#icon` for file icons.

This PR should fix #160

### DIFF
--- a/lua/lib/renderer.lua
+++ b/lua/lib/renderer.lua
@@ -42,7 +42,7 @@ if icon_state.show_file_icon then
 
     if icon then
       if hl_group then
-        table.insert(hl, { hl_group, line, depth, depth + #icon })
+        table.insert(hl, { hl_group, line, depth, depth + #icon + 1 })
       end
       return icon.." "
     else


kyazdani42:
well actually this is weird that st doesn't render properly, i'll merge this for you but i think the issue actually comes from somewhere else (you wouldn't need to actually highlight the space after the icon right ?).

smartding:
> well actually this is weird that st doesn't render properly, i'll merge this for you but i think the issue actually comes from somewhere else (you wouldn't need to actually highlight the space after the icon right ?).

Thanks.
I tested with my zsh prompt in st, in order to highlight a wide glyph, I had to highlight both the glyph and the space after it. Highlighting the space after the glyph is not required by other terminals such as alacritty. 
I went through the st patches, there seems none that fix this behavior.

PR #170, opened 2021-01-07
Add dir_up function to expose navigating up one directory

jdelkins:
Just a suggestion. I like to use the '-' key to open the tree, then inside the tree, to use '-' to quickly navigate up through the directory structure. I couldn't figure out how to do that second part without adding a function to the main package. If there's a better way to do this, I'd love to hear it. Thanks

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -105,6 +105,19 @@ function M.on_keypress(mode)
   end
 end
 
+function M.dir_up()
+    local node = lib.get_node_at_cursor()
+    if not node then
+        return lib.change_dir('..')
+    else
+        local newdir = vim.fn.fnamemodify(node.absolute_path, ':h')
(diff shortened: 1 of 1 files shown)


kyazdani42:
Hi thanks for the PR :)
i guess you could bind the keymap through lua like this:
```vim
au BufNewFile,BufRead NvimTree lua vim.api.nvim_buf_set_keymap(0, 'n', '-', 'lua require"nvim-tree".dir_up()', { silent = true })
```

kyazdani42 (approved):

jdelkins:
That's exactly right. If you think this would be generally useful, I'd be happy to add a default key map for it in another PR.

kyazdani42:
There is a mapping table in `lib.lua` (`local mappings`) and a keybindings table in `config.lua` that could be used for a default key map

PR #173, opened 2021-01-10
Move dir_up functionality into lib and make '-' a default keymap for it

jdelkins:
As discussed, I made this into a default keymap & also updated the documentation. Thanks!

### DIFF
--- a/lua/lib/lib.lua
+++ b/lua/lib/lib.lua
@@ -252,6 +252,7 @@ local function set_mappings()
     [bindings.paste] = 'on_keypress("paste")';
     [bindings.prev_git_item] = 'on_keypress("prev_git_item")';
     [bindings.next_git_item] = 'on_keypress("next_git_item")';
+    [bindings.dir_up] = 'on_keypress("dir_up")';
     gx = "xdg_open()";
   }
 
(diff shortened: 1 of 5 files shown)


lua/lib/lib.lua
```diff
@@ -366,4 +367,16 @@ function M.toggle_dotfiles()
...
 
+function M.dir_up(node)
+  if not node then
+    return M.change_dir('..')
+  else
+    local newdir = vim.fn.fnamemodify(node.absolute_path, ':h')
+    if newdir == M.Tree.cwd then
+      M.change_dir('..')
+    end
+    return M.set_index_and_redraw(newdir)
```
davidsu, line 378: I find it confusing when going up a dir it closes the expanded path to the current node and jumps the cursor to the top of the tree, i'd rather see it this way
```suggestion
    return M.set_index_and_redraw(node.absolute_path)
```
what do you think?

davidsu, line 378: I find it confusing when going up a dir it closes the expanded path to the current node and jumps the cursor to the top of the tree, i'd rather see it this way
```suggestion
    local newdir = vim.fn.fnamemodify(M.Tree.cwd, ':h')
    if newdir == M.Tree.cwd then
      M.change_dir('..')
    end
    return M.set_index_and_redraw(node.absolute_path)
```
what do you think?

jdelkins: Yes, I like that better. I got rid of the `if newdir..end` part as that was blocking anything from happening.

One small problem with this approach is that there is no way to navigate back down to the current working directory, so I also added a small patch to `lib.open` to change directories back to the invoking window's cwd when the tree window is reopened. I tried a few approaches to accomplish this, and this seemed to work the best, but certainly open to input!

Thanks for considering!

kyazdani42 (approved):
thanks again :)

PR #174, opened 2021-01-11
Execute action in target window if possible (fixes #171)

### LINKED ISSUE #171: Preserve focused window
Hi, I really enjoy your plugin, the best file-manager for me atm!
If I have at least one split and currently focused window is **not** top-left, then after opening and closing nvim-tree window, the focus gets re-setted to the top-left window (and if I open file with `o`, it also opens in the top-left). I tried to restore position using `line("'\"")`, but it always gives me `1`, which means cursor position isn't saved? Is there such option to solve this, or maybe some workaround?

jdelkins:
Hi, this is for your consideration. If you don't buy my arguments, I understand.

I think this addresses the concern in issue #171 and maybe a few others

Rationale: I prefer somewhat different semantics than the existing code when opening a node, as described below. To me, this is more intuitive and more or less follows what NERD_Tree does (as I recall -- been a while since I used it). For example, some things that are better in my view with this approach:

* If I start nvim-tree from the leftmost window while that buffer is modified, the default edit command would fail to open the file. With this approach, if the window contains a modified buffer, it will split automatically.

* if I open nvim-tree from the rightmost window, when I open a node it will replace my leftmost buffer, which i most likely didn't want to change.

With this approach, we open files in the window from which the tree was opened, if possible. I call this window the target. If the buffer in the target window is modified, then execute the configured split (split or vsplit) relative to the target. If the target is unmodified, open the selected node in the target, hiding original the buffer. Other windows are untouched. If the selected node is already visible in a window, don't do any of the foregoing; instead, just jump to that window. If the target has been closed, the preserve the original semantics, except that the modified check is still done.

### DIFF
--- a/lua/lib/lib.lua
+++ b/lua/lib/lib.lua
@@ -20,6 +20,7 @@ M.Tree = {
   win_width_allow_resize = vim.g.nvim_tree_width_allow_resize,
   loaded = false,
   bufnr = nil,
+  target_winid = nil,
   winnr = function()
     for _, i in ipairs(api.nvim_list_wins()) do
       if api.nvim_buf_get_name(api.nvim_win_get_buf(i)):match('.*/'..M.Tree.buf_name..'$') then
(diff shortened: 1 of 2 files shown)


lua/lib/lib.lua
```diff
@@ -189,17 +190,47 @@ local function check_and_open_split()
 end
 
 function M.open_file(mode, filename)
+  local target_winnr = api.nvim_eval(string.format('win_id2win(%d)', M.Tree.target_winid))
```
akinsho, line 193: just a quick note that you can much more easily use these vim functions by doing
`vim.fn.win_id2win(value)` directly in lua rather than the longwinded `api.nvim_eval` calls
the same goes for `vim.fn.winbufnr`. Essentially any vim functions can be accessed using `vim.fn.func_name` (if it is a lua keyword you'd have to do `vim.fn['function'](value)`

jdelkins: Yes, of course. Fixed.

Thanks for pointing it out.

kyazdani42 (approved):
looks good to me, the file opening behavior was indeed a bit clunky at times.

cloggier:
> With this approach, we open files in the window from which the tree was opened, if possible. I call this window the target. If the buffer in the target window is modified, then execute the configured split (split or vsplit) relative to the target. If the target is unmodified, open the selected node in the target, hiding original the buffer.

I don't think this is what everyone wants. I have `hidden` set and don't want automatic splits, there are mappings for intentional splitting. Perhaps it either should be configurable or depend on users' `hidden` setting.

jdelkins:
> I don't think this is what everyone wants. I have `hidden` set and don't want automatic splits, there are mappings for intentional splitting. Perhaps it either should be configurable or depend on users' `hidden` setting.

Thanks for pointing this out. Submitted a new PR that honors the `hidden` setting. Can you try it and see if that does what you expect?

PR #177, opened 2021-01-13
fixes #175 | early return if internal state is invalid(does not match…

### LINKED ISSUE #175: keep getting `Invalid buffer id`
I get the following message `Error executing lua ...ite/pack/packer/start/nvim-tree.lua/lua/lib/renderer.lua:284: Invalid buffer id: 7`, do you know what this is about?

davidsu:
… nvim's state

### DIFF
--- a/lua/lib/renderer.lua
+++ b/lua/lib/renderer.lua
@@ -279,9 +279,10 @@ end
 
 local M = {}
 
-function is_bufnr_valid(bufnr)
+local function is_bufnr_valid(bufnr)
   return vim.fn.bufexists(bufnr) == 1
 end
+
 function M.draw(tree, reload)
   if not is_bufnr_valid(tree.bufnr) then return end
   api.nvim_buf_set_option(tree.bufnr, 'modifiable', true)
--- a/lua/lib/renderer.lua
+++ b/lua/lib/renderer.lua
@@ -279,8 +279,11 @@ end
 
 local M = {}
 
+function is_bufnr_valid(bufnr)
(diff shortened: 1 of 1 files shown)


lua/lib/renderer.lua
```diff
@@ -300,7 +303,7 @@ function M.draw(tree, reload)
 end
 
 function M.render_hl(bufnr)
-  if not bufnr then return end
+  if not is_bufnr_valid(bufnr) then return end
```
davidsu, line 306: this was throwing on re-source of init.vim

akinsho:
@davidsu I'm definitely keen for a fix, looking at this does it stop the tree from being opened or drawn? I'm guessing not as when I tend to get this error nvim tree isn't even open so I'm guessing its during the `BufEnter` autocommand where this is failing (This is me just asking out of curiosity).

davidsu:
The tree opens properly.

lua/lib/renderer.lua
```diff
@@ -279,8 +279,11 @@ end
 
 local M = {}
 
+function is_bufnr_valid(bufnr)
+  return vim.fn.bufexists(bufnr) == 1
+end
 function M.draw(tree, reload)
-  if not tree.bufnr then return end
+  if not is_bufnr_valid(tree.bufnr) then return end
```
kyazdani42, line 286: could you just add an empty line between these two function. Also declaring is_bufnr_valid as local to avoid conflicts :)

kyazdani42 (changes requested):

kyazdani42 (approved):

PR #179, opened 2021-01-16
Honor the 'hidden' option when deciding whether to split the target window

jdelkins:
- Honor `hidden` setting when deciding whether to split window
- Remove unused `check_and_open_split()`

Refer to [this comment](https://github.com/kyazdani42/nvim-tree.lua/pull/174#issuecomment-761575598)

### DIFF
--- a/lua/lib/lib.lua
+++ b/lua/lib/lib.lua
@@ -183,12 +183,6 @@ function M.set_index_and_redraw(fname)
   end
 end
 
-local function check_and_open_split()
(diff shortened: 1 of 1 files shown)


cloggier:
Ah, that's amazing! Thanks a bunch!

cloggier:
> Thanks for pointing this out. Submitted a new PR that honors the hidden setting. Can you try it and see if that does what you expect?

Seems to be working fine. Thanks!

kyazdani42 (approved):

PR #181, opened 2021-01-17
Don't do `git status -u` if repo is configured not to show untracked files

jdelkins:
Some git use cases, specifically, in my case, where $HOME is a git worktree
with a bare git repo somewhere else, really get bogged down if you do `git
status -u`. It was making the plugin unusable when `GIT_WORK_TREE=$HOME`
environment variable is set as such. This overcomes that issue.

### DIFF
--- a/lua/lib/git.lua
+++ b/lua/lib/git.lua
@@ -6,7 +6,11 @@ local roots = {}
 local not_git = 'not a git repo'
 
 local function update_root_status(root)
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
i'm a little worried it might impact performance but i'll give it a go anyway, thanks again :)

jdelkins:
> i'm a little worried it might impact performance but i'll give it a go anyway, thanks again :)

I don't think a one call to `git config` will be noticeable compared to what happens during `git ls-files`, however I could be wrong. Thanks for merging!

PR #187, opened 2021-01-22
fix: escape filename before opening it

elianiva:
I have a weird case where I need to open a file with `$` in its filename. ex. `$layout.svelte`, if we don't escape it, it will open a buffer with `.svelte` as its filename instead of `$layout.svelte`


kyazdani42 (approved):
thanks !

PR #195, opened 2021-02-08
Give users the option to enable netrw

dorgnarg:
I wound up wanting to use netrw alongside nvim-tree, but I realized I couldn't as-written. I saw that you said it wasn't something you were interested in in #47 because you didn't find it useful, but since there's at least some demand for it (and I'd like to use it), I thought it would be nice to include it as an option :)


kyazdani42:
yes, see #169

dorgnarg:
Updated this PR a little bit, there are now two new variables for dealing with netrw:
`g:nvim_tree_disable_netrw` (default: 1) disables netrw altogether (default behavior as it is now)
`g:nvim_tree_hijack_netrw` (default: 1) stops netrw from opening when you open nvim on a directory when activated, but if netrw is enabled, it doesn't prevent any of its other utilities.

As of right now, based on @CantoroMC 's comment in #169:
1. If the user bypasses disabling netrw, then `nvim dir` will function the same as it does currently. That way you get the netrw tools some people have been wanting, without worrying about funky netrw behavior
2. If the user bypasses disabling netrw, *and* bypasses hijacking, then `nvim dir` will open the directory with netrw, and will prevent nvim-tree from auto-opening, so it will be as if you're just using netrw normally

Nothing is using `VimEnter` still, but the hijacking is using `FileExplorer` events to suppress netrw

kyazdani42:
That's a smart move, but i'm not too sure about leaving this as option. To me, the best way to use this would be nvim-tree +netrw functionnalities. 
Could you provide a use case for not hijacking netrw always ?

dorgnarg:
The main use-case I'm considering is a full-window `:Explore` option. Sometimes when I'm browsing I'm cool with a drawer but sometimes I'd prefer to have it take up a full window, but nvim-tree doesn't currently have that option and I don't know if I know my way around lua well enough to implement it. As a side-note, I think this may have been what #200 was talking about as well.

This doesn't necessarily solve the problem of nvim-tree handling it entirely like NERDTree does when hijacking netrw, but it does still allow for full-window browsing while using the plugin. If it doesn't seem like something super kosher to the project, or if you'd rather plan to implement it within nvim-tree and not give the option to let netrw handle it ever, I can remove the hijack option.

akinsho:
I think the ability to edit a directory would be super useful, there was another similar issue to #200 where we discussed that and I briefly had a go at trying to add that to nvim-tree but didn't get too far short of doing a similar thing and hijacking netrw as this PR does but actually opening nvim-tree in a full window caused some issues. I think this would be a really useful thing to have for the time being for anyone else who misses/needs that netrw behaviour till it's implemented here

kyazdani42:
indeed, let's merge this and we'll update the behavior if needed.

PR #199, opened 2021-02-09
Add option to don't auto open tree on specific filetypes

sbulav:
When you set `vim.g.nvim_tree_auto_open = 1` , tree opens when vim is without any arguments.
But this behavior  is conflicts with plugins like `vim-startify` or `dashboard-nvim` as tree is opening and gain focus.

This PR adds new option `nvim_tree_auto_ignore_ft` which disables auto open on specific filetypes


kyazdani42 (approved):
Thanks a lot !

Issue #193, opened 2021-02-05
Intro disappeared after installed


redbug312:
The minimal config installed with packer:

```lua
vim.cmd [[packadd packer.nvim]]

return require('packer').startup(function()
    use {'wbthomason/packer.nvim', opt = true}
    use 'kyazdani42/nvim-tree.lua'
end)
```

And the expected intro screen will disappear after installed.
![intro](https://user-images.githubusercontent.com/11537681/107015678-fb7f8500-67d7-11eb-9dd6-c62fb1cd8f9f.png)

It seems highlight commands in [`colors.setup()`][code] triggers redraw and cleans the intro screen.
Can that part be separated into a `syntax/` folder?

[code]: https://github.com/kyazdani42/nvim-tree.lua/blob/b2852578769b53430af678ae7a99482f2a4be0da/lua/lib/colors.lua#L75

kyazdani42:
i try to avoid running vimscript, so maybe this issue should be addressed in another way. Maybe loading the syntax through a lua call in the syntax folder would help. Not sure about that, i'm really busy atm so if you feel like doing this go for it :)

osa1:
Came here to report this. The intro shows for maybe half a second, then disappears.

kyazdani42:
closed in #202

gegoune:
This is still an issue, not sure whether root cause is the same, but with nvim-tree installed intro screen shows very briefly at start and then disappears.

kyazdani42:
not sure when it came back, but i sure should take a look

kyazdani42:
i think this is not related to nvim-tree because when disabled, the intro still is erased. Might be caused by another plugin.

z0xyz:
The exact same problem is still persistent !
@kyazdani42 I guess The bug should be reopened in order to address this issue

kyazdani42:
It is not with my config, as i said before, it must be caused by another plugin.

z0xyz:
I've already narrowed down the factors that might potentially cause that by disabling all of the additional plugins that i have, left nvim-tree the only one remaining. Yet the problem still persist .

kyazdani42:
can you provide at least the minimal configuration with nvim-tree where it happens ? I cannot reproduce this.

z0xyz:
All i have is the exact default [configuration](https://www.lifewire.com/best-free-sports-streaming-sites-4169519) without any modifications .

tobiassen:
[...]

PR #202, opened 2021-02-13
Avoid intro screen cleared

RedBug312:
This solves #193 as I posed.
I try to avoid `colors.setup()` being deferred by moving it into `plugin/tree.vim`.

### DIFF
--- a/lua/lib/colors.lua
+++ b/lua/lib/colors.lua
@@ -62,6 +62,7 @@ local function get_links()
     FileMerge = 'NvimTreeGitMerge',
     FileStaged = 'NvimTreeGitStaged',
     FileDeleted = 'NvimTreeGitDeleted',
+    Popup = 'Normal',
   }
 end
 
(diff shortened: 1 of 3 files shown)


kyazdani42:
this PR breaks icon display somehow. I use packer and i don't lazy load nvim-tree (because it's a tree, eh), and i think variable assignation is made after the call to reset_highlight, which prevents calling `nvim-web-devicons.setup`

kyazdani42:
this PR breaks icon display somehow. I use packer and i don't lazy load nvim-tree (because it's a tree, eh), and i think variable assignation is made after the call to reset_highlight, which prevents calling `nvim-web-devicons.setup`

RedBug312:
Oh, I mean deferred to be the line `defer-fn` at [lua/nvim-tree.lua](https://github.com/kyazdani42/nvim-tree.lua/blob/48cd6a1ad31754aa722b53d45d5d923236de4083/lua/nvim-tree.lua#L219).
Now the call `colors.setup` is moved from `plugin/tree.lua` to `lua/nvim-tree.lua`. I'm not sure about nvim loading order, but this resolves the icon problem.

kyazdani42:
indeed, looks like it works now, thanks :)

PR #206, opened 2021-02-18
change behaviour of rename

seblj:
Changes the behaviour of renaming a file or folder. This will remove the current filename from the absolute path, so it's not necessary to backspace the current name before typing in the new name.

### DIFF
--- a/lua/lib/fs.lua
+++ b/lua/lib/fs.lua
@@ -241,7 +241,8 @@ end
 function M.rename(node)
   if node.name == '..' then return end
 
-  local abs_path = string.gsub(node.absolute_path, node.name, '')
+  local namelen = node.name:len()
+  local abs_path = node.absolute_path:sub(0, namelen * (-1) -1)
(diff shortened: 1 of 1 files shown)


lua/lib/fs.lua
```diff
@@ -241,7 +241,8 @@ end
 function M.rename(node)
   if node.name == '..' then return end
 
-  local new_name = vim.fn.input("Rename " ..node.name.. " to ", node.absolute_path)
+  local abs_path = string.gsub(node.absolute_path, node.name, '')
```
kyazdani42, line 244: I'm not sure if this would work for every case, could you try renaming a file with a `%` or `.` inside ? I'm not too sure but i've encountered this issues with lua regexes a lot, and i believe gsub does apply a regex as 2nd argument, which would lead to undefined behavior or bugs

seblj: You are correct, I will look into this!

seblj:
I think this should work. Find the length of current filename to figure out path without the filename

kyazdani42 (approved):
seems right ;) thanks

PR #216, opened 2021-02-26
Fix: wrong dictionary format example for empty & empty_open on doc

bandithijo:
There is some wrong example on `g:nvim_tree_icons` dictionary value for folder `empty` & `empty_open`.
I guess that example is for lua.


kyazdani42:
indeed thanks

Issue #208, opened 2021-02-22
Windows support


ZenoArrow:
Was just curious what could be done to add Windows support. Does this depend on changes being implemented in some of the nvim-tree dependencies?

kyazdani42:
Hi, windows support would need a full proper path handling (resolving paths with `\` instead of `/`), i don't think it would be too hard, but it could raise new issues.
There might be some libuv calls that might differ from unix based systems. I'm not too sure and cannot test it out, i don't possess any windows machine.
I would accept a PR, that i would test properly on linux to make sure nothing breaks, and i'd trust windows support to anybody who wishes to maintain it.

PR #222, opened 2021-03-06
Patch windows

pudymody:
Hi, i found your plugin on Reddit and i liked its simplicity! Good job.
Sadly in my try to test it, i found out that it wasnt able to execute on windows.

Here is a patch for that.
I dont know if EVERYTHING is working, but basic navigation, creation of folder/files, copy/cut and deletion was working on my tests.

It also checks that the current version of Neovim is at least 0.5, otherwise it doesnt load itself.

Fixes: #208

### DIFF
--- a/lua/nvim-tree/git.lua
+++ b/lua/nvim-tree/git.lua
@@ -21,10 +21,10 @@ local function update_root_status(root)
       body = body:gsub('^.* %-> ', '')
     end
 
-		--- Git returns paths with a forward slash wherever you run it, thats why i have to replace it only on windows
-		if is_win then
-			body = body:gsub("/", "\\")
-		end
+    --- Git returns paths with a forward slash wherever you run it, thats why i have to replace it only on windows
+    if is_win then
+      body = body:gsub("/", "\\")
+    end
 
     roots[root][body] = head
   end
--- a/lua/nvim-tree/git.lua
+++ b/lua/nvim-tree/git.lua
@@ -4,6 +4,7 @@ local M = {}
 local roots = {}
 
 local not_git = 'not a git repo'
+local is_win = vim.api.nvim_call_function("has", {"win32"}) == 1
 
 local function update_root_status(root)
   local untracked = ' -u'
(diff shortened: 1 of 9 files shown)


pudymody:
I've just found out that git integration doesnt work, so i think its not ready to merge yet.
Im sorry i havent tested it enough.

kyazdani42:
cool thanks :) tell me when you're done i'll review the code still works on linux too !

pudymody:
Now everything should work as expected.

lua/nvim-tree/git.lua
```diff
@@ -53,6 +60,11 @@ local function create_root(cwd)
     return false
   end
 
+	local is_win = vim.api.nvim_call_function("has", {"win32"})
+	if is_win then
+		git_root = git_root:gsub("/", "\\")
+	end
+
```
kyazdani42, line 67: this file is not properly indented

kyazdani42 (changes requested):
I think the path_separator code is called many times, it could be loaded only once and accessed through it's own module, providing functions like `join_path`. Also why package.config instead of `vim.fn.has('win32') and '\\' or '/'` ? could windows have `/` path separators ?

pudymody:
> Also why package.config instead of `vim.fn.has('win32') and '\\' or '/'`

No particular reason, i just prefer lua to tell me what the separator is, instead of defining it myself.

> could windows have / path separators ?

In windows the '/' character is used to separate folders/files, but i liked the consistency of having all the paths with '\\' and not mix between both.

As per the module, is it okay if i put it as a function inside the utils module?

kyazdani42:
allright, yes go for it

pudymody:
Done :)

lua/nvim-tree/git.lua
```diff
@@ -60,10 +60,10 @@ local function create_root(cwd)
     return false
   end
 
-	local is_win = vim.api.nvim_call_function("has", {"win32"})
-	if is_win then
-		git_root = git_root:gsub("/", "\\")
-	end
+  local is_win = vim.api.nvim_call_function("has", {"win32"})
```
kyazdani42, line 63: you could call this function only once, not on very create_root call :)

kyazdani42 (changes requested):
one last small thing and we're good to go. I'm testing this PR on linux in the meanwhile to make sure nothing was broke

kyazdani42:
This break git integration on linux not sure why

pudymody:
You were absolutely right, i was silly and forgot to check that the condition was true. (i thought that it returned an integer).
Now everything should be working again.

kyazdani42:
looks good, thanks for working on that. You can also update the README and the doc file to remove the `not working on windows` information and i'll merge this :)

kyazdani42:
thanks for the contribution !

PR #232, opened 2021-03-14
Fix g:nvim_tree_git_hl is not disabling Git status reloading

levirs565:
g:nvim_tree_git_hl is not disabling Git status reloading even when g:nvim_tree_show_icons.git is 0.


kyazdani42 (approved):
thanks :)

PR #247, opened 2021-03-25
feat: Option for grouping empty directories

### LINKED ISSUE #191: Group empty folders
Is there a way to accomplish this?
VSCode handles it pretty well and I believe it may be a good improvement especially for those, like me, who have to deal with java.

nvim:
[2021-01-29T091151](url)

vscode:
2021-01-29T091245

sindrets:
Fixes #191.
This adds an option `g:nvim_tree_group_empty` (disabled by default) that compacts dirs that only contain 1 dir / symlink dir, into one node in the file tree.

2021-03-25-012655_maim

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ let g:nvim_tree_width_allow_resize  = 1 "0 by default, will not resize the tree 
 let g:nvim_tree_disable_netrw = 0 "1 by default, disables netrw
 let g:nvim_tree_hijack_netrw = 0 "1 by default, prevents netrw from automatically opening when opening directories (but lets you keep its other utilities)
 let g:nvim_tree_add_trailing = 1 "0 by default, append a trailing slash to folder names
+let g:nvim_tree_group_empty = 1 " 0 by default, compact folders that only contain a single folder into one node in the file tree
 let g:nvim_tree_show_icons = {
     \ 'git': 1,
     \ 'folders': 0,
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -203,6 +203,11 @@ functionnalities.
 Can be 0 or 1. When 1, appends a trailing slash to folder names.
 0 by default.
 
+|g:nvim_tree_group_empty|                         *g:nvim_tree_group_empty*
+
+Can be 0 or 1. When 1, folders that contain only one folder are grouped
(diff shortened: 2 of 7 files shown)


kyazdani42:
thanks a lot ! The overall algorithm is perfect and the feature works perfectly.
I must note that toggling this behavior would be useful if you want to add files in between grouped folders :)
Merging this anyway :)

cloggier:
Using lua to configure this, is setting this option to `true` suppose to work?

```lua
vim.g.nvim_tree_group_empty = true
```

does not work, while

```lua
vim.g.nvim_tree_group_empty = 1
```

does.

Some other options I have set to true do work as expected.

kyazdani42:
that's strange, if you find issues refering to the documentation please leave an issue.
I also have to work on that setup flow, but i'd like to finish the buffer thing first. It will make the plugin configuration more lua like.

Lux3mb0urg:
Is there a callback function to toggle this feature? 
Changing value of ` vim.g.nvim_tree_group_empty` does not work.

kyazdani42:
it does work for me though, just assign it to 1 (`vim.g.nvim_tree_group_empty` must be set in lua not in vimscript)

sindrets:
@cloggier I implemented the option the same way all the other boolean options are currently implemented. The reason why only some options work when you set them to the lua value `true` is that comparing vim booleans and lua booleans is strange. 

`1 == true` is a false lua statement. However, `if 1 then ... end` is a true lua statement. Which is why it will work sometimes to set vim variables to the lua value `true`. And to make things even more confusing `true == vim.v["true"]` is a true lua statement.

But when using vim variables in lua, you should treat them like vim variables. That means you should use 1 for true and 0 for false until proper lua configuration is implemented.

@Lux3mb0urg no, I did not implement a toggle function.

cloggier:
@sindrets thank you for detailed response and amazing PR!

kyazdani42:
indeed until i don't make the setup refactoring, this option thing will keep being clunky.

chardskarth:
Hi everyone, I was trying to do this (group empty directories) but setting `vim.g.nvim_tree_group_empty = 1` doesn't seem to work. 

This works however by setting this config
https://github.com/nvim-tree/nvim-tree.lua?tab=readme-ov-file#setup

PR #249, opened 2021-03-25
feat: preserve the file tree buffer

sindrets:
This makes it so that the file tree buffer is only created if a previously created buffer cannot be found.

- What does this solve?
  - When the tree window is closed, the tree buffer is wiped. When the tree window is opened again, the tree buffer is recreated and the state of the previous tree is lost. All previously opened folders are now closed.

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -269,7 +269,6 @@ end
 
 local function create_buf()
   local options = {
-    -- bufhidden = 'wipe';
     buftype = 'nofile';
     modifiable = false;
   }
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -239,6 +239,10 @@ function M.open_file(mode, filename)
 end
 
 function M.change_dir(foldername)
+  if vim.fn.expand(foldername) == M.Tree.cwd then
+    return
+  end
+
   api.nvim_command('cd '..foldername)
   M.Tree.entries = {}
   M.init(false, M.Tree.bufnr ~= nil)
@@ -265,7 +269,7 @@ end
 
 local function create_buf()
   local options = {
-    bufhidden = 'wipe';
+    -- bufhidden = 'wipe';
     buftype = 'nofile';
     modifiable = false;
   }
@@ -291,7 +295,6 @@ function M.close()
     return vim.cmd ':q!'
   end
   api.nvim_win_close(M.Tree.winnr(), true)
-  M.Tree.bufnr = nil
 end
 
 function M.set_target_win()
@@ -300,7 +303,11 @@ end
 
 function M.open()
   M.set_target_win()
-  create_buf()
+
+  if not M.buf_exists() then
+    create_buf()
+  end
+
   create_win()
   api.nvim_win_set_buf(M.Tree.winnr(), M.Tree.bufnr)
 
@@ -367,6 +374,13 @@ function M.win_focus(winnr)
   end
 
   api.nvim_set_current_win(wnr)
+end
+
+function M.buf_exists()
+  return ( M.Tree.bufnr
+    and vim.fn.bufexists(M.Tree.bufnr) == 1
+    and vim.fn.bufname(M.Tree.bufnr) == M.Tree.buf_name
+    )
 end
 
 function M.toggle_ignored()
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -275,6 +275,7 @@ local function create_buf()
 
   M.Tree.bufnr = api.nvim_create_buf(false, true)
   api.nvim_buf_set_name(M.Tree.bufnr, M.Tree.buf_name)
+  api.nvim_buf_set_var(M.Tree.bufnr, "nvim_tree_buffer_ready", 1)
 
   for opt, val in pairs(options) do
     api.nvim_buf_set_option(M.Tree.bufnr, opt, val)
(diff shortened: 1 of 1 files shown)


kyazdani42:
hum thanks but i'm currently working on the buffer/window refactoring which will make this PR obsolete (i believe).
Also with this PR, closing the tree with `bd!` then reopening makes the tree completely unusable.
Not sure if the behavior could be fixed, but i need to finish the refactoring to make the tree usable on multiple tabs due to one neovim commit creating double frees in these situations...

sindrets:
@kyazdani42 Alright. I made this pull a bit quickly, but I realized later that the reason why the state of the tree is lost is because of the cwd change, which happens every time `open()` is called, which causes the tree to be reinitialized. Would you be willing to merge this if I dropped all the changes aside from those I made to `change_dir()`? That way the state of the tree is preserved. Cursor position is lost however, because the buffer is wiped, which is still quite annoying in large directories.

> closing the tree with `bd!` then reopening makes the tree completely unusable.

It becomes usable again after properly wiping the buffer with `:bw!`. But this is also solvable by improving my `buf_exists()` function a little bit. For instance we can assign a buffer-local variable to the tree-buffer and check if that exists. Buffer-local variables are removed upon `:bd`.

kyazdani42:
I'm not sure about something though:
when i call the open/close commands, the tree does not loose state. Could you elaborate where your issue comes up so i can understand the purpose of the PR ?

sindrets:
@kyazdani42  That is strange, because the tree entries are wiped every time `change_dir()` is called, and that is called from `open()`.

In `open()`:
https://github.com/kyazdani42/nvim-tree.lua/blob/709d6b968bd905db03521ea3f44bfcb56e4a2c03/lua/nvim-tree/lib.lua#L320-L322

In `change_dir()`:
https://github.com/kyazdani42/nvim-tree.lua/blob/709d6b968bd905db03521ea3f44bfcb56e4a2c03/lua/nvim-tree/lib.lua#L250-L254

Screencast from current master:
Peek 2021-03-27 13-37

Traceback after calling `:NvimTreeToggle`:
```log
stack traceback:
    ...ndrets/Documents/git/nvim-tree.lua/lua/nvim-tree/lib.lua:48: in function 'init'
    ...ndrets/Documents/git/nvim-tree.lua/lua/nvim-tree/lib.lua:254: in function 'change_dir'
    ...ndrets/Documents/git/nvim-tree.lua/lua/nvim-tree/lib.lua:322: in function <...ndrets/Documents/git/nvim-tree.lua/lua/nvim-tree/lib.lua:311>
```

```
▶ nvim --version                                        
NVIM v0.5.0-dev+1158-gf31b8dabf
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/cc -D_FORTIFY_SOURCE=2 -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/sindrets/.cache/yay/neovim-git/src/build/config -I/home/sindrets/.cache/yay/neovim-git/src/neovim-git/src -I/usr/include -I/home/sindrets/.cache/yay/neovim-git/src/build/src/nvim/auto -I/home/sindrets/.cache/yay/neovim-git/src/build/include
Compiled by sindrets@ArchDT

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/share/nvim"

```

kyazdani42:
hum strange my version never calls change_dir. Could you provide me with a minimal config please ? This looks like a bug to me but might be related to something else

sindrets:
@kyazdani42 It happens on my end even with nothing but this config:

```vim
call plug#begin("~/.vim/plug")
Plug 'kyazdani42/nvim-web-devicons'
call plug#end()
```

But am I missing something here? Following the logic of the `open()` function this behavior is to be expected. `change_dir` is called whenever the tree is loaded. Which is always, after the tree has been opened once, or something calls the refresh function:

https://github.com/kyazdani42/nvim-tree.lua/blob/e5bd242003c285ccfd0153efb13d8cdd7173d0cf/lua/nvim-tree/lib.lua#L144-L148

And sure enough: if I call `:NvimTreeRefresh` while the tree is closed, `Tree.loaded` is set to false and the state of the tree is preserved when I open it again because `change_dir` is never called.

And further, in `change_dir`, why do you want to rebuild the tree, even when `foldername` is the same as `M.Tree.cwd`? Seems redundant.

sindrets:
The latest commit now also detects when the tree buffer has been deleted with `:bd!`. After using this for a short while, I think it is largely desirable to preserve the buffer while it still exists, instead of abandoning it for no reason on every tree close. So regardless, I will be using this in my own fork. I personally feel that file trees are mostly useful in larger projects, and in such projects it's very annoying to lose the cursor position in the tree every time I close it. 
Feel free to close this pull if it's not something you want. I'll look forward to see what you do with the buffer/window refactoring.

kyazdani42 (approved):
i think you are right, i'm gonna merge this in the meantime as this work just better than before :)
thanks for your work !

PR #250, opened 2021-03-25
feat: Add command to focus tree window regardless of window state

sindrets:
This adds command `:NvimTreeFocus`, that opens, and focuses the tree if it's closed. Otherwise, bring focus to the already open window. When using multiple windows it's a lot easier use a mapping to a command like this, as I don't need to consider whether or not the tree is already open. This command always brings you to the tree whether it is closed or not.

- How does this differ from `:NvimTreeOpen`?
  - If you call `:NvimTreeOpen` from a different window while the tree window is already open, nothing happens.
- How does this differ from `lua require'nvim-tree.lib'.win_focus()`?
  - `win_focus()` doesn't open the tree if it's closed.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -29,6 +29,11 @@ opens the tree
 
 closes the tree
 
+|:NvimTreeFocus| 	                          *:NvimTreeFocus*
+
+Opens, and focuses the tree if it's closed. Otherwise, bring focus to the
+already open window.
+
 |:NvimTreeToggle| 	                          *:NvimTreeToggle*
 
 open or close the tree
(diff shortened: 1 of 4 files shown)


kyazdani42:
maybe win_focus should open the tree if it's not open ? just extending the function would be better imo, or maybe passing an option to it like `{ open_if_closed = boolean }`

sindrets:
@kyazdani42 Sure. Do you want me to remove the `:NvimTreeFocus` command as well, and just let users use the lua api? Or just make the command call `win_focus()`?

kyazdani42:
Maybe just leave the user to bind the command, i think i'll remove all the vim commands and make a little description in the readme / docs to explain how it works :)

sindrets:
@kyazdani42 Requesting re-review.

kyazdani42 (approved):

Issue #180, opened 2021-01-17
Add option to use .gitignore files to ignore paths


peach-lasagna:

kyazdani42:
would be great indeed, behind an option because i believe it might impact performance a little

afonsocarlos:
I can take this

peach-lasagna:
first
![изображение](https://user-images.githubusercontent.com/59787538/112474955-d3a5aa00-8da2-11eb-92a6-2e7d33832489.png)

afonsocarlos:
Haha, didn't know you were developing this already. That's fine, maybe I can help with something else then

peach-lasagna:
Done. thanks, kyazdani!

PR #251, opened 2021-03-25
add Gitignore

peach-lasagna:
fixes (trying) #180

### DIFF
--- a/lua/nvim-tree/populate.lua
+++ b/lua/nvim-tree/populate.lua
@@ -80,6 +80,21 @@ end
 
 local function gen_ignore_check()
   local ignore_list = {}
+
+  local function add_toignore(path)
+    local content = utils.read_file(path)
+    for s in content:gmatch("[^\r\n]+") do
+      ignore_list[s] = true
+    end
+  end
+
+  if vim.g.nvim_tree_gitgnore > 0 then
+    add_toignore('.gitignore')
+  end
+  if vim.g.nvim_tree_gitgnore == 2 then
+    add_toignore(git.get_path_gitexclude())
+  end
+  
   if vim.g.nvim_tree_ignore and #vim.g.nvim_tree_ignore > 0 then
     for _, entry in pairs(vim.g.nvim_tree_ignore) do
       ignore_list[entry] = true
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -11,6 +11,14 @@ function M.echo_warning(msg)
   api.nvim_command('echohl None')
 end
 
+function M.read_file(path)
+  local file = io.open(path, "r")
+  if not file then return '' end
+  local content = file:read "*a"
+  file:close()
+  return content
+end
+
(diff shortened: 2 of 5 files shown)


lua/nvim-tree/populate.lua
```diff
@@ -80,6 +80,21 @@ end
...
   local ignore_list = {}
+
+  local function add_toignore(path)
+    local content = utils.read_file(path)
+    for s in content:gmatch("[^\r\n]+") do
+      ignore_list[s] = true
+    end
+  end
+
+  if vim.g.nvim_tree_gitgnore > 0 then
```
kyazdani42, line 91: typo `gitIgnore`
also if you don't set the value, we get an error because comparing nil with number :)
`if (vim.g.nvim_tree_gitignore or 0) > 0 then`

lua/nvim-tree/utils.lua
```diff
@@ -11,6 +11,14 @@ function M.echo_warning(msg)
   api.nvim_command('echohl None')
 end
 
+function M.read_file(path)
```
kyazdani42, line 14: i'd prefer using `vim.loop` here:
[docs](https://github.com/luvit/luv/blob/master/docs.md)
code from documentation:
```
local function readFileSync(path)
  local fd = assert(uv.fs_open(path, "r", 438))
  local stat = assert(uv.fs_fstat(fd))
  local data = assert(uv.fs_read(fd, stat.size, 0))
  assert(uv.fs_close(fd))
  return data
end
```
without the asserts ;)
you could access `uv` as `vim.loop` (i like to call it `luv`)

lua/nvim-tree/populate.lua
```diff
@@ -80,6 +80,21 @@ end
...
+    add_toignore('.gitignore')
+  end
+  if vim.g.nvim_tree_gitgnore == 2 then
```
kyazdani42, line 94: typo

kyazdani42 (changes requested):
thanks for the PR but some adjustements are in order :)
I'm gonna test this more extensively once you've made the appropriate changes !

peach-lasagna:
sorry, i'm beginner in lua

lua/nvim-tree/utils.lua
```diff
@@ -11,6 +12,14 @@ function M.echo_warning(msg)
...
+  local fd = uv.fs_open(path, "r", 438)
+  local stat = uv.fs_fstat(fd)
+  local data = uv.fs_read(fd, stat.size, 0)
```
kyazdani42, line 18: these functions fails if a file is not found. Doing checks on each line would be necessary imo:
```lua
if not fd then return '' end

if not stat then return '' end

return data or ''
```

kyazdani42 (changes requested):
Don't worry about being a beginner in lua (we all went this path), the language is easy, neovim's api needs some time to be learned but there are great resources on github if you want :)

peach-lasagna:
Done

kyazdani42 (approved):
Thanks for the PR :)

Issue #194, opened 2021-02-05
Advanced buffer navigation


Chaitanyabsprip:
I am coming from coc-explorer and the only thing I miss is the `hjkl` navigation in the file tree. for example `l` to go into a directory, `h` to come out of it or "fold" it. I could map `l` to `<cr>` but then to get out of the node/folder binding `h` will need some kind of condition which make me go up and not into a directory. I am not familiiar with your code or lua, is this something you can implement?

cloggier:
I use following mappings, which bring some of the functionality you mentioned:

```lua
  vim.g.nvim_tree_bindings = {
    edit = {"<CR>", "o", "l"},
    close_node = {"<S-CR>", "<BS>", "h"}
  }
```

kyazdani42:
Hi, i'm quite busy atm and will not be able to work on this plugin. We'll keep this open as i think other users talked about this in the past and i think it would be a great feature to implement.

ahmedkhalf:
> I use following mappings, which bring some of the functionality you mentioned:
> 
> ```lua
>   vim.g.nvim_tree_bindings = {
>     edit = {"<CR>", "o", "l"},
>     close_node = {"<S-CR>", "<BS>", "h"}
>   }
> ```

@cloggier Im getting this error when I use this:
```
Error executing vim.schedule lua callback: ....config/nvim/plugged/nvim-tree.lua/lua/nvim-tree/lib.lua:247: Expected lua string
```

ahmedkhalf:
Okay I managed to fix this error by doing this:
```nvim
lua << EOF

local function get_lua_cb(cb_name)
  return string.format(":lua require'nvim-tree'.on_keypress('%s')<CR>", cb_name)
end

vim.g.nvim_tree_bindings = {
  ["<cr>"] = get_lua_cb("edit"),
  ["<s-cr>"] = get_lua_cb("close_node"),
  ["o"] = get_lua_cb("edit"),
  ["<BS>"] = get_lua_cb("close_node"),
  ["l"] = get_lua_cb("edit"),
  ["h"] = get_lua_cb("close_node"),
}

EOF
```
It's not the most elegant, but it works.

kyazdani42:
could we consider this issue fixed then ?

Chaitanyabsprip:
I tried what @ahmedkhalf suggested, but `h` is still not working as required. pressing h takes the cursor to the top of the file-browser buffer, does not "close-node".

[...]

PR #257, opened 2021-03-29
Feat/add advanced navigation

afonsocarlos:
Implements #194 

New features added:
- [x] Move to parent directory without closing
- [x] Navigate through sibling directory/files
- [x] Go to the first/last sibling of the current node


lua/nvim-tree/lib.lua
```diff
@@ -78,6 +78,31 @@ local function get_node_at_line(line)
   return iter
 end
 
+local function get_line_from_node(node, find_parent)
+  local sep = package.config:sub(1,1)
+  local node_path = node.absolute_path
+
+  if find_parent then
+    node_path = node.absolute_path:match("(.*)"..sep)
```
kyazdani42, line 86: there are utils to handle the path separator in `utils.lua`

kyazdani42 (changes requested):
Thanks for the PR, but please use existing utils code before i properly review it :)

lua/nvim-tree/config.lua
```diff
@@ -67,9 +67,14 @@ function M.get_bindings()
     ["<C-v>"]          = M.nvim_tree_callback("vsplit"),
     ["<C-x>"]          = M.nvim_tree_callback("split"),
     ["<C-t>"]          = M.nvim_tree_callback("tabnew"),
+    ["<C-k>"]          = M.nvim_tree_callback("prev_sibling"),
+    ["<C-j>"]          = M.nvim_tree_callback("next_sibling"),
```
kyazdani42, line 71: hum i use those two for moving between up and down windows and i think many people would too. maybe something like `<` or `>` ?

kyazdani42 (changes requested):
this little change and i'll merge this :)

lua/nvim-tree/config.lua
afonsocarlos: Great!

kyazdani42:
thanks ;) but i got an error when doing that on the first line (the `..` thing)

Issue #229, opened 2021-03-12
Suggestion: folders open and close different colors


hrqmonteiro:
Hi, it would be nice to have different colors for the opened folders and the default folder glyphs, like we could do in Nerdtree, like this:

![2021-03-12_18-17](https://user-images.githubusercontent.com/17055027/110999481-3046a500-835f-11eb-9980-1d9db1bed26b.png)

A different highlight group for the opened ones and for the closed ones.

kyazdani42:
you could implement this if you wish, you've got to modify the color table (to add another group), and the highlight choice in renderer.lua :)

hrqmonteiro:
I'm not a lua programmer, i tried to do it but i couldn't... Even forked the repo, so if i could manage to do it i would PR. But i am not able to.

Thanks to the attention anyway.

kyazdani42:
Hi, do not close suggestions that might be realized by other people :) maybe i'll have time at some point

hrqmonteiro:
Oh sorry, didn't realized that, my mistake

vuki656:
Hey @hrqmonteiro you could do something like this 

Lua variant
```
vim.g.nvim_tree_icons = {
    folder = {
        default = "",
        open = "",
        empty_open = "",
        empty = "",
        symlink = "",
    }
}
```


Viml variant
```
let g:nvim_tree_icons = {
    \ 'folder': {
    \   'default': "",
    \   'open': "",
    \   'empty_open': "",
    \   'empty': "",
    \   'symlink': "",
    \   }
    \ }
```

It will do this
![image](https://user-images.githubusercontent.com/8104698/112717276-e4ae0280-8eeb-11eb-97c3-ccda24ce4135.png)

Is this what you had in mind?

hrqmonteiro:
> Hey @hrqmonteiro you could do something like this
> 
> Lua variant
> 
> ```
> vim.g.nvim_tree_icons = {
>     folder = {
>         default = "",
>         open = "",
>         empty_open = "",
>         empty = "",
>         symlink = "",
>     }
> }
> ```
> 
> Viml variant
> 
> ```
> let g:nvim_tree_icons = {
>     \ 'folder': {
>     \   'default': "",
>     \   'open': "",
>     \   'empty_open': "",
>     \   'empty': "",
>     \   'symlink': "",
>     \   }
>     \ }
> ```
> 
> It will do this
[...]

PR #258, opened 2021-03-29
Feat/icon highlighting for opened files and directories

afonsocarlos:
Implements #229 

Add color highlight for both files and directories opened.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -95,6 +95,11 @@ to render the icons. The `icons` key can only work if `nvim-web-devicons`
 is installed and in your |runtimepath|
 (https://github.com/kyazdani42/nvim-web-devicons)
 
+|g:nvim_tree_highlight_icons|			        *g:nvim_tree_highlight_icons*
+
+Can be `0` or `1`. When `1`, will highlight icons for opened files and directories
+Default is 0
+
 |g:nvim_tree_icons|                                *g:nvim_tree_icons*
 
 You can set icons for:
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -361,7 +361,9 @@ NvimTreeFolderName
 NvimTreeRootFolder
 NvimTreeFolderIcon
 NvimTreeEmptyFolderName
+NvimTreeOpenedFolderName
 NvimTreeExecFile
+NvimTreeOpenedFile
 NvimTreeSpecialFile
 NvimTreeImageFile
 NvimTreeMarkdownFile
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -95,7 +95,7 @@ to render the icons. The `icons` key can only work if `nvim-web-devicons`
 is installed and in your |runtimepath|
 (https://github.com/kyazdani42/nvim-web-devicons)
 
-|g:nvim_tree_highlight_icons|			        *g:nvim_tree_highlight_icons*
+|g:nvim_tree_highlight_opened_files|			        *g:nvim_tree_highlight_opened_files*
 
 Can be `0` or `1`. When `1`, will highlight icons for opened files and directories
 Default is 0
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -106,8 +106,6 @@ function M.on_keypress(mode)
   else
     lib.open_file(mode, node.absolute_path)
   end
-
-  renderer.draw(lib.Tree, true)
 end
 
 function M.refresh()
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -106,6 +106,8 @@ function M.on_keypress(mode)
   else
     lib.open_file(mode, node.absolute_path)
   end
+
+  renderer.draw(lib.Tree, true)
 end
 
 function M.refresh()
--- a/lua/nvim-tree/colors.lua
+++ b/lua/nvim-tree/colors.lua
@@ -39,7 +39,7 @@ local function get_hl_groups()
     ExecFile = { gui = 'bold', fg = colors.green },
     SpecialFile = { gui = 'bold,underline', fg = colors.yellow },
     ImageFile = { gui = 'bold', fg = colors.purple },
-    OpenedFile = { gui = 'bold', fg = colors.red },
+    OpenedFile = { gui = 'bold', fg = colors.green },
 
     GitDirty = { fg = colors.dark_red },
     GitDeleted = { fg = colors.dark_red },
@@ -54,7 +54,7 @@ local function get_links()
   return {
     FolderName = 'Directory',
     EmptyFolderName = 'Directory',
-    OpenedFolderName = 'ErrorMsg',
+    OpenedFolderName = 'NvimTreeOpenedFile',
     Normal = 'Normal',
     EndOfBuffer = 'EndOfBuffer',
     CursorLine = 'CursorLine',
--- a/lua/nvim-tree/colors.lua
+++ b/lua/nvim-tree/colors.lua
@@ -39,6 +39,7 @@ local function get_hl_groups()
     ExecFile = { gui = 'bold', fg = colors.green },
     SpecialFile = { gui = 'bold,underline', fg = colors.yellow },
     ImageFile = { gui = 'bold', fg = colors.purple },
+    OpenedFile = { gui = 'bold', fg = colors.red },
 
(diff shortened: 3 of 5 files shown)


afonsocarlos:
One thing to notice though is that when editing a file its highlight doesn't update right away, I must either refresh the tree manually or trigger it by opening/closing a folder in the tree. And I couldn't find a way to get it working neatly. Here's an example of how it is now:

https://user-images.githubusercontent.com/13454744/112913917-251aa580-90d1-11eb-899a-b1e5a624cba0.mp4

kyazdani42:
i wonder how it would behave with the git highlighting activated ? I though the issue was about highlighting the icon only and not the filenames ?

afonsocarlos:
> i wonder how it would behave with the git highlighting activated ? I though the issue was about highlighting the icon only and not the filenames ?

By highlighting the icon only you mean this?

https://user-images.githubusercontent.com/13454744/113213321-67b5bc80-924e-11eb-817e-961f044fa77c.mp4

kyazdani42:
I think it would be better yes. Also the issue you wanted to tackle was about setting a color for open folder and close folder specifically.

afonsocarlos:
Yes I know. Actually I was hoping to get both file and directory highlighted, but since file highlight gives this out of sync feeling I think I'll remove it for now, maybe open a new issue and implement it later. Thanks for the feedback

kyazdani42:
too many highlights at the same time might also get confusing. Maybe adding markers in the sign column ?

kyazdani42:
or bolden / italicize the filenames

afonsocarlos:
> or bolden / italicize the filenames

This sounds great actually! Only one thing though, do you know how I could make files highlight update immediately after opening?

kyazdani42:
not sure what you mean, there are no files opening after opening right ? And if there were, you've added your logic in the renderer so it should happen immediately.

afonsocarlos:
The problem is the color highlight for files specifically doesn't trigger any rendering in NvimTree. The highlight only works after taking some action that triggers NvimTree render (e.g Opening/Closing a directory) or triggering NvimTree render update manually

hrqmonteiro:
@afonsocarlos  why did you implemented the highlight only for the name of the current opened folder, and not for the folder icon itself? Wouldn't that be possible?

kyazdani42:
i kind of lost the purpose of this PR. Could you:
1) hide the new feature behind an option and document it
2) edit the title to better describe what this PR really is about :)

afonsocarlos:
@kyazdani42 I could manage to highlight only icons and have file opening rendering properly! Guess this PR is tidier now.

lua/nvim-tree.lua
```diff
@@ -112,6 +112,8 @@ function M.on_keypress(mode)
   else
     lib.open_file(mode, node.absolute_path)
   end
+
+  renderer.draw(lib.Tree, true)
```
kyazdani42, line 116: is this required ? if it wasn't there i guess it was for a reason :)

lua/nvim-tree/colors.lua
```diff
@@ -38,6 +38,7 @@ local function get_hl_groups()
     ExecFile = { gui = 'bold', fg = colors.green },
     SpecialFile = { gui = 'bold,underline', fg = colors.yellow },
     ImageFile = { gui = 'bold', fg = colors.purple },
+    OpenedFile = { gui = 'bold', fg = colors.red },
```
kyazdani42, line 41: i'd chose something lighter, like green or yellow :)

lua/nvim-tree/colors.lua
```diff
@@ -52,6 +53,7 @@ local function get_links()
   return {
     FolderName = 'Directory',
     EmptyFolderName = 'Directory',
+    OpenedFolderName = 'ErrorMsg',
```
kyazdani42, line 56: yeah don't link error to opened folder too, this is disturbing. by default you should link to Directory just like the other two

doc/nvim-tree-lua.txt
```diff
@@ -95,6 +95,11 @@ to render the icons. The `icons` key can only work if `nvim-web-devicons`
 is installed and in your |runtimepath|
 (https://github.com/kyazdani42/nvim-web-devicons)
 
+|g:nvim_tree_highlight_icons|			        *g:nvim_tree_highlight_icons*
```
kyazdani42, line 98: i'd call the option `nvim_tree_highlight_opened_files`

lua/nvim-tree/renderer.lua
```diff
@@ -296,6 +299,19 @@ local function update_draw_data(tree, depth, markers)
...
       end
 
+      if vim.g.nvim_tree_highlight_icons then
+        if vim.fn.bufloaded(node.absolute_path) > 0 then
+          table.insert(hl, {'NvimTreeOpenedFile', index, offset, offset+#icon })
+        else
+          for i, value in pairs(hl) do
+            if value[1] == 'NvimTreeOpenedFile' and value[2] == index then
+              table.remove(hl, i)
+              break
```
kyazdani42, line 309: is this loop required ? i don't think this highlight would stay as it's regenerated each time :)

kyazdani42 (changes requested):

lua/nvim-tree.lua
afonsocarlos: This is what makes file highlighting work when opening a file from nvim-tree. Otherwise, the highlighting doesn't take place right after editing a file (it only happens with files specifically though).

lua/nvim-tree/renderer.lua
afonsocarlos: You're right! Thanks! :+1: :rocket:

lua/nvim-tree.lua
kyazdani42: i see, maybe placing this in the open_file call would be more appropriate then.

afonsocarlos: It makes sense. I moved it, just not sure I placed it into the right place in `open_file`

lua/nvim-tree/renderer.lua
```diff
@@ -260,6 +260,9 @@ local function update_draw_data(tree, depth, markers)
       end
       if not has_children then folder_hl = "NvimTreeEmptyFolderName" end
       set_folder_hl(index, offset, #icon, #name+#git_icon, folder_hl)
+      if vim.g.nvim_tree_highlight_opened_files and node.open then
+        set_folder_hl(index, offset, 0, #icon, "NvimTreeOpenedFolderName")
+      end
```
kyazdani42, line 265: now that i think of this, maybe instead of using the variable to highlight the folder icon when opened, maybe just doing
```lua
if not has_children..... end
if node.open then folder_hl = "NvimTreeOpenedFolderName" end
```
would suffice, and highlighting NvimTreeOpenenedFolderName as directory by default

kyazdani42 (changes requested):

lua/nvim-tree/renderer.lua
kyazdani42: what i mean is that you would only need to override the highlight yourself to get highlighting on folder, not based on the variable value.

afonsocarlos: Ok, I think I understood, but that would imply highlighting everything (not only the icon) including folder name. Is that a problem? If not I can simplify it as you suggested

kyazdani42: hum i guess it would make more sense indeed even though i didn't think about this. We should keep consistency so i'd say yes.

afonsocarlos: Agreed. Also, the Highlight color now makes it more discrete even highlighting the whole foldername. Check it out

lua/nvim-tree/colors.lua
```diff
@@ -53,6 +54,7 @@ local function get_links()
   return {
     FolderName = 'Directory',
     EmptyFolderName = 'Directory',
+    OpenedFolderName = 'NvimTreeOpenedFile',
```
kyazdani42, line 57: this should be Directory too, because people might not want opened folders to be highlighted differently by default, users that wishes to benefit that feature will set the hl themselves :)

kyazdani42 (changes requested):
i should've mentionned that, sorry

lua/nvim-tree/colors.lua
afonsocarlos: Hmmm, I see. Fixing that

kyazdani42 (approved):
it's good now ! thanks for working on this :)

PR #260, opened 2021-03-30
feat: nvim lsp integration

### LINKED ISSUE #227: LSP Diagnostic integration
Not sure if this has been posted but it would be cool to show errors in the tree.

kyazdani42:
first attempt at integrating lsp diagnostics in the tree !
should fix #227


williamboman:
Nice! Getting some this error: 

```
Error executing vim.schedule lua callback: ...packer/start/nvim-tree.lua/lua/nvim-tree/diagnostics.lua:24: attempt to perform arithmetic on local 'idx' (a nil value)
```

Otherwise seems to work perfect (I'm still getting the highlight)! Would be cool if it was possible to register keymaps to jump to next/prev item

kyazdani42:
just pushed a 'maybe' fix, i did not take the time to reproduce the error :) is it ok now ?

kyazdani42:
i think it's ready and i've updated the documentation. Maybe we'll make more integrations in the future (like severity level could highlight with other colors that we would link to lsp base hl groups)

PR #262, opened 2021-03-31
Fix setup in readme

luathn:
It just a tiny change.
I think newbie will put all setup code to their configuration file and do not realize why it was faulty.


kyazdani42 (approved):
ha thanks :)

PR #265, opened 2021-04-02
add event API

williamboman:
Opening a WIP draft PR just to get a discussion going.

I think it'd be really neat being able to hook into the different actions nvim-tree does. For example - when renaming a node I'd like to inform my LSP client that a rename just happened:

```lua
local events = require'nvim-tree.events'
local typescript_ls = require'lsp.clients.typescript'

events.register_callback('renamed', function (old_fname, new_fname)
    typescript_ls.rename_file(old_fname, new_fname)
end)
```

I went with quite a vanilla event system where any amount of callbacks can be registered for any given event name.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -436,6 +436,7 @@ For example, registering a handler for when a node is renamed is done like this:
 ==============================================================================
 Lua module: nvim-tree.events                           *nvim-tree.events*
 
+                                     *nvim-tree.events.on_node_renamed()*
 on_node_renamed({handler})
                 Registers a handler for when a node is renamed.
                 • Note: A node can either be a file or a directory.
@@ -444,12 +445,15 @@ on_node_renamed({handler})
                     {handler}    (function) Handler function, with the
                                  signature `function(payload)`.
 
+                                     *nvim-tree.events.on_file_created()*
 on_file_created({handler})
                 Registers a handler for when a file is created.
 
                 Parameters: ~
                     {handler}    (function) Handler function, with the
                                  signature `function(payload)`.
+
+                                     *nvim-tree.events.on_file_removed()*
 on_file_removed({handler})
                 Registers a handler for when a file is removed.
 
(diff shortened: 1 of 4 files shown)


kyazdani42:
Hi i think that would be a great feature to add indeed, thanks for implementing the first draw, you could go for it i'd accept a fully made PR :)

williamboman:
Question: When deleting a directory, would you as a consumer expect to receive a) a single "folder deleted" event, or b) an event for each file and dir deleted within the deleted folder?

kyazdani42:
I guess i'd prefer having the main folder and also all deleted files. As a consumer, i wouldn't expect reproducing what the plugin already did :)

williamboman:
I went with a simpler approach where events are structured around the user action itself. This might be a bit too opaque for certain use cases, but I figured it's a good start.

With the current implementation, you can't really figure out which files were deleted or "renamed" if you are deleting or renaming a folder, you just get the FolderRenamed/FolderRemoved events. At least not out of the box.

doc/nvim-tree-lua.txt
```diff
@@ -1,6 +1,6 @@
 *nvim-tree.lua* A file explorer written in lua
 
-Minimum version of neovim: 0.4.0
+Minimum version of neovim: 0.5.0
```
kyazdani42, line 3: good catch ^^'

lua/nvim-tree/events.lua
```diff
@@ -0,0 +1,94 @@
+local M = {}
+
+local global_handlers = {}
+
+local Event = {
+    NodeRenamed = 'NodeRenamed',
+    FileCreated = 'FileCreated',
+    FileRemoved = 'FileRemoved',
+    FolderCreated = 'FolderCreated',
+    FolderRemoved = 'FolderRemoved',
```
kyazdani42, line 10: could you indent this file with the `.editorconfig` settings please :) ? i use 2 spaces for indenting lua :)

kyazdani42 (changes requested):
thanks for the PR, your lua code is quite pleasant to read. Just a small change and i'll merge this

lua/nvim-tree/events.lua
williamboman: :+1: There doesn't seem to be an `.editorconfig` so I added one, let me know if you want me to remove it

kyazdani42: woups i thought i had one ^^' thanks for adding it

kyazdani42 (approved):
LGTM, thanks :)

PR #269, opened 2021-04-03
Use platform path separator when setting tree index.

p3rl:
Find file on windows doesn't work when using hard coded path separator '/'. 
Updated `set_tree_index_and_redraw` to use `utils.path_separator`.


kyazdani42 (approved):
good catch, thanks :)

PR #277, opened 2021-04-08
chore: remove async to avoid unecessary complexity

### LINKED ISSUE #270: nvim-tree is too fast, following wincmd does not work right
While this is a 'serious' issue, the underlying cause may actually be what's in the title, or more likely async issues.

I'm on neovim latest and nvim-tree is up-to-date, and I observe the following behavior, but first, here's what i'm trying to achieve:

- I want to toggle the tree to have a bird's eye view on my project, but not focus the nvim tree buffer. On NERDTree for example, this is achieved simply by mapping to "NERDTreeToggle | wincmd p" which returns to the previous windows after opening the tree one.

Then the issue: I couldn't reproduce this behavior with :NvimTreeToggle, or even lua require'nvim-tree'.toggle()  (has not effect or errors out, but that could be my bad syntax), but then I came across this (simplified version).

```vim
function ! Nvimtreetogglecustom()
  lua require'nvim-tree'.toggle()
  " sleep 1ms
  wincmd p
endfunction
```

If the `sleep 1ms' is commented, calling the function toggles nvim-tree but the focus stay on the tree.

If the sleep 1ms is uncommented, calling the function toggle the tree and switches to the previous windows (so at least on 'opening') this achieve the desired behavior

What do you think of this?

kyazdani42:
I've looked around and seen that asynchronous stuff was probably overkill for this plugin. I'm not sure if this might also improve performance so if people can test it i'd be much happy.
related: #172 
fixes #270

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -14,9 +14,9 @@ function M.toggle()
     lib.close()
   else
     if vim.g.nvim_tree_follow == 1 then
-      vim.schedule(function() M.find_file(true) end)
+      M.find_file(true)
     else
-      vim.schedule(lib.open)
+      lib.open()
     end
   end
 end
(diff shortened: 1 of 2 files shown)


cloggier:
I was trying  it out - didn't notice any differences.

Issue #248, opened 2021-03-25
Question? How to open a full screen tree when doing "nvim ."


alvarogalloc:
I used to use fern but while i am changing most of my vimscript config to lua, when i tried to change to nvim_tree-lua i encountered that when i put the nvim_tree_auto_open = 1 it opens the tree just in the width i specified, the thing is that i want that full width when it is the only buffer opened.

kyazdani42:
hi ! this has been asked in #214 and #200 but not possible yet, i have to finish the buffer / window rework that i started last week before this is made possible.

Alvaroalloc:
Should I close this since this problem is adressed to previous??

kyazdani42:
don't worry, i'll close them alltogether once this is implemented

kyazdani42:
closed in bac5633e1f7e7 `1.1.0`

Issue #238, opened 2021-03-16
Only auto-open on directory


AckslD:
I would like the tree to auto open when I do `vim $DIR` but not when I do `vim`, is that currently possible?

kyazdani42:
Hi, it is not yet possible, although if you use something like 'startify', you could set `let g:nvim_tree_auto_ignore_ft = { 'startify' }`

sindrets:
@AckslD This can be set up with an autocommand:

```vim
augroup NvimTreeConfig
  au!
  au BufEnter * if isdirectory(expand('%')) | exec("cd " . expand('%')) | exec('NvimTreeOpen') | endif
augroup END

lua << EOF
  require'nvim-tree.events'.on_nvim_tree_ready(function ()
    vim.cmd("NvimTreeRefresh")
  end)
EOF
```

kyazdani42:
i like your hacks @sindrets :) but i think it should be a feature anyway, the auto_open could be easily extended i just didn't spend the time to do that.

sindrets:
Sounds good! But in the meantime, this auto command works fine:)

AckslD:
Thanks for sharing @sindrets :)

Issue #214, opened 2021-02-23
Fullscreen


lucax88x:
Is there a way to have tree opened as full screen, without a "right" window? 

I tried nvim_tree_width=100 but it's always 50% of the screen.

kyazdani42:
i think this is a kind of duplicate of #200. I don't know why it would take only 50%, but i believe the tree width is not expressed as % but as columns. this is why it doesn't open fullscreen, a PR to fix that would be gladly welcome.

lucax88x:
Dunno, I saw this one, but despite the title and confusing "description" I cannot relate to my problem.

What I want is that we don't open the "tree" as a side window, but as a full window, like opening a new file.

Exactly like the old netrw or dirvish are doing.

kyazdani42:
for me its kind of the same thing. I remember this request was made quite some time ago, but im not that interested in doing that but ill accept a PR that whishes to emulate this behavior. I also need to make a full rewrite of buffer/window handling I think itll make things easier.

kyazdani42:
closed in 3bac5633e1f7e7 `1.1.0`

lucax88x:
@kyazdani42 so full screen is now working?

kyazdani42:
when you open it by default yes it should be fullscreen

enheit:
@kyazdani42 Is it possible to open nvim-tree in full screen when using `:NvimTreeOpen`?

I would expect to have something like this in the nvim-tree configuration.

```lua
config = {
  view = {
    fullscreen = true
  }
}
```

And it would look like this.

![image](https://github.com/nvim-tree/nvim-tree.lua/assets/8645216/b4b66bfa-2437-4e3d-8561-743cede55bff)

But currently I can set only the width to "100%"

```lua
config = {
  view = {
    width = "100%"
  }
}
```

So, it 'kinda' solves the problem, but it's not opening the nvim-tree in a separate window; rather, it tries to expand it to full size (note the vertical line which divides between the current nvim-tree and the opened file).

![image](https://github.com/nvim-tree/nvim-tree.lua/assets/8645216/63eeb4a2-6026-4972-888f-4f843885e6ad)

So, if I drag the terminal window, it behaves like this.

![image](https://github.com/nvim-tree/nvim-tree.lua/assets/8645216/074747be-9971-4dfc-93ec-263dff848804)

[...]

Issue #200, opened 2021-02-12
Fullscreen tree


pianocomposer321:
I was sure there would already be an issue open about this, but I didn't see one, so...

Is there a way to edit directories with this plugin, similar to netrw or nerdtree when hijack netrw is set? If not, is this feature planned?

Thanks.

kyazdani42:
hi, what do you mean by 'edit directories' ? by typing `r` you can already rename it if that's what you mean.

pianocomposer321:
Nope, sorry I should have been more specific.

When this plugin is not installed and netrw is enabled (which it is by default in all recent versions of [neo]vim), you can do `nvim .`, for example, and it will open the directory full-screen in netrw. If vim is already open, you can do `:e .` to open netrw in the current split window. With nerdtree installed, you can do the same thing but it will open in nerdtree instead of neovim. But with nvim-tree installed, not only does it not open the directory in the current split in nvim-tree, but it also breaks netrw.

pianocomposer321:
http://vimcasts.org/blog/2013/01/oil-and-vinegar-split-windows-and-project-drawer/ for more info

kyazdani42:
I'm not sure but a PR for fixing these netrw issues has been made (see #195).

kyazdani42:
think this is ok now

qgadrian:
@kyazdani42 I believe this issue is more of opening the tree in the current window (and explained [here](https://github.com/kyazdani42/nvim-tree.lua/issues/200#issuecomment-778416552)) and not a new side window with the file explorer.

I see there are _hijack_current_window_ functions and tested this behavior myself with little success. I'm not familiar with the code but I believe the overall architecture does not work well with this behavior, am I right? Is this feasible and/or something you have planned?

kyazdani42:
opening directories through `:e dir` or `:vsplit dir` should change the tree base dir to this dir

PR #288, opened 2021-04-13
feat: open tree on directory

kyazdani42:
This PR enables opening the tree fullscreen when editing a directory. It also enables opening the tree fullscreen on enter with `vim DIR` or `vim`

Related: #238 #436 #200 #214 #248 #131 #188 

@sindrets i'll leave this open for a week or so, i think it should work, but i need people to test it before i merge it.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -194,6 +194,9 @@ function M.on_enter()
     and ((is_dir and netrw_disabled) or bufname == '')
     and not vim.tbl_contains(ft_ignore, buftype)
   lib.init(should_open, should_open)
+  if bufname == '' or is_dir then
+    vim.cmd ":only"
+  end
 end
 
 local function is_file_readable(fname)
@@ -256,6 +259,23 @@ local function update_root_dir()
   if lib.Tree.cwd == new_cwd then return end
 
   lib.change_dir(new_cwd)
+end
+
+function M.open_on_directory()
+  local buf = api.nvim_get_current_buf()
+  local bufname = api.nvim_buf_get_name(buf)
+  if vim.fn.isdirectory(bufname) ~= 1 or bufname == lib.Tree.cwd then
+    return
+  end
+
+  if not view.win_open() then
+    view.open()
+  end
+  lib.change_dir(bufname)
+  lib.set_index_and_redraw(bufname)
+  view.focus()
+  vim.cmd ":only"
+  vim.cmd(":bd "..bufname)
 end
 
 function M.buf_enter()
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -175,6 +175,21 @@ function M.print_clipboard()
   fs.print_clipboard()
 end
 
+function M.hijack_current_window()
+  local View = require'nvim-tree.view'.View
+  if not View.bufnr then
+    View.bufnr = api.nvim_get_current_buf()
+  end
+  local current_tab = api.nvim_get_current_tabpage()
+  if not View.tabpages then
+    View.tabpages = {
+      [current_tab] = { winnr = api.nvim_get_current_win() }
+    }
+  else
+    View.tabpages[current_tab] = { winnr = api.nvim_get_current_win() }
+  end
+end
+
 function M.on_enter()
   local bufnr = api.nvim_get_current_buf()
   local bufname = api.nvim_buf_get_name(bufnr)
@@ -193,10 +208,10 @@ function M.on_enter()
   local should_open = vim.g.nvim_tree_auto_open == 1
     and ((is_dir and netrw_disabled) or bufname == '')
     and not vim.tbl_contains(ft_ignore, buftype)
-  lib.init(should_open, should_open)
   if bufname == '' or is_dir then
-    vim.cmd ":only"
+    M.hijack_current_window()
   end
+  lib.init(should_open, should_open)
 end
 
 local function is_file_readable(fname)
@@ -268,14 +283,14 @@ function M.open_on_directory()
     return
   end
 
-  if not view.win_open() then
-    view.open()
+  if view.win_open() and #api.nvim_list_wins() > 1 then
+    view.close()
   end
+  M.hijack_current_window()
   lib.change_dir(bufname)
   lib.set_index_and_redraw(bufname)
   view.focus()
-  vim.cmd ":only"
-  vim.cmd(":bd "..bufname)
+  view.replace_window()
 end
 
 function M.buf_enter()
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -134,7 +134,8 @@ function M.setup()
     vim.bo[M.View.bufnr][opt.name] = opt.val
   end
 
-  vim.cmd "au! BufWinEnter * lua require'nvim-tree.view'._prevent_buffer_override()"
+  vim.cmd "au! BufWinEnter,BufWinLeave * lua require'nvim-tree.view'._prevent_buffer_override()"
+  vim.cmd "au! BufEnter,BufNewFile * lua require'nvim-tree'.open_on_directory()"
   if vim.g.nvim_tree_disable_keybindings == 1 then
     return
   end
@@ -180,9 +181,18 @@ function M._prevent_buffer_override()
   vim.schedule(function()
     local curwin = a.nvim_get_current_win()
     local curbuf = a.nvim_win_get_buf(curwin)
-    if curwin ~= M.get_winnr() or curbuf == M.View.bufnr then return end
+
+    if curwin ~= M.get_winnr() or curbuf == M.View.bufnr then
+      return
+    end
 
     vim.cmd("buffer "..M.View.bufnr)
+
+    local bufname = a.nvim_buf_get_name(curbuf)
+    local isdir = vim.fn.isdirectory(bufname) == 1
+    if isdir or not bufname or bufname == "" then
+      return
+    end
 
     if #vim.api.nvim_list_wins() < 2 then
       vim.cmd("vsplit")
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -184,7 +184,9 @@ function M._prevent_buffer_override()
       return
     end
 
-    vim.cmd("buffer "..M.View.bufnr)
+    if a.nvim_buf_is_loaded(M.View.bufnr) and a.nvim_buf_is_valid(M.View.bufnr) then
+      vim.cmd("buffer "..M.View.bufnr)
+    end
 
     local bufname = a.nvim_buf_get_name(curbuf)
     local isdir = vim.fn.isdirectory(bufname) == 1
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -88,12 +88,6 @@ local function find_rogue_buffer()
   return nil
 end
 
----Check if the tree buffer is valid and loaded.
----@return boolean
-local function is_buf_valid()
-  return a.nvim_buf_is_valid(M.View.bufnr) and a.nvim_buf_is_loaded(M.View.bufnr)
-end
-
 ---Find pre-existing NvimTree buffer, delete its windows then wipe it.
 ---@private
 function M._wipe_rogue_buffer()
@@ -118,12 +112,16 @@ local function warn_wrong_mapping()
   require'nvim-tree.utils'.echo_warning(warn_str)
 end
 
+local HAS_LOADED = false
 -- set user options and create tree buffer (should never be wiped)
 function M.setup()
   M.View.side = vim.g.nvim_tree_side or M.View.side
   M.View.width = vim.g.nvim_tree_width or M.View.width
 
-  M.View.bufnr = a.nvim_create_buf(false, false)
+  if not HAS_LOADED then
+    M.View.bufnr = a.nvim_create_buf(false, false)
+    HAS_LOADED = true
+  end
 
   if not pcall(a.nvim_buf_set_name, M.View.bufnr, 'NvimTree') then
     M._wipe_rogue_buffer()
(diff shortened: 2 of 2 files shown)


akinsho:
@kyazdani42 I just tried this out and don't see the tree being opened when I run `nvim <dir-name>` I do get a message that says `Already only one window`? What settings should a user have to trigger this behaviour. I've set disable netrw and hijack netrw to true.

sindrets:
@kyazdani42 This looks great! Unfortunately I'm unavailable for a while and I don't think I'll have an opportunity to test it out.

akinsho:
Forgot to mention another issue I've noticed is that when opening, a file from the fullscreen window it seems to open into an empty file e.g. the file contents aren't loaded if you leave and come back to the buffer that then prompts me to load the file.

lua/nvim-tree.lua
```diff
@@ -205,6 +208,23 @@ local function update_root_dir()
...
+    return
+  end
+
+  if not view.win_open() then
+    view.open()
+  end
+  lib.change_dir(bufname)
+  lib.set_index_and_redraw(bufname)
+  view.focus()
+  vim.cmd ":only"
```
wincent, line 224: This seems like it will break [the "vinegar" use case](https://github.com/tpope/vim-vinegar), which is not to open in "full screen" takeover mode, but rather to open in the current window.

I think the "vinegar" mode is what most of the related issues are actually asking for; only #248 and #214 mention "full-width" display.

kyazdani42: i think i see what you mean. I did not test that with multiple windows and buffers.
The 'vinegar' use case seems like the right way to do it, i'll implement that soon :)

vladdoster:
Can confirm it works.

```
$ grep -A 6 -B 1 'nvim-tree.lua' lua/pluginList.lua
        use {
            "kyazdani42/nvim-tree.lua",
            branch = "feat/open-on-dir",   -- Specifies a git branch to use
	   after="nvim-web-devicons",
            config = function()
                require "plugins.nvimtree"
            end
        }
```

vladdoster:
well.... it did. I cleared everything to make sure and have hit this edge case.


https://user-images.githubusercontent.com/10052309/127737515-72286bfa-d6f5-4d19-a5ac-52b461bba92e.mov

kyazdani42:
you do not seem to hijack netrw, this can cause potential issue.
As for the 2nd one, haven't seen that either :/

mrossinek:
I'm interested in switching from vim-dirvish to your plugin. However, I got really used to the "drivish"-way of using `-` in normal mode to jump into a fullscreen-view of the directory tree. I checked out this branch but have not gotten that to work. I attempted the following map but that does not really work:
```vim
nnoremap - <cmd>lua require'nvim-tree'.open_on_directory()<CR>
```
Do you have any plans to support such a workflow?

ExpandingMan:
This seems to have stalled out, any chance of it getting merged?

kyazdani42:
i'll merge this after refactoring the setup i guess, because i'm changing quite a lot of things and this PR might be improved a little bit, specially for implementing the window takeover instead of doing a simple `:only`

kyazdani42:
update note:
i'll merge this in the next days after people made some testing. I have actually implemented the vinegar use case instead of doing `:only`, was easier than expected. I want to be sure i did not break anything, so please let me know in the comment if there is something wrong !

akinsho:
@kyazdani42 I've just tried testing this out in my current vim session and I get the following error when I try and open a directory. 

```
Error executing vim.schedule lua callback: ...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:187: Vim(buffer):E86: Buffer 2 does not exist
```

When I try to toggle nvim tree is shows
`[NvimTree] this is the last open window, are you sure you want to quit nvim ? y/n:`

This seems to relate to using a vim session as when I open without a session it seems to work, although it does still show some strangeness i.e. after I open a buffer from the directory window whenever I visit that buffer again it automatically opens nvim-tree again after I closed it. When I delete my session and start a new one from scratch, it seems to work fine, but think anyone who tries to use this with an existing vim session will hit this error.

My nvim tree config uses the following (if that helps debug)
```lua
  vim.g.nvim_tree_special_files = {}
  vim.g.nvim_tree_lsp_diagnostics = 1
  vim.g.nvim_tree_indent_markers = 1
  vim.g.nvim_tree_group_empty = 1
  vim.g.nvim_tree_git_hl = 1
  vim.g.nvim_tree_auto_close = 0 -- closes the tree when it's the last window
  vim.g.nvim_tree_follow = 1 -- show selected file on open
  vim.g.nvim_tree_auto_resize = 1
  vim.g.nvim_tree_width = '20%'
  vim.g.nvim_tree_width_allow_resize = 1
  vim.g.nvim_tree_disable_netrw = 0
  vim.g.nvim_tree_hijack_netrw = 1
  vim.g.nvim_tree_root_folder_modifier = ':t'
  vim.g.nvim_tree_ignore = { '.DS_Store', 'fugitive:', '.git' }
  vim.g.nvim_tree_highlight_opened_files = 1
```

kyazdani42:
this is not related to that I think, nvim tree is broken when using sessions I'm not sure why, I'll take a look

kyazdani42:
@akinsho i've partially fixed the session issue, but the `buffer with this name already exist` is almost impossible to fix until i've made the setup refactoring. you could follow the discussion here #591

zetashift:
I'm on the latest master but I'm still getting ```Error executing vim.schedule lua callback: ...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:187: Vim(buffer):E86: Buffer 2 does not exist``` error, I'm using `auto-session` though

jghauser:
It works for me but seems to set `nowrap` so that all further buffers opened have it set too (even if i have it set to `wrap` in my init.vim). It would be great if `nowrap` were set for only the tree buffer.

ExpandingMan:
I just wanted to thank the author for this PR, this makes nvimtree vastly more useful for me.  I'll probably use it significantly more often now.

akinsho:
> I'm on the latest master but I'm still getting `Error executing vim.schedule lua callback: ...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:187: Vim(buffer):E86: Buffer 2 does not exist` error, I'm using `auto-session` though

I had the same issue @zetashift if you remove the session and start a new one then the issue goes away.

kyazdani42:
should be allright now, i've fixed the buffer 2 does not exist issue and the plugin should work as expected with nvim sessions

FabioAntunes:
Need clarification on this one. If I do `nvim .` I would expect  it to open nvim-tree lua instead I get an empty buffer:
Do I need any specific configs? These are my configs:

```
vim.g.nvim_tree_gitignore = 0
vim.g.nvim_tree_hide_dotfiles = 0
```

kyazdani42:
you need to set `vim.g.nvim_tree_auto_open=1`

jghauser:
What I wrote above about this plugin setting nowrap still applies. 

> It works for me but seems to set nowrap so that all further buffers opened have it set too (even if i have it set to wrap in my init.vim). It would be great if nowrap were set for only the tree buffer.

I really like the plugin and would love to be able to have it open automatically on directories. If I could help troubleshoot this somehow, I'd be more than happy to! Thanks again for the excellent work!

vladdoster:
I had this working at one point, but seem to have lost the feature. What am I overlooking?

Big ups on working through all the issues! `nvimtree` is by far the best file explorer plugin.

```
g.nvim_open_on_directory = 1
g.nvim_tree_add_trailing = 1
g.nvim_tree_allow_resize = 1
g.nvim_tree_auto_ignore_ft = { "dashboard" }
g.nvim_tree_auto_open = 1
g.nvim_tree_disable_netrw = 0
g.nvim_tree_follow = 1
g.nvim_tree_git_hl = 1
g.nvim_tree_gitignore = 1
g.nvim_tree_hide_dotfiles = 0
g.nvim_tree_highlight_opened_files = 0
g.nvim_tree_hijack_netrw = 1
g.nvim_tree_ignore = { ".git", "node_modules", ".cache" }
g.nvim_tree_indent_markers = 0
g.nvim_tree_lsp_diagnostics = 0
g.nvim_tree_respect_buf_cwd = 1
g.nvim_tree_root_folder_modifier = table.concat { ":t:gs?$?/..", string.rep(" ", 1000), "?:gs?^??" }
g.nvim_tree_side = "left"
g.nvim_tree_tab_open = 0
g.nvim_tree_update_cwd = 1
g.nvim_tree_width = 25
```

```
      use {
         "kyazdani42/nvim-tree.lua",
         after = "nvim-web-devicons",
         branch = "feat/open-on-dir",
         cmd = { "NvimTreeToggle", "NvimTreeFocus" },
         config = function()
            require "plugins.configs.nvimtree"
         end,
         setup = function()
            require("core.mappings").nvimtree()
         end,
      }
```

kyazdani42:
@jghauser i do not set wrap or nowrap in the plugin at all, i think the issue comes from your config.
If you wish to have a custom option for nvim-tree you could just `au FileType NvimTree setlocal nowrap`

kyazdani42:
@vladdoster what do you mean by 'lost' ? works fine for me

jghauser:
@kyazdani42, thanks for the info! Are you sure, though? If I comment out [this line](https://github.com/kyazdani42/nvim-tree.lua/blob/34e70ec69eba60ae210464ef0b058ac59cd70cd3/lua/nvim-tree/renderer.lua#L459) then the `wrap` setting from my init.lua doesn't get overridden.

kyazdani42:
well yes actually the lua api for setting options is kind of broken i haven't seen this wrap setting being put there. I'll fix this.

akinsho:
@kyazdani42 I'm in the same boat as @vladdoster I seem to get very sporadic behaviour when I use this feature, my primary use case is to open a directory after startup using `:edit <directory>`. Currently the issues I have are.

1. When I try to edit a directory if I haven't previously opened the tree, then it doesn't work. If this happens it also causes the original window to split vertically.
2. If you get round the first step and the nvim tree full width buffer does open, then you select a file it opens the buffer then opens the tree beside it.
3. This last one is similar to how netrw behaves but is confusing tbh which is if you run `edit <directory>/` it will open a buffer called `<directory>` as well as the nvim tree.

kyazdani42:
@akinsho pushed an option and fixes for this feature, are your use case fixed ?

akinsho:
@kyazdani42 seems not to open the tree buffer at all when I enter `:edit directory_path/` am I missing a setting I have to configure? I've enabled `hijack_netrw` and `disable_netrw` in setup

kyazdani42:
okay fixed the impl in `1.12.13` (1fe211eda019746c), now the whole implementation should be perfect

vladdoster:
Okay, testing now! 🤞🏻

vladdoster:
Eh, doesn't seem to work? Most likely am overlooking a setting, though.

```lua
local present, tree_c = pcall(require, "nvim-tree.config")
if not present then
    return
end

local g = vim.g

require("nvim-tree").setup {
    disable_netrw = true,
    hijack_netrw = true,
    open_on_setup = true,
    ignore_ft_on_setup = {},
    auto_close = true,
    open_on_tab = false,
    update_to_buf_dir = true,
    hijack_cursor = false,
    update_cwd = true,
    lsp_diagnostics = false,
    update_focused_file = {
        enable = false,
        update_cwd = true,
        ignore_list = {
            ".git",
            "node_modules",
            ".cache",
            "build",
            "__pycache__",
        },
    },
    system_open = {
        cmd = nil,
        args = {},
    },
    view = {
        width = 30,
        side = "left",
        auto_resize = true,
        mappings = {
            custom_only = false,
            list = {},
        },
    },
}

g.nvim_tree_ignore = { ".git", "node_modules", ".cache" }
g.nvim_tree_gitignore = 1
g.nvim_tree_quit_on_open = 1
g.nvim_tree_indent_markers = 1
g.nvim_tree_hide_dotfiles = 1
g.nvim_tree_git_hl = 1
g.nvim_tree_highlight_opened_files = 1
g.nvim_tree_root_folder_modifier = ":~"
g.nvim_tree_add_trailing = 1
g.nvim_tree_group_empty = 1
g.nvim_tree_disable_window_picker = 1
g.nvim_tree_icon_padding = " "
g.nvim_tree_symlink_arrow = " >> "
g.nvim_tree_respect_buf_cwd = 0
g.nvim_tree_create_in_closed_folder = 0
g.nvim_tree_refresh_wait = 500
```

vladdoster:
Yay!! it worked. I ended up using your nvim-tree config and it works flawless!!

Thank you so, so much!

kyazdani42:
@vladdoster yes it is, specified in the doc and the readme. I'm going to add a note for that in the migration issue

akinsho:
Hmm what settings do I need to enable this, I still can't get it to work 😞 i.e. `:edit directory/`

it just opens a blank buffer.
My config is in the details:

<details>

```lua
    vim.g.nvim_tree_icons = {
    default = '',
    git = {
      unstaged = '',
      staged = '',
      unmerged = '',
      renamed = '',
      untracked = '',
      deleted = '',
    },
  }
  vim.g.nvim_tree_special_files = {}
  vim.g.nvim_tree_indent_markers = 1
  vim.g.nvim_tree_group_empty = 1
  vim.g.nvim_tree_git_hl = 1
  vim.g.nvim_tree_width_allow_resize = 1
  vim.g.nvim_tree_root_folder_modifier = ':t'
  vim.g.nvim_tree_ignore = { '.DS_Store', 'fugitive:', '.git' }
  vim.g.nvim_tree_highlight_opened_files = 1

  require('nvim-tree').setup {
    view = {
      width = '20%',
      auto_resize = true,
      list = {
        { key = 'cd', cb = action 'cd' },
      },
    },
    lsp_diagnostics = true,
    disable_netrw = true,
    hijack_netrw = true,
    open_on_setup = false,
    hijack_cursor = true,
    update_cwd = true,
    update_focused_file = {
      enable = true,
      update_cwd = true,
    },
  }
```

</details>

kyazdani42:
it does not open automatically if `open_on_setup` is false. i'll improve the configuration a bit

akinsho:
@kyazdani42 thanks, that fixed it, and it now opens as expected. It's a bit of a minor point, but the second point I raised still happens. 
> If you get round the first step and the nvim tree full width buffer does open, then you select a file it opens the buffer then opens the tree beside it.

Say you open a directory using `:e dir/` and then open a file with `CR` or `o` the buffer opens but it opens a tree beside it. The use case I'm imagining is a bit like dirvish where you can navigate to a directory open a buffer and just edit with no side panel opening so you can do it over and over again without the tree split opening.

kyazdani42:
something like `vim.g.nvim_tree_close_on_open` you mean ? I don't like when the tree is closing without me explicitly doing it so i prefer not setting this by default as it would break for everybody not using the close_on_open feature.

akinsho:
@kyazdani42 I just tried `nvim_tree_close_on_open` and it doesn't stop this happening. It also seems like it would make every tree close when I open the file, whereas what I mean is when I do `e <dir>/` the whole window becomes a nvim tree buffer and when I open I file I expect to replace it.

I don't expect it to then open itself afterwards. I appreciate this all probably seems like the same thing i.e. the tree buffer in a split or full window has one set of behaviour but this is largely based on how netrw and dirvish work ([here's a video showing this workflow](https://youtu.be/ILpdtL-4gv0?t=189)). You can edit a directory move to a file open it do some more editing, edit another dir find the file you want edit that and move around a few dirs making edits like that without another split appearing.

kyazdani42:
i guess this is a feature request because the intended behavior was not to replicate dirvish or netrw behavior :) i like the tree being open most of the time. Actually it wouldn't be very hard to do, if you wished you could submit a PR.

akinsho:
@kyazdani42 being able to open a directory was the main aim, didn't mean to go down a tangent, it's not a huge deal, definitely just a nice to have. I honestly thought it was a bug since it creates a new window without my doing anything specific to trigger that 🤷🏿‍♂️

kyazdani42:
i can make a PR this behavior should not be hard to implement anyway

PR #295, opened 2021-04-14
fix (lsp diagnostics highlight): find_node index, bufnr

Nodman:
When at least one root file with lsp diagnostics problems is present and at least one  folder with more then 0 child nodes is expanded - NvimTree would  throw an error:

```
Error executing vim.schedule lua callback: ...lua/nvim-tree/diagnostics.lua:23: Column value outside range
```

This happens because:
* `bufnr` was taken from the `lib.Tree` instead of `view.View`.
* wrong `linenr` being passed  to `highlight_node` from `find_node` where every step of recursion would produce + 1 index offset for each child node without highlight.

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -14,7 +14,7 @@ local function get_severity(diagnostics)
 end
 
 local function highlight_node(node, linenr)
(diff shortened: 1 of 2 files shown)


kyazdani42 (approved):
good catch thanks :)

PR #300, opened 2021-04-15
fix(view): Don't create a new empty buffer on every open.

### LINKED ISSUE #301: Everytime i toggle NvimTree, it creates a tab called [buffer]
Since the last update, it is behaving like this. And it wasn't like this before.

https://user-images.githubusercontent.com/17055027/114947194-0e8b8280-9e23-11eb-8a92-f884ebd813ef.mp4

Is there any way to fix it?

sindrets:
Fixes #301.

Calling `vnew` creates a new empty buffer every time the tree opens. The window then changes to the correct buffer only a couple lines later, leaving the unused, unneeded, empty buffer behind.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -130,7 +130,7 @@ local move_tbl = {
 }
 
 function M.open()
-  a.nvim_command("vnew")
+  a.nvim_command("vsp")
   local move_to = move_tbl[M.View.side]
   a.nvim_command("wincmd "..move_to)
   a.nvim_command("vertical resize "..M.View.width)


Shatur95:
After this PR I have an error:

```
E5108: Error executing lua ...im/pack/plugins/opt/nvim-tree.lua/lua/nvim-tree/view.lua:142: Vim(buffer):E86: Buffer 2 does not exist
```

Steps to reproduce:

```vimscript
:NvimTreeToggle
:setlocal bufhidden=wipe
:bd
:NvimTreeToggle
```

sindrets:
@Shatur95 Not related to this PR. That happened before this change. But I have created a draft for a new PR that would solve your described problem #307.

But why do you want to set `bufhidden=wipe` on the tree buffer?

kyazdani42:
do not set wipe, I do not recreate the buffer two time

Shatur95:
> Not related to this PR. That happened before this change.

Sorry, I was wrong with the commit number :(
And thanks for the fix!

> But why do you want to set bufhidden=wipe on the tree buffer?
> do not set wipe, I do not recreate the buffer two time

I just have a custom function that closes the buffers. Should I add nvim-tree to exceptions or should I not use `bufhidden = wipe` for other buffers too?

kyazdani42:
depends the use case, but it would be better to make nvim-tree an exception because you don't want to reload the whole buffer each time when it doesn't change and doesn't consume too much ram :)

Shatur95:
> depends the use case, but it would be better to make nvim-tree an exception because you don't want to reload the whole buffer each time when it doesn't change and doesn't consume too much ram :)

Thanks for the tip, will do!

LoydAndrew:
I have the same error if I close tree with :bd or q. But I don't have buffhiden set to anything, :set buffhiden? shows buffhiden=
What do I missing? Sorry to I ask here, but this thread is kind of related.

kyazdani42:
after some time, i agree that trying to keep the tree bufnr always the same was quite a bad idea. I didn't think about users enough :smile: Maybe try to not `:bd` the tree then, NvimTreeClose handles that by itself already

PR #303, opened 2021-04-16
feat(diagnostics): More responsive and configurable diagnostics signs.

sindrets:
This makes the diagnostic signs in the tree more responsive by using the `LspDiagnosticsChanged` auto command event rather than `BufWritePost`, as well as updating diagnostics where it made sense (`unroll_dir`, `refresh`). It also adds configurable icons for the four types of diagnostics.

```lua
lsp = {
  hint = "",
  info = "",
  warning = "",
  error = "",
},
```

I also fixed an error where diagnostics were sorted by highest severity. As far as I can tell, the severity property works kind of backwards in that severity 1 is `error`, and severity 4 is `hint`. This seemed consistent across all the language servers I tested.

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -33,7 +33,7 @@ local signs = {}
 
 local function add_sign(linenr, severity)
   local buf = view.View.bufnr
-  if not vim.fn.bufexists(buf) or not vim.fn.bufloaded(buf) then return end
+  if not a.nvim_buf_is_valid(buf) or not a.nvim_buf_is_loaded(buf) then return end
   local sign_name = sign_names[severity][1]
   table.insert(signs, vim.fn.sign_place(1, 'NvimTreeDiagnosticSigns', sign_name, buf, { lnum = linenr+1 }))
 end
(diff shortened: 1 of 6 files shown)


lua/nvim-tree/diagnostics.lua
```diff
@@ -58,7 +63,7 @@ function M.update()
     signs = {}
   end
   for bufname, severity in pairs(buffer_severity) do
-    if severity > 0 then
+    if 1 <= severity and severity <= 4 then
```
kyazdani42, line 66: i'd do if 0 < severity and severity < 5 then.
the 2nd check shouldn't happen though right ?

kyazdani42 (changes requested):
good catch :) thanks a lot !
could you just update the documentations where i inform the lsp diagnostic is updated on BufWritePost please ?

lua/nvim-tree/diagnostics.lua
sindrets: Yes, the second expression is needed because `severity` might be `math.huge`. Possibly it could be a value higher than 4 as well, I don't know if severity 1-4 is just convention or if it is enforced by lsp. So we need to make sure `severity` is in the correct range otherwise `add_sign` will fail.

kyazdani42 (approved):
thanks !

PR #307, opened 2021-04-16
fix(view): Recreate tree buffer if deleted, and handle scenario where buffer already exists.

### LINKED ISSUE #296: Session Restore Problems
Hello,

Initially, thank you for your continuous work on this beautiful approach to file explorer.

I have recently started to encounter problems with session management after the commit sha `bbb8d60`. When I load it with `nvim .` the nvim-tree works fine but when I save that session and reload it, it throws out a buffer id unknown error (not always) and just toggles a split of the current buffer. I have searched through the issues but this buffer id unknown seems to be a rather old issue [here](https://github.com/kyazdani42/nvim-tree.lua/issues/175) which might be unrelated.

sindrets:
Fixes #296.

This makes the tree able to recover when the tree buffer gets deleted or wiped. It also tackles scenarios where the NvimTree buffer already exists before initialization, as a result of i.e. a session being loaded. This is what I believe is the root of the problems described in 296.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -66,6 +66,17 @@ M.View = {
   }
 }
 
+---Find a rogue NvimTree buffer that might have been spawned by i.e. a session.
+---@return integer|nil
+local function find_rogue_buffer()
+  for i = 1, vim.fn.bufnr("$"), 1 do
+    if vim.fn.bufname(i) == "NvimTree" then
+      return i
+    end
+  end
+  return nil
+end
+
 -- set user options and create tree buffer (should never be wiped)
 function M.setup()
   M.View.auto_resize = vim.g.nvim_tree_auto_resize or M.View.auto_resize
@@ -73,10 +84,22 @@ function M.setup()
   M.View.width = vim.g.nvim_tree_width or M.View.width
 
   M.View.bufnr = a.nvim_create_buf(false, false)
+
+  if not pcall(a.nvim_buf_set_name, M.View.bufnr, 'NvimTree') then
+    -- Find the pre-existing NvimTree buffer, delete its windows then wipe it.
+    local bn = find_rogue_buffer()
+    if bn then
+      local win_ids = a.nvim_eval("win_findbuf(" .. bn .. ")")
+      for _, id in ipairs(win_ids) do
+        a.nvim_win_close(id, true)
+      end
+      a.nvim_command("bw " .. bn)
+    end
+  end
+
   for k, v in pairs(M.View.bufopts) do
     a.nvim_buf_set_option(M.View.bufnr, k, v)
   end
-  a.nvim_buf_set_name(M.View.bufnr, 'NvimTree')
 
   if not vim.g.nvim_tree_disable_keybindings then
     M.View.bindings = vim.tbl_extend(
@@ -130,6 +153,10 @@ local move_tbl = {
 }
 
 function M.open()
+  if not a.nvim_buf_is_valid(M.View.bufnr) then
+    M.setup()
+  end
+
   a.nvim_command("vsp")
   local move_to = move_tbl[M.View.side]
   a.nvim_command("wincmd "..move_to)
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -77,21 +77,22 @@ local function find_rogue_buffer()
   return nil
 end
 
----Check if the tree buffer is valid. The buffer is only valid if it exists and
----hasn't been deleted or wiped.
+---Check if the tree buffer is valid and loaded.
 ---@return boolean
 local function is_buf_valid()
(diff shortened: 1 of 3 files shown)


sindrets:
This seems to fix the problems described in 296, but I need to test this more thoroughly.

lua/nvim-tree/view.lua
```diff
@@ -66,17 +66,60 @@ M.View = {
...
+      and a.nvim_buf_get_var(M.View.bufnr, "nvim_tree_buffer_ready") == 1
+      )
+  end)
+
+  if not success then
+    return false
+  else
+    return valid
+  end
+end
```
kyazdani42, line 96: i think you could just return `a.nvim_buf_is_valid(M.View.bufnr) and a.nvim_buf_is_loaded(M.View.bufnr)`

lua/nvim-tree/view.lua
```diff
@@ -66,17 +66,60 @@ M.View = {
...
+
 -- set user options and create tree buffer (should never be wiped)
 function M.setup()
   M.View.auto_resize = vim.g.nvim_tree_auto_resize or M.View.auto_resize
   M.View.side = vim.g.nvim_tree_side or M.View.side
   M.View.width = vim.g.nvim_tree_width or M.View.width
 
   M.View.bufnr = a.nvim_create_buf(false, false)
+
+  if not pcall(a.nvim_buf_set_name, M.View.bufnr, 'NvimTree') then
```
kyazdani42, line 106: can you not set the same name for 2 buffers ?

lua/nvim-tree/view.lua
```diff
@@ -66,17 +66,60 @@ M.View = {
...
+    -- Find the pre-existing NvimTree buffer, delete its windows then wipe it.
+    local bn = find_rogue_buffer()
+    if bn then
+      local win_ids = a.nvim_eval("win_findbuf(" .. bn .. ")")
+      for _, id in ipairs(win_ids) do
+        a.nvim_win_close(id, true)
+      end
+      a.nvim_command("bw " .. bn)
```
kyazdani42, line 114: use `a.nvim_buf_delete`

kyazdani42:
i was going to do it but thanks for doing it first :)

lua/nvim-tree/view.lua
sindrets: Ah, right. That's better. I'm too used to the vim script way.

lua/nvim-tree/view.lua
sindrets: See, I didn't think that was a problem either, but try it. It throws `Vim:E95: Buffer with this name already exists`.

lua/nvim-tree/view.lua
sindrets: Nice. Again, I'm too stuck in the vim script headspace, where "delete" and "wipe" do not mean the same thing.

lua/nvim-tree/view.lua
kyazdani42: ha didn't know, good to know :)

lua/nvim-tree/view.lua
kyazdani42: no worries :) i've always hated vimscript and learned neovim development alltogether with lua and its api so it's easier for me to think that way

sindrets:
This seems to work quite well on my end. So if we can get some of the people over in 296 to test this, I think we're good to go.

lua/nvim-tree/view.lua
```diff
@@ -66,17 +66,52 @@ M.View = {
...
+local function is_buf_valid()
+  return a.nvim_buf_is_valid(M.View.bufnr) and a.nvim_buf_is_loaded(M.View.bufnr)
+end
+
+---Find pre-existing NvimTree buffer, delete its windows then wipe it.
+---@private
+function M._wipe_rogue_buffer()
+  local bn = find_rogue_buffer()
+  if bn then
+    local win_ids = a.nvim_eval("win_findbuf(" .. bn .. ")")
```
kyazdani42, line 91: `vim.fn.win_findbuf`

kyazdani42 (changes requested):
almost :D

lua/nvim-tree/view.lua
sindrets: Wow, I was just looking for this, and it didn't show up on tab completion for me so I didn't realize it was implemented.

Shatur95:
> This seems to work quite well on my end. So if we can get some of the people over in 296 to test this, I think we're good to go.

Tested, works as expected!

lua/nvim-tree/view.lua
kyazdani42: most of the time, you can just try it `:lua vim.fn.win_findbuf()` and it will crash saying either the function doesn't exist or the function misses parameters :)

kyazdani42:
thanks again !

PR #314, opened 2021-04-18
fix(view): Fix error thrown when trying to wipe buffer in autocmd window.

### LINKED ISSUE #309: Conflitcs with :mksession
**Steps to reproduce**

```vim
:mksession test.vim
:source test.vim
```

**Current result**

Session loaded with the following error:

```
|| Error detected while processing SessionLoadPost Autocommands for "*":
|| E5108: Error executing lua ...im/pack/plugins/opt/nvim-tree.lua/lua/nvim-tree/view.lua:92: Vim:E813: Cannot close autocmd window
```

**Expected result**

Session loaded without any errors.

sindrets:
Fixes #309.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -95,7 +95,7 @@ function M._wipe_rogue_buffer()
     end
     -- Do not use `nvim_buf_delete` here as it throws an uncacthable error if
     -- the buffer is in an autocmd window.
-    vim.api.nvim_command("silent! bw " .. bn)
+    a.nvim_command("silent! bw " .. bn)
   end
 end
 
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -68,9 +68,9 @@ M.View = {
 ---Find a rogue NvimTree buffer that might have been spawned by i.e. a session.
 ---@return integer|nil
 local function find_rogue_buffer()
-  for i = 1, vim.fn.bufnr("$"), 1 do
-    if vim.fn.bufname(i) == "NvimTree" then
-      return i
+  for _, v in ipairs(a.nvim_list_bufs()) do
+    if vim.fn.bufname(v) == "NvimTree" then
+      return v
     end
   end
   return nil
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/view.lua
```diff
@@ -89,9 +89,13 @@ function M._wipe_rogue_buffer()
...
     for _, id in ipairs(win_ids) do
-      a.nvim_win_close(id, true)
+      if vim.fn.win_gettype(id) ~= "autocmd" then
+        a.nvim_win_close(id, true)
+      end
     end
-    a.nvim_buf_delete(bn, {})
+    -- Do not use `nvim_buf_delete` here as it throws an uncacthable error if
+    -- the buffer is in an autocmd window.
+    a.nvim_command("silent! bw " .. bn)
```
sindrets, line 98: I guess the reason its not caught by `pcall` is that the window isn't closed directly as a consequence of `nvim_buf_delete`, but vim tries to close it because we delete the last buffer in the window. Maybe upstream bug?

lua/nvim-tree/view.lua
```diff
@@ -68,9 +68,9 @@ M.View = {
 ---Find a rogue NvimTree buffer that might have been spawned by i.e. a session.
 ---@return integer|nil
 local function find_rogue_buffer()
-  for i = 1, vim.fn.bufnr("$"), 1 do
-    if vim.fn.bufname(i) == "NvimTree" then
-      return i
+  for _, v in ipairs(a.nvim_list_bufs()) do
+    if vim.fn.bufname(v) == "NvimTree" then
```
kyazdani42, line 72: you can also use `a.nvim_buf_get_name(v)` i think :)

lua/nvim-tree/view.lua
kyazdani42: did you try to schedule the buf_delete call ?

sindrets: Good call, scheduling seems to work quite well!

lua/nvim-tree/view.lua
sindrets: `nvim_buf_get_name` returns a path, so I thought this was cleaner.

lua/nvim-tree/view.lua
kyazdani42: yep, because it'll send the command in the event loop, thus all the other synchronous stuff must be done before triggering this :)

PR #315, opened 2021-04-18
Apply `NvimTreeOpenedFile` highlight to filename instead of icon

morgsmccauley:
Using a bold highlight is not visible when applied to the icon alone. I feel like this highlight makes more sense when applied to the filename text. Thoughts?

### DIFF
--- a/lua/nvim-tree/renderer.lua
+++ b/lua/nvim-tree/renderer.lua
@@ -302,7 +302,7 @@ local function update_draw_data(tree, depth, markers)
 
       if vim.g.nvim_tree_highlight_opened_files then
         if vim.fn.bufloaded(node.absolute_path) > 0 then
-          table.insert(hl, {'NvimTreeOpenedFile', index, offset, offset+#icon })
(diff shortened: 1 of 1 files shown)


kyazdani42:
we discussed this already, i was worried about hl conflicts between git highlighting and this. Also it looked very bold to me (although i don't use this feature anyway). 
@afonsocarlos what do you think (related #229 #258)

afonsocarlos:
I've been thinking about this, maybe a solution for this would be to have something like `g:nvim_tree_gitignore`

```vim
g:nvim_tree_highlight_opened_files = 0  " highlight fully deactivated (default)
                                        " 1 enable icon highlight
                                        " 2 enable line (icon + filename) highlight
``` 

Another option would be to have a highlight definition for the icon and another for the filename (though I prefer the first approach)

afonsocarlos:
PS: I was sending a PR to fix the last problem related to #229
I could implement this too, but in case @morgsmccauley wishes to contribute I can leave this for him :grin:

morgsmccauley:
I'll leave this to the experts! Thanks @afonsocarlos @kyazdani42

kyazdani42:
go for it @afonsocarlos seems good to me

PR #326, opened 2021-04-20
feat: add statusline winhighlight to explorer tree

akinsho:
This a small PR to add the ability to highlight the statusline differently in the nvim-tree buffer using `NvimTreeStatusLine` and `NvimTreeStatusLineNC` similar to the other `winhighlight` for other highlight groups.

NOTE: I converted the highlights to a table and concat them for readability sake so the line isn't so long and hard to read. Happy to change it back if this is undesirable 🤷🏿‍♂️


kyazdani42 (approved):
so much better, didn't though of that ;)

PR #330, opened 2021-04-20
chore(colors): Set highlight links for NvimTreeStatusLine

### LINKED ISSUE #332: Statusline not behaving properly
Did the following to set the statusline `au BufWinEnter NvimTree let &l:stl=' NvimTree'`
This is the outcome:
test

sindrets:
This sets some default highlight links for the new `NvimTreeStatusLine` highlight groups so that vim doesn't fill the statusline with carrots.

Example with minimal rc:

Before:
2021-04-20-234738_maim

After:
2021-04-20-234839_maim

fixes #332

### DIFF
--- a/lua/nvim-tree/colors.lua
+++ b/lua/nvim-tree/colors.lua
@@ -71,6 +71,8 @@ local function get_links()
     LspDiagnosticsWarning = "LspDiagnosticsDefaultWarning",
     LspDiagnosticsInformation = "LspDiagnosticsDefaultInformation",
     LspDiagnosticsHint = "LspDiagnosticsDefaultHint",
+    StatusLine = "StatusLine",
+    StatusLineNC = "StatusLineNC",
   }
 end
 


bandithijo:
Nice catch!

This is the standard behaviour of Vim.
If `StatusLine` and `StatusLineNC` have the same highlighting, the program will use the caret character (^) in the status line of the current window (in order to highlight it).

```
:help hl-StatusLineNC
---------------------

Note: if this is equal to "StatusLine" Vim will use "^^^" in the status line
of the current window.
```

Before this fix PR,
```
:verbose hi NvimTreeStatusLine
NvimTreeStatusLine xxx cleared
```
```
:verbose hi NvimTreeStatusLineNC
NvimTreeStatusLineNC xxx cleared
```
Both of them have `cleared` which are same.

I have manually solved with,
```
hi link NvimTreeStatusLine   StatusLine
hi link NvimTreeStatuslineNC StatusLineNC
```

But your solution is better.
Thank you 😄

PR #331, opened 2021-04-21
Improve highlight options

afonsocarlos:
This PR Is intended to improve highlighting options so it can suit most, if not all, of the need for tweaking file/folder highlights

### DIFF
--- a/lua/nvim-tree/renderer.lua
+++ b/lua/nvim-tree/renderer.lua
@@ -44,7 +44,9 @@ if icon_state.show_folder_icon then
   end
   set_folder_hl = function(line, depth, icon_len, name_len, hl_group)
     table.insert(hl, {hl_group, line, depth+icon_len, depth+icon_len+name_len+get_trailing_length()})
-    table.insert(hl, {'NvimTreeFolderIcon', line, depth, depth+icon_len})
+    local hl_icon = 'NvimTreeFolderIcon'
+    if vim.g.nvim_tree_highlight_opened_files then hl_icon = hl_group end
+    table.insert(hl, {hl_icon, line, depth, depth+icon_len})
   end
 end
 
--- a/lua/nvim-tree/renderer.lua
+++ b/lua/nvim-tree/renderer.lua
@@ -46,7 +46,7 @@ if icon_state.show_folder_icon then
   end
   set_folder_hl = function(line, depth, icon_len, name_len, hl_group)
     table.insert(hl, {hl_group, line, depth+icon_len, depth+icon_len+name_len+get_trailing_length()})
-    local hl_icon = vim.g.nvim_tree_highlight_opened_files and hl_group or 'NvimTreeFolderIcon'
+    local hl_icon = (vim.g.nvim_tree_highlight_opened_files or 0) ~= 0 and hl_group or 'NvimTreeFolderIcon'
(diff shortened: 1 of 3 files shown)


lua/nvim-tree/renderer.lua
```diff
@@ -44,7 +44,9 @@ if icon_state.show_folder_icon then
   end
   set_folder_hl = function(line, depth, icon_len, name_len, hl_group)
     table.insert(hl, {hl_group, line, depth+icon_len, depth+icon_len+name_len+get_trailing_length()})
-    table.insert(hl, {'NvimTreeFolderIcon', line, depth, depth+icon_len})
+    local hl_icon = 'NvimTreeFolderIcon'
+    if vim.g.nvim_tree_highlight_opened_files then hl_icon = hl_group end
+    table.insert(hl, {hl_icon, line, depth, depth+icon_len})
```
kyazdani42, line 49: `local hl_icon = vim.g.nvim_tree_highlight_opened_files and hl_group or 'NvimTreeFolderIcon'`

kyazdani42 (changes requested):

afonsocarlos:
@kyazdani42 what do you think of this approach? This way you can choose between highlighting the icon, the filename, or both. The other way, I guess, would be creating separate OpenedIcon and OpenedFile highlights.

If this is good for you I can work on the docs

lua/nvim-tree/renderer.lua
```diff
@@ -302,9 +301,13 @@ local function update_draw_data(tree, depth, markers)
         table.insert(hl, {'NvimTreeImageFile', index, offset+#icon+#git_icons, -1 })
       end
 
-      if vim.g.nvim_tree_highlight_opened_files then
-        if vim.fn.bufloaded(node.absolute_path) > 0 then
+      if vim.fn.bufloaded(node.absolute_path) > 0 then
```
kyazdani42, line 304: hum i wouldn't call bufloaded for each iteration, could you maybe do a should_hl_opened_files variable based on the value of nvim_tree_highlight_opened_files ? Done once at module initialisation (outside of the function), it would be faster i think. Maybe just make a function generator like i did at the start to handle this code.

afonsocarlos: Hmm, I see. That's neat! I'll work on that

lua/nvim-tree/renderer.lua
```diff
@@ -44,7 +44,8 @@ if icon_state.show_folder_icon then
   end
   set_folder_hl = function(line, depth, icon_len, name_len, hl_group)
     table.insert(hl, {hl_group, line, depth+icon_len, depth+icon_len+name_len+get_trailing_length()})
-    table.insert(hl, {'NvimTreeFolderIcon', line, depth, depth+icon_len})
+    local hl_icon = vim.g.nvim_tree_highlight_opened_files and hl_group or 'NvimTreeFolderIcon'
```
sindrets, line 47: The config check here should be `(vim.g.nvim_tree_highlight_opened_files or 0) ~= 0`, because 0 is a truthy value in lua. And everywhere else we treat 0 as meaning disabled.

kyazdani42 (approved):

kyazdani42:
rebase this and i'll merge :)

afonsocarlos:
Done man! :grin:

PR #334, opened 2021-04-21
feat: Sort nodes sensibly

sindrets:
Currently nodes don't always appear in conventional order in the file tree, and they are sometimes inserted in the wrong position when multiple nodes are added. This is a limitation of how insertion of new nodes is currently implemented. Further, symlinks are "sorted" as a separate group as opposed to how it is conventionally done: treating symlinks as the type of file/folder it points to. I.e a symlink pointing to a folder should be sorted alongside other folders.

Proposed solution: sort folders when new nodes are added. This is reasonable and does not have much notable impact on performance with my current implementation. Lua's `table.sort` is backed by an unstable quick sort algorithm, so I opted to instead implement a stable merge sort. Running some tests i measured a ~6.2ms runtime while sorting a folder with 5000 nodes.

Example behavior on current master while toggling dotfiles:

https://user-images.githubusercontent.com/2786478/115632673-874d7b80-a308-11eb-9dca-f000039bcab8.mp4

`.editorconfig` and `.luacheckrc` are inserted at the wrong index and symlinks are sorted incorrectly.

Example behavior with this new implementation, while toggling dotfiles:

https://user-images.githubusercontent.com/2786478/115632839-e27f6e00-a308-11eb-9356-dc4809e5b0fe.mp4

All files retain their position in the tree after toggle and symlinks are sorted correctly.

### DIFF
--- a/lua/nvim-tree/populate.lua
+++ b/lua/nvim-tree/populate.lua
@@ -106,6 +106,16 @@ local function should_group(cwd, dirs, files, links)
   return false
 end
 
+local function node_comparator(a, b)
+  if a.entries and not b.entries then
+    return true
+  elseif not a.entries and b.entries then
+    return false
+  end
+
+  return a.name:lower() <= b.name:lower()
+end
+
 local function gen_ignore_check(cwd)
   if not cwd then cwd = luv.cwd() end
   local ignore_list = {}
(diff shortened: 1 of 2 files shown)


kyazdani42 (approved):
oh but by default it should be sorted i think. anyway this looks better :)

sindrets:
It's sorted *almost* correctly after the first `populate`. Except for symlinks, because you group directories, links and files and then add them in that order.

PR #335, opened 2021-04-22
fix(git): Fix the gitignore implementation.

sindrets:
After using my previous gitignore implementation for a while I discovered that it has some problems. I.e. it does not list ignored files unless the files' parent directory is checked in to git. But after reading through some git man-pages, I found out that we can vastly simplify the implementation by adding the `ignored=matching` option to the `git status` command. This is just a lot more elegant overall, as then setting the ignore status is done exactly the same way we set all the other git statuses.

The `gitignore_map` is still needed for the `g:nvim_tree_gitignore` feature, but I'm now just constructing it as part of the `update_root_status` function.

### DIFF
--- a/lua/nvim-tree/git.lua
+++ b/lua/nvim-tree/git.lua
@@ -31,7 +31,10 @@ local function update_root_status(root)
     end
 
     roots[root][body] = head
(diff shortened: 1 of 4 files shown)


kyazdani42 (approved):
looks good and works well ! Thanks again :) just rebase and fix the conflicts

PR #337, opened 2021-04-22
Allow window width to be set dynamically

terrortylor:
This retains the current behaviour of setting the width to either the global var nvim_tree_width or the default width of 30.
However, allows the global var to be set prior to opening the the tree window. Allowing a user to set the width to say 1/3 of their screen, or a minimum of 30 which ever is biggest.

I wanted a 1/3 of the terminal to be used, and my terminal resizes when I have a monitor plugged in and when I don't.
So this is purely driven by my needs/wants... BUT YAY!

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -73,6 +73,10 @@ M.View = {
   }
 }
 
+local function get_width()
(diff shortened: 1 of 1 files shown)


terrortylor:
Ha jsut say this: :lua require'nvim-tree.view'.View.width = 50
Closing!

PR #338, opened 2021-04-22
Change filesystem actions to be immediate, minor changes to UX

CKolkey:
I was a little bothered that the filesystem actions needed you to press `<cr>` after every command. When overwriting files, it's `y<cr>` or `r<cr>` or `n<cr>`. You get it. I thought it would be a lot nicer to have the actions execute instantly, so I made those changes.

I've also made some minor changes to the UX, just to simplify the experience. I didn't think there needed to be text saying a file was created, since you just told the plugin to do it, and you can see it in the tree view. Same with removing files. You can see the feedback in the tree, so the text felt unneeded.

Additionally, asking if I wanted to copy a file/folder after I told it to paste said file/folder felt a bit "extra", so I got rid of it.

If there are any changes you want, I'd be happy to make them :)

And if you don't like this idea, just go ahead and close the pr ;)

Thanks for the great work!

### DIFF
--- a/lua/nvim-tree/fs.lua
+++ b/lua/nvim-tree/fs.lua
@@ -19,9 +19,14 @@ local function refresh_tree()
   vim.api.nvim_command(":NvimTreeRefresh")
 end
 
+local function get_user_input()
+  return vim.fn.nr2char(vim.fn.getchar())
+end
+
 local function create_file(file)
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
Hi thanks for the PR, i'll test these UX changes for some time and if its too troubling i'll revert it, but i guess your argument is valid :)

Issue #290, opened 2021-04-14
Creating new window from nvim-tree breaks default window settings


vhyrro:
I have these settings set for tabs:
```lua
vim.bo.expandtab = true
vim.bo.copyindent = true
vim.bo.shiftwidth = 4
vim.bo.tabstop = 4
vim.bo.softtabstop = 4
vim.bo.autoindent = true
vim.bo.smartindent = true
```
Whenever I open a file using `:e` or just `nvim file` my tab settings are respected, yet whenever I open a file through nvim-tree (the `o` keybinding), my tab settings seem to be erased or overwritten only for that file. Having to work around this using autocommands is rather tedious :)

Vhyrro:
After investigating a bit more, it seems to be a neovim internal problem, as this plugin's code does not seem to do any modification in that regard

The only way to fix it seems to be to set up an autocommand

```lua
vim.cmd [[ autocmd FileType * :lua setup_tabs() ]]
```

Closing, but maybe double checking wouldn't hurt just in case :)

kyazdani42:
i will investigate soon anyway, maybe this is related to the way the plugin is creating it's windows/buffer

kyazdani42:
humm i've seen that issue too, i might need help because it's an area that i cannot find documentation for. 
But for what i remember, vim/nvim will use current window settings when opening another window. this is quite confusing in this situation.

kyazdani42:
should be fixed in 4388338a6a1

PR #345, opened 2021-04-24
refactor: `open_file`

### LINKED ISSUE #341: If you close all visible buffers, nvimtree will open new buffers in new splits (not in current split)
Hi! I love your plugin.

I found a strange bug, I needed to test a lot of things to replicate it but here how to be sure to replicate it!

1 Take this minimal vimrc
```vim
call plug#begin()

call plug#end()
```

2 Open nvim with an empty buffer with `nvim`.

3 type `:NvimTreeToggle`

4 Open a file with the `enter` key.
note: If you open a  new file after this, it will replace the previous file in the split (I think it's a wanted behavior).

5 close the split with a `:q`

6 Open a new file with the `enter` key (a split appear on the left now)

7 Go to the NvimTree Panel (ctrl+w and L) and open a new file again (a split will open on the middle now).

sindrets:
Fixes #341.

@kyazdani42 Please take a look at this when you have time. While I was implementing the window picker I had to interact with the `open_file` function, and I noticed that it was strangely obscure. It has some hard to read, 150+ character long, chained ternary expressions, and variables that change type depending on several conditions.

I refactored the function, focusing on improving the readability of the logic flow, and avoiding the long chained ternary expressions. I did this while carefully trying to ensure that the behavior is exactly the same as before, while also fixing some of the oddities, like i.e. those described in 341.

One thing I *did* change was what windows its allowed to interact with. Specifically I don't think it should interact with windows outside the current tabpage. That's sort of the point of tabpages: they're completely separate window layouts. This actually fixes a problem that exists with the current implementation: if you try to open a file that is already open in another tabpage, nothing happens.

### DIFF
--- a/lua/nvim-tree/config.lua
+++ b/lua/nvim-tree/config.lua
@@ -73,11 +73,11 @@ function M.window_options()
   if vim.g.nvim_tree_side == 'right' then
     opts.open_command = 'h'
     opts.preview_command = 'l'
-    opts.split_command = 'nosplitright'
+    opts.split_command = 'aboveleft'
   else
     opts.open_command = 'l'
     opts.preview_command = 'h'
-    opts.split_command = 'splitright'
+    opts.split_command = 'belowright'
   end
 
   return opts
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -235,7 +235,7 @@ function M.open_file(mode, filename)
     if not target_winid or not vim.tbl_contains(win_ids, target_winid) then
       -- Target is invalid, or window does not exist in current tabpage: create
       -- new window
-      api.nvim_command("belowright vsp")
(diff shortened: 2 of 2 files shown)


sindrets:
@gaetan-puleo This should fix the problems you described in 341. Please give it a try if you have time, and let me know if you encounter any problems.

gaetan-puleo:
I just tried @sindrets , it works!

We have an other bug. 
When you close the last window and open a new one the linenumbers are disabled.
I think I saw this one before too but I forgot to mention it.

 Not sure if it's related, maybe I can create a new issue?

sindrets:
Thanks for taking the time! The other problem is another open issue that we haven't resolved yet #290.

kyazdani42 (approved):
thanks for rewriting this !

gaetan-puleo:
YEAH It's merged thanks @sindrets and @kyazdani42

PR #351, opened 2021-04-26
fix(renderer): Ensure buffer is loaded.

### LINKED ISSUE #350: Invalid buffer id
When I execute `:set background = light #or dark` before opening NvimTree (e.g. no valid NvimTree buffers exists), I get the following error:

```
E5108: Error executing lua ...ack/plugins/opt/nvim-tree.lua/lua/nvim-tree/renderer.lua:344: Invalid buffer id: 2
```

sindrets:
Fixes #350.


kyazdani42 (approved):
:)

PR #385, opened 2021-05-18
feat: Support copy name and path

iwfan:
- type `y` will copy name to system clipboard
- type `Y` will copy relative path to system clipboard
- type `gy` will copy absolute path to system clipboard


lua/nvim-tree/fs.lua
```diff
@@ -317,4 +317,32 @@ function M.print_clipboard()
   return api.nvim_out_write(table.concat(content, '\n')..'\n')
 end
 
+local function copy_to_clipboard(content)
+  vim.fn.setreg('+', content);
+  vim.fn.setreg('"', content);
+  return api.nvim_out_write(string.format('Copy %s to system clipboard! \n', content))
```
kyazdani42, line 323: `Copied` would be more appropriate here

lua/nvim-tree/fs.lua
```diff
@@ -317,4 +317,32 @@ function M.print_clipboard()
...
+end
+
+function M.copy_filename(node)
+  return copy_to_clipboard(node.name)
+end
+
+function M.copy_path(node)
+  local absolute_path = node.absolute_path
+  local relative_path = utils.path_relative(absolute_path, lib.Tree.cwd)
+  local content = path_is_dir(absolute_path) and utils.path_add_trailing(relative_path) or relative_path
```
kyazdani42, line 338: `path_is_dir` can be replicated by checking `node.entries`, you don't need to call fs_stat and all :)

kyazdani42 (changes requested):

lua/nvim-tree/fs.lua
iwfan: Okay, I renamed this function name to `Copied` .

lua/nvim-tree/fs.lua
iwfan: I removed `path_is_dir` function. But by the way, `node.entries` is it a bit unobscure, I can't directly understand what it means.

lua/nvim-tree/fs.lua
```diff
@@ -317,4 +317,27 @@ function M.print_clipboard()
   return api.nvim_out_write(table.concat(content, '\n')..'\n')
 end
 
+local function Copied(content)
```
kyazdani42, line 320: i did not mean the function, the name before was ok, but the string.format should say `Copied %s to system clipboard` instead of Copy

kyazdani42 (changes requested):

lua/nvim-tree/fs.lua
kyazdani42: i guess so, but it still is part of this `node` api which is not really well defined :)

lua/nvim-tree/fs.lua
iwfan: Sorry, I misunderstood.

kyazdani42 (approved):

PR #386, opened 2021-05-19
fix(lib): Fix the `tabnew` callback.

### LINKED ISSUE #381: C-t doesn't seem to open on a new tab?
I just installed the extension, and also managed to reproduced this bug with a minimal configuration with only nvim-tree.lua loaded. I'm using default setting for all the variables and keybindings.

When pressing C-t on a file, the file is opened in the same tab where the nvim-tree is open (effectively same behaviour as <CR>. This is different than what happens when pressing C-t in FZF or t in NERDTree. I verified that the other tools correctly open a new tab, and also that :tabnew works as intended.

Am I doing something wrong or is this a recent bug?
NVIM v0.5.0-dev+1326-g93aea8cee

sindrets:
Fixes #381.

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -282,6 +282,18 @@ function M.pick_window()
 end
 
 function M.open_file(mode, filename)
+  if mode == "tabnew" then
+    -- Switch window first to ensure new window doesn't inherit settings from
+    -- NvimTree
+    if M.Tree.target_winid > 0 and api.nvim_win_is_valid(M.Tree.target_winid) then
+      api.nvim_set_current_win(M.Tree.target_winid)
+    else
+      vim.cmd("wincmd p")
+    end
+    vim.cmd("tabe " .. vim.fn.fnameescape(filename))
+    return
+  end
+
   local tabpage = api.nvim_get_current_tabpage()
   local win_ids = api.nvim_tabpage_list_wins(tabpage)
 


kyazdani42 (approved):
you could've led me to the fix for this window setting issue

sindrets:
@kyazdani42 huh? What do you mean?

kyazdani42:
```
   -- Switch window first to ensure new window doesn't inherit settings from
    -- NvimTree
    ```
    check out #290 i've been struggling for a while since i've made the 'view' refactoring...

sindrets:
Oh, sorry. I thought you knew the cause of the problem. The problem for me is that I can't seem to find a good solution for the problem. Every solution is very situational. Take for instance the case where nvimtree is the last open window. There doesn't seem to be any way to create a new window without inheriting the window settings from nvimtree.

kyazdani42:
but before this issue did not happen at all. I'm not sure why, but this new way of handling the window is making new windows open with its properties instead of the global window properties defined by the users :/

sindrets:
Hm, I don't know how this wasn't a problem before, but this is how window setting inheritance has always worked in vim for as long as I can remember. There are no "global" window settings. Window settings are always set only for the current window, and when you create a new window, it inherits all the window settings from the current window.

Actually it's more complicated than that, because there are also exceptions. Like for instance, it seems like if you open a new window with a buffer that is already loaded, the new window does *not* inherit settings.

This is all rather complicated and contrived, and I've seen that even the core devs struggle to maintain a mental model of how all this works.

kyazdani42:
ah thanks for the explanation, wasn't sure after reading the docs 3 times ...

PR #394, opened 2021-05-21
fix(lib): Handle nvim_win_get_option correctly when it fails.

### LINKED ISSUE #392: Statusline issue with multiple windows
After the recent update, my statusline messes up once I select a window.

Steps to reproduce:
* open 2 split windows
* Try to open some file, select for example `B`
* Check first window statusline, it prints an error.

Here's the screenshot:
screenshot

And here's my statusline configuration: https://github.com/kristijanhusak/neovim-config/blob/bleeding-edge/nvim/lua/partials/statusline.lua

sindrets:
Fixes #392.

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -245,12 +245,12 @@ function M.pick_window()
   -- Setup UI
   for _, id in ipairs(selectable) do
     local char = chars:sub(i, i)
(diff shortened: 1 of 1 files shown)


kristijanhusak (approved):

kristijanhusak:
It would be great to merge this if there's no reason to wait for anything else.

PR #395, opened 2021-05-21
feat(lib): Improved tabnew behavior.

sindrets:
Close the tree if nvim_tree_quit_on_open is enabled, otherwise retain
focus on the tree.


sindrets:
@ysdexlic Give this a try if you have time.

kyazdani42:
I will not be available until Monday evening, so I'll review properly on my computer when I'm back from my long weekend :)

kyazdani42 (approved):
thanks :)

PR #399, opened 2021-05-25
fix(lib): Refresh tree if it was opened before init.

### LINKED ISSUE #383: Tree empty if entering quickly after starting neovim
This is split from the discussions in #280.

I do not use packer lazy loading for nvim-tree. Yet, when starting neovim and quickly entering the tree (with a leader mapping), the tree is often initially empty and I have to re-toggle or refresh it to get contents.

image

This was the result of starting neovim and directly afterwards using my leader mapping leader-n. Setting debug="throw" did not surface anything.

My config is available here: https://github.com/languitar/config-nvim

Neovim is compiled daily and the plugins are updated along with that.

sindrets:
Fixes #383.

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -34,6 +34,8 @@ function M.init(with_open, with_reload)
 
   if with_open then
     M.open()
+  elseif (view.win_open()) then
+    M.refresh_tree()
   end
 
   if with_reload then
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/lib.lua
```diff
@@ -34,6 +34,8 @@ function M.init(with_open, with_reload)
 
   if with_open then
     M.open()
+  elseif (view.win_open()) then
```
kyazdani42, line 37: could you just remove the parens around the condition to keep the codestyle please ?

kyazdani42 (changes requested):

kyazdani42 (approved):
thanks man :)

PR #401, opened 2021-05-27
feat: Option for excluding buffers from the window picker.

### LINKED ISSUE #400: don't ask which window to open a new file to if we only have edit window (A) and quickfix (B)
I have 3 windows open: editor, file tree and quickfix (with some LSP results), when I open a new file in the tree - it asks where do I want the file to end up. 

it would be good to filter out the quickfix window from the suggestions and just open the file.

sindrets:
Fixes #400.

Adds a new option `g:nvim_tree_window_picker_exclude`, that lets you exclude certain buffers from the window picker.

Example:

```vim
let g:nvim_tree_window_picker_exclude = {
    \   'filetype': [
    \     'packer'
    \     'qf',
    \   ],
    \   'buftype': [
    \     'terminal'
    \   ]
    \ }
" Dictionary of buffer option names mapped to a list of option values that
" indicates to the window picker that the buffer's window should not be
" selectable.
```


kyazdani42 (approved):
thanks again :) i was going to ask you if you wanted to tackle the issue, but you were faster than me !

PR #405, opened 2021-05-28
Shift focus to previous window before closing tree

booperlv:
This is more of a fallback than anything - as in normal cases everything would work. 
This may save a few more use cases and doesn't really destroy any to my knowledge though (Tested out a few). 

I used lualine (evil line config, modified) and every time I closed it, it left the focus in the tree. Had to reopen and then select the normal window manually.

This was a simple solution I came up with.


booperlv:
Saw issue #416, this is probably not the optimal solution. Closing :)

PR #411, opened 2021-05-29
Add function to open current node with OS default app to README tips section

cseelus:
Adds a function `NvimTreeOSOpen()` to the `README` tips section, which opens the node under the cursor in the OSs default application.


kyazdani42 (approved):
good idea thanks

PR #415, opened 2021-05-30
fix: Gitignore rules not being applied on new nodes.

### LINKED ISSUE #412: Not taking .gitignore into account for files created after starting up vim
# Problem description

The gitignore settings are only taken into account for the files that existed when starting vim, not for files created afterwards. 

# Steps to reproduce

Setup a new repo for the tests:
```
mkdir test
cd test
touch a a~ b b~ c
git init
echo "*~" > .gitignore
```

## Test 1

1. Start vim
1. `:set backup` (to make sure backup files get created in the same directory with `~` appended to the end of the filename)
1. Open the file browser, observe that you can toggle with `Shift-I` whether the files `a~` and `b~` are shown.
1. Open file `c` and save it while the file browser is open. Note that `c~` gets created and it gets shown in the browser, but `Shift+I` won't hide it like the rest of the backup files; it is _always_ shown.
1. `NvimTreeRefresh` or opening/closing the browser doesn't help; `c~` will be always shown.
1. Restart vim and open the file browser again. Now `c~` gets toggled correctly too with `Shift-I`.

## Test 2

1. Start vim
2. Open the file browser, observe that you can toggle with `Shift-I` whether the files `a~` and `b~` are shown.
3. Create a new file called `x~` from another shell while vim is open.
4. Press `Shift+R` in the file browser to refresh the view; note that `x~` is now shown.
5. Similarly to Test1, `Shift+I` will toggle the backup files that existed at startup, but not `x~`.

# Conclusion

Somehow the gitignore is taken into account only for the state of the filesystem at startup (?), but not for files created from vim.

sindrets:
Fixes #412.

Reload git roots before refreshing nodes so that the gitignore map is
updated, and ignore rules are applied correctly on refresh_entries.

### DIFF
--- a/lua/nvim-tree/config.lua
+++ b/lua/nvim-tree/config.lua
@@ -64,6 +64,12 @@ function M.get_icon_state()
   }
 end
 
+function M.use_git()
+  return M.get_icon_state().show_git_icon
+      or vim.g.nvim_tree_git_hl == 1
+      or vim.g.nvim_tree_gitignore == 1
+end
+
 function M.nvim_tree_callback(callback_name)
   return string.format(":lua require'nvim-tree'.on_keypress('%s')<CR>", callback_name)
 end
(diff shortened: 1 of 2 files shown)


kyazdani42 (approved):
thanks !

Issue #419, opened 2021-05-31
Issue with Galaxyline in combination with Nvim-Tree


Artem-Schander:
Hi

Firstly, thank you for this package.
I created an [issue in the Galaxyline repo](https://github.com/glepnir/galaxyline.nvim/issues/177), but I think here is also the right place (at least) to share the workaround.

When I open Nvim-Tree, Galaxyline goes into short_list mode while Nvim-Tree is open (as expected), but it does not change back when I close it.

Here is a workaround for everyone experiencing the same behaviour.
```
if pcall(require, 'galaxyline') then
    vim.cmd('command! NvimTreeToggle lua require("galaxyline").load_galaxyline(); require"nvim-tree".toggle()')
end
```

kyazdani42:
fixed by #421

Issue #418, opened 2021-05-31
Unable to open splits when focus in nvim-tree


petertriho:
https://github.com/kyazdani42/nvim-tree.lua/commit/be184bd94e558b302461b359d02ed1b01f7154a1 seems to have broken the ability to open splits when I'm focused in nvim-tree i.e. I use to be able to run :Git (from vim-fugitive) or :Neogit kind=split (from neogit) to open the git status split

rekaerst:
be184bd also causes nvim-tree to clone itself when opening a new window while focused in nvim-tree.
![image](https://user-images.githubusercontent.com/52642831/120210242-28154c00-c262-11eb-9df9-c62d68415649.png)

And messed up floating window.
![image](https://user-images.githubusercontent.com/52642831/120209497-57778900-c261-11eb-9963-15cfe9e4eeef.png)

```lua
function M.save_previous_buf()
  vim.cmd "wincmd p"
  old_win = api.nvim_get_current_win()
  vim.cmd "wincmd p"
end

function M.set_previous_buf()
  if old_win then
    api.nvim_set_current_win(old_win)
  end
end
```

kyazdani42:
sorry for the inconvenience... a mistake on my end. Fixed by #421

petertriho:
@kyazdani42 No worries! Thanks you so much for the awesome plugin

PR #421, opened 2021-05-31
Revert "feat: reset old window when leaving the tree"

### LINKED ISSUE #416: Bug: NvimTree doesn't revert highlight active for statusline when closing it.
Nvim version: NVIM v0.5.0-dev+1359-g43956dea5
NvimTree: updated to latest commit. 

When toggling NvimTree the statusline doesn't go back to be active, and it stays with the highlight colors of inactive mode even though NvimTree is closed. Tested with feline and lualine, both have the same behavior. 

nvim

In this video see how the statusline is inactive when NvimTree is closed, the only way to fixed is by moving to another file/buffer.  

### Minimal init.lua to reproduce.

```lua
-- install packer
local fn = vim.fn

local install_path = "/tmp/nvim/site/pack/packer/start/packer.nvim"

vim.cmd [[set packpath=/tmp/nvim/site]]

local function load_plugins()
    local use = require("packer").use
    require("packer").startup(
        {
            function()
                use "wbthomason/packer.nvim"
                use "hoob3rt/lualine.nvim"
                use "kyazdani42/nvim-tree.lua"
            end,
            config = {package_root = "/tmp/nvim/site/pack"}
        }
    )
end

_G.load_config = function()
    require("lualine").setup()
end

if fn.isdirectory(install_path) == 0 then
    fn.system({"git", "clone", "https://github.com/wbthomason/packer.nvim", install_path})
    load_plugins()
    require("packer").sync()
    vim.cmd "autocmd User PackerComplete ++once lua load_config()"
else
    load_plugins()
    _G.load_config()
end
```

### Steps to reproduce:
1. Use minimal init.lua above.
2. Open a file.
3. Toggle NvimTree
4. Statusline will be inactive even after NvimTree is closed.
5. Move to another buffer/file and the statusline will get fixed.

### More context:
This has been happening for couple of days, I thought it was something wrong with my statusline, but no. 
1. It only happens when I toggle NvimTree. 
2. Doesn't happen with any other plugin with UI elements that I have (like undotree). 
3. Installed lualine just to test and it has the same behavior. 

Above a minimal init.lua to reproduce. And obviously this is a new thing, wasn't happening before.

sindrets:
Fixes #416, #418, #419.

This reverts commit be184bd94e558b302461b359d02ed1b01f7154a1.

@kyazdani42 I still don't fully understand what this commit is supposed to accomplish, but it is certainly not working correctly, and it's currently causing all sorts of strange behavior. I think we should revert this until we can figure out a better way to do it.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -194,20 +194,6 @@ function M.reset_highlight()
   renderer.render_hl(view.View.bufnr)
 end
 
-local old_win = nil
-
-function M.save_previous_buf()
-  vim.cmd "wincmd p"
-  old_win = api.nvim_get_current_win()
-  vim.cmd "wincmd p"
-end
-
-function M.set_previous_buf()
-  if old_win then
-    api.nvim_set_current_win(old_win)
-  end
-end
-
 view.setup()
 colors.setup()
 vim.defer_fn(M.on_enter, 1)
--- a/plugin/tree.vim
+++ b/plugin/tree.vim
@@ -17,8 +17,6 @@ augroup NvimTree
     au User LspDiagnosticsChanged lua require'nvim-tree.diagnostics'.update()
   endif
   au BufEnter * lua require'nvim-tree'.buf_enter()
-  au BufEnter NvimTree lua require'nvim-tree'.save_previous_buf()
-  au BufLeave NvimTree lua require'nvim-tree'.set_previous_buf()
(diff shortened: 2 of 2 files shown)


kyazdani42:
agreed, i've been experiencing many of these issues today. It was supposed to refocus the window from which the tree was opened after closing the tree but obviously it launches all sort of autocmd because of window switching. Not sure how to do that in a better way :/

PR #423, opened 2021-05-31
Improve icon fetching by file extension

luissimas:
This adds the support for different icons when dealing with files that have multiple extensions (like `users.test.js`).

The motivation for this is based on [this issue from nvim-web-devicons](https://github.com/kyazdani42/nvim-web-devicons/issues/36).

Given this sample config file for [nvim-web-devicons](https://github.com/kyazdani42/nvim-web-devicons):
```lua
require("nvim-web-devicons").setup {
  override = {
    ["test.js"] = {
      icon = "ﭧ",
      color = "#cbcb41",
      name = "JsTest"
    },
    ["spec.js"] = {
      icon = "",
      color = "#cbcb41",
      name = "JsSpec"
    },
    ["test.lua"] = {
      icon = "ﭧ",
      color = "#51a0cf",
      name = "LuaTest"
    },
    ["spec.lua"] = {
      icon = "",
      color = "#51a0cf",
      name = "LuaSpec"
    }
  },
  default = true
}
```

## Original behavior
image

## New behavior
image

### DIFF
--- a/lua/nvim-tree/populate.lua
+++ b/lua/nvim-tree/populate.lua
@@ -46,7 +46,7 @@ local function file_new(cwd, name)
     name = name,
     absolute_path = absolute_path,
     executable = is_exec,
-    extension = vim.fn.fnamemodify(name, ':e') or "",
+    extension = string.match(name, "%.(.*)") or "",
     match_name = path_to_matching_str(name),
     match_path = path_to_matching_str(absolute_path),
   }
--- a/lua/nvim-tree/populate.lua
+++ b/lua/nvim-tree/populate.lua
@@ -46,7 +46,7 @@ local function file_new(cwd, name)
     name = name,
     absolute_path = absolute_path,
     executable = is_exec,
-    extension = string.match(name, "%.(.*)") or "",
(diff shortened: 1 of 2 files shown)


kyazdani42:
this solution looks pretty good to me, i'm gonna test it out later today, and if it works as expected i'll merge it :) thanks

lua/nvim-tree/populate.lua
```diff
@@ -46,7 +46,7 @@ local function file_new(cwd, name)
     name = name,
     absolute_path = absolute_path,
     executable = is_exec,
-    extension = vim.fn.fnamemodify(name, ':e') or "",
+    extension = string.match(name, "%.(.*)") or "",
```
sindrets, line 49: This pattern will return the full file name for dotfiles like `.gitignore`, which is incorrect. I think the pattern needs to be changed to something like `".?[^.]+%.(.*)"`.

luissimas: Yes, you're right. I didn't think about that case. I did some tests with this pattern that you suggested and it seems to work fine. Thanks!

kyazdani42 (approved):
looks like it works fine :) thanks for the contribution !

PR #427, opened 2021-06-02
add small arrows before folders

### LINKED ISSUE #426: Option to put an arrow in front of the folder
I want to display an arrow in front of the folder icon like atom or vscode.
image

If I use nvim_tree_icons to set two characters(e.g. '> □'), it will be out of alignment with other than the folder.

kyazdani42:
closes #426

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,13 @@ let g:nvim_tree_show_icons = {
     \ 'git': 1,
     \ 'folders': 0,
     \ 'files': 0,
+    \ 'folder_arrows': 0,
     \ }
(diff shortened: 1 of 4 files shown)


yamatsum:
Thank you for your quick implementation!
I tried it immediately, but there were two problems.
- Indentation of folder and other icons is not aligned
- The color of the arrow letters is hard to see

CleanShot 2021-06-03 at 01 58 13@2x

kyazdani42:
fixed the first one indeed.
For the second one, open a new issue, i'll implement this when i have some more time.

PR #438, opened 2021-06-08
Import diagnostics iff it's enabled

lanza:


lua/nvim-tree/lib.lua
```diff
@@ -4,7 +4,9 @@ local luv = vim.loop
 local renderer = require'nvim-tree.renderer'
 local config = require'nvim-tree.config'
 local git = require'nvim-tree.git'
-local diagnostics = require'nvim-tree.diagnostics'
+if vim.g.nvim_tree_lsp_diagnostics == 1 then
+  local diagnostics = require'nvim-tree.diagnostics'
+end
```
kyazdani42, line 9: sorry but this will error out when using `diagnostics.somefunc` because diagnostics will be nil outside of the `if` statement.

kyazdani42 (changes requested):

kyazdani42:
i'm not sure what's the purpose of this PR ?

lanza:
> i'm not sure what's the purpose of this PR ?

Sorry for the delay -- this plugin errors out without the diagnostics plugin installed and active.

lua/nvim-tree/lib.lua
lanza: I'm guarding the import here with the same guard you use throughout the rest of the file. All uses of and defs of `local diagnostics` are guarded in this file by `vim.g.nvim_tree_lsp_diagnostics`.

sindrets:
> this plugin errors out without the diagnostics plugin installed and active.

@lanza NvimTree does not depend on  any diagnostics plugin. If you somehow get an error from this `require` statement, chances are you have something misconfigured, your neovim is outdated, or your plugin installation is broken.

PR #445, opened 2021-06-11
Remove cursorcolumn and colorcolumn

lvim-tech:
Disable cursorcolumn and colorcolumn

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -22,6 +22,8 @@ M.View = {
     signcolumn = 'yes',
     foldmethod = 'manual',
     foldcolumn = '0',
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/view.lua
```diff
@@ -22,6 +22,8 @@ M.View = {
     signcolumn = 'yes',
     foldmethod = 'manual',
     foldcolumn = '0',
+    cursorcolumn = false,
+    colorcolumn = '0',
```
kyazdani42, line 26: not sure what those this change implies, these options are disabled by default as per the docs right ?

lvim-tech: nvim-tree

PR #462, opened 2021-06-21
add option to use width as percentage

booperlv:
adds option to use nvim_tree_width as a percentage (reads the width as a percentage and converts to decimal, multiply to vim total columns)

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -226,9 +226,9 @@ function M.open()
   a.nvim_command("wincmd "..move_to)
 
   local width
-  if (vim.g.nvim_tree_width_as_percent == true) then
-    local percentAsDecimal = M.View.width/100
-    width = math.floor( vim.o.columns * percentAsDecimal )
+  if vim.g.nvim_tree_width_as_percent == 1 then
+    local percent_as_decimal = M.View.width / 100
+    width = math.floor(vim.o.columns * percent_as_decimal)
   else
     width = M.View.width
   end
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -224,7 +224,16 @@ function M.open()
   a.nvim_command("vsp")
   local move_to = move_tbl[M.View.side]
   a.nvim_command("wincmd "..move_to)
-  a.nvim_command("vertical resize "..M.View.width)
+
+  local width
+  if (vim.g.nvim_tree_width_as_percent == true) then
+    local percentAsDecimal = M.View.width/100
+    width = math.floor( vim.o.columns * percentAsDecimal )
+  else
+    width = M.View.width
+  end
+  a.nvim_command("vertical resize "..width)
+
   local winnr = a.nvim_get_current_win()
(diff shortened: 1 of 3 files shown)


lua/nvim-tree/view.lua
```diff
@@ -224,7 +224,16 @@ function M.open()
   a.nvim_command("vsp")
   local move_to = move_tbl[M.View.side]
   a.nvim_command("wincmd "..move_to)
-  a.nvim_command("vertical resize "..M.View.width)
+
+  local width
+  if (vim.g.nvim_tree_width_as_percent == true) then
```
kyazdani42, line 229: no parens in lua conditions please :)

lua/nvim-tree/view.lua
```diff
@@ -224,7 +224,16 @@ function M.open()
...
+    local percentAsDecimal = M.View.width/100
```
kyazdani42, line 230: space around /. Also use snake_case

lua/nvim-tree/view.lua
```diff
@@ -224,7 +224,16 @@ function M.open()
...
+    width = math.floor( vim.o.columns * percentAsDecimal )
```
kyazdani42, line 231: remove spaces after opening parens and before closing parens

kyazdani42 (changes requested):
thanks for the feature, a few formatting issues to fix before merging :)

lua/nvim-tree/view.lua
gegoune: Should that not be ` == 1`? Aren't all other option in nvim-tree `0/1` rather than `true/false`?

kyazdani42 (changes requested):
oh yes and please add documentation for the option

lua/nvim-tree/view.lua
gegoune: Could just add some formatter to CI checking each PR for style issues. It's good to have consistent styling throughout the project.

booperlv:
Sorry for the odd code, whipped this up as a quick fix while on a break from other projects.
Will clean up in a few, Thanks!

booperlv:
One more question by the way, Where do you think it would be best to place the option in the documentation? Would right under g:nvim_tree_width be better or is it good as is right now?

sindrets:
Some thoughts: do we really need another option for this? Seems really unnecessary. Can we not just make it so that `nvim_tree_width` also accepts a string, and then we can determine whether or not it's a percentage depending on the value.

```vim
let g:nvim_tree_width = 30 " indicates width of 30 columns.
let g:nvim_tree_width = '30%' " indicates width of 30% of '&columns'.
```

gegoune:
That is very reasonable.

kyazdani42:
agreed with sindrets proposal, looks better this way :)

booperlv:
alright, going to start working on it :)

booperlv:
Opened a new pull request ( #473 ) with the suggested changes, both because I think it's different, and because I messed up resetting the fork's master branch and ended up deleting the fork. The more I learn about git and not messing around with it at 2AM. Closing this one now :D

PR #466, opened 2021-06-24
[Readme] Use correct example for special files

### LINKED ISSUE #410: nvim_tree_special_files miss icons, setting doesn't work correctly
Don't know if its intentional, but `nvim_tree_special_files` are shown without a file icon:

Screen Shot 2021-05-29 at 13 20 48

Just copying the setting from the NvimTree README (`let g:nvim_tree_special_files = [ 'README.md', 'Makefile', 'MAKEFILE' ]`) into my NvimTree settings, and my projects `README.md` strangely has no special highlight anymore.

If I wanted to disable any special highlighting for files, using `let g:nvim_tree_special_files = [ ]` works as expected though.

baahrens:
With the original example I couldn't get it to work except for when you use `let g:nvim_tree_special_files = []`.

This would probably fix #410

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ let g:nvim_tree_window_picker_exclude = {
 " Dictionary of buffer option names mapped to a list of option values that
 " indicates to the window picker that the buffer's window should not be
 " selectable.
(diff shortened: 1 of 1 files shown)


kyazdani42:
thanks for that, i didn't realize the documentation was wrong :/

PR #473, opened 2021-06-28
add option to make width a percentage of "&columns"

booperlv:
Allows nvim_tree_width to accept a string, that will determine whether or not it's a percentage of width of entire editor in columns.
Following @sindrets proposal from the previous pull request.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -202,12 +202,23 @@ function M.focus(winnr, open_if_closed)
   a.nvim_set_current_win(wnr)
 end
 
+local function get_width()
+  if type(M.View.width) == "string" then
+    if vim.endswith(M.View.width, "%") then
+      local width_as_number = tonumber(M.View.width:match('[^%d]*([%d%.]+)%%'))
+      local percent_as_decimal = width_as_number / 100
+      return math.floor(vim.o.columns * percent_as_decimal)
+    end
+  end
+  return M.View.width
+end
+
 function M.resize()
   if not a.nvim_win_is_valid(M.get_winnr()) then
     return
   end
 
-  a.nvim_win_set_width(M.get_winnr(), M.View.width)
+  a.nvim_win_set_width(M.get_winnr(), get_width())
 end
 
 local move_tbl = {
@@ -225,7 +236,7 @@ function M.open()
   a.nvim_command("vsp")
   local move_to = move_tbl[M.View.side]
   a.nvim_command("wincmd "..move_to)
-  a.nvim_command("vertical resize "..M.View.width)
+  a.nvim_command("vertical resize "..get_width())
   local winnr = a.nvim_get_current_win()
   local tabpage = a.nvim_get_current_tabpage()
   M.View.tabpages[tabpage] = vim.tbl_extend("force", M.View.tabpages[tabpage] or {help = false}, {winnr = winnr})
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -203,14 +203,13 @@ function M.focus(winnr, open_if_closed)
 end
 
 local function get_width()
-  if type(M.View.width) == "string" then
-    if vim.endswith(M.View.width, "%") then
-      local width_as_number = tonumber(M.View.width:match('[^%d]*([%d%.]+)%%'))
-      local percent_as_decimal = width_as_number / 100
-      return math.floor(vim.o.columns * percent_as_decimal)
-    end
+  if type(M.View.width) == "number" then
+    return M.View.width
+  elseif type(M.View.width) == "string" and vim.endswith(M.View.width, "%") then
+    local width_as_number = tonumber(M.View.width:sub(0, -2))
+    local percent_as_decimal = width_as_number / 100
+    return math.floor(vim.o.columns * percent_as_decimal)
   end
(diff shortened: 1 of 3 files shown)


lua/nvim-tree/view.lua
```diff
@@ -202,12 +202,23 @@ function M.focus(winnr, open_if_closed)
   a.nvim_set_current_win(wnr)
 end
 
+local function handle_width()
+  if type(M.View.width) == "string" then
+    if string.sub(M.View.width, -1) == "%" then
+      local width_as_number = math.floor(string.sub(M.View.width, 0, -2))
```
sindrets, line 208: This substring is not gonna work for single digit, or triple digit percentage. Please use a pattern for this instead. And don't floor the number here. Allow users to use decimal percentage. You are flooring the final result anyway.

```lua
if vim.endswith(M.View.width, "%") then
  local width_as_number = tonumber(M.View.width:match('[^%d]*([%d%.]+)%%') or 30)
```

lua/nvim-tree/view.lua
```diff
@@ -202,12 +202,23 @@ function M.focus(winnr, open_if_closed)
   a.nvim_set_current_win(wnr)
 end
 
+local function handle_width()
```
sindrets, line 205: This function name is somewhat ambiguous. Maybe `get_width` would be more descriptive?

sindrets (changes requested):

lua/nvim-tree/view.lua
booperlv: Awesome pattern here!
Would this suffice?
```lua
local function get_width()
  if vim.endswith(M.View.width, "%") then
    local width_as_number = tonumber(M.View.width:match('[^%d]*([%d%.]+)%%'))
    local percent_as_decimal = width_as_number / 100
    return math.floor(vim.o.columns * percent_as_decimal)
  end
  return M.View.width
end
```

sindrets: Yes, that's great. But you still need the type check, because `vim.endswith` expects a string and will throw an error otherwise.

booperlv:
*Facepalm* did the same mistake again. I am very sorry - first time actually doing anything on github other than messing around with my own repo. Reopening this.

sindrets (approved):

sindrets:
No worries, we've all been through this at some point:)

lua/nvim-tree/view.lua
```diff
@@ -202,12 +202,23 @@ function M.focus(winnr, open_if_closed)
...
 
+local function get_width()
+  if type(M.View.width) == "string" then
+    if vim.endswith(M.View.width, "%") then
+      local width_as_number = tonumber(M.View.width:match('[^%d]*([%d%.]+)%%'))
+      local percent_as_decimal = width_as_number / 100
+      return math.floor(vim.o.columns * percent_as_decimal)
+    end
+  end
+  return M.View.width
```
kyazdani42, line 213: when you have a logic that makes the whole code indented this way, it's better to do it in the other way:
```lua
if type(M.View.width) == "number" then
  return M.View.width
end
...rest of the code
```

lua/nvim-tree/view.lua
```diff
@@ -202,12 +202,23 @@ function M.focus(winnr, open_if_closed)
   a.nvim_set_current_win(wnr)
 end
 
+local function get_width()
+  if type(M.View.width) == "string" then
+    if vim.endswith(M.View.width, "%") then
+      local width_as_number = tonumber(M.View.width:match('[^%d]*([%d%.]+)%%'))
```
kyazdani42, line 208: i prefer the sub instead of the match, easier to understand.

kyazdani42 (changes requested):

lua/nvim-tree/view.lua
booperlv: Going to leave this open just in case: I may have not understood this fully. Please check the most recent commit!

lua/nvim-tree/view.lua
```diff
@@ -203,14 +203,13 @@ function M.focus(winnr, open_if_closed)
...
 local function get_width()
-  if type(M.View.width) == "string" then
-    if vim.endswith(M.View.width, "%") then
-      local width_as_number = tonumber(M.View.width:match('[^%d]*([%d%.]+)%%'))
-      local percent_as_decimal = width_as_number / 100
-      return math.floor(vim.o.columns * percent_as_decimal)
-    end
+  if type(M.View.width) == "number" then
+    return M.View.width
+  elseif type(M.View.width) == "string" and vim.endswith(M.View.width, "%") then
```
kyazdani42, line 208: i think you might not need this check. If the user badly setup the tree, let it fail.
```lua
if type(M.View.width) == "number" then
  return M.View.width
end
local width_as_number = ...
local percent...
return math.floor(...)
```

kyazdani42 (changes requested):

lua/nvim-tree/view.lua
booperlv: I see, I initially thought that maybe I should only accept things that are actually valid inputs: though this is equally as reasonable of a solution.

booperlv:
Assuming this is all good now, Should I cleanup the commit history before it gets merged?

kyazdani42 (approved):
thanks ;)

PR #486, opened 2021-07-02
Use relative path when possible

eric-hansen:
Introduces a new variable: `vim_tree_load_git_file_relatively` to not impact previous behavior.  If non-nil then will attempt to use the absolute path, assuming within a Git project.  I couldn't find a way to do it otherwise and didn't want to spend too much time hacking this about.

Resolves https://github.com/kyazdani42/nvim-tree.lua/issues/482

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -80,7 +80,7 @@ local keypress_funcs = {
   close = function() M.close() end,
   preview = function(node)
     if node.entries ~= nil or node.name == '..' then return end
-    return lib.open_file('preview', node.absolute_path)
+    return lib.open_file('preview', node.filename_to_open)
   end,
 }
 
@@ -106,7 +106,7 @@ function M.on_keypress(mode)
   elseif node.entries ~= nil then
     lib.unroll_dir(node)
   else
-    lib.open_file(mode, node.absolute_path)
+    lib.open_file(mode, node.filename_to_open)
   end
 end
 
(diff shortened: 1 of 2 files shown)


smartding:
please don't assume we're always within a git project, I use nvim to edit non-git-managed files too

eric-hansen:
> please don't assume we're always within a git project, I use nvim to edit non-git-managed files too

Fixed. :)

sindrets:
This won't work when the git root is not the cwd of the tree. Also, this will be enabled even when:
```vim
let g:nvim_tree_load_git_file_relatively = 0
```
Which goes against how we treat all the other boolean options. #487 is a better approach.

eric-hansen:
Gotcha.  Since you believe #487 is the better approach I've went and closed this out.

Hubro:
For some reason this change was reverted:

https://github.com/kyazdani42/nvim-tree.lua/commit/099f905e5ca93e12f455deb98a0d820c69ab5dd7

And nvim-tree is once again opening all files using an absolute path, which is a huge pain for me.

Why was the change reverted? Is there something we can do to re-introduce it?

smartding:
@Hubro 
I don't seem to experience the bug described in that video in #495, so I just kept my patch in my own fork, the patch only changes two lines of code.

fragov:
Same issue. Btw, Telescope opens files with relative paths.
Can this be fixed?

alex-courtis:
> Same issue. Btw, Telescope opens files with relative paths. Can this be fixed?

There is little consistency between plugins and whether they open files relatively or absolutely.

It should be possible to open relatively; please raise a feature request.

PR #487, opened 2021-07-03
open file with relative path when possible

smartding:
nvim-tree currently opens file with its absolute path, so ctrl-g and echo expand('%') can return a really long path that's hard to read.

This PR make nvim-tree open/preview file with its relative path. It works for both files and links.
If the current file links to a another file which is outside current working directory, nvim-tree opens the file with its absolute path.

Should fix https://github.com/kyazdani42/nvim-tree.lua/issues/482


kyazdani42 (approved):
lgtm, thanks :)

PR #497, opened 2021-07-07
Add `NvimTreeResize size` to resize the NvimTree window

kassio:
Related to: https://github.com/kyazdani42/nvim-tree.lua/issues/468


kyazdani42 (approved):
thanks :)

PR #508, opened 2021-07-11
fix(doc) neovim required version

n3wborn:
Now that nvim 0.5 is stable (and nightly >= 0.6), replace **neovim nightly (>= 0.5.0)**  by **neovim >= 0.5.0** in doc


kyazdani42 (approved):
thanks for the contrib :)

PR #510, opened 2021-07-12
Rename buffers on cut.

diogo464:
Currently cutting files from one place to another leaves their buffers with the old name and saving that buffer will create a new file in the old location instead of saving to the new one. This change just renames the buffers the same way as if you renamed the file.


kyazdani42 (approved):
thanks !

PR #512, opened 2021-07-14
add new config option g:nvim_tree_symlink_arrow (fixes #283)

### LINKED ISSUE #283: "➛" for symlink should be configurable
https://github.com/kyazdani42/nvim-tree.lua/blob/82b20f5b5ed741d2e6360990ee11a50f0cd253a4/lua/nvim-tree/renderer.lua#L279

The "➛" symbol when showing a symlink should be configurable. If you agree I can make a PR :)

cs:


lua/nvim-tree/renderer.lua
```diff
@@ -335,8 +335,9 @@ local function update_draw_data(tree, depth, markers)
     elseif node.link_to then
       local icon = get_symlink_icon()
       local link_hl = git_hl or 'NvimTreeSymlink'
+      local arrow = vim.g.nvim_tree_symlink_arrow or ' >> '
       table.insert(hl, { link_hl, index, offset, -1 })
-      table.insert(lines, padding..icon..node.name.." ➛ "..node.link_to)
```
kyazdani42, line 339: i rather use this icon as default

kyazdani42 (changes requested):

lua/nvim-tree/renderer.lua
kyazdani42: (looks better imo than `>>` with the right font)

cs: Done ;)

kyazdani42 (approved):

PR #529, opened 2021-07-24
fix: guarantee order of buffer options

jose-elias-alvarez:
Hello! I've run into a small issue with nvim-tree caused by the fact that currently, the order in which buffer options are set isn't fixed because of `pairs`. Specifically, setting `filetype` before `buftype` can cause issues for `FileType` event listeners that rely on checking `buftype`. This small PR switches to `ipairs` to make sure that the options are set in a predictable order. Let me know if this is OK!

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -35,11 +35,11 @@ M.View = {
     }, ',')
   },
   bufopts = {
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
lgtm ;)

PR #533, opened 2021-07-29
Added new option, respect_buf_cwd

ucchiee:
Added a new option that respect the current buffer's cwd. Nvim-tree changes its cwd to that of buffer's when this option is enabled and nvim-tree is opened.

This ensures that the cwd of nvim-tree is same as that of buffer's.


ucchiee:
For those who want to use this option:

Install from my repo
```
Plug 'ucchiee/nvim-tree.lua', {'branch': 'ucchiee/feature/option/respect_buf_cwd'}
```
and enable the option
```
let g:nvim_tree_respect_buf_cwd = 1
```

kyazdani42 (changes requested):
missing a doc description :)

ucchiee:
Thank you, updated the doc.

kyazdani42 (approved):

PR #539, opened 2021-07-30
Feature: open file/folder with default system application

xeluxee:
This feature is equivalent to "reveal" in nerdtree or "execute_system" in defx.

Open a file with default system application by pressing `s`
Open a folder with system file manager by pressing `s`

Tested on Linux.
Should work also on MacOS, Windows, Windows Cygwin and systems using `xdg-open`, like FreeBSD.

PR #543, opened 2021-08-04
open system application in background

mortezadadgar:
so nvim will detach itself from system process.

this was missed from #539.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -110,7 +110,7 @@ local keypress_funcs = {
       return
     end
 
-    local command_output = vim.fn.system(system_command .. '"' .. vim.fn.substitute(node.absolute_path, '"', '\\\\"', 'g') .. '"')
+    local command_output = vim.fn.system(system_command .. '"' .. vim.fn.substitute(node.absolute_path, '"', '\\\\"', 'g') .. '" &')
 
     if vim.v.shell_error ~= 0 then
       vim.cmd('echohl ErrorMsg')


xeluxee:
Hi
First of all your change would make this feature incompatible with some operating systems, because `&` only works on Unix-like systems.
Then I can't understand the need of detaching, because the launchers used in #539 (`xdg-open`, `open`, `start ""`, `cygstart`) automatically detach the default system application after opening a file.

mortezadadgar:
> Hi
> First of all your change would make this feature incompatible with some operating systems, because `&` only works on Unix-like systems.
> Then I can't understand the need of detaching, because the launchers used in #539 (`xdg-open`, `open`, `start ""`, `cygstart`) automatically detach the default system application after opening a file.

in my case xdg-open froze nvim after opening a system app not running in background like opening a html file with firefox
and I didn't know "&" is not compatible with every operation system, closing now.

xeluxee:
Sorry I can't reproduce this, if I press `s` on a html file it is opened with Firefox and everything goes fine.
Is this happening only with html files and Firefox?

mortezadadgar:
> Sorry I can't reproduce this, if I press `s` on a html file it is opened with Firefox and everything goes fine.
> Is this happening only with html files and Firefox?

yeah I can also reproduce with sxiv

xeluxee:
@mortezadadgar please try out `feat/system_open` branch. It works asynchronously now.
https://github.com/xeluxee/nvim-tree.lua/tree/feat/system_open
Tell me what happens and if everything goes fine I'll make a PR.

mortezadadgar:
> @mortezadadgar please try out `feat/system_open` branch. It works asynchronously now.
> https://github.com/xeluxee/nvim-tree.lua/tree/feat/system_open
> Tell me what happens and if everything goes fine I'll make a PR.

Yes works perfectly!

PR #545, opened 2021-08-04
Feature: add the possibility to not focus tree on TabEnter when g:nvim_tree_tab_open=1

marene:
Following me opening https://github.com/kyazdani42/nvim-tree.lua/issues/544, I had a look inside the code (lua seems very cool), and tried the following thing to allow to not autofocus the tree when `g:nvim_tree_tab_open` is set to 1 and a new tab is opened.

I added an `option.focus_tree` (that defaults to true to keep default behaviour of the method) to `open()` in the view module that, if set to `false`, will trigger a `wincmd p` after the tree's window has been opened.

Then in `tab_change()` (I checked: `tab_change` is _only_ call on the `TabEnter` autocmd),  `open` is called with the `focus_tree` option set to false.

This way when auto-opening the tree in new tabs, the document of the new tab is automatically focused instead of the tree.

If you want to make that behaviour parametrised, no problem, I'll add an option to it!

I hope this contribution is valuable!

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -46,7 +46,7 @@ function M.tab_change()
       if bufname:match("Neogit") ~= nil or bufname:match("--graph") ~= nil then
         return
       end
-      view.open()
+      view.open({ focus_tree = false })
     end
   end)
 end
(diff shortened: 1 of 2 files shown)


kyazdani42 (approved):
LGTM, i don't feel strongly about that because i don't use tabs, if users request for an option we will add it.

PR #551, opened 2021-08-09
system_open: use asynchronous vim.loop to spawn process + let user customize the command to be used to open a file/folder with default system application

xeluxee:
This is an improvement of #539
Related: #543

Added 2 new variables that are automatically set if not edited by user:
- `g:nvim_tree_system_open_command`: a string containing the command to be used to open a file/folder with default system application
- `g:nvim_tree_system_open_command_args`: array of strings containing arguments to be passed to the command

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -90,7 +90,7 @@ local keypress_funcs = {
     return lib.open_file('preview', node.absolute_path)
   end,
   system_open = function(node)
-    if vim.fn.exists("g:nvim_tree_system_open_command") == 0 then
+    if vim.g.nvim_tree_system_open_command == nil then
       if vim.fn.has('win32') == 1 or vim.fn.has('win32unix') == 1 then
         vim.g.nvim_tree_system_open_command = 'cmd'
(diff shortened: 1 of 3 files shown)


lua/nvim-tree.lua
```diff
@@ -90,34 +90,47 @@ local keypress_funcs = {
     return lib.open_file('preview', node.absolute_path)
   end,
   system_open = function(node)
-    local system_command
-    if vim.fn.has('win16') == 1 or vim.fn.has('win32') == 1 or vim.fn.has('win64') == 1 then
-      system_command = 'start "" '
-    elseif vim.fn.has('win32unix') == 1 then
-      if vim.fn.stridx(vim.fn.system('uname'), 'CYGWIN') ~= -1 then
-        system_command = 'cygstart '
+    if vim.fn.exists("g:nvim_tree_system_open_command") == 0 then
```
kyazdani42, line 93: `vim.g.nvim_tree_system_open_command ~= nil`

xeluxee: Fixed

kyazdani42 (approved):

Issue #567, opened 2021-08-16
unexpected behaviour of `nvim_tree_respect_buf_cwd`


Frefreak:
Hi, not sure if its an issue of misunderstanding. From the description I assume if I open a new file in other directory `a/b/c/file`, then with this option enabled when nvim tree is opened the folder should be set to `a/b/c`? However it seems currently its not the case:
![Peek 2021-08-16 17-44](https://user-images.githubusercontent.com/7122156/129544541-ed3ccada-546b-4474-abf8-10f9a93255ee.gif)

From the code I see its using `getcwd`, not something like `expand('%:p:h')`, so I'm not sure if current behavior is expected or not.

kyazdani42:
is this fixed with #583 ?

Frefreak:
It seems the behavior is still not what I thought how it would be. It now opens the tree with root set to cwd yes, but if I open a new buffer, the cwd would not be automatically set to the folder containing the file (which is correct behavior), so the tree root is not what I want.
I guess if currently its working as expected then I propose we can add a new command to support that (which is, a new `NvimTree<Something>` command that opens the tree with root set as the folder containing the file). I can try to help implement this if you think that's reasonable.

Frefreak:
> shouldn't the behavior be: whenever a buffer is entered, the tree cwd is set to it's containing folder ?
That's what I was thinking, and at the time I posted this issue it seems to behave differently.

However I just tried updating to the master and now the behavior seems to be what I expect. Can you please confirm this and if it is ok we can close this issue :smile:

Frefreak:
> > shouldn't the behavior be: whenever a buffer is entered, the tree cwd is set to it's containing folder ?
> > That's what I was thinking, and at the time I posted this issue it seems to behave differently.
> 
> However I just tried updating to the master and now the behavior seems to be what I expect. Can you please confirm this and if it is ok we can close this issue smile

kyazdani42:
closing this in the meantime

PR #583, opened 2021-08-21
fix: respect_buf_cwd #567

nguyenvukhang:
moved the command to get cwd to before opening the file tree.
fixes execution of this config:
```
let g:nvim_tree_respect_buf_cwd = 1
```


kyazdani42 (approved):
indeed, thanks

PR #589, opened 2021-08-25
Add temporary lua config function

taybart:
I know changes are in the works, it would be nice to add this in so we can use a lua config now.

This is in reference to: https://github.com/kyazdani42/nvim-tree.lua/issues/557

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,97 @@ set termguicolors " this variable must be enabled for colors to be applied prope
 highlight NvimTreeFolderIcon guibg=blue
 ```
 
+### Using the setup function
+
+```lua
+require('nvim-tree').setup{
+  side = 'right', --left by default
+  width = 40, --30 by default, can be width_in_columns or 'width_in_percent%'
+  ignore = { '.git', 'node_modules', '.cache' }, --empty by default
+  gitignore = 1, --0 by default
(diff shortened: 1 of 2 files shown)


kyazdani42:
Hi, thanks for the contribution. I wont merge this because the setup refactoring should come soon enough, but i'll leave this open in case some people want to use it.

taybart:
Sounds good! I’ll close when those updates land.

taybart:
I think this was a fun "lua learning" experience for me. Here is the solution that was originally suggested that I will move to:

```lua
  use {
    'kyazdani42/nvim-tree.lua',
    requires = { 'kyazdani42/nvim-web-devicons' },
    cmd = {'NvimTreeToggle', 'NvimTreeFindFile'},
    setup = function()
      local c = {
        width = '30%',
        auto_close = true,
        hide_dotfiles = true,
        group_empty = true,
        highlight_opened_files = true,
        auto_resize = false,
        window_picker_exclude = {
          filetype = { 'packer' },
        },
      }
      for opt, value in pairs(c) do
        if type(value) == 'boolean' then
          value = value and 1 or 0
        end
        vim.g['nvim_tree_' .. opt] = value
      end
    end
  }
```

PR #597, opened 2021-08-28
Add augroup to view setup

Krafi2:
One of the recent commits has introduced a few autocommands inside the setup function for view. The exclamation mark after the commands resets all user autocommands, so I've placed them inside an augroup to prevent this from happening.


kyazdani42 (approved):
good catch :)

Issue #515, opened 2021-07-16
Invalid window id when trying to open nvimtree on same file multiple times or sometimes on the first time.


devins2518:
Issue can be seen [here](https://0x0.st/-OFs.png). This sometimes happens right after I open a file, or it works for a few times and then does this. Not sure what's causing this since my config doesn't seem to be too different from some of the premade ones. The part of my config:
```lua
local g = vim.g

g["nvim_tree_add_trailing"] = 1
g["nvim_tree_allow_resize"] = 1
g["nvim_tree_auto_close"] = 0
g["nvim_tree_auto_open"] = 0
g["nvim_tree_disable_default_keybindings"] = 1
g["nvim_tree_follow"] = 1
g["nvim_tree_git_hl"] = 1
g["nvim_tree_group_empty"] = 1
g["nvim_tree_hide_dotfiles"] = 1
g["nvim_tree_ignore"] = {
    ".git", "target", "node_modules", ".cache", "Cargo.lock"
}
g["nvim_tree_indent_markers"] = 1
g["nvim_tree_quit_on_open"] = 1
g["nvim_tree_root_folder_modifier"] = ":~"
g["nvim_tree_show_icons"] = {git = 1, folders = 1, files = 1}
g["nvim_tree_side"] = "left"
g["nvim_tree_tab_open"] = 1
g["nvim_tree_update_cwd"] = 1
g["nvim_tree_width"] = 25
g["nvim_tree_width_allow_resize"] = 1
g["nvim_tree_icons"] = {
    default = " ",
    symlink = " ",
    git = {
        unstaged = '✗',
        staged = "✓",
        unmerged = "",
        renamed = "➜",
        untracked = "★"
    },
    folder = {
        default = '',
        open = '',
        empty = '',
        empty_open = '',
        symlink = '',
        symlink_open = ''
    }
}

-- Mappings for nvimtree
local tree_cb = require'nvim-tree.config'.nvim_tree_callback

g["nvim_tree_bindings"] = {
    {key = {"<CR>", "o", "<2-LeftMouse>"}, cb = tree_cb("edit")},
    {key = {"<2-RightMouse>", "c"}, cb = tree_cb("cd")},
    {key = "vs", cb = tree_cb("vsplit")}, {key = "sp", cb = tree_cb("split")},
    {key = "tn", cb = tree_cb("tabnew")},
    {key = {"<BS>", "<S-CR>"}, cb = tree_cb("close_node")},
[...]

Issue #184, opened 2021-01-18
Not opening in repository root


gegoune:
Have noticed that tree does not open in repository's root in case such as:

1. `~/ $ nvim` 
  a. open file with `Startify` that is part of git repository
  b. invoke `:NvimTreeFindFile` — tree is opened with root still being `~` (I use vim `vim-rooter` to change nvim's `pwd`)
  c. close tree buffer with `q` (or any other way)
  d. invoke `:NvimTreeFindFile` again — tree correctly opened in buffer's repository root directory

1.a correctly changes nvim's directory; `:pwd` return repository's root directory for opened file so `vim-rooter` does its job.

Other than that - the best tree plugin there is! Thanks!

kyazdani42:
thanks :)
well yes i've disabled git-rooter because it conflicted with this and other plugins too, but i know cwd startup is a bit wonky, a refactoring is definitely needed here.

cloggier:
I have noticed while making your plugin lazy load that it would not work at all and had to put

```lua
vim.cmd "doautocmd NvimTree BufEnter *"
```
into packer's `config` function for *nvim-tree*. With that in place I had no problem with tree opening in repository root. Perhaps that info will help you somewhat.

Thanks for all your work on that plugin and treesitter!

kyazdani42:
your welcome :) well lazy loading is not working because of the way i setup nvim tree. I have to rework the setup tu be more lua like which would probably allow lazy loading

cloggier:
With what you are doing in `plugin/` I don't really need to lazy load it at all, just tried ink the past and came over this 'accidental fix' for root issue.

maxigaz:
`:NvimTreeToggle` still opens in my home folder the first time; I have to close and reopen the file viewer to make it open the directory the file is at. (This happens when I'm trying to open a session from Startify, with buffers inside `~/.config/nvim`.)

@cloggier I've tried your workaround, but I get the following error:

```
Error running config for nvim-tree.lua
Error in packer_compiled: Vim(lua):E5108: Error executing lua [string ":lua"]:52: [string "..."]:0: Vim(doautocmd):E216: No such group or event: NvimTree BufEnter *
Please check your config for correctness
```

Here's how I've set it up using Packer:

```
use {
  'kyazdani42/nvim-tree.lua',
  requires = {'kyazdani42/nvim-web-devicons', opt = true},
  config = function()
    vim.cmd([[doautocmd NvimTree BufEnter *]])
[...]

Issue #93, opened 2020-08-23
Slows neovim on startup


carlitux:
Package manager: packer.nvim

Using as a start plugin slows the first rendering, adding as a lazy plugin using as opt plugin do not render anything

![Screenshot from 2020-08-23 15-45-22](https://user-images.githubusercontent.com/40538/90987272-af0cf980-e557-11ea-991b-b287af0b3d9a.png)
.

kyazdani42:
hello,
I do not know what lazy rendering is (i basically just add my plugin to my runtimepath manually so...), i've never had slow startups especially compared to other trees, this tree is MUCH faster and doesn't make the whole ide run slowly. I tested on a 40K file directory and well, it just works under 1sec, which is like 10 times faster than other tree implementations (maybe not all of them but feel free to test them out :) ). I also use it everyday for work.
So please describe your issue in a better way

carlitux:
@kyazdani42 yes, it is faster but startup is slow as you can see in log files, with the plugin it is 3x slower than without it.

[vim-without-lua-tree.log](https://github.com/kyazdani42/nvim-tree.lua/files/5147066/vim-without-lua-tree.log)
[vim-with-lua-tree.log](https://github.com/kyazdani42/nvim-tree.lua/files/5147067/vim-with-lua-tree.log)

Sorry if I explained wrong but I didn't talk about lazy rendering, I mean that when package (**:help packages**) feature is used where **nvim-tree.lua** is put in **opt** file and running this commands:

```vim
packadd nvim-tree.lua
LuaTreeOpen
```

Do not work, as the image I sent before.

akinsho:
I think some of this might relate to how `packages` handles lua plugins, I'm no expert though. I profile my vim startup a fair bit and lua tree is never anywhere near the top :shrug:.

kyazdani42:
I'm not using this feature, and i don't have much time lately trying to fix other people issues :/
If you could find a way to fix this, open a PR, i would be happy to accept it if it makes other people happy :)

comfortablynick:
I also attempt to lazy load as many plugins as possible using the `opt` directory in `&packpath`. This plugin initially worked doing that, but then stopped. I think I figured out why: it now uses the `VimEnter` event to initialize the plugin..

https://github.com/kyazdani42/nvim-tree.lua/blob/640d147d706aa33f096f5585176fd3c76303377b/plugin/tree.vim#L17

[...]

PR #603, opened 2021-08-31
chore: refacto setup part 1

kyazdani42:
refacto setup for code entrypoint
following options switched to boolean values as options to the setup function:
- `nvim_tree_disable_netrw` -> `disable_netrw`
- `nvim_tree_hijack_netrw` -> `hijack_netrw`
- `nvim_tree_auto_open` -> `open_on_setup`
- `nvim_tree_auto_close` -> `auto_close`
- `nvim_tree_tab_open` -> `tab_open`
- `nvim-tree-update-cwd` -> `update_cwd`
- `nvim_tree_hijack_cursor` -> `hijack_cursor`
- `nvim_tree_system_open_command` -> `system_open.cmd`
- `nvim_tree_system_open_command_args` -> `system_open.args`
- `nvim_tree_follow` -> `update_focused_file.enable`
- `nvim_tree_follow_update_path` -> `update_focused_file.update_cwd`
Also added new option `update_focused_file.ignore_list` which will
ignore filepath or filetypes that matches one entry of the list when
updating the path if update_cwd is true.

After a few try of refactoring the setup in one go, i decided to move options bit by bit to the setup because it was too too complexe to do it in one go.
I'll do a step by step refactoring to remove all the `g:` options.

I'll leave this open for a week or so, so user can test this before merging to make sure nothing was broken.

This PR should enable lazy loading. All the others `g:` options should be set _BEFORE_ running `nvim-tree.setup`

## TODO:
- [x] update documentation (missing docs file)
- [x] fix the `man` hijack issue.
- [x] write a deprecation function for options in the runtime

should fix: #515 #184 #93 
and maybe more but i'm not sure


gegoune:
Have noticed some slow downs, running with

```lua
tree.setup {
  auto_close = true,
  update_focused_file = {
    enable = true,
  },
}
```

gives rally slow startup times:

```
    191.036  000.164  000.164: sourcing /Users/<me>/.config/nvim/after/plugin/nvim-notify.lua
    310.851  119.705  119.705: sourcing /Users/<me>/.config/nvim/after/plugin/nvim-tree.lua
    317.488  006.489  006.489: sourcing /Users/<me>/.config/nvim/after/plugin/nvim-treesitter.lua
 ```

The same file with `config()` commented out:

```
255.593  046.949  046.949: sourcing /Users/<me>/.config/nvim/after/plugin/nvim-tree.lua
```
(which is still super long) but it does not properly initialise nvim-tree on that branch. So setup is mandatory now I assume?

With removed setup those are the lines that contribute to those ~50ms:

```lua
local tree = require 'nvim-tree'
local tree_callback = require('nvim-tree.config').nvim_tree_callback
```

kyazdani42:
@gegoune could not reproduce the slowdown, on basic folders (10 to 500 folder/files), it opens in under 20ms in my machine (i debugged the setup part specifically with `vim.loop.hrtime` and `vim.loop.now` to be sure, i've managed to get it running to 80ms to 100ms in /usr/lib which contains ~4000 entries, and it looks even faster than running `exa` so i'm suprised it's running so slow on your machine. Maybe this comes from some place else in the code ?

gegoune:
Just to be sure, I am not talking about opening nvim-tree window but time it takes to run `setup` + couple of those requires when opening nvim. I have it set up in `after/plugin/nvim-tree.lua` file.

kyazdani42:
i understand what you said before, but i could definitely not reproduce, on my machine it's really fast, opening the tree or not during setup does not change thing much, because well:
```
setup took 15.944164ms
rendering took 0.222216ms
rendering took 0.155439ms
rendering took 0.301504ms
rendering took 0.760211ms
rendering took 0.412551ms
rendering took 0.716118ms
rendering took 0.367504ms
```
not sure where this happens, i'm going to make some more debugging

kyazdani42:
@gegoune could you try [this profiler](https://github.com/norcalli/profiler.nvim) to see if nvim-tree really is that slow compared to other plugins ? If so, could you send me your complete config for nvim-tree for the setup ?

gegoune:
I will try with the profiler in a minute, although I have no idea how to use it.

From your output, isn't 15ms still quite a lot? Does nvim-tree actually retrieves directory's content on setup? I don't think it should do anything other than configure itself and only read directory when its window gets opened.

kyazdani42:
it does initialize the tree with basic content. This code is quite old :smile:

doc/nvim-tree-lua.txt
```diff
@@ -58,6 +58,107 @@ Print clipboard content for both cut and copy
...
+Here is a list of the options available in the setup call:
+- |disable_netrw|: completely disable netrw
+  type: `boolean`
+  default: `true`
+
+- |hijack_netrw|: hijack netrw windows (overriden if |disable_netrw| is `true`)
+  type: `boolean`
+  default: `true`
+
+- |open_on_setup|: will automatically the tree when running setup if current
```
gegoune, line 101: Missed 'open' after 'automatically'.

doc/nvim-tree-lua.txt
```diff
@@ -58,6 +58,107 @@ Print clipboard content for both cut and copy
...
+  open. You can use this option if you don't want the tree to open in some
+  scenarios (eg using vim startify).
+  type: `{string}`
+  default: `{}`
+
+- |auto_close|: force closing neovim when the tree is the last window in the view.
+  type: `boolean`
+  default: `false`
+
+- |tab_open|: opens the tree automatically when switching tabpage or opening a new
```
gegoune, line 116: `open_on_setup` and `tab_open` are very different, perhaps they should both start with `open_` and for example be like `open_on_startup|setup` and `open_on_tab`?

kyazdani42: that's what i told mysefl while writing the docs, thanks for confirming this!

gegoune: or even
```lua
auto_open = {
  startup = bool,
  tab = bool,
}
```
?

gegoune:
Have also noticed that nvim-tree doesn't change directory (`:pwd`) to git repository. It used to work very nicely with [ygm2/rooter.nvim](https://github.com/ygm2/rooter.nvim) and although rooter changes pwd correctly it is not reflected in nvim-tree. All directories all the way to opened file fill be unfold, but root of tree will be my home directory rather than repository's root.

Does not work properly:
1. Open nvim in `~`: `nvim`; without any arguments.
2. Use [ibhagwan/fzf-lua](https://github.com/ibhagwan/fzf-lua) to open file with MRU (oldfiles).

File opens, but after opening nvim-tree its root is still in `~`

Does work properly:
1. Open nvim in `~`: `nvim file/in/first/repo repo2/some/other/file/`.
2. Open nvim-tree and switch between both buffers.

nvim-tree's root nicely updates to repositories root. Works as expected.

@ibhagwan, do you think it might be caused by fzf-lua somehow? Although if I open another file using fzf-lua's oldfiles after second scenario then directory in nvim-tree updates nicely.

<details>
  <summary>after/plugin/nvim-tree.lua</summary>
  
```lua
local tree = require 'nvim-tree'
local tree_callback = require('nvim-tree.config').nvim_tree_callback

vim.g.nvim_tree_bindings = {
  { key = '<C-s>', cb = tree_callback 'split' },
  { key = '<C-x>', cb = '' },
  { key = 'h', cb = tree_callback 'close_node' },
  { key = { '<CR>', 'l' }, cb = tree_callback 'edit' },
}

require('nvim-tree.view').View.winopts.cursorline = true

tree.setup {
  auto_close = true,
  update_cwd = true,
  update_focused_file = {
    enable = true,
    update_cwd = true, -- ?
  },
}
```
</details>

<details>
  <summary>plugin/nvim-tree.lua</summary>
  
```lua
local g = vim.g
local bar = '▍'

-- g.nvim_tree_auto_close = 1
-- g.nvim_tree_follow = 1
g.nvim_tree_git_hl = 1
g.nvim_tree_group_empty = 1
g.nvim_tree_highlight_opened_files = 2
-- g.nvim_tree_hijack_cursor = 0
g.nvim_tree_icons = {
  git = {
    unstaged = 'M',
    staged = 'S',
    unmerged = 'U',
    renamed = 'R',
    untracked = '?',
    deleted = 'D',
    ignored = 'I',
  },
  folder = {
    default = '+',
    open = '-',
    empty = '+',
    empty_open = '-',
    symlink = '+',
    symlink_open = '-',
  },
  lsp = {
    error = bar,
    warning = bar,
    info = bar,
    hint = bar,
  },
}
g.nvim_tree_ignore = { '.git', 'node_modules', '.cache' }
g.nvim_tree_lsp_diagnostics = 1
g.nvim_tree_root_folder_modifier = ':t:r'
g.nvim_tree_show_icons = {
  files = 0,
  folders = 1,
  git = 1,
}
g.nvim_tree_special_files = {}
g.nvim_tree_width = 40
```
</details>

gegoune:
@kyazdani42 I have also tried to get profiler working but couldn't. Do you mind showing me how to do that when you get a sec, please?

ibhagwan:
> @ibhagwan, do you think it might be caused by fzf-lua somehow? Although if I open another file using fzf-lua's oldfiles after second scenario then directory in nvim-tree updates nicely.

@gegoune, I don't think so, all `fzf-lua` does when you open a selected file is `:e <filepath>`, why don't you try your first scenario's 2nd step as `:e <a recently edited file outside of your cwd>` and see if you can reproduce the issue without `fzf-lua`?

gegoune:
Ah, of course, didn't think of testing it like that. Case you described above does not update nvim-tree's root, so it's not fzf-lua's problem, forgive me for mentioning you for no good reason.

doc/nvim-tree-lua.txt
kyazdani42: they are not really related. auto_open will open on startup automatically, tab will only keep the window open as if it was the same screen. I guess i'd like to rename the tab_open one because it's not really explicit.

kyazdani42:
@gegoune maybe this is caused by `vim.cmd "au DirChanged * lua require'nvim-tree.lib'.change_dir(vim.loop.cwd())"` ? what does `vim.loop.cwd` shows when you change dir ? I'm not sure vim.loop follows internal vim path.

gegoune:
Oh, maybe, do you see anything wrong with their logic?

Just checked `vim.loop.cwd` - it returns git repository's root dir, same as `:pwd`.

kyazdani42:
just did some debugging, the `DirChanged` event does not seem to be called at all and i'm not sure why, although vim directory was changed as expected. Not sure if this is a bug within neovim or rooter.nvim. I'm going to check wether other dirchanged event are well sent to the other plugins i have and i'll be back here with more infos.

gegoune:
I have tried replacing rooter with [ahmedkhalf/project.nvim](https://github.com/ahmedkhalf/project.nvim) and it doesn't work in the same way so probably rooter isn't at blame here.

kyazdani42:
Did not understand your last comment. If you mean `it doesnt work in the same way`, does this mean `it fails in the same way` ?
Also just tried the DirChanged disabling nvim-tree, and it still does not trigger the event so i think it comes from something else.

Ok, just found out... the rooter.nvim only run `cd` when `g:loaded_tree` is 1. So no change dir events are set otherwise, i think it might be a plugin issue, a rooter should change the base dir no matter what.
As for project.nvim i'm not sure.
Did you try https://github.com/airblade/vim-rooter ?

gegoune:
Yes, sorry, meant exactly what you assumed, that it fails the same way.

Didn't try vim-rooter, might have some time later tonight.

But if rooter.nvim requires that var and project.nvim also doesn't work does it mean that there is some other issue at play here since rooter.ncim has no `loaded_tree` string anywhere in its code.

kyazdani42:
ok so, tested with a very minimal init.lua:
```lua
vim.cmd "set rtp+=~/project.nvim"
require'project_nvim'.setup {
  manual_mode = false,
  patterns = { ".git", "_darcs", ".hg", ".bzr", ".svn", "Makefile", "package.json" },
  silent_chdir = false,
}

vim.cmd 'au DirChanged * lua print(2)'
```
and it does change the directory as expected, but does not trigger the `DirChanged` event. I think it will be the same for every plugin then. This is definitely an `neovim` issue.

kyazdani42:
or maybe because autocmd cannot be nested ?

ahmedkhalf:
Fixed, sorry for the long wait.

PR #613, opened 2021-09-06
feat(config): add notify to nvim_tree_window_picker_exclude

gegoune:
With https://github.com/rcarriga/nvim-notify/ becoming more and more popular it's likely that issue like https://github.com/rcarriga/nvim-notify/issues/18 will pop up more often. I therefore propose adding that plugin's filetype to default `nvim_tree_window_picker_exclude`.


kyazdani42 (approved):
good idea

Issue #622, opened 2021-09-16
Make new files stay in the parent folder, when the folder on the cursor is closed


booperlv:
Creating a file here:
```
parent
  folder (cursor on this line)
  file1
  file2
```
results in
```
parent
  folder (cursor on this line)
  file1
  file2
  created_file
```

but when the folder is opened:
```
parent
  folder (cursor on this line)
    folder_file1
    folder_file2
  file1
  file2
```
creating a file would result in:
```
parent
  folder (cursor on this line)
    folder_file1
    folder_file2
    created_file
  file1
  file2
```

Already created something that implements this same functionality:
But before I open a pull request, what should the option be called?

booperlv:
Feature added #628 :)

PR #628, opened 2021-09-18
feat: option to set path destination to parent folder when cursor is on a closed folder while creating files

booperlv:
a more thorough description of the feature in #622 
Seems like this is the first instance of an option for the `fs.lua` module, so I decided to go with directly adding the variable there.
This was the only name I could think of that made sense, but I'm not sure if it makes _enough_ sense, so please do help me think of a better name if so.

Please do tell me if there's anything that I should change! Thanks :)

### DIFF
--- a/lua/nvim-tree/fs.lua
+++ b/lua/nvim-tree/fs.lua
@@ -57,15 +57,14 @@ function M.create(node)
     }
   end
 
+  local node_is_open = (
+    vim.g.nvim_tree_create_in_closed_folder == 0
+    and node.open or false
+  )
+
   local add_into
-  if node.entries ~= nil then
+  if node.entries ~= nil and node_is_open then
     add_into = utils.path_add_trailing(node.absolute_path)
-    local create_in_closed_folder = vim.g.nvim_tree_create_in_closed_folder or 1
-    if create_in_closed_folder == 0 then
-      if node.open == false then
-        add_into = node.absolute_path:sub(0, -(#node.name + 1))
-      end
-    end
   else
     add_into = node.absolute_path:sub(0, -(#node.name + 1))
   end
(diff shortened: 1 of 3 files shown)


CKolkey:
I've had something similar in a fork for a while:

```lua
  local add_into
  if node.entries ~= nil and node.open then
    add_into = utils.path_add_trailing(node.absolute_path)
  else
    add_into = node.absolute_path:sub(0, -(#node.name + 1))
  end
```

booperlv:
> I've had something similar in a fork for a while:
> 
> ```lua
>   local add_into
>   if node.entries ~= nil and node.open then
>     add_into = utils.path_add_trailing(node.absolute_path)
>   else
>     add_into = node.absolute_path:sub(0, -(#node.name + 1))
>   end
> ```

This does seem like a more robust solution! I'll change it accordingly :)

lua/nvim-tree/fs.lua
```diff
@@ -57,8 +57,13 @@ function M.create(node)
     }
   end
 
+  local node_is_open = (
+    vim.g.nvim_tree_create_in_closed_folder == 0
+    and node.open or false
```
kyazdani42, line 62: this feature is not working and breaks default behavior.
`local node_is_open = vim.g.nvim_tree_create_in_closed_folder == 1 or node.open` should be more appropriate if i understand the feature correctly.
With you condition, setting `0` or `1` for `nvim_tree_create_in_closed_folder` does not change anything and always creates the file outside of the folder.

kyazdani42 (changes requested):

lua/nvim-tree/fs.lua
booperlv: Apologies, sleep deprivation has gotten the better of me again. Pushing a fix in a few!

kyazdani42 (approved):

Issue #632, opened 2021-09-24
The tree always renders the icon padding, even when the icon is defined as an empty string


saviocmc:
First of all, thank you very much for your great work, this plugin is awesome!

My issue is that even when I set the icon to be an empty string, like the example below (lua config), the tree still renders the icon padding.
```
    git = {
        unstaged = '•',
        staged = '•',
        unmerged = '',
        renamed = '➜',
        untracked = '',
        deleted = '',
        ignored = '',
    },
```
![2021-09-23-23-32-31](https://user-images.githubusercontent.com/9142154/134609562-6a49b504-1f5b-42d3-9fbb-6da0aabb4de5.png)

For the example above, the ```deleted``` and ```ignored``` are set as empty string, but the padding still renders.
The ```tags``` file is ignored by .gitignore, and the icon padding is rendered, as the image shows, making it misaligned with the other files.
In my opinion, the expected behavior should be to act as there were no icon, and skip the icon rendering altogether.

PR #633, opened 2021-09-24
Only renders the git icon if the icon is not empty

saviocmc:
This is mainly to avoid rendering a padding for a non-existent icon
making the nodes misaligned for no apparent reason

Fix issue #632


kyazdani42 (approved):
indeed

PR #647, opened 2021-09-27
fix: help after refactoring view

qqshfox:


kyazdani42 (approved):
auto renaming missed this :smile: thanks for doing the lsp job :)

PR #649, opened 2021-09-27
fix: add missing icons for `default` and `symlink`

numToStr:
I think those are missing as the docs say that those icons are present by default but it isn't. Below are the icons
image


gegoune:
@kyazdani42 you have fixed an issue where `default` had to be set to `''` to get rid of that icon when no icons are set to be displayed at all, but symlink icons still get displayed. Would you want me to open an issue for that?

kyazdani42:
shouldn't you set `''` for the symlink icon too ? I don't see icons for symlink icons on my end

gegoune:
I though https://github.com/kyazdani42/nvim-tree.lua/blob/29e5b754b7250aded609e8c071804053db5765c6/README.md#L148-L151 meant that by default it's empty? I am not setting it to anything.

Issue #653, opened 2021-09-27
System open duplicates passed arguments and kills spawned processes on window close


piotr-machura:
When opening files with the `system_open` program all of the **previously** open files are passed as arguments to the `cmd`.
Eg. If I press `s` on `fileA` it will open, and if I press it again the file will open **twice** (three times if I press it again etc.)
I have tested it using a `myopen` script:
```bash
#!/bin/bash
echo "$@" >> ~/args.txt
```
Drop that in your `$PATH` and set the options like so:
```lua
require('nvim-tree').setup {
    system_open = { cmd  = 'myopen', args = {} },
}
```
Try opening a couple of files and you will see the list of args "grow", with the file names appended instead of being the only passed argument.

On another note (I believe this is related, or at least also originates [here](https://github.com/kyazdani42/nvim-tree.lua/blob/1b746744ad04c14d2cf260641750f77ec5f62635/lua/nvim-tree.lua#L112)), the spawned proceses seem to get killed on terminal window close, even if the opening program disowns them.

This can be reproduced with another script `myopen2`:
```bash
#!/bin/bash
nohup xdg-open "$1" &>/dev/null & disown # you may replace xdg-open $1 with any GUI program
```
If I invoke this script manually with `myopen2 somefile.pdf` and then close the terminal window the spawned pdf viewer remains alive (as expected). But if I set it as system open:
```lua
require('nvim-tree').setup {
    system_open = { cmd  = 'myopen2', args = {} },
}
```
then open somefile.pdf by clicking `s` in the tree and close the terminal window (the one with neovim in it) the pdf viewer gets killed. Stranglely, if I exit neovim first (with a simple `:qa`) and *then* close terminal window the pdf viewer will *not* get killed.

**Edit:** the  pdf viewer process is not owned by the terminal/shell/neovim (`ps -l <pdf viewer PID>` shows the PPID is 1), so the disowns/nohups are working. What is happening here?

saviocmc:
@piotr-machura can you test if this solves the duplication issue for you?
https://github.com/kyazdani42/nvim-tree.lua/pull/655/files

piotr-machura:
PR #655 does indeed fix the duplication issue, thank you so much! The process ownership remains an issue though.

saviocmc:
My experience on linux using ```xdg-open``` is exactly the same as you reported for your ```myopen2```.
If I close the terminal **without** exiting nvim first, all spawned process are killed.
[...]

PR #655, opened 2021-09-28
fix 'system_open' keypress callback

saviocmc:
I found two issues in the ```system_open``` keypress callback (at least on linux).

The first one was easy, the functions was using an undefined flag ```_config.is_linux``` (that was always returning false) instead of the correct flag ```_config.is_unix```.

The second issue I didn't quite understand yet. From the 'xdg-open' error message I manage to realize that the table ```process.args``` was persisting its state throughout the callback calls, and each time the current file path was being appended to it, causing errors after the first call.

I don't know why this table is not redefined after each call, but I could solve the issue removing the file path from the end of the table right after spawn the process.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -118,7 +118,7 @@ local keypress_funcs = {
         }
       elseif _config.is_macos then
         _config.system_open.cmd = 'open'
-      elseif _config.is_linux then
+      elseif _config.is_unix then
         _config.system_open.cmd = 'xdg-open'
       else
         require'nvim-tree.utils'.echo_warning("Cannot open file with system application. Unrecognized platform.")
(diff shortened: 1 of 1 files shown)


gegoune:
Related #653

lua/nvim-tree.lua
```diff
@@ -145,6 +145,7 @@ local keypress_funcs = {
         end
       end
     )
+    table.remove(process.args)
```
kyazdani42, line 148: i think the process table being used inside the callback is some kind of memory leak maybe, at least i think the GC is not properly cleaning up because of this. Thanks for the fixes though, i do not use this feature myself so couldn't see the issues :)

kyazdani42 (approved):

Issue #663, opened 2021-09-29
`autocmd BufReadPost * lua require'nvim-tree'.refresh()` causes error


gavynriebau:
### Steps to reproduce:

Add the following autocommand to your init.vim / init.lua file:

```vim
autocmd BufReadPost * lua require'nvim-tree'.refresh()
```

Start neovim with an the name of a file to open immediately, e.g.

```bash
$ nvim ~/.config/init.lua
```

### Expected result

The file should open without errors.

### Actual result

The following error will be printed when vim starts:

```
Error detected while processing BufReadPost Autocommands for "*":
E5108: Error executing lua ...ck/packer/start/nvim-tree.lua/lua/nvim-tree/populate.lua:162: bad argument #1 to 'fs_scandir' (string expected, got nil)
Press ENTER or type command to continue
```

### Background Info

I added the autocmd described above after finding that the nvim-tree's open file highlighting sometimes did not work (specifically when I opened a file from telescope plugin file finder).

The purpose of the autocmd is to automatically call `refresh()` when a new buffer has been opened and therefore ensure that the open file highlighting is correct.

After adding the autocmd I started seeing the error as described above.

kyazdani42:
should be fixed in `1.4.11` (c0cd3aeb657313), reopen if it's not the case :)

PR #664, opened 2021-09-29
fix: add check to prevent error on refresh entries (#663)

gavynriebau:
Fix for issue #663

### DIFF
--- a/lua/nvim-tree/populate.lua
+++ b/lua/nvim-tree/populate.lua
@@ -159,6 +159,10 @@ end
 local should_ignore = gen_ignore_check()
 
 function M.refresh_entries(entries, cwd, parent_node)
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/populate.lua
```diff
@@ -159,6 +159,10 @@ end
 local should_ignore = gen_ignore_check()
 
 function M.refresh_entries(entries, cwd, parent_node)
+  if cwd == nil then
```
kyazdani42, line 162: This is not right, the cwd should always be defined when calling this function. The patch should come a calling function, or a calling calling function, this will probably lead to undefined behavior.

kyazdani42:
closing this, i'll merge a proper patch for your issue

PR #708, opened 2021-10-08
Don't close nvim if current buffer is deleted

### LINKED ISSUE #703: Deleting a file that is the active buffer closes neovim
When deleting a file in the tree while having other files open, neovim closes if the deleted file is the current buffer.

I think this is related to `auto_close` but I am not sure.

image

Expected behavior: Switch to a different (potentially random) buffer

taybart:
Fixes #703 

Let me know if there is any refactor/optimization needed

Explanation by line:
```lua
local function clear_buffer(absolute_path)
  local bufs = vim.fn.getbufinfo({bufloaded = 1, buflisted = 1}) -- get loaded and listed bufs
  for _, buf in pairs(bufs) do
    if buf.name == absolute_path then -- this is the buffer we want
      if buf.hidden == 0 and #bufs > 1 then -- buffer visible and not alone
        local winnr = api.nvim_get_current_win() -- store old winnr
        api.nvim_set_current_win(buf.windows[1]) -- switch to win
        vim.cmd(':bn') -- switch to next buffer
        api.nvim_set_current_win(winnr) -- go back to original
      end
      vim.api.nvim_buf_delete(buf.bufnr, {}) -- delete the buffer in question
      return -- done
    end
  end
end
```

### DIFF
--- a/lua/nvim-tree/colors.lua
+++ b/lua/nvim-tree/colors.lua
@@ -79,7 +79,7 @@ local function get_links()
 end
 
 function M.setup()
-  if config.get_icon_state().show_file_icon then
+  if config.get_icon_state().show_file_icon and config.get_icon_state().has_devicons then
     require'nvim-web-devicons'.setup()
   end
   local higlight_groups = get_hl_groups()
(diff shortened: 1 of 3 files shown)


taybart:
@kyazdani42 I can use more `vim.api` functions, I thought `vim.fn.getbufinfo` was a pretty good one to use for this. I also see that the last api call has `vim` which it should probably not. Let me know if you want a refactor

PR #738, opened 2021-10-22
feat: if pick window is nil return -1 this way it will create a new one

alexnix:
what made me add this change was that using C-v would only allow me to pick one of the (max) two windows (other than the  tree), but i quite often work with 3-4 splits

this allows me to press any key that is not mapped to a window, and thus get a new split. i can easily just press Enter and i get a new split


kyazdani42:
@alexnix ping :)

Lopi-py:
@kyazdani42 I'm not alexnix, but I have the same issue, https://gyazo.com/3b7ae3f32f1936f98c79ba4845d48487 (still opens the file if I press `escape` in windows picker) I'm on Windows

alexnix:
now it should abort if the pressed key was Esc (asci 27)

kyazdani42:
@alexnix when a said escape, i was saying basically everything that could trigger the close (such a `C-c` also). Does `C-c` work ?
Also just tested your issue on current master, and it seems like when pressing any other key than the one shown by the picker, it still opens a window. What is the purpose of this PR exactly ?

Issue #549, opened 2021-08-07
Performance / Lag issues


kyazdani42:
Hello, please don't make yet another issue or comment an existing issue regarding `huge lag` in nvim-tree.

Please see [performance-issues](https://github.com/kyazdani42/nvim-tree.lua#performance-issues) in the README.

If your editor start to lag when using nvim-tree, it might be because of several things:
- you are using fish as an editor shell (which might be fixed in the future), try `set shell=/bin/bash` in your vim config.
- you are running in a git directory with lots of changes:
  - try running `git status --porcelain=v1 --ignored=matching -u` and see if it hangs for a while. If it does, try disabling the git integration all by itself:
```lua
-- this is lua code (in `init.lua`)
require'nvim-tree'.setup {
  git = {
    enable = false
  }
}
```

I have to disclaim that the git status handling will never be fast on huge directories (trying to run the command above with GIT_DIR set did take a few 16seconds on my machine).
The data returned is really heavy and slows down the editor just by itself (loaded 40 to 80mb just of git status text, which can represent 300 or more MB of data in ram which slows down the nvim process).

Usually, it works fine on most repositories, but will probably slow down sometimes.

Alternatively, each git process is now discarded after a configured timeout which is `400ms` by default. This is configurable in
```lua
require'nvim-tree'.setup {
  git = {
    enable = true,
    timeout = 400 -- (in ms)
  }
}
```

Thank you all for reading this.

yqlbu:
> @kyazdani42 works perfectly, it's lightning fast now, thank you!
> 
> One quick note: the commas are commented out so cause a syntax error
> 
> ```
> vim.g.nvim_tree_show_icons = {
>   git = 0,
>   folders = 1, -- or 0,
>   files = 1, -- or 0,
>   folder_arrows = 1 -- or 0
> }
> vim.g.nvim_tree_gitignore = 0
> vim.g.nvim_tree_git_hl = 0
> ```

After adding the configs above, it does solve the lagging issue.

pedropmedina:
I'm experiencing this issue following the latest update. The lag seems more prominent when writing the buffer.

Here's my setup:
```lua
local tree_cb = require("nvim-tree.config").nvim_tree_callback

vim.g.nvim_tree_indent_markers = 1
vim.g.nvim_tree_hide_dotfiles = 0
vim.g.nvim_tree_git_hl = 0
vim.g.nvim_tree_gitignore = 0
[...]

PR #743, opened 2021-10-24
feat/chore: rewrite git with job and some other fixes

### LINKED ISSUE #727: .gitignore regression
The gitignore issue I reported was fixed a while ago, but now it's back again. Please refer to my original issue on how to test:
https://github.com/kyazdani42/nvim-tree.lua/issues/412

There's some weird new behaviour too, btw. If you keep re-saving the same file, the backup file (ending with `~`) will _sometimes_ disappear after the 3rd save...

kyazdani42:
This PR is rewriting the git integration to avoid blocking the whole UI with `vim.fn.systemlist` on huge projects. It will now run in a separate process, and will be disabled if it takes more time than than the `timeout` configuration.

It also includes some fixes around refreshing and such things, which should be better.
It also fixes #727 and maybe more.
Also addresses partially #549


nyngwang:
Thanks for your work!

PR #747, opened 2021-10-25
Add "Open buffers only mode"

### LINKED ISSUE #744: Add view to show open buffers only
I have this idea I'd like to implement, which is to have a file explorer which can be toggled between showing all files and only the files which are currently open. The purpose of this is to provide a good representation of the workspace for the current task you are focused on, both as a visual reminder and a way to switch buffers.

This an alternative to showing buffers as tabs (which I find confusing for some reason) or having a flat list of buffers that can be referenced on command with BufExplorer or :ls.  I have also tried leaving all my buffers open in small splits, but I don't always have the space for this. I think the best way to list your open buffers is as a vertical tree within their folder structure.

The workflow would be to:

1. show the full tree, 
2. open a few files, 
3. toggle back to "Open Buffers Only" mode
4. as buffers are opened via Go to Definition, Find References, Telescope, etc, they are added to the "Open Buffers Only" view
4. work...
5. occasionally jump to nvim-tree sidebar to switch buffers
5. toggle back to "All Files" mode to browse and open another file via folder navigation
6. back to "Open Buffers Only" mode
7. work...

I would be willing to code this and put in a PR. My questions are:

1. Are you interested in having this as a feature?
2. I gathered from other issues that there is a lot of refactoring going on right now, should I work off of a particular branch or wait for something else to be completed?
3. Any advice on the best way to go about this? I was thinking about injecting an if clause to filter the tree just before it renders if a flag is set from the toggle function.

cseickel:
closes #744

This also includes updates to the refresh limiting code to make it more reliable. As it was, extra refresh calls were just being dropped which sometimes resulted with the view being out of sync. This was particularly noticeable with any toggle calls like the one I added. I turned it into a more of a debounce strategy, where the refresh is guaranteed to run after every call, but not more than once per vim.g.nvim_tree_refresh_wait period. I also set that default to 100 ms, let me know if that is not acceptable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,8 @@ highlight NvimTreeFolderIcon guibg=blue
 - `<BS>` will close current opened directory or parent
 - type `a` to add a file. Adding a directory requires leaving a leading `/` at the end of the path.
   > you can add multiple directories by doing foo/bar/baz/f and it will add foo bar and baz directories and f as a file
+- type `b` to toggle opened buffers only mode
+  > This will filter the tree to show only the opened buffers and the folders which contain them. Press 'b' again to go back to showing all files and folders.
 - type `r` to rename a file
 - type `<C-r>` to rename a file and omit the filename on input
 - type `x` to add/remove file/directory to cut clipboard
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -58,6 +58,17 @@ similar to combination of |:NvimTreeToggle| and |:NvimTreeFindFile|
 
 Show the tree in "opened buffers only" mode.
 
+|:NvimTreeEditNextFile| 	                  *:NvimTreeEditNextFile*
+
+Edits the next file visible in the tree, moving from top to bottom. If the
+tree is currently focused, this will switch to the previously used window.
+This is is particularly useful in conjuction with the "opened buffers only"
+mode.
+
+|:NvimTreeEditPrevFile| 	                  *:NvimTreeEditPrevFile*
+
+Same as NvimTreeEditNextFile, but in reverse.
+
 |:NvimTreeClipboard|			          *:NvimTreeClipboard*
 
 Print clipboard content for both cut and copy
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -54,6 +54,10 @@ It will also open the leafs of the tree leading to the file in the buffer
 close the tree or change the cursor in the tree for the current bufname,
 similar to combination of |:NvimTreeToggle| and |:NvimTreeFindFile|
 
+|:NvimTreeBuffers| 	                          *:NvimTreeBuffers*
+
+Show the tree in "opened buffers only" mode.
+
 |:NvimTreeClipboard|			          *:NvimTreeClipboard*
 
 Print clipboard content for both cut and copy
@@ -479,6 +483,7 @@ INFORMATIONS				        *nvim-tree-info*
 - typing 'P' will move cursor to the parent directory
 
 - type 'a' to add a file
+- type 'b' to toggle opened buffers only mode
 - type 'r' to rename a file
 - type `<C-r>` to rename a file and omit the filename on input
 - type 'x' to add/remove file/directory to cut clipboard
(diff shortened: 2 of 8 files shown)


kyazdani42:
hi ! Thanks for the contribution.

I've noted a few issues with your PR:
- open nvim-tree, expand a directory, do `bb`, folders are all opened and empty
- open nvim-tree, open a file, go back to the tree `bb`, folders are all opened and empty, going back to the file loads all folders that were previously closed
- some files are not properly indented, use `luacheck .` to fix these issues :) PS: indenting is 2 spaces

cseickel:
I can't recreate the issue with your steps, are you using all default options are is anything configured in your setup?

I was able to see a similar problem flash by while holding down the `b` key. I have an idea about what the issue is, but it would be nice to be able to recreate it more reliably if I can.

cseickel:
OK @kyazdani42, I think I fixed everything now. Take a look and let me know if you see any more issues.

khalidchawtany:
@cseickel I tries your branch it is a nice idea. I think there is a bug. New opened buffers does not show up in the tree till you  manually refresh using "R".

cseickel:
@khalidchawtany this is actually an existing bug that I have always had a problem with, the tree does not refresh on it's own when buffers are opened and closed. I've noticed this before because I use a different highlight for open files. There are also issues with the option to highlight the currently active file.

I've worked around these issues by adding this to my own config:

```
            function! UpdateNvimTreeBuffers(timerId) abort
                lua require('nvim-tree').refresh()
                lua require('nvim-tree.lib').redraw()
                lua require('nvim-tree').find_file()
            endfunction

            augroup nvim_tree_autocmds
                autocmd!
                autocmd BufWinEnter * silent! call UpdateNvimTreeBuffers(0)
                autocmd BufDelete * silent! call timer_start(10, 'UpdateNvimTreeBuffers')
            augroup END
            ]])
```

I had thought about suggesting this setup in the documentation or just adding this in the init of nvim-tree, but it might be a heavy handed solution. What do you think?

khalidchawtany:
@cseickel thanks for the suggestion. I will borrow that :).

I don't know whether it is a heavy handed solution or not. However, I am going to use it.

cseickel:
For closing buffers, I am just using `:bd` on the buffer itself. I haven't added specific commands to nvim-tree to handle that, but I suppose it would be a good idea and I could do that. 

I also have vim script functions that let me jump to the next/previous buffers as they appear in a file tree (alphabetical, grouped by folder). They do compliment this feature nicely but I wasn't sure those functions belong here or not. Do you think they should be included? Of course, I would need to convert this to lua first: https://github.com/cseickel/dotfiles/blob/d904e5e28a16e5b1bc29efdefc7b3258dd89c023/config/nvim/core-mappings.vim#L268

As far as an indicator of what mode you are in, I'm not sure how to handle that. Perhaps it would be good to put some help text at the bottom in case someone gets there accidentally? Maybe just an italics text that says:

"Currently displaying open buffers only, press 'b' to display all files.'

khalidchawtany:
> I also have vim script functions that let me jump to the next/previous buffers as they appear in a file tree (alphabetical, grouped by folder). They do compliment this feature nicely but I wasn't sure those functions belong here or not. Do you think they should be included? Of course, I would need to convert this to lua first: https://github.com/cseickel/dotfiles/blob/d904e5e28a16e5b1bc29efdefc7b3258dd89c023/config/nvim/core-mappings.vim#L268
I would love to have them
> "Currently displaying open buffers only, press 'b' to display all files.'
This is a good idea

kyazdani42:
hi, sorry for the late answer. Could you rebase before i do another review please ?

cseickel:
I am working on the rebase but the conflicts are pretty tangled up and I don't have much free time right now. This might take a while...

cseickel:
Hey @kyazdani42, sorry to leave this hanging for so long. I've decided to abandon this PR because I am working on my own tree plugin based on NuiTree. 

Please feel free to use any code you might want from this PR, and thanks for the great plugin!

khalidchawtany:
> Please feel free to use any code you might want from this PR, and thanks for the great plugin!

Can you share the link to the new repo?

cseickel:
Sure, here it is: https://github.com/nvim-neo-tree/neo-tree.nvim

PR #760, opened 2021-10-30
chore: renderer rewrite

kyazdani42:
complete rewrite of renderer, allow for more options. Still some work to do.

The new modular approach will allow user to override components maybe.
I'll list all the modifications here once i'm done.


kyazdani42:
too big to continue, i'll refactor this one with small steps

PR #762, opened 2021-10-31
feat: allow number and relativenumber settings

JavierPoduje:
This PR contains one commit that allows us to configure the line number, just like we would do with our regular buffers. 

Default behaviour:
default

Just number:
just_number

Just relative number:
just_relative

Number and relative number:
number_and_relative

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -106,14 +106,16 @@ function.
         height = 30,
         side = 'left',
         auto_resize = false,
+        number = false,
+        relativenumber = false,
         mappings = {
           custom_only = false,
           list = {}
         }
       },
       filters = {
-	dotfiles = false,
-	custom = {}
+        dotfiles = false,
+        custom = {}
       }
     }
 <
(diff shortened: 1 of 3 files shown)


kyazdani42 (changes requested):
doc file should be updated please :)

JavierPoduje:
> doc file should be updated please :)

@kyazdani42 doc file updated. My bad!

doc/nvim-tree-lua.txt
```diff
@@ -106,6 +106,8 @@ function.
         height = 30,
         side = 'left',
         auto_resize = false,
+	number = false,
+	relativenumber = false,
```
kyazdani42, line 110: wrong indent ^^' i believe i use spaces to indent help files.

kyazdani42 (approved):

doc/nvim-tree-lua.txt
JavierPoduje: Fixed! I also change another indentation with the same problem just below mine :mage_man: 

before:

before

after:

after

kyazdani42 (approved):

PR #784, opened 2021-11-16
Using "vim.env.PATHEXT" to determine Windows executables.

UnderCooled:
This [SU answer](https://superuser.com/questions/228680/on-windows-what-filename-extensions-denote-an-executable/228693) is not perfect, but at least it works without marking every file in Windows executable.

Before:
image

After:
image


lua/nvim-tree/populate.lua
```diff
@@ -39,12 +39,18 @@ end
 
 local function file_new(cwd, name)
   local absolute_path = utils.path_join({cwd, name})
-  local is_exec = luv.fs_access(absolute_path, 'X')
+  local ext = string.match(name, ".?[^.]+%.(.*)") or ""
+  local is_exec = false
+  if vim.fn.has('win32') == 1 then
```
kyazdani42, line 44: could you store this in the global scope so it's not called on each iteration ?

lua/nvim-tree/utils.lua
```diff
@@ -165,4 +165,21 @@ function M.merge_sort(t, comparator)
...
+    return false
+  end
+  ext = '.' .. string.upper(ext)
+  local pathexts = vim.split(vim.env.PATHEXT, ';')
+  for _, exe_ext in ipairs(pathexts) do
+    if ext == exe_ext then
+      return true
+    end
+  end
+  return false
```
kyazdani42, line 182: on small suggestion to make this faster, why not loading a map of executable on startup instead of looping over them each time.
This would result in something like:
```lua
local wexe = vim.split(vim.env.PATHEXT, ';')
local pathexts = {}
for _, v in pairs(wexe) do
  pathexts[v] = true
end
function M.is_windows_exe(ext)
...
return pathexts[ext]
end
```

kyazdani42 (approved):

kyazdani42:
please fix the linting issue and i'll merge this :)

kyazdani42 (approved):

PR #786, opened 2021-11-17
feat: add signcolumn setting

NiYanhhhhh:
I think there is someone doesn't like the blank signcolum, so I add an option to change that.
As it is 'auto', there won't be a signcolumn before every line. :)

config:
image
when no errors and no warnings:
image
when there are something wrong:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ require'nvim-tree'.setup {
     hide_root_folder = false,
     side = 'left',
     auto_resize = false,
+    signcolumn = 'yes',
     mappings = {
       custom_only = false,
       list = {}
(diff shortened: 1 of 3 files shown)


gegoune:
But there already is
```lua
require('nvim-tree.view').View.winopts.signcolumn = 'yes:5'
```
which seems to work.

NiYanhhhhh:
Have you update to the newest version? This seems not to work on my machine..

gegoune:
Works fine for me on 5d8453dfb.

NiYanhhhhh:
Oh, I see. This should be after the setup function.

PR #787, opened 2021-11-18
Typo on setup code

duarteocarmo:


kyazdani42:
it's not a typo, in lua you can omit the parens when there is only one argument to the function.

duarteocarmo:
You are absolutely right! My bad.. Still figuring out the lua syntax.. Keep up the great work!

PR #788, opened 2021-11-18
fix: colors setup highlight

NiYanhhhhh:
I find there are something wrong when getting highlight from current colorscheme. On the right way hlID should be passed to synIDtrans() function first. :)
image


NiYanhhhhh:
Here I also fix the git icon color problem.
I find the folder highlight unexpectedly covers the git icon like the picture below
image
Now it' acts like this:
image

kyazdani42 (approved):
strange i've never seen anything with this. Merging, if someone complains, i'll revert but it should not be the case :)

PR #798, opened 2021-11-26
fix: executable recognition on Windows

magicmonty:
In Windows, the function `vim.loop.fs_access(<filename>, 'X')` always returns true. So all files will be marked as executable.
In this PR I fix this, by getting the list of executable extension off the environment variable `PATHEXT` and comparing the file name extension with it, if neovim runs under Windows.

### DIFF
--- a/lua/nvim-tree/populate.lua
+++ b/lua/nvim-tree/populate.lua
@@ -37,9 +37,32 @@ local function dir_new(cwd, name)
   }
 end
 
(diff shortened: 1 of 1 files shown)


kyazdani42:
i believe this PR is duplicate of https://github.com/kyazdani42/nvim-tree.lua/pull/784, closing this.

magicmonty:
Sure, sorry, missed that one ;)

PR #799, opened 2021-11-27
fix: close_node behavior on file

MunifTanjim:
If `close_node` was called on a file, the parent folder is collapsed, but the ui was not being redrawn.


kyazdani42 (approved):

kyazdani42:
thanks :)

Issue #779, opened 2021-11-09
NvimTree does not open the file when it is open standalone


wookayin:
When nvim-tree is working when "hijacking" directory buffers, it cannot open the file.

Reproduce:

1. `vim .`   (or after entering vim, `e .`, `vs .`, etc.)
2. At this point, NvimTree will open in place of netrw (because it hijacks netrw upon FileExplorer)
3. Press `<CR>` to open any file
4. Nothing happens

It should work as Nerdtree and netrw would do.

Expected behavior:

1. Open the file
2. The current window (directory explorer) will be replaced by the file

wookayin:
This is because `nvim-tree.lib.Tree.target_winid` is nil

i1skn:
so, what one could do to fix it?

msva:
for example, I've "hacked" it around by `update_to_buf_dir=false` and `au VimEnter * :bwipeout` :D 
But proper fix would be to bisect the code and revert the change when it become broken.
Or to fix it another way and create a PR.

msva:
well, and yes, I've disabled the picker.

wookayin:
@msva That would be a great workaround, at least the default netrw or nerdtree would work well.

Ideally we would keep NvimTree handle the directory buffer, but this seems to require quite lines of changes. 

@i1skn  I tried adding to L349 `target_winid = api.nvim_get_current_win()` to force nvim-tree to open the file in the "current" window (the same one as the directory buffer), but the nvim-tree window will still appear in the left (with some blinking UI glitches), which should be fixed by improving other relevant logics.

wookayin:
@kyazdani42 Any updates?

wookayin:
#815 works to fix the bug itself, but it still doesn't work same as the netrw explorer when opening a directory buffer (e.g., only one nvim-tree buffer can present). I will open another issue about this. Thank you for the fix.

PR #815, opened 2021-12-01
fix: NvimTree does not open the file when it is open standalone

eightpigs:
fix: #779


kyazdani42 (approved):

kyazdani42:
i'm not sure why this was added though, but this seems strange indeed. Thanks for the fix

PR #817, opened 2021-12-02
feat: add keybindings for launching telescope

kylo252:
Add keybindings for launching Telescope in the directory under the cursor

- `<c-f>` will trigger `find_files`
- `<c-g>` will trigger `live_grep`

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -590,6 +590,8 @@ Defaults to:
       { key = "-",                            cb = tree_cb("dir_up") },
       { key = "s",                            cb = tree_cb("system_open") },
       { key = "q",                            cb = tree_cb("close") },
+      { key = "<c-f>",                        cb = M.nvim_tree_callback("launch_find_files") },
+      { key = "<c-g>",                        cb = M.nvim_tree_callback("launch_live_grep") },
       { key = "g?",                           cb = tree_cb("toggle_help") },
     }
 <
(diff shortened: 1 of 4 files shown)


kylo252:
I'm not sure if you want to have something like this in the wiki instead, but I figured it could be useful for others to see. 

Here's the full version to use from the user-config

```lua
function M.launch_live_grep(opts)
  return M.launch_telescop("live_grep", opts)
end

function M.launch_find_files(opts)
  return M.launch_telescope("find_files", opts)
end

function M.launch_telescope(func_name, opts)
  local telescope_status_ok, _ = pcall(require, "telescope")
  if not telescope_status_ok then
    return
  end
  local lib_status_ok, lib = pcall(require, "nvim-tree.lib")
  if not lib_status_ok then
    return
  end
  local is_folder = node.has_children and true
  local basedir = is_folder and node.absolute_path or vim.fn.fnamemodify(node.absolute_path, ":h")
  vim.api.nvim_out_write(string.format("Got: %s \n", basedir))
  opts = opts or {}
  opts.cwd = basedir
  return require("telescope.builtin.files")[func_name](opts)
end
```

And then to use it, add something like this depending on your configuration structure
```lua

local function custom_callback(callback_name)
    return string.format(":lua require('core.nvimtree').%s()<CR>", callback_name)
end

-- this is passed to  ("nvim-tree").setup(opts)
 local mappings_list = {
      { key = "<c-f>", cb = custom_callback "launch_find_files" },
      { key = "<c-g>", cb = custom_callback "launch_live_grep" },
}
```

kyazdani42:
Hi ! Thanks for the contribution, but i'd prefer keeping integration with other plugins into users hands :) 
I guess we could add wiki pages indeed, would be more appropriate for this kind of thing :)

kyazdani42:
https://github.com/kyazdani42/nvim-tree.lua/wiki/Open-file-in-telescope
closing this :)

Issue #803, opened 2021-11-28
Error when call the `NvimTreeToggle` after update Neovim nightly.


nyngwang:
## Neovim version

NVIM v0.6.0-dev+644-ge4976cbd2
Build type: Release
LuaJIT 2.1.0-beta3

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/HEAD-e4976cb_1/share/nvim"

Run :checkhealth for more info

## Setup

I'm using `packer.nvim` with the following settings:

```
use {
  'kyazdani42/nvim-tree.lua',
  requires = {
    'kyazdani42/nvim-web-devicons'
  },
  config = function()
    require'nvim-tree'.setup {
      hijack_cursor = true,
      update_cwd = true,
      open_on_tab = true,
      view = {
        side = 'right',
        width = 45,
        auto_resize = false,
      },
      update_focused_file = {
        enable      = true,
        update_cwd  = true,
      },
      filters = {
        custom = { '.git', '__pycache__', '.DS_Store' }
      }
    }
    vim.g.nvim_tree_show_icons = {
      ['git'] = 1,
      ['folders'] = 0,
      ['files'] = 0,
      ['folder_arrows'] = 0,
    }
    vim.g.nvim_tree_highlight_opened_files = 1
    vim.g.nvim_tree_respect_buf_cwd = 1
    vim.g.nvim_tree_indent_markers = 1
    vim.g.nvim_tree_add_trailing = 1
  end
}
vim.api.nvim_set_keymap('n', ';', '<cmd>NvimTreeToggle<CR>', NOREF_NOERR_TRUNC)
```

## Error message

```
Error detected while processing CursorMoved Autocommands for "NvimTree":
E5108: Error executing lua ...m/site/pack/packer/start/nvim-tree.lua/lua/nvim-tree.lua:341: Column value outside range
stack traceback:
        [C]: in function 'nvim_win_set_cursor'
        ...m/site/pack/packer/start/nvim-tree.lua/lua/nvim-tree.lua:341: in function 'place_cursor_on_node'
        [string ":lua"]:1: in main chunk
```

nyngwang:
After some testing this error will occur when I run `nvim path/to/my/file` but if I start from a folder `nvim path/to/a/folder/` then there will be no problem. Very weird.

imSafdarAwan:
[...]

PR #821, opened 2021-12-03
fix: idx passed to api.nvim_win_set_cursor may be -1

### LINKED ISSUE #783: :NvimTreeFindFile not working
nvim-tree only opens the window but does not go to the file when I run `:NvimTreeFindFile`. I remember it working before. But it stopped working quite some time ago.

kunish:
This causes out of range error on CursorMoved event is fired if hijack_cursor option is set to true.
Although I don't know why it fixes the issue, but it has been working fine for me.

Potentially fix #783, #803


kyazdani42 (approved):
indeed, thanks for the fix

PR #822, opened 2021-12-03
Focus the parent node instead of its first child node

RobertAudi:
This change make the `parent_node` (`P` mapping) repeatable, this way by pressing `P` multiple times one can navigate to the parent directory on each key press.

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -556,6 +556,7 @@ function M.parent_node(node, should_close)
       parent.open = false
       altered_tree = true
     end
+    line = math.max(unpack({line - 1, 1}))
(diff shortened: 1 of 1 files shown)


kyazdani42:
not sure what is the purpose of this PR ? running `P` multiple times does bring me to the parent dir on each keypress already

RobertAudi:
@kyazdani42 I created that PR because `P` brings me to the first child of the parent instead of placing the cursor on the parent directory, so pressing `P` multiple times didn't do anything since the first child did not change.

kyazdani42:
P brings to the parent as expected, with this change, it brings to the parent of the parent which is not expected behavior.
I'm closing this then.

PR #834, opened 2021-12-09
Fixing deprecated LSP diagnostic getter (fixes https://github.com/kyazdani42/nvim-tree.lua/issues/833)

GaetanLepage:
TODO:
- [ ] check if `vim.diagnostic.get` really is the same as `vim.lsp.diagnostic.get_all`

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -1,7 +1,7 @@
 local a = vim.api
 local utils = require'nvim-tree.utils'
 local view = require'nvim-tree.view'
-local get_diagnostics = vim.lsp.diagnostic.get_all
+local get_diagnostics = vim.diagnostic.get
 
 local M = {}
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/diagnostics.lua
```diff
@@ -1,7 +1,7 @@
 local a = vim.api
 local utils = require'nvim-tree.utils'
 local view = require'nvim-tree.view'
-local get_diagnostics = vim.lsp.diagnostic.get_all
```
gegoune, line 4: That changes minimal required version of nvim supported by nvim-tree. It should be guarded I think.
```lua
local has_07 = vim.fn.has('nvim-0.7') == 1
```

jasonrhansen:
I opened #840 to fix this since `vim.diagnostic.get` and `vim.lsp.diagnostic.get_all` return different things.

Issue #831, opened 2021-12-07
NvimTreeFindFileToggle command  does not find the file


numbcoder:
**OS**
macOS 12.0.1

**Neovim version**
NVIM v0.7.0-dev+676-g4306b395d

**Nvim-tree version**
commit 2e33b1654384921ec1cc9656a2018744f3f1ce81

**Describe the bug**
Execute `NvimTreeFindFileToggle` command, nvim-tree opened, but the tree is not expanded, and  the cursor is not focused on the file. 
If I expand the nvim-tree dir manually before execute `NvimTreeFindFileToggle` command, the command will be works well.

**To Reproduce**
1. Open a file with Telescope
2. Execute `NvimTreeFindFileToggle` command

Nvim-tree will be opened, but the cursor is not focused on the file.

**Expected behavior**
Nvim-tree will be opened, the dir will be expanded and the cursor will be focused on the file.

kuznetsss:
The issue already exists (#783) but it was closed for some reason

PR #836, opened 2021-12-09
Fix redrawing tree when not inside git repo

### LINKED ISSUE #783: :NvimTreeFindFile not working

kuznetsss:
Fixes #783 and #831

Redraw was not called when cwd is not a git repo.


kyazdani42 (approved):
looks like i'm going to need to rewrite this at some point, the redrawing logic is a bit too complex.

PR #838, opened 2021-12-10
preserve previous focus and update cwd on find option

escorponox:
solves #171

added an option to change root dir on findFile. This mimics `:NERDTree %`

Let me know if this approach fits. It is my first Lua code.

Thank you for this awesome plugin!!!


gegoune (commented):
This is way too opinionated and user configuration specific. Not every one uses gruvbox.

escorponox:
sorry I messed it up, the colors were not added initially, I will reopen later

PR #840, opened 2021-12-10
Fix for deprecated vim.lsp.diagnostic.get_all

### LINKED ISSUE #833: Bug: vim.lsp.diagnostic.get_count() is deprecated
**OS**
Arch Linux

**Neovim version**
`NVIM v0.7.0-dev+697-g6063e07a0`

**Describe the bug**
The plugin seems to rely on `vim.lsp.diagnostic.get_count()` but the latter has been deprecated.

**To Reproduce**
Nothing special, just loading the plugin with 'diagnostics' enabled.

jasonrhansen:
Fixes #833

Continues to work on older version of neovim by using old API if new one doesn't exist.

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -34,7 +34,10 @@ end
 local function from_nvim_lsp()
   local buffer_severity = {}
 
+  -- vim.lsp.diagnostic.get_all was deprecated in nvim 0.7 and replaced with vim.diagnostic.get
+  -- This conditional can be removed when the minimum required version of nvim is changed to 0.7.
   if vim.diagnostic then
+    -- nvim version >= 0.7
     for _, diagnostic in ipairs(vim.diagnostic.get()) do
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/diagnostics.lua
```diff
@@ -35,12 +34,25 @@ end
 local function from_nvim_lsp()
   local buffer_severity = {}
 
-  for buf, diagnostics in pairs(get_diagnostics()) do
-    if a.nvim_buf_is_valid(buf) then
-      local bufname = a.nvim_buf_get_name(buf)
-      if not buffer_severity[bufname] then
-        local severity = get_lowest_severity(diagnostics)
-        buffer_severity[bufname] = severity
+  if vim.diagnostic then
```
gegoune, line 37: Might be worth adding note here stating that this conditional should be removed when minimal required version of nvim is 0.7. Probably won't happen any time soon but it's a shame to leave dead code in.

kyazdani42 (approved):

PR #847, opened 2021-12-15
restore-feat: highlight-git-ignored-files

### LINKED ISSUE #813: How to highlight items that is ignored by git with `git.ignore` set to `false`?
```lua
local nvim_tree = require('nvim-tree')
nvim_tree.setup({
  git = {
    ignore = false
  }
})
```

With `git.ignore` set to `false`, as far as I know there is no now way to set highlights for items that is ignored by git. Previously, I had known if certain item is ignored by looking at its highlight.

# Before

image

# HEAD

image

saviocmc:
See #813 for more details.

The commit 6662b60a2bb55f0c1eca4db54cba5c57c2b2daff introduced a breaking change in the plugin, as the git ignored folders/files could not be highlight anymore. This was caused by the decision to not include the parameter `--ignored=matching` in the git runner command when the user wants to *show* the ignored files, resulting in a `nil` git status for this files and making impossible to the renderer functions to set the proper highlights.

This PR aims to restore the previous behavior by  including `--ignored=matching` in all git commands. I think this is the only way to get the old behavior without changing the current config.

In order to be possible to get this behaviour *and* keep the possibility to call git without the  `--ignored=matching` (for possible performance gains) I think a new config option must be added, like `git.highlight_ignored`. But I would prefer to not have to make such change.

Fix #813

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -255,9 +255,9 @@ Here is a list of the options available in the setup call:
     type: `boolean`
     default: `true`
 
-  - |git.ignore|: ignore files based on `.gitignore`.
-    will add `ignored=matching` to the integration when `true`. Otherwise will
-    add `ignored=no` to the integration which can lead to better performance.
+  - |git.ignore|: ignore files based on `.gitignore`
+    type: `boolean`
+    default: `true`
 
   - |git.timeout|: kills the git process after some time if it takes too long
     type: `number`
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -19,7 +19,7 @@ function M.reload(callback)
     Runner.run {
       project_root = project_root,
       list_untracked = git_utils.should_show_untracked(project_root),
-      list_ignored = M.config.ignore,
+      list_ignored = true,
       timeout = M.config.timeout,
       on_end = function(git_status)
         M.projects[project_root] = {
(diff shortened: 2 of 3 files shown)


kyazdani42 (approved):
looks good to me :)
Would you mind adding a new config option for this ? Otherwise i'll do it

saviocmc:
> Would you mind adding a new config option for this ? Otherwise i'll do it

Hello! I would like to clarify what config option you ment.
This PR **does not** introduce any behavior or option, it just restores the old behavior.
What I sad about adding an option was to let the user choose if the argument `--ignored=matching` should be added to the git call by the plugin. When git is called without this argument, there is less lines to parse and therefore a potential performance optimization, but we lose the feat of highlight the ignored files/folders. With that in mind, there are two possibilities of config option:

* Option 1:
Add an option like `git.parse_ignored` to include of not the arg `--ignored=matching` to the git command? It could be true by default, as it is necessary for either highlighting or hiding the ignored files and the user could set it to false to possibly gain performance.

* Option 2:
Rename the option `git.ignore` to `git.hide_ignored_files` and add option `git.highlight_ignored_files` so the plugin can decide when the arg `--ignored=matching` is necessary, in this case, when either option is true. If both is false, the is no need for the git arg and therefore the possibility for the performance gain.

The first one is simpler, but not very clear for the user I think. The second is more clear in the intention, but would introduce a breaking change. Which of the two you think is better (if any)?

kyazdani42 (approved):
let's merge this for now, we'll fix it properly in a second time

PR #855, opened 2021-12-20
custom node keymaps

escorponox:
This allows to define custom keybindings in configuration that receive the node as a parameter. Something like:
```lua
local tree_cb = require'nvim-tree.config'.nvim_tree_callback

local function print_node_path(node) {
  print(node.absolute_path)
}

require'nvim-tree'.setup {
  view = {
    mappings = {
      list = {
        { key = "s", cb = tree_cb("vsplit") }, -- deprecated
        { key = "t", action = "tabnew" },
        { key = "p", action = "print_path", action_cb = print_node_path },
      }
    }
  }
}

```

I use it to run a script in the node absolute path. Let me know if this could be useful for others and I will write down the docs.

Thank you so much for this awesome plugin!!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,50 +219,60 @@ highlight NvimTreeFolderIcon guibg=blue
 The `list` option in `view.mappings.list` is a table of
 ```lua
 -- key can be either a string or a table of string (lhs)
--- cb is the callback that will be called
+-- action is the name of the action
+-- action_cb is the function that will be called, it receives the node as a parameter. Optional for default actions
 -- mode is normal by default
+
+local tree_cb = require'nvim-tree.config'.nvim_tree_callback
+
+local function print_node_path(node) {
+  print(node.absolute_path)
+}
+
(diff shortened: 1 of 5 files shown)


kyazdani42:
looks good to me !
Something that i wonder though, couldn't this completely replace the existing keymap implementation ? 
if you could spend some time on that, i'd be happy to accept this PR :) And if you do, don't remove the existing implementation, just deprecate it.
If you want this to be merged as is, just write the documentation for it and we'll keep both implementations together.

escorponox:
I will do it during this week whenever I have a couple of hours :wink:

escorponox:
I think it is ready now. I tried to make a mappings module but I hit require loops (I'm new to lua)...

kyazdani42 (approved):
good enough for me now, i will extract the whole mappings and actions configurations and setup to their own module soon, no need to worry about this :)

PR #862, opened 2021-12-23
Show diagnostics on directories

revyakin:
This PR adds the ability to show diagnostics on a closed directory if it has child nodes with diagnostics.

The feature can be enabled/disabled via `diagnostics.show_on_dirs` option.

img
img
img


kyazdani42 (approved):

Issue #859, opened 2021-12-22
Nvim-tree cannot close notify popup when closing with :NvimTreeClose


dkeithdj:
**OS**
Windows 10 (WSL)

**Neovim version**
NVIM v0.7.0-dev+716-g9a0196d24
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

**Nvim-tree version**
master

**Describe the bug**
`Vim:E444: Cannot close last window` doing :NvimTreeClose while notify popup is still open
![nvimRecord](https://user-images.githubusercontent.com/42316655/147091098-89d2b551-3e06-41a2-8aa7-8b33f4a54ab0.gif)

**To Reproduce**
nvim .
do`:NvimTreeClose` or other commands related to closing nvim-tree

**Expected behavior**
no errors and popup will close

**minimal Nvim-Tree config**
```
vim.g.nvim_tree_window_picker_exclude = {
  filetype = { "notify", "packer", "qf" },
buftype = { "terminal" },
}
vim.g.nvim_tree_quit_on_open = 1
vim.g.nvim_tree_indent_markers = 1

nvim_tree.setup({
  -- ignore_ft_on_setup = { "notify" },
  open_on_setup = false,

  auto_close = false,
  -- update_to_buf_dir = { enable = false, auto_open = false },
  view = {
    width = 30,
  },
  update_focused_file = { enable = true },

  -- disable_window_picker = 0,
})
```

kyazdani42:
should be fixed now by #863, reopen if it isn't

dkeithdj:
It works fine now, although there's some slight bug where the popup will freeze when a new buffer is opened
but you can just `CTRL-W-O` it to close the popup. Not that I get errors often but there's that.

NOTE: This bug only occurs when you start nvim with nvim-tree such as `nvim .`

**Added gif for visualization**
![nvimRecordRecord](https://user-images.githubusercontent.com/42316655/147380469-5cccb3b3-fea8-412b-b0f5-dd24fd7023d2.gif)

PR #863, opened 2021-12-23
fix(view): check wins are floating

rcarriga:
Just adds a check to see if any of the other windows are not floating before closing nvim-tree.

Raised in nvim-notify in https://github.com/rcarriga/nvim-notify/issues/45 and also fixes the issue #859


kyazdani42 (approved):
thanks for the fix :)

PR #874, opened 2021-12-25
Add a config option to use :cd instead of :lcd

### LINKED ISSUE #829: Can we update `change_dir` to use `cd` instead of `lcd`?
I have some really unexpected behavior and tracked it down to the fact that `lcd` in used instead of `cd` [here](https://github.com/kyazdani42/nvim-tree.lua/blob/master/lua/nvim-tree/lib.lua#L458). Would it be possible to change that or to make it optional to change dirs using `cd`?

I mainly use the keymapping to change dirs in the tree view, but then the view starts to switch which dirs to show when I just trough buffers that were already open. Think #776 is also mentioning something along these lines.

If there is a way I can add a custom mapping to fix it for myself I would also be happy, but maybe even better to not use `lcd` at all?

svanharmelen:
Fixes #829

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -467,7 +467,11 @@ function M.change_dir(name)
   end
   current_tab = new_tab
 
-  vim.cmd('lcd '..vim.fn.fnameescape(foldername))
+  if vim.g.nvim_tree_change_dir_global == 1 then
+    vim.cmd('cd '..vim.fn.fnameescape(foldername))
+  else
+    vim.cmd('lcd '..vim.fn.fnameescape(foldername))
+  end
   M.init(false, foldername)
 end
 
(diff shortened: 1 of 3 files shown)


lua/nvim-tree/lib.lua
```diff
@@ -467,7 +467,11 @@ function M.change_dir(name)
   end
   current_tab = new_tab
 
-  vim.cmd('lcd '..vim.fn.fnameescape(foldername))
+  if vim.g.nvim_tree_change_dir_global == 1 then
+    vim.cmd('cd '..vim.fn.fnameescape(foldername))
```
kyazdani42, line 471: did you try this change with `opts.update_cwd` to true ? it might create bugs.

kyazdani42 (commented):
Can you just check that this doesn't create issues with the DirChanged event ? i'll merge this if its not the case :)

lua/nvim-tree/lib.lua
svanharmelen: I have `opts.update_cwd` enabled by default in my setup. I used this branch since I created the PR 27 days ago and haven't noticed any issues or unexpected behavior.

svanharmelen:
@kyazdani42 please see the inline comment. I tested this quite a lot the last 4 weeks already (using `update_cwd=true`.

kyazdani42 (approved):
thanks for the PR :)

Issue #856, opened 2021-12-20
Add option to exclude files that are hidden


sarmong:
I have dotfiles hidden by default.  However, I don't want my `.config` directory be hidden.

Would be nice to add an option to exclude some files and folders by name or glob

kyazdani42:
Not hard but not trivial to do, you could give it a shot :)

alex-courtis:
`filters.exclude` is live and working.

PR #878, opened 2021-12-26
feat: add exclude filter list

sarmong:
See #856 

Not sure if the naming is suitable, would appreciate other suggestions


kyazdani42 (approved):
looks good to me :) thanks for implementing this.

PR #893, opened 2022-01-12
A more modern way to raise warning message

anuvyklack:
A better way to raise warnings without VimL hacks.


kyazdani42 (approved):
good idea :)

PR #901, opened 2022-01-19
docs: Warn about dirbuf.nvim in `update_to_buf_dir` docs

elihunter173:
Hello! I'm the author of the file manager plugin [dirbuf.nvim](https://github.com/elihunter173/dirbuf.nvim).

It seems dirbuf is incompatible with nvim-tree's `update_to_buf_dir` option in the same way that `vim-dirvish` is, so I added a warning to the documentation about this alongside the warning about vim-dirvish. I'm also adding similar warnings to dirbuf's documentation

Related: https://github.com/elihunter173/dirbuf.nvim/issues/2


kyazdani42 (approved):
thanks :)

PR #909, opened 2022-01-21
fix(action): resolve nil error

defpis:
image


kyazdani42:
ah haven't seen that, fixed all issues alltogether :) thanks for the fix anyway

Issue #881, opened 2022-01-03
The option `cb = tree_cb("trash")` broken


nyngwang:
**OS**
macOS

**Neovim version**
NVIM v0.7.0-dev+797-gb218d02c4

**Nvim-tree version**
lastest

**Describe the bug**
The option `cb = tree_cb("trash")` broken. After I pressed `y` the file is not moved to the trash can.

harrisoncramer:
I thought this as well, but it turns out the implementation relies on a separate command line tool. You can install it with NPM, here: `https://github.com/sindresorhus/trash-cli`

hermitmaster:
It would be nice to have this documented.

kyazdani42:
I think it's already documented :
```
  - |trash.cmd|: the command used to trash items (must be installed on your system)
      type: `string`
      default: `"trash"`
```
closing this because its not broken :smile:

PR #916, opened 2022-01-26
fix(actions) trashing

### LINKED ISSUE #913: After refacto `trash` get broken
**OS**
MacOS

**Neovim version**
0.6.0

Before 1248bce refacto: create actions folder and move some code to it, `D` works. After the exact refactor commit, `D` just open the file/dir.

RaafatTurki:
it seems like when the actions system was implemented the trash feature was not tested, this fixes #913  and #881


kyazdani42:
thanks :)

PR #924, opened 2022-02-02
fix: close windows properly

Lopi-py:
remove-file [before](https://gyazo.com/de97a8661cc438669381b2790cdb309f),
[example](https://gyazo.com/97f0a8ab23a3dfeeeb9c1f30394d295d) with [nvim-notify](https://github.com/rcarriga/nvim-notify)

### DIFF
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -1,3 +1,5 @@
+local has_notify, notify = pcall(require, 'notify')
+
 local a = vim.api
 local uv = vim.loop
 
@@ -11,7 +13,11 @@ end
 
 function M.warn(msg)
   vim.schedule(function()
-    vim.notify("[NvimTree] "..msg, vim.log.levels.WARN)
+    if has_notify then
+      notify(msg, vim.log.levels.WARN, { title = "NvimTree" })
+    else
+      vim.notify("[NvimTree] "..msg, vim.log.levels.WARN)
+    end
   end)
 end
 
--- a/lua/nvim-tree/actions/remove-file.lua
+++ b/lua/nvim-tree/actions/remove-file.lua
@@ -6,6 +6,14 @@ local events = require'nvim-tree.events'
 
 local M = {}
 
+local function close_windows(windows)
+  for _, window in ipairs(windows) do
+    if a.nvim_win_is_valid(window) then
+      a.nvim_win_close(window, true)
+    end
+  end
+end
+
 local function clear_buffer(absolute_path)
   local bufs = vim.fn.getbufinfo({bufloaded = 1, buflisted = 1})
   for _, buf in pairs(bufs) do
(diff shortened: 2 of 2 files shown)


lua/nvim-tree/utils.lua
```diff
@@ -1,3 +1,5 @@
+local notify = vim.F.npcall(require, 'notify')
```
kyazdani42, line 1: what is this F thing :smile: ? it's not documented on my end, is it an internal lua api ?

lua/nvim-tree.lua
```diff
@@ -200,14 +200,8 @@ function M.reset_highlight()
...
 
-local prev_line
+-- removing `prev_line` since does not work when pressing `$` or `0`
 function M.place_cursor_on_node()
-  local l = vim.api.nvim_win_get_cursor(0)[1]
-  if l == prev_line then
-    return
-  end
-  prev_line = l
-
```
kyazdani42, line 210: this is used so the hijacking does not hijack when moving on the line. `$` and `0` were expected to be working normally.

lua/nvim-tree/utils.lua
Lopi-py: I did not see documentation, maybe because it's very simple, and yes, it's from the neovim lua internal api https://github.com/neovim/neovim/blob/master/runtime/lua/vim/F.lua

lua/nvim-tree.lua
Lopi-py: mm, I think hijacking `$` and `0` might be a nice addition, or at least make it optional

lua/nvim-tree/utils.lua
kyazdani42: I like the api, but if it's not documented i'd rather not use it in the plugin because it might be unstable :)

lua/nvim-tree.lua
kyazdani42: Well if you have an overflowing filename, and you want to go to the end of the line, you should be able to move, thats why i implemented this. Better leave it like this for now, or it might create issues for other people.

lua/nvim-tree/utils.lua
Lopi-py: I changed it to use `pcall`

lua/nvim-tree.lua
Lopi-py: Removed

Lopi-py:
Sorry, I renamed the branch and this was closed

kyazdani42:
could you rebase please :) ?

Lopi-py:
What do you mean? I'm a bit new using github + git

kyazdani42:
rebasing means taking your branch and moving the tail commit on top of another branch (in this case master), because your branch is conflicting with master right now.
see [this tutorial](https://git-rebase.io/) which is pretty great for understanding rebasing.

Lopi-py:
> rebasing means taking your branch and moving the tail commit on top of another branch (in this case master), because your branch is conflicting with master right now. see [this tutorial](https://git-rebase.io/) which is pretty great for understanding rebasing.

Thanks, I rebased correctly? 🤔

kyazdani42:
looks like you did not :smile:
you should:
- 1: update your master origin branch to the upstream master branch
- 2: checkout this branch
- 3: `git rebase -i master`
- 4: drop the uneeded commits
- 5: `git push -f origin fix/hijack_cursor_properly`

kyazdani42:
you forgot to drop those i guess ?

Lopi-py:
Sorry the late replies, It's done I guess

kyazdani42 (approved):

Issue #937, opened 2022-02-06
Race Condition when plugin is autoloaded.


asaidimu:
**OS**
```Linux 5.16.5-arch1-1```

**Neovim version**
```
NVIM v0.6.1
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
```

**Nvim-tree version**
Commit ```8a6c7ba```

**Describe the bug**
The following error message is displayed when plugin is auto loaded with packer.

```
E5108: Error executing lua ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:165:
Vim(echomsg):E114: Missing quote: "Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:83:
Vim(lua):E5108: Error executing lua ...site/pack/packer/opt/nvim-tree.lua/lua/nvim-tree/lib.lua:114: attempt to index field 'Tree' (a nil value)
stack traceback:
        [C]: in function 'cmd'
        ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:165: in function <...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:161>
        [string ":lua"]:1: in main chunk
```

**To Reproduce**
Have packer autoload the plugin like so..
```
use  {
    'kyazdani42/nvim-tree.lua',
    cmd = { 'NvimTreeOpen', 'NvimTreeToggle' },
    -- requires lua script in which I call set up  
}
```
Call either `:NvimTreeOpen` or `:NvimTreeToggle`

**Expected behavior**
Nvim-tree to open without an error.

**Comment by asaidimu** [comment]
Naturally I dug into the source to find out what exactly was happening.

I find this line a bit problematic in my case. It delays the call to `nvim-tree.lib.init()`. 
```
vim.schedule(function() M.on_enter(opts) end)
```
So that when I call `nvim-tree.open()`, init has not yet run.

kyazdani42:
upgrade nvim-tree again, it's been fixed earlier on

PR #940, opened 2022-02-06
fix: fix race condition when plugin is auto-loaded

asaidimu:
Proposed fix for issue #937


kyazdani42:
closing because it's fixed in latest master normally :)

Issue #945, opened 2022-02-08
Working directory is wrong when launching Neovim by specifying a directory.


guilhermerodz:
**OS**
Ubuntu 20.04.3 LTS (Focal Fossa)

**Neovim version**
0.7.0-dev

**Nvim-tree version**
ea92e7b

**Describe the bug**
Since 8a6c7ba, the working directory is no longer synced EVEN if I specify a directory when launching Neovim.

(dropped a few guesses of mine at the end of this issue)

**To Reproduce**

On Terminal:
```bash
$ cd ~/a # navigate into folder A
$ nvim ~/b # attempt to launch nvim on folder B
```

Inside Neovim:
```
:pwd
/home/guilhermerodz/a << Should be "b" not "a".
```

NOTE: This behaviour wasn't happening before this commit: 8a6c7ba.

**Expected behavior**
- [ ] When launching Neovim, `nvim-tree` should automatically set the **current working directory** if specified in the command line.

---

- MAIN GUESS: I know there are very recent refactors. I think this issue is related to `nvim-tree/explorer/init.lua`. The `Explorer.new(cwd)` method is called. A new Explorer instance is created, then the `cwd` variable is set. But this logic causes Explorer's cwd to be unsynced with Neovim's current working directory, which remains unchanged.
- A NOTE: I took some time to compare our updated version to an older version. In the current version, at `nvim-tree/actions/change-dir`, the `M.fn` method stopped being called at startup.

PR #946, opened 2022-02-08
fix: change directory after setting Explorer.cwd

guilhermerodz:
Workaround for #945

### DIFF
--- a/lua/nvim-tree/explorer/init.lua
+++ b/lua/nvim-tree/explorer/init.lua
@@ -13,6 +13,7 @@ Explorer.__index = Explorer
 
 function Explorer.new(cwd)
   cwd = cwd or uv.cwd()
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/explorer/init.lua
```diff
@@ -13,6 +13,7 @@ Explorer.__index = Explorer
 
 function Explorer.new(cwd)
   cwd = cwd or uv.cwd()
+  vim.cmd('cd '..cwd)
```
kyazdani42, line 16: I'm not sure nvim-tree should manage your base directory, also this would loop when `:cd dir` and also crash when doing `nvim dir`

kyazdani42:
this is not the right place to put that :)

PR #956, opened 2022-02-09
Added a new option `auto_reload_on_write`

svermeulen:
I was finding that this plugin was causing a weird delay every time I saved a file, and traced it down to the `au BufWritePost` autocommand, which was reloading the explorer every time.  So I added support for a new option to disable this feature, for users like myself that prefer the extra performance when saving files.

I actually think the default should be to have this off, but since this would be a breaking change, I set the default as true.


kyazdani42 (approved):
Thanks for the PR. You are right, i'm aiming at using fs watchers instead to avoid reloading the whole tree each time to remove this autocmd.
Thanks for updating the docs :)

PR #978, opened 2022-02-14
fix: use fs_realpath to normalize path

younger-1:
I notice that `path_normalize` here is to change path like `/home/young/foo/..` to `/home/young`.

But it may not handle paths like `/home/young/../xavier` correctly.

 https://github.com/kyazdani42/nvim-tree.lua/blob/3f4ed9b6c2598ab8304186486a05ae7a328b8d49/lua/nvim-tree/utils.lua#L225-L229

### DIFF
--- a/lua/nvim-tree/explorer/init.lua
+++ b/lua/nvim-tree/explorer/init.lua
@@ -13,7 +13,7 @@ local Explorer = {}
 Explorer.__index = Explorer
 
 function Explorer.new(cwd)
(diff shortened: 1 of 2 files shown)


kyazdani42:
could you remove the utils code then please ? :) i didn't know about this, searching for path in the help did not give me this :(

younger-1:
> could you remove the utils code then please ? :)

Done.

kyazdani42 (approved):

kyazdani42:
thanks :)

PR #994, opened 2022-02-17
fix: also hijack netrw on VimEnter

zeertzjq:
If setup() is called before netrw is loaded, netrw needs to be hijacked on VimEnter instead.


kyazdani42 (approved):
thanks :)

PR #999, opened 2022-02-19
fix: add config for preserving window proportions on open file

rlch:
[Before](https://asciinema.org/a/kUQ5jCy8oujceqTtZFye5tAym)
[After](https://asciinema.org/a/D4jZgcfvyysKBmtqfcquBtadj)

Please let me know if there's anything I haven't considered in this PR

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -537,6 +537,10 @@ selectable. The default table is
       }
     }
 <
+|g:nvim_tree_preserve_window_proportions|     *g:nvim_tree_preserve_window_proportions*
+
+Can be 0 or 1. When 1, will preserve window proportions when opening a file. 
+Otherwise, the height and width of windows other than nvim-tree will be equalized.
 
 |g:nvim_tree_icon_padding|                *g:nvim_tree_icon_padding*
 
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -113,6 +113,7 @@ function.
         height = 30,
         side = 'left',
         auto_resize = false,
+        preserve_window_proportions = false,
         number = false,
         relativenumber = false,
 	signcolumn = "yes",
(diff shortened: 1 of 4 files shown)


lua/nvim-tree/view.lua
```diff
@@ -179,7 +179,10 @@ function M.resize(size)
   else
     a.nvim_win_set_height(M.get_winnr(), get_size())
   end
-  vim.cmd ":wincmd ="
+
+	if vim.g.nvim_tree_preserve_window_proportions ~= 1 then
```
kyazdani42, line 183: use a setup option please, i'm removing all vim global options iteratively.

kyazdani42: something like `view.preserve_window_propertions`

doc/nvim-tree-lua.txt
```diff
@@ -537,7 +543,6 @@ selectable. The default table is
       }
     }
 <
-
```
kyazdani42, line 540: uneeded change

kyazdani42 (commented):
Can you also add the default inside the default mappings in nvim-tree.lua ?
Also, add the default inside the help doc setup :)

rlch:
> Can you also add the default inside the default mappings in nvim-tree.lua ? 

The defaults are also in `view.lua`, is this fine or should they only be defined in one file?

kyazdani42:
Hum i made a mistake sorry, could you remove the view setup you added and add it in the view.lua file directly ? Sorry for the inconvenience

kyazdani42 (approved):
PR is good now, but i realized when opening a file, it will still equalize the windows because of the resize function.

rlch:
>  but i realized when opening a file, it will still equalize the windows because of the resize function.

What do you mean by this? When I open a file with `preserve_window_proportions=true` it doesn't equalize the windows. Is there an edge-case I haven't considered in the [demo](https://asciinema.org/a/D4jZgcfvyysKBmtqfcquBtadj)?

kyazdani42:
in some case, line 214 open-file will call view.resize which does `wincmd =`

rlch:
> in some case, line 214 open-file will call view.resize which does `wincmd =`

```lua
if not M.View.preserve_window_proportions then
  vim.cmd(":wincmd =")
end
```
Given the `wincmd =` in `view.resize` is guarded by the option, this won't be called form `open-file` right?

kyazdani42 (approved):
jeez i thought it was another function this was guarded :smile:

PR #1011, opened 2022-02-20
fix: use "" as flag to remove default mappings

xiyaowong:
Before, It will remove all user mappings which only have `key` and `cb`.  such as
```lua
mappings = {
      custom_only = false,
      list = {
        { key = 'h', cb = tree_cb 'close_node' },
        { key = { '<CR>', 'o', '<2-LeftMouse>', 'l' }, cb = tree_cb 'edit' },
        { key = 's', cb = tree_cb 'vsplit' },
        { key = 'i', cb = tree_cb 'split' },
      },
}
```

### DIFF
--- a/lua/nvim-tree/actions/init.lua
+++ b/lua/nvim-tree/actions/init.lua
@@ -2,6 +2,7 @@ local a = vim.api
 
 local lib = require'nvim-tree.lib'
 local view = require'nvim-tree.view'
+local util = require'nvim-tree.utils'
 local nvim_tree_callback = require'nvim-tree.config'.nvim_tree_callback
 
 local M = {
@@ -126,24 +127,34 @@ local function merge_mappings(user_mappings)
     return M.mappings
   end
 
+  local function is_empty(s)
+    return s == ''
+  end
+
   local user_keys = {}
   local removed_keys = {}
+  -- remove default mappings if action is a empty string
   for _, map in pairs(user_mappings) do
     if type(map.key) == "table" then
       for _, key in pairs(map.key) do
         table.insert(user_keys, key)
-        if not map.action then
+        if is_empty(map.action) then
           table.insert(removed_keys, key)
         end
       end
     else
       table.insert(user_keys, map.key)
-      if not map.action then
+      if is_empty(map.action) then
         table.insert(removed_keys, map.key)
       end
     end
+
     if map.action and type(map.action_cb) == "function" then
-      M.custom_keypress_funcs[map.action] = map.action_cb
+      if not is_empty(map.action) then
+        M.custom_keypress_funcs[map.action] = map.action_cb
+      else
+        util.warn("action can't be empty if action_cb provided")
+    end
     end
   end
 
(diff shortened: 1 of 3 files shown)


xiyaowong:
Although tree_cb and the cb property are deprecated, they're not really removed and many people including myself are still using them:)

kyazdani42 (changes requested):
can you allow nil to be passed too ?

xiyaowong:
In lua table, fields that don't exist are always nil. If set up in the same way as the example above, this line will remove all custom mappings  https://github.com/kyazdani42/nvim-tree.lua/blob/a1937ca079c785cb76d7470a9f248bced0459f05/lua/nvim-tree/actions/init.lua#L165

kyazdani42:
action should not be nil when user passes custom mappings though. The check could involve the empty string and the nil check. RN, if map.action is nil, the is_empty check will return false.

lua/nvim-tree/actions/init.lua
```diff
@@ -126,24 +127,34 @@ local function merge_mappings(user_mappings)
...
-      if not map.action then
+      if is_empty(map.action) then
         table.insert(removed_keys, map.key)
       end
     end
+
     if map.action and type(map.action_cb) == "function" then
-      M.custom_keypress_funcs[map.action] = map.action_cb
+      if not is_empty(map.action) then
+        M.custom_keypress_funcs[map.action] = map.action_cb
```
kyazdani42, line 154: here it will try to set [nil] = map.action_cb if the user forgets to pass in map.action and is nil

xiyaowong:
you can try this configuration
```lua
mappings = {
      custom_only = false,
      list = {
        { key = 'h', cb = tree_cb 'close_node' },
        { key = { '<CR>', 'o', '<2-LeftMouse>', 'l' }, cb = tree_cb 'edit' },
        { key = 's', cb = tree_cb 'vsplit' },
        { key = 'i', cb = tree_cb 'split' },
      },
}
```
then you will kown what's the problem

lua/nvim-tree/actions/init.lua
xiyaowong: You forgot the previous condition `if mal.action and...`.

kyazdani42:
you should then add a special check for `cb` which is deprecated. I really don't like the way it's been done anyway...
Did you try passing `{ key = "l", action_cb = function() end }` ?

xiyaowong:
> Although tree_cb and the cb property are deprecated, they're not removed.



> you should then add a special check for `cb` which is deprecated. I really don't like the way it's been done anyway... Did you try passing `{ key = "l", action_cb = function() end }` ?

Yes, displaying a warning message is the best way, I'm a long-time user of nvim-tree, but I just learned today that cb has been deprecated. I already changed my configuration.

kyazdani42:
let's merge this, we'll improve this bit by bit i suppose, this code needs a refresher.

alex-courtis:
Whoops, I didn't realise that keys could be specified without action, but with a valid callback.

Thanks for fixing that.

alex-courtis:
> this code needs a refresher

We could simplify things by denormalising the mapping lists when processing the mappings e.g.

```lua
      { key = {"<CR>", "o", "<2-LeftMouse>"}, action = "edit" },
```

becomes

```lua
      { key = "<CR>", action = "edit" },
      { key = "o", action = "edit" },
      { key = "<2-LeftMouse>", action = "edit" },
```

`merge_mappings` is then greatly simplified.

PR #1014, opened 2022-02-20
feat: Add a new keybinding to search and expand a certain path

be-ndee:


lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,41 @@
...
+      if matches then
+        if node.nodes then
+          if not node.open then
+            node.open = true
+            TreeExplorer:expand(node)
+            renderer.draw()
+          end
+          search_node(node.nodes)
+        else
+          -- TODO open file
```
kyazdani42, line 28: is this PR finished :smile: ?

be-ndee: You are too accurate ;-) I removed the TODO.

be-ndee:
I've merged the master branch into the feature branch, so there are no conflicts to resolve.

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,39 @@
+local utils = require'nvim-tree.utils'
+local git = require"nvim-tree.git"
+local renderer = require'nvim-tree.renderer'
+
+local M = {
+}
```
kyazdani42, line 6:
```suggestion
local M = {}
```
just a little thing :smile: not really important

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,39 @@
...
+
+function M.fn()
+  if not TreeExplorer then return end
+
+  local input_path = vim.fn.input("Search node: ", "", "file")
+  utils.clear_prompt()
+
+  absolute_input_path = TreeExplorer.cwd .. utils.path_separator .. input_path
```
kyazdani42, line 14: missing `local`.
Also you could use `utils.path_join` instead

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,39 @@
...
+
+  local function search_node(nodes)
+    for _, node in ipairs(nodes) do
+      matches = utils.str_find(absolute_input_path, node.absolute_path .. utils.path_separator)
+      if matches then
+        if node.nodes then
```
kyazdani42, line 20: could be done in one line `if matches and node.nodes then`

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,39 @@
...
+  if not TreeExplorer then return end
+
+  local input_path = vim.fn.input("Search node: ", "", "file")
+  utils.clear_prompt()
+
+  absolute_input_path = TreeExplorer.cwd .. utils.path_separator .. input_path
+
+  local function search_node(nodes)
+    for _, node in ipairs(nodes) do
+      matches = utils.str_find(absolute_input_path, node.absolute_path .. utils.path_separator)
```
kyazdani42, line 18: missing local. Also why adding a path separator ?

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,39 @@
...
+      if matches then
+        if node.nodes then
+          if not node.open then
+            node.open = true
+            TreeExplorer:expand(node)
+            renderer.draw()
```
kyazdani42, line 24: expand already does `renderer.draw` once it is finished. But actually, it's not a good idea to use this yet because it's asynchronous (i have to make it synchronous). Thus the following search_node could give an empty table.

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,39 @@
...
+        end
+      end
+    end
+  end
+
+  node = search_node(TreeExplorer.nodes)
+end
+
+function M.setup(options)
+end
```
kyazdani42, line 36: uneeded.

kyazdani42 (changes requested):
This look very similar to find-file. You could copy some code from there. I'm going to refacto the explorer and git to allow for using synchronously and redraw manually.

lua/nvim-tree/actions/search-node.lua
be-ndee: Done.

be-ndee:
I updated the code. Hope now everything looks fine :) Sorry, this was my first lua code I've ever written, so I'm new to it..

be-ndee (commented):
I made all the requested changes.

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,33 @@
...
+        if not node.open then
+          node.open = true
+          TreeExplorer:expand(node)
+        end
+        search_node(node.nodes)
+      end
+    end
+  end
+
+  search_node(TreeExplorer.nodes)
```
kyazdani42, line 29: you should redraw after that :)
also, might be interesting to copy more of the find-file algorithm and set the cursor on the found node ?

kyazdani42:
No worries code is clean enough for a first time :)

be-ndee:
The cursor position will be updated. Because of some index problems, I had to improve the whole search algorithm.

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,72 @@
...
+  if index > 0 and view.is_visible() then
+    if TreeExplorer.cwd ~= '/' and not view.View.hide_root_folder then
+      index = index + 1
+    end
+
+    view.set_cursor({index, 0})
+  end
+
+  if tree_altered then
+    renderer.draw()
```
kyazdani42, line 67: i think you should draw before setting the cursor, otherwise it might get an index out of bound error

kyazdani42 (commented):
almost good :) can you fix the lint issue ? (run `luacheck .`) to see the lint issue.

be-ndee:
Fixed the lint issue and prevent the out of index exception :) Hope we are good now ;)

lua/nvim-tree/actions/search-node.lua
```diff
@@ -0,0 +1,71 @@
...
+
+  local input_path = vim.fn.input("Search node: ", "", "file")
+  utils.clear_prompt()
+
+  local absolute_input_path = utils.path_join({
+    TreeExplorer.cwd,
+    input_path
+  })
+
+  print(absolute_input_path)
```
kyazdani42, line 18: unecessary print statement :smile:

be-ndee: absolutely right... 🤦

be-ndee (commented):
Hope it's all done now.

kyazdani42 (approved):
it is ! good job :) i think this will need to be extended incrementally to allow some `*` or `.` kind of thing. Completion for path also could be a good idea. We'll see ! thanks for the contribution :)

be-ndee:
Your welcome! Yes, and something like fuzzy finding would be helpful. Or the possiblity to open the path from "outside" of the plugin, so I could use fzf or telescope to search for a path and then focus it in the nvim-tree plugin.

Thanks for merging and thanks for the plugin. I did not regret switching from NERDTree to your plugin :)

gegoune:
How is that feature suppose to work? I tried `S` and typing in partial or full filename (that exists in repository) with cursor on top of the nvim-tree window and nothing happened after `<CR>`.

be-ndee:
I made some changes in #1022.

@gegoune you have to type in a full path. Than it jumps to that path in the tree. Example:

```
foo/
  bla/
    bar/
bar/
```

When you type in `bar`, then it should match the last directory on line 4. If you type in `foo/bla/bar` it will match the directory on line 3. Does this answer your qustion?

gegoune:
Yes, it does, thanks!

PR #1015, opened 2022-02-20
fix(doc): nvim_tree_create_in_closed_folder default value

ilkecan:
https://github.com/kyazdani42/nvim-tree.lua/blob/15afe3a8b4a058e7b841a67da1e61b420465e40f/lua/nvim-tree/fs.lua#L60

Note that the actual default value is `nil` but this is currently equivalent to `0` from the user's perspective.


kyazdani42 (commented):
thanks :)

PR #1019, opened 2022-02-22
feat: allow not changing the cwd when changing directories

Gelio:
Allow keeping the cwd constant when changing directories inside the tree. `TreeExplorer.cwd` is the source of truth for the current root directory of the tree.

With `enable = false`, this matches the behavior of netrw.

Not changing the cwd can break some other plugins, e.g. [the lualine extension for NERDTree](https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/extensions/nerdtree.lua) (which [the nvim-tree extension uses](https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/extensions/nvim-tree.lua)) relies on `cwd` to show the directory in the status line. Those should be modified to rely on `TreeExplorer.cwd` if someone disables `change_dir`. I did exactly that change in https://github.com/Gelio/ubuntu-dotfiles/pull/1 by creating my own lualine extension that reads the directory from `TreeExplorer.cwd`.

Related to https://github.com/kyazdani42/nvim-tree.lua/issues/860

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ require'nvim-tree'.setup {
   },
   actions = {
     change_dir = {
+      enable = true,
       global = false,
     },
     open_file = {
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -132,6 +132,7 @@ function.
       },
       actions = {
         change_dir = {
+          enable = true,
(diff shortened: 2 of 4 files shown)


kyazdani42 (approved):
thanks !

PR #1024, opened 2022-02-25
feat: add new keybinding "b" to run a command in the focused node

be-ndee:
Press "b" when you are in the tree. User can enter a shell command like "ls -la". That command is then run in the directory on which your cursor is, at the moment.


be-ndee:
Thanks for the idea, sounds good to me! But what do you mean with "you don't have to implement running the command yourself."?

Gelio:
Thanks!

> you don't have to implement running the command yourself.

I mean [this part](https://github.com/kyazdani42/nvim-tree.lua/pull/1024/files#diff-5833be1c49e26015af8e5a920739c0fd030c363a048f7abaf221c6559e29a178R25-R32):

image

Vim will take care of running the shell command if you prefix it with `!` (see `:h :!`)

kyazdani42:
I can make a quick review here by first saying:
1. gelio is right, you don't need the custom code for running a shell command
2. gelio idea is great
3. just remove the comments please :)

be-ndee:
@Gelio thanks, I will have a look at that.

@kyazdani42 will do that. I've though some comments would make the code easier to understand :)

Gelio:
> Do you know how to do that? I did not find anything to enter the command-line. I just found vim.cmd(), which executes a command. But what we want to do is to only enter the command-line with a predefined command.

Looking at the result of `:verbose map .` in vim-fugitive, I found out how to do it from Lua:

```lua
local path_to_file = "index.html"
vim.api.nvim_input(": " .. path_to_file .. "<Home>")

```

Now you only need to set the `path_to_file` to a real value.

be-ndee:
@Gelio thank you! That works perfectly!

I created to different keybindings:

`b` -> enters the command line with the directoryname
`B` -> enters command line with the file name

So if your cursor is on a file, than you could use `b` to run a command with the directory name or use `B` for the filename.

be-ndee:
@kyazdani42 I think this should be finished now.

lua/nvim-tree/actions/run-command.lua
```diff
@@ -0,0 +1,17 @@
+local M = {}
+
+function M.run_file_command(node)
+  vim.api.nvim_input(": " .. node.absolute_path .. "<Home>")
+end
+
+function M.run_directory_command(node)
```
Gelio, line 7: I'm not sure if we need a dedicated `run_directory_command`. It seems to correspond to `Pb` (`P` to go to the parent node and then `b`

Removing it would lower the complexity of this function a bit.

I'm not the maintainer, so don't take my word for it :smile: I'm just stating my point of view

lua/nvim-tree/actions/run-command.lua
```diff
@@ -0,0 +1,17 @@
...
+  local path = node.absolute_path
+  if not node.nodes then
+    path = string.sub(path, 1, string.len(path) - string.len(node.name) - 1)
```
Gelio, line 10: Clever! I suppose this method works for both Windows and Linux/OSX paths :+1:

README.md
```diff
@@ -221,6 +221,7 @@ require'nvim-tree'.setup {
 - Double right click acts like `<C-]>`
 - `W` will collapse the whole tree
 - `S` will prompt the user to enter a path and then expands the tree to match the path
+- `b` will prompt the user to enter a shel command which is the run inside the focused node (where the cursor is)
```
Gelio, line 224: Minor:

Looks like we are missing an explanation for `B`

Gelio, line 224: Minor, typo and a slight reword:

```suggestion
- `b` will prompt the user to enter a vim command which is run on the directory of the focused node (where the cursor is)
```

Gelio (commented):
Looks promising :+1:

lua/nvim-tree/actions/run-command.lua
be-ndee: Yepp, I removed it :) So we end with a function containing one line of code :-D

README.md
```diff
@@ -285,7 +286,8 @@ local list = {
   { key = "q",                            action = "close" },
   { key = "g?",                           action = "toggle_help" },
   { key = "W",                            action = "collapse_all" },
-  { key = "S",                            action = "search_node" }
+  { key = "S",                            action = "search_node" },
+  { key = "b",                            action = "run_file_command" }
```
Gelio, line 290: Minor:

I know it's late and it's another round of review, but personally I would rather have this mapping be under `.` to match vim-fugitive :smile: What do you think?

Gelio (approved):
LGTM! I would only change the mapping to be under `.` instead of `b` to match vim-fugitive, but I understand some people may not use that plugin

lua/nvim-tree/actions/run-command.lua
Gelio: Nice, simple is beautiful :smile:

README.md
kyazdani42: I agree it would make more sense. `b` is not really a good default for `run command`. I'd have set something like `x` or `e` (execute), but `.` would be ok too.

be-ndee: You could achieve that with creating a custom mapping when using nvim-tree :P

`{key = '.', action = 'run_file_command'}`

kyazdani42: what we mean here is having a default like this would make more sense :smile:

be-ndee: I was just joking ;) I've changed it to `.`

Gelio (approved):
Nice! LGTM

I'd only squash the commits into one before merging the PR so the diff is nicer for anyone following commit-by-commit

kyazdani42 (approved):

Issue #972, opened 2022-02-13
[bug] prev_git_item with hijack_cursor selects icon instead of previous file


alex-courtis:
**OS**
`Arch Linux 5.16.8-arch1-1`

**Neovim version**
```
NVIM v0.6.1
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

**Nvim-tree version**
0f7da14

**Describe the bug**
Executing `prev_git_item` when on a dirty file moves the cursor to that file's icon, instead of the previous file.

**To Reproduce**
empty nvim configuration, no plugins othen than nvim-tree and chriskempson/base16-vim

```
tree.setup {
	hijack_cursor = true,
}
```

Open a directory with many dirty files. Move the cursor to the last dirty file with `]c`. `[c` takes you to that last file's icon, instead of the previous file.

[![asciicast](https://asciinema.org/a/468658.svg)](https://asciinema.org/a/468658)

**Expected behavior**
Icon should not be selected; the previous dirty file should be.

PR #1025, opened 2022-02-26
#972 bug: prev_git_item with hijack_cursor selects icon instead of previous file

alex-courtis:
Rewrote `find_git_item`

Repurposed `get_node_at_line` as `get_nodes_by_line`

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -22,21 +22,23 @@ function M.init(foldername)
   end
 end
 
-local function get_node_at_line(line)
-  local index = view.View.hide_root_folder and 1 or 2
+function M.get_nodes_by_line(nodes_all, line_start)
+  local nodes_by_line = {}
+  local line = line_start
   local function iter(nodes)
     for _, node in ipairs(nodes) do
-      if index == line then
-        return node
-      end
-      index = index + 1
+      nodes_by_line[line] = node
+      line = line + 1
       if node.open == true then
         local child = iter(node.nodes)
-        if child ~= nil then return child end
+        if child ~= nil then
+          return child
+        end
       end
     end
   end
-  return iter
+  iter(nodes_all)
+  return nodes_by_line
 end
 
 function M.get_node_at_cursor()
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/lib.lua
```diff
@@ -50,7 +52,7 @@ function M.get_node_at_cursor()
   local line = cursor[1]
   if view.is_help_ui() then
     local help_lines = require'nvim-tree.renderer.help'.compute_lines()
-    local help_text = get_node_at_line(line+1)(help_lines)
+    local help_text = M.get_nodes_by_line(help_lines, 1)[line]
```
alex-courtis, line 55: This wasn't working properly when `hide_root_folder` set. Now it always starts at 1.

lua/nvim-tree/lib.lua
```diff
@@ -60,7 +62,7 @@ function M.get_node_at_cursor()
     if TreeExplorer.cwd == "/" then
       line = line + 1
     end
-    return get_node_at_line(line)(TreeExplorer.nodes)
+    return M.get_nodes_by_line(TreeExplorer.nodes, view.View.hide_root_folder and 1 or 2)[line]
```
alex-courtis, line 65: This will be a little slower, as get_nodes_by_line iterates all nodes, rather than returning early as `get_node_at_line` did.

kyazdani42 (approved):
A bit slower indeed but not that much.
The search algorithm could be simpler with a filter function but looks good enough like this.
could be a 3liner if implementing a filter for tables with keys :)

alex-courtis:
> A bit slower indeed but not that much. The search algorithm could be simpler with a filter function but looks good enough like this. could be a 3liner if implementing a filter for tables with keys :)

Yes. It doesn't seem prudent to try and optimise this now. keys / indexes are the better solution.

What key are you considering? absolute_path? Indexes would be very useful, perhaps line, file type, ignored...

These could be added without affecting existing code (apart from maintaining their state) and then existing code gradually refactored to use them.

kyazdani42:
could be indexed by line like now, and filtering out based on the line value and the existence of a git_status i suppose :) More specifically for this algorithm. Don't think it matters too much anyway !

PR #1027, opened 2022-02-26
add more sensible defaults for window_picker.excludes

alex-courtis:
There are many common windows - mostly plugin windows - that users will not want to open files in.

buftype `nofile` covers many vim and plugin cases such as man page, tagbar, bufexplorer etc. This saves the user from having to manually exclude all those filetypes.

filetype `diff` covers regular vim diff, as well as many plugins that use diffs such as vim-gitgutter and fugitive.

filetype `fugitive` is explicitly present as some of its buffers have an acwrite buftype. I added this as it is very popular, but happy to remove.


kyazdani42 (approved):
sensible choice :)

PR #1035, opened 2022-03-01
remove semicolons

be-ndee:


gegoune:
Perhaps @kyazdani42 could be convinced to add stylua ('s configuration)? :)

kyazdani42 (approved):

kyazdani42:
there is already luacheck, should i also add stylua :smile: ?

PR #1040, opened 2022-03-02
feat: add sort_by option to sort files by modification time

### LINKED ISSUE #1039: Change the sort order
I use nvim to edit my notes with ~/notes. This dir has dozens of files and it'd be nice to sort by recently updated instead of always alphabetically.

An option passed to setup like `sort_order = "modification_time"` or similar. A second option to control if dirs are always ordered first or not might be good, too.

None

bdwiel:
Fixes #1039 for files only. I'll wait for feedback on whether you like this feature or not before investigating why it's not working on directories. I was thinking that directories will always be sorted above files at the same level. So for each node, show `sorted(directories)` followed by `sorted(files/symlinks)`. What do you think?

### DIFF
--- a/lua/nvim-tree/explorer/node-builders.lua
+++ b/lua/nvim-tree/explorer/node-builders.lua
@@ -5,6 +5,16 @@ local M = {
   is_windows = vim.fn.has('win32') == 1
 }
 
+local function get_last_modified(absolute_path)
+  local stat = uv.fs_stat(absolute_path)
+
+  local last_modified = 0
+  if stat ~= nil then
+    last_modified = stat.mtime.sec
+  end
+  return last_modified
+end
+
 function M.get_dir_git_status(parent_ignored, status, absolute_path)
   if parent_ignored then
     return '!!'
@@ -30,6 +40,7 @@ function M.folder(absolute_path, name, status, parent_ignored)
     name = name,
     nodes = {},
     open = false,
+    last_modified = get_last_modified(absolute_path),
   }
 end
 
@@ -49,6 +60,7 @@ function M.file(absolute_path, name, status, parent_ignored)
     extension = ext,
     git_status = M.get_git_status(parent_ignored, status, absolute_path),
     name = name,
+    last_modified = get_last_modified(absolute_path),
   }
 end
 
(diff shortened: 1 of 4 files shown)


lua/nvim-tree/explorer/utils.lua
```diff
@@ -17,7 +17,16 @@ function M.node_comparator(a, b)
...
+  if M.config.sort_by == "modification_time" then
+      -- newer files have higher mtime
+      local a_mtime = uv.fs_stat(a.absolute_path, nil).mtime.sec
+      local b_mtime = uv.fs_stat(b.absolute_path, nil).mtime.sec
+      return b_mtime <= a_mtime
+  elseif M.config.sort_by == "name" then
+    return a.name:lower() <= b.name:lower()
+  else
+    return a.name:lower() <= b.name:lower()
+  end
```
kyazdani42, line 29: can you maybe just generate a different node_comparator functions based on the config instead of doing checks inside this ?

bdwiel: Sure. Any changes requested for the config names?

kyazdani42: Config name is good enough :)

lua/nvim-tree/explorer/node-builders.lua
```diff
@@ -42,13 +49,20 @@ end
 
 function M.file(absolute_path, name, status, parent_ignored)
   local ext = string.match(name, ".?[^.]+%.(.*)") or ""
+  local stat = uv.fs_stat(absolute_path)
+
+  local last_modified = 0
+  if stat ~= nil then
+    last_modified = stat.mtime.sec
+  end
```
kyazdani42, line 57: duplicated code, could you extract this in a get_last_modified function or something like that ?

lua/nvim-tree/explorer/utils.lua
```diff
@@ -20,6 +21,20 @@ function M.node_comparator(a, b)
...
+  if not (a and b) then
+    return true
+  end
+  if a.nodes and not b.nodes then
+    return true
+  elseif not a.nodes and b.nodes then
+    return false
+  end
+
+  -- sort newer files first
```
kyazdani42, line 34: uneeded comment :)

kyazdani42 (changes requested):
almost good, some code smell.

kyazdani42 (approved):
thanks, nice feature. would be nice to add a toggle action for this too :)

siduck:
@kyazdani42 
`sort_by = "modification_time"`

This option sorts by the last modified files. I want the opposite, like showing last modified files/dirs at the end

alex-courtis:
> sort_by = "modified_time_new" or "modified_time_old"

That's pretty specific. We could be more generic with something like `sort_reverse = false`. Please raise a feature request.

siduck:
> > sort_by = "modified_time_new" or "modified_time_old"
> 
> That's pretty specific. We could be more generic with something like `sort_reverse = false`. Please raise a feature request.

i doubt if sort_reverse would be helpful for other sort options like sort by name, extension etc

alex-courtis:
> i doubt if `sort_reverse` would be helpful for other sort options like sort by name, extension etc
> 
> The current `modification_time` can be renamed to `modified_time_new` and a new sort option could be added `modified_time_old` ( old files on top )
> 
> or `new_files` , `old_files`

Please raise a feature request, with a PR gratefully appreciated.

`sort_reverse` may not make sense for the current options (although it does work) however can be used for future sort options.

PR #1043, opened 2022-03-03
Close help on q/close instead of ignore keypress

smhc:
I believe this is better matches expectations when pressing 'q' on the help ui.

### DIFF
--- a/lua/nvim-tree/actions/init.lua
+++ b/lua/nvim-tree/actions/init.lua
@@ -81,6 +81,9 @@ local keypress_funcs = {
 }
 
 function M.on_keypress(action)
+  if view.is_help_ui() and action == 'close' then
+    action = 'toggle_help';
+  end
   if view.is_help_ui() and action ~= 'toggle_help' then return end
   local node = lib.get_node_at_cursor()
   if not node then return end


lua/nvim-tree/actions/init.lua
```diff
@@ -81,6 +81,9 @@ local keypress_funcs = {
 }
 
 function M.on_keypress(action)
+  if view.is_help_ui() and action == 'close' then
+    action = 'toggle_help';
+  end
```
kyazdani42, line 86: you toggle the help with `g?`, why does this need another binding ?

smhc: I think it's more natural to think of it as opening help, then using the generic close to close the current view. Most other "views" in vim works like this.  e.g :help vs :q. I hit 'q' every time I am on the help view to close it and feel stuck on that screen until I remember it's the same command to close it as open it.

I don't really see a reason why 'close' should stop working just because the help view is open - having an escape out that works as expected on both screens seems pretty helpful to the user.

No doubt you are comfortable with the current behaviour - but I'm fairly sure new and casual users would agree with the 'q to close either screen' sentiment.

kyazdani42 (approved):
makes sense. I don't use this feature actually i know the tree pretty well already :smile:

PR #1048, opened 2022-03-04
feature: add feature to increase/decrease size

be-ndee:


gegoune:
I never used `NvimTreeResize` before but how about bundling new functionality into that command?

`NvimTreeResize 50` - as of now: resize to 50
`NvimTreeResize +10` - increment by 10
`NvimTreeResize -10` - decrement by 10

be-ndee:
Good idea @gegoune ! I changed to code so you can use the NvimTreeResize function like you suggested it.

lua/nvim-tree/utils.lua
```diff
@@ -245,4 +245,10 @@ function M.table_create_missing(tbl, sub)
   return t
 end
 
+-- @param value string
+-- @return string
+function M.trim_string(value)
+  return (string.gsub(value, "^%s*(.-)%s*$", "%1"))
+end
```
kyazdani42, line 252: uneeded, there is `vim.trim` function already :)

be-ndee: Ah, perfect. Searched for a similar function in native lua...

kyazdani42 (approved):

PR #1051, opened 2022-03-05
New Bug Report Template

alex-courtis:
Try it out here: https://github.com/alex-courtis/nvim-tree.lua/issues

Heavily based on [telescope's](https://github.com/nvim-telescope/telescope.nvim).

Minimal config contains defaults. https://github.com/kyazdani42/nvim-tree.lua/issues/1050 will make synchronising these easier. We could use a completely minimal setup `require'nvim-tree'.setup {}` however that may make it difficult for the user to add settings and identify the defaults.

Works on my machine, I would be grateful for some testing of `nvt-min.lua`


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,171 @@
...
+        vim.fn.system { 'git', 'clone', '--depth=1', 'https://github.com/wbthomason/packer.nvim', install_path }
+        end
+        load_plugins()
+        require('packer').sync()
+        vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
+        vim.opt.termguicolors = true
+        vim.opt.cursorline = true
+
+        -- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
+        _G.setup = function()
```
alex-courtis, line 94: This could be split into a separate file, but that complicates things for the user.

alex-courtis:
/ping @gegoune as I cannot add reviewers to the PR ;)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,171 @@
...
+        end
+        if vim.fn.isdirectory(install_path) == 0 then
+        print("Installing nvim-tree and dependencies.")
+        vim.fn.system { 'git', 'clone', '--depth=1', 'https://github.com/wbthomason/packer.nvim', install_path }
+        end
+        load_plugins()
+        require('packer').sync()
+        vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
+        vim.opt.termguicolors = true
+        vim.opt.cursorline = true
```
alex-courtis, line 91: Not strictly necessary, however it makes it easier to identify issues around file focus etc.

.github/ISSUE_TEMPLATE/bug_report.yml
gegoune: I am wondering if it should list all available options or should users put required one by themselves.

alex-courtis: That is a very good question.

1. All defaults specified
\- Very verbose, large file
\+ Helps highlight which defaults the user has changed

1. User's changes only
\+ Clean setup, small file
\- User must know which defaults they have changed

The README and help directs the user to use 1.

I [prefer 2](https://github.com/alex-courtis/arch/blob/master/config/nvim/lua/nvt.lua) hovewer from bug reports it appears that many user prefer 1. They are likely just continuing with the setup that they have been directed to use.

The intent of the "clean setup" is to have the user replicate their issue with few config changes. However, if they completely replace the default config with their own, we should still be able to replicate.

Thoughts @kyazdani42 ?

kyazdani42: i'm not too sure, i'm a bit afraid that users just skip this because it's too lengthy. We could try and see how people behave with this new issue template, if it's too much, we can simplify this part.

kyazdani42 (approved):
let's try that for a while :) thanks

.github/ISSUE_TEMPLATE/bug_report.yml
gegoune: Let's see but I am leaning towards reports describing reproducible steps as: "I used `setup()` function with only option `A` set to a value which results in unexpected behaviour". Other defaults are relevant I think.

Thanks for doing leg work here! It should make @kyazdani42's life way easier now.

alex-courtis: It will be interesting to see what users actually do with the minimal config.

PR #1053, opened 2022-03-06
File logging

alex-courtis:
Add the ability to log to a file. One file per nvim session is created.

Some logging has been added:

Configuration
* default options merged with user
* merged mappings

Git runner
* git command
* stderr and stdout
* result: success, timeout or failure code

I have elected to put the logs in the user's home rather than /tmp, so that they don't forget about them. They may forget to turn logging off and many large files will accumulate.

### DIFF
--- a/lua/nvim-tree/git/runner.lua
+++ b/lua/nvim-tree/git/runner.lua
@@ -1,4 +1,5 @@
 local uv = vim.loop
+local log = require'nvim-tree.log'
 local utils = require'nvim-tree.utils'
 
 local Runner = {}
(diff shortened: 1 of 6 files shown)


lua/nvim-tree/git/runner.lua
```diff
@@ -100,11 +122,20 @@ function Runner.run(opts)
     list_ignored = opts.list_ignored,
     timeout = opts.timeout or 400,
     output = {},
-    _done = false
+    rc = nil, -- -1 indicates timeout
```
alex-courtis, line 125: Care has been taken to not alter the functionality or flow of control here. _done -> rc is the only significant change.

lua/nvim-tree/log.lua
```diff
@@ -0,0 +1,40 @@
...
+    return
+  end
+
+  M.raw(typ, string.format("[%s] [%s] %s\n", os.date("%H:%M:%S"), typ, fmt), ...)
+end
+
+function M.setup(opts)
+  M.config = opts.log
+  if M.config and M.config.enable and M.config.types then
+	  M.path = string.format("%s/nvim-tree-%s-%s.log", vim.env.HOME, os.date("%H:%M:%S"), vim.env.USER)
```
alex-courtis, line 35: Time is enough to uniquely identify the file. Date is too long.

kyazdani42, line 35: could we store that in `vim.stdpath('data')` ? I don't like programs not respecting XDG :)

alex-courtis: Using `vim.fn.stdpath("cache")` which is where nvim, packer and lsp put their logs.

Also printed this path at startup so that the user is aware.

alex-courtis:
I would be grateful for a successful test (notably the git bits) by @gegoune and @kyazdani42 before we merge this. There are a lot of moving parts.

kyazdani42 (approved):
great work thanks !

PR #1056, opened 2022-03-06
feat(live-filter): add ability to live filter out nodes in the tree

kyazdani42:
This PR adds the ability to filter out nodes visually in the tree. It is filtering and redrawing while typing, which should provide a good user experience.

~it is very much unfinished, there are some caveats.
I'm leaving this PR open until everything is properly finished to let people discuss what would be nice to implement and potential issues that this implementation would introduce.~

TODO:
- [x] fix filtering on grouped nodes
- [x] allow configuration for prompt
- [x] documentation

TODO in followup PR:
- filter opened buffers
- allow user to pass custom init / deinit / matching functions.
- filter with automatic depth search

### DIFF
--- a/lua/nvim-tree/actions/find-file.lua
+++ b/lua/nvim-tree/actions/find-file.lua
@@ -29,35 +29,37 @@ function M.fn(fname)
 
   local function iterate_nodes(nodes)
     for _, node in ipairs(nodes) do
-      i = i + 1
+      if not node.hidden then
+        i = i + 1
 
-      if not node.absolute_path or not uv.fs_stat(node.absolute_path) then
-        break
-      end
-
-      -- match against node absolute and link, as symlinks themselves will differ
-      if node.absolute_path == fname_real or node.link_to == fname_real then
-        return i
-      end
-      local abs_match = vim.startswith(fname_real, node.absolute_path .. utils.path_separator)
-      local link_match = node.link_to and vim.startswith(fname_real, node.link_to .. utils.path_separator)
-      local path_matches = node.nodes and (abs_match or link_match)
-      if path_matches then
-        if not node.open then
-          node.open = true
-          tree_altered = true
+        if not node.absolute_path or not uv.fs_stat(node.absolute_path) then
+          break
         end
 
-        if #node.nodes == 0 then
-          core.get_explorer():expand(node)
-        end
-
-        if iterate_nodes(node.nodes) ~= nil then
+        -- match against node absolute and link, as symlinks themselves will differ
+        if node.absolute_path == fname_real or node.link_to == fname_real then
           return i
         end
-        -- mandatory to iterate i
-      elseif node.open then
-        iterate_nodes(node.nodes)
+        local abs_match = vim.startswith(fname_real, node.absolute_path .. utils.path_separator)
+        local link_match = node.link_to and vim.startswith(fname_real, node.link_to .. utils.path_separator)
+        local path_matches = node.nodes and (abs_match or link_match)
+        if path_matches then
+          if not node.open then
+            node.open = true
+            tree_altered = true
+          end
+
+          if #node.nodes == 0 then
+            core.get_explorer():expand(node)
+          end
+
+          if iterate_nodes(node.nodes) ~= nil then
+            return i
+          end
+          -- mandatory to iterate i
+        elseif node.open then
+          iterate_nodes(node.nodes)
+        end
       end
     end
   end
--- a/lua/nvim-tree/actions/movements.lua
+++ b/lua/nvim-tree/actions/movements.lua
@@ -6,34 +6,25 @@ local lib = require "nvim-tree.lib"
 
 local M = {}
 
-local function get_line_from_node(node, find_parent)
+local function get_index_of(node, nodes)
   local node_path = node.absolute_path
+  local line = 1
 
-  if find_parent then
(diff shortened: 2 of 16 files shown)


suessflorian:
You rock @kyazdani42, keep it up :)

be-ndee:
Great feature!

An idea: at the moment it only filters the visible nodes. It would be cool to filter also not visible trees as well.

kyazdani42:
@be-ndee yep, there are some todos in the "live-filter.lua" file at the top. I think i'm happy with this implementation.
I still have some modules to fix before improving the filtering and searching logic. 
The issue with the in depth search would make it very slow. I'm not sure yet what the right solution would be for that scenario.

alex-courtis:
This is very good news. My use case:  vim-fugitive likes to open a buffer named `.git/index` which gets picked up by `update_focused_file`.

lua/nvim-tree/actions/movements.lua
```diff
@@ -17,19 +16,21 @@ local function get_line_from_node(node, find_parent)
     node_path = node.absolute_path:match("(.*)" .. utils.path_separator)
   end
 
-  local line = 2
+  local line = core.get_nodes_starting_line()
```
alex-courtis, line 19: I like this change.

alex-courtis:
An easy enhancement will be a static filter, say, `view.ignore_list` like `update_focused_file`.

Use case: vim-fugitive like to open a buffer named `.git/index`, which `update_focused_file` picks up and focuses, opening the .git directory, which annoys me. I could add `.git/*` to the `ignore_list` to remove the whole directory from the tree.

kyazdani42:
@alex-courtis i think this is good enough for a first implementation :) I've worked out all the issues i believe so this should be okay for a first release.
This will need to be extended (see the TODO in followup pr on the PR Comment).

alex-courtis:
Testing Notes, successful cases not mentioned.

Nice to have:
- action to clear the filter

Limitations:
- `always_show_folders = true` hides folders that contain matches unless that folder has already been opened

Bugs:
- `prefix = ""` shows the filter dialogue on the previous line 
20220515_160119 1606x1024
20220515_160148 1606x1024

lua/nvim-tree/actions/find-file.lua
```diff
@@ -29,35 +29,37 @@ function M.fn(fname)
 
   local function iterate_nodes(nodes)
     for _, node in ipairs(nodes) do
-      i = i + 1
+      if not node.hidden then
```
alex-courtis, line 32: This definitely works.

lua/nvim-tree/live-filter.lua
```diff
@@ -0,0 +1,127 @@
...
+  a.nvim_buf_set_keymap(overlay_bufnr, "i", "<CR>", "<cmd>stopinsert<CR>", {})
+end
+
+local function create_overlay()
+  configure_buffer_overlay()
+  overlay_winnr = a.nvim_open_win(overlay_bufnr, true, {
+    col = 1,
+    row = 0,
+    relative = "cursor",
+    width = math.max(20, a.nvim_win_get_width(view.get_winnr()) - #M.prefix - 2),
```
alex-courtis, line 101: This doesn't look so good for narrow nvim-tree windows.

The prefix can be cut off (right side tree) or cause the nvim-tree window to scroll right (left side tree).

There will be a nice solution; it may come to me in my sleep :P

alex-courtis (changes requested):
It works. Let's fix these before merge:

* action to clear the filter
* `prefix = ""` using wrong line

Later:

The input window works. My edge cases are overcritical. Act only if someone complains or finds significant bugs.

Hiding matching files when `always_show_folders = true` is a problem, but can be addressed in the future. It will need a filesystem search to resolve.

kyazdani42:
@alex-courtis i've managed to fix the empty filter issue, and added a keybinding to clear the filter (i completely forgot to add it :smile:)
I'm not sure what you mean by 
> Hiding matching files when always_show_folders = true is a problem, but can be addressed in the future. It will need a filesystem search to resolve.

always show folders should not change behavior of the filter. It only displays folders even without matches to allow opening folders manually which should show matching files if there are any.

Although the search is mentioned in my comment, i decided to not add it here because i would make the feature more complex, as a first draft, i decided to keep it simple for now. This will be added in a future version, once we find a proper way to do it without bursting the CPU ^^

gegoune:
Didn't try this feature yet, but thinking if implementation could be done is such a way that it would allow for more than just filtering by name - filter out not open buffers or filter out not git modified/staged/etc buffers (if has git mark) would be very useful.

kyazdani42:
@gegoune see the comment on PR :smile:

gegoune:
@kyazdani42 Ah, awesome, did not see updates to original comment. Sorry for the noise!

alex-courtis:
> I'm not sure what you mean by

First we use `always_show_folders = true` and can find `view.lua`

Then we use `always_show_folders = false` and cannot find `view.lua` until _after_ we have opened `lua/nvim-tree` once.

This isn't a blocker, just a limitation of the filter. We can `S` for `view.lua` as it does a filesystem search.

[asciicast](https://asciinema.org/a/494523)

alex-courtis (approved):
Tested some cases when searching filtered trees. No surprises.

PR #1057, opened 2022-03-08
feat: make it possible to collapse the tree but keep the directories open which are in used in buffers

be-ndee:
Sometimes I want to "clean up" my nvim-tree but keep the directories open, which are opened in buffers at the moment.


be-ndee:
Don't say anything - I've forgot the documentation...

kyazdani42 (approved):

Issue #504, opened 2021-07-10
Thinks a symlinked directory is empty


d-miketa:
Symlinked directories are highlighted with the empty folder colour, even if the directories they point to are non-empty. I'd expect the symlink dirs to have a completely independent colour, regardless of content, follow the colour scheme of normal folders or have their own, content-dependent scheme.

kyazdani42:
sorry for the delays, looks like a bug to me, doesn't happen with my config. Could you provide a small sample config so i can reproduce this please ?

d-miketa:
Sure! The issue looks like this:
<img width="217" alt="image" src="https://user-images.githubusercontent.com/320321/131383666-c1ef67db-be9e-4575-9907-b654c92dde01.png">
The top folder is symlinked, the bottom folder is empty.

The following is with a minimal init file:
<img width="192" alt="image" src="https://user-images.githubusercontent.com/320321/131387043-aca3082e-1020-4e3b-9151-83d93207fe48.png">

Notice no colour distinction between the three types of dictionary: populated, empty and symlinked. It's possible to define different colours for populated on the one hand and empty or symlinked on the other, but it makes more sense to distinguish all three categories.

`nvim-tree-debug.lua`
```
# install packer
local execute = vim.api.nvim_command
local fn = vim.fn

local install_path_start = fn.stdpath('data')..'/site/pack/packer/start/packer.nvim'
local install_path_opt = fn.stdpath('data')..'/site/pack/packer/opt/packer.nvim'

if fn.empty(fn.glob(install_path_start)) + fn.empty(fn.glob(install_path_opt)) > 0 then
    fn.system({'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path_start})
    execute 'packadd packer.nvim'
end

# install nvim-tree
return {
    require('packer').startup(function(use)
        use 'wbthomason/packer.nvim'
        use 'kyazdani42/nvim-tree.lua'
    end)
}
```
`nvim -u nvim-tree-debug.lua -U NONE -N -i NONE`
(potentially running `:PackerSync` once and then rebooting if neovim picks up on a pre-existing set of plugins installed via Packer.)

alex-courtis:
We no longer have colours to distinguish empty etc.

[...]

PR #1089, opened 2022-03-19
#504 determine whether symlinked directories are empty when building

alex-courtis:
Only issue might be performance, however we are already scanning directories...


kyazdani42 (approved):
i don't think it matters much :) thanks

Issue #837, opened 2021-12-09
Use relative path to find file


kuznetsss:
I think using absolute path in searching for file is wrong because there could be symlinks.

For example in my system I open file from home directory:
`~$ nvim Documents/vimwiki/index.md`
But `Documents` is symlink `Documents -> /mnt/Data/home_user/Documents`
So the real absolute file path is `/mnt/Data/home_user/Documents/vimwiki/index.md` which nvim-tree tries to find being in `/home/user` directory. So the file can't be found.

I think it would be better to search file by relative path from nvim's cwd.

kyazdani42:
you are correct, we should be searching by relative path. You could open a PR if i didn't fix this by the end of the day.

alex-courtis:
Current behaviour:

```
mkdir /tmp/root1
mkdir /tmp/root2
cd /tmp/root2
mkdir dir3
touch dir3/file31
cd /tmp/root1
mkdir dir1
mkdir dir2
touch file1
touch dir1/file11
cd /tmp/root1
ln -s /tmp/root2/dir3 dir31
ln -s ../root2/dir3 dir32
nvim -nu /tmp/nvt-min.lua
```

Success:
dir1/file11
dir31/file31
dir32/file31
dir1
dir2
dir31
dir32
file1

Fail:
file11
file31

It appears that the issues with the relative path of symlinks, be they absolute or relative. Search is successful for file11 and file31 if we C-] into dir1 dir31 or dir32 first.

alex-courtis:
This does highlight a problem, which might be in my understanding: we can only search for files/dirs which match a path under cwd, not for files/dirs of a name.

Proposal: search matches full file/dir names under cwd, as per `:find` behaviour.

alex-courtis:
There are still some lingering issues around symlinked rel vs absolute:

find-file absolute path: node will be opened, however it will not be focussed
find-file path relative to cwd: node will not be found

alex-courtis:
Resolved via #1092 #1107

PR #1092, opened 2022-03-20
#837 search subdirs

alex-courtis:
It is possible that I am missing the point, however it seems to me that search is not working.

Changes:
- rewrote to use filesystem search; node search was problematic and complex
- matches input that exactly matches the end of a path e.g. `file1` `dir1/` `dir1/file11` but not `ir1/file11` or `le11`
- adds `:find` style command line completion for all files under cwd

Resolves:
- bug in which searching would result in two nodes would be created for the same file

Does not:
- search under the current node; searches under cwd; this could be added

Exposed:
- find-file bug around symlinks #1093

### DIFF
--- a/lua/nvim-tree/actions/search-node.lua
+++ b/lua/nvim-tree/actions/search-node.lua
@@ -1,95 +1,75 @@
+local api = vim.api
+local uv = vim.loop
 local utils = require "nvim-tree.utils"
-local view = require "nvim-tree.view"
-local renderer = require "nvim-tree.renderer"
 local core = require "nvim-tree.core"
+local find_file = require("nvim-tree.actions.find-file").fn
 
 local M = {}
+
+local function search(dir, input_path)
+  local path, name, stat, handle, _
+
+  if not dir then
+    return
+  end
+
+  handle, _ = uv.fs_scandir(dir)
+  if not handle then
+    return
+  end
+
+  name, _ = uv.fs_scandir_next(handle)
+  while name do
+    path = dir .. "/" .. name
+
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
indeed seemed like it wasn't working. Good idea to fallback to find_file

be-ndee:
Okay, maybe I gave the action a miss leading name :-D A better name would have been "go to file". So my intention was that you can enter a path to jump to that directory/file in the tree. So you don't have to click or navigate through the whole tree. But "your" search is also a good feature. But it would be good, that the search checks the options for ignored files. Because I have a node_modules directory, which is in the gitignore and not visible for me in the tree, but the search searches in the node_modules, so it's quite slow.

kyazdani42:
why not just using `/` then ? I don't understand the point of just moving the cursor to a file ? The search should not search inside ignored directories right ?

alex-courtis:
> search checks the options for ignored files

That is a great idea, as the file system search indeed does not obey .gitignore etc.

Probably as an option: there are currently three "flavours" of ignored: git_ignored, dotfiles and custom. Perhaps obeying all three of those would be a "good enough" solution. There will be an elegant an simple solution, however will take some thinking about.

alex-courtis:
Raised #1099

2 seems like a workable solution, however I have a feeling that there is something simpler.

alex-courtis:
> why not just using `/` then

That matched the functionality of S when I was testing. Finding files in subdirectories was not working.

It seems that I was misunderstanding the purpose of S. Please accept my apologies @be-ndee

PR #1093, opened 2022-03-20
#837 find file symlinks fixes

### LINKED ISSUE #837: Use relative path to find file

alex-courtis:
#837 and #1092 exposed some remaining issues with symlink resolution. This covers the find-file cases.

### DIFF
--- a/lua/nvim-tree/actions/find-file.lua
+++ b/lua/nvim-tree/actions/find-file.lua
@@ -1,3 +1,4 @@
+local uv = vim.loop
 local view = require "nvim-tree.view"
 local utils = require "nvim-tree.utils"
 local renderer = require "nvim-tree.renderer"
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/actions/find-file.lua
```diff
@@ -13,17 +14,35 @@ function M.fn(fname)
...
   local function iterate_nodes(nodes)
     for _, node in ipairs(nodes) do
       i = i + 1
-      if node.absolute_path == fname then
-        return i
+
+      local stat, _ = uv.fs_stat(node.absolute_path)
+      if not stat then
+        break
+      end
```
alex-courtis, line 33: Extra safety check cannot hurt.

lua/nvim-tree/actions/find-file.lua
```diff
@@ -13,17 +14,35 @@ function M.fn(fname)
...
+        break
       end
 
-      local path_matches = node.nodes and vim.startswith(fname, node.absolute_path .. utils.path_separator)
+      -- match against node absolute and real, for the case of symlinks, which will differ
+      if node.absolute_path == fname_real or real_path == fname_real then
+        return i
+      end
+      local abs_match = vim.startswith(fname_real, node.absolute_path .. utils.path_separator)
+      local real_match = vim.startswith(fname_real, real_path .. utils.path_separator)
```
alex-courtis, line 44: We could keep a "real path" on the node, however that seems too complicated.

kyazdani42: There is a `link_to` attributes for symlinks on node, maybe this is what you would like to use here ?

alex-courtis: > There is a `link_to` attributes for symlinks on node, maybe this is what you would like to use here ?

Goot thinking. I'll rework on the weekend.

alex-courtis: It is now simpler, thank you.

kyazdani42 (approved):
thanks :)

PR #1098, opened 2022-03-22
fix: window picker bug on global statusline

### LINKED ISSUE #1096: Window picker messing with global statusline
### Description

Because the window picker changes the statusline layout, when using the new `global statusline` (https://github.com/neovim/neovim/issues/9342) the statusline appears _weird_.

image

Suggestion:
Show the window picker information in floating windows.

### Neovim version

```markdown
NVIM v0.7.0-dev+1292-ge9b53f3fb
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by kassioborges@kassios-mbp.lan

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/local/Cellar/neovim/HEAD-e9b53f3/share/nvim"

```
```


macOS 12.3

### nvim-tree version

e87ee0e3c30368fa97085125ffe13ee37c0daf70

### Steps to reproduce

1. Set global statusline
2. Open multiple splits
3. Open window picker

### Expected behavior

_No response_

### Actual behavior

_No response_

### Minimal config

```Lua
-
```

ichigozero:
This PR attempts to fix #1096. 

I set the NvimTree status line to blank when global status is set and window picker is activated. When window picker is deactivated, I restores NvimTree status line to its previous state. 

See below screenshots for reference.

Before patch
before

After patch
after

### DIFF
--- a/lua/nvim-tree/actions/open-file.lua
+++ b/lua/nvim-tree/actions/open-file.lua
@@ -58,6 +58,19 @@ local function pick_window()
   local laststatus = vim.o.laststatus
   vim.o.laststatus = 2
 
+  if laststatus == 3 then
+    local ok_status, statusline = pcall(api.nvim_win_get_option, tree_winid, "statusline")
+    local ok_hl, winhl = pcall(api.nvim_win_get_option, tree_winid, "winhl")
+
+    win_opts[tree_winid] = {
+      statusline = ok_status and statusline or "",
+      winhl = ok_hl and winhl or "",
+    }
+
+    api.nvim_win_set_option(tree_winid, "statusline", " ")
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):

aspeddro:
Why set the NvimTree status line to blank?


This is changing the colors of `NvimTree`
image

The version prior to this PR is more consistent with user experience.

kyazdani42:
thing is, if you go from laststatus 3 to laststatus 2, the bar will appear in nvim-tree in a different way than you are used to be, which would be less consistent imo. The window picker should be a simple "can you select" or "cant you select", not displaying personal buffer information i guess. But maybe this could also be expanded as if the buffer cannot be selected, the statusline display another color that is disabled.

Issue #791, opened 2021-11-19
Slowness when setting update_cwd to true on windows


kishikaisei:
So the title says it all, for some reason, when I am setting `update_cwd` to `true` makes `nvim-tree` slow, movement, expanding nodes, and the first couple of seconds of opening a file, movement in that file is slow as well (moving up and down for instance, takes a moment to register).

OS:      Windows 11 (21H2) (but had the same issue on Windows 10)
NVIM: v0.6.0-dev+156

Settings: nvim-tree.lua
```
local g   = vim.g
local cmd = vim.cmd

g.nvim_tree_gitignore = 1               -- 0 by default
g.nvim_tree_indent_markers = 0          -- 0 by default, this option shows indent markers when folders are open
g.nvim_tree_git_hl = 1                  -- 0 by default, will enable file highlight for git attributes (can be used without the icons).
g.nvim_tree_highlight_opened_files = 0  -- 0 by default, will enable folder and file icon highlight for opened files/directories.
g.nvim_tree_root_folder_modifier = ':t' -- This is the default. See :help filename-modifiers for more options
g.nvim_tree_add_trailing = 1            -- 0 by default, append a trailing slash to folder names
g.nvim_tree_group_empty = 0             -- 0 by default, compact folders that only contain a single folder into one node in the file tree
g.nvim_tree_disable_window_picker = 0   -- 0 by default, will disable the window picker.
g.nvim_tree_icon_padding = ' '          -- one space by default, used for rendering the space between the icon and the filename. Use with caution, it could break rendering if you set an empty string depending on your font.
g.nvim_tree_symlink_arrow = ' ➛ '      -- defaults to ' ➛ '. used as a separator between symlinks' source and target.
g.nvim_tree_respect_buf_cwd = 1         -- 0 by default, will change cwd of nvim-tree to that of new buffer's when opening nvim-tree. -- set to 1
g.nvim_tree_create_in_closed_folder = 1 -- 1 by default, When creating files, sets the path of a file when cursor is on a closed folder to the parent folder when 0, and inside the folder when 1.
g.nvim_tree_refresh_wait = 500          -- 1000 by default, control how often the tree can be refreshed, 1000 means the tree can be refresh once per 1000ms.
g.nvim_tree_window_picker_exclude = {
	['filetype'] = {
		'notify',
		'packer',
		'qf'
	},
	['buftype'] = {
		'terminal'
	}
}
-- default shows no icon by default
[...]

PR #1108, opened 2022-03-26
#791 add profiling for some operations

alex-courtis:
Add timing to some operations. We can add more if more slow operations are identified.

Should help us with #791

### DIFF
--- a/lua/nvim-tree/actions/find-file.lua
+++ b/lua/nvim-tree/actions/find-file.lua
@@ -1,3 +1,4 @@
+local log = require "nvim-tree.log"
 local view = require "nvim-tree.view"
 local utils = require "nvim-tree.utils"
 local renderer = require "nvim-tree.renderer"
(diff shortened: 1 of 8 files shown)


lua/nvim-tree/actions/find-file.lua
```diff
@@ -13,6 +14,8 @@ function M.fn(fname)
   end
   running[fname] = true
 
+  local ps = log.profile_start("find file %s", fname)
```
alex-courtis, line 17: This is being invoked (with draw) twice when
```lua
		update_focused_file = {
			enable = true,
			update_cwd = true,
		},
```

That might explain some of the slowness in #791.

Either way, we probably should fix that ;)

kyazdani42 (approved):
nice addition thanks :) I think the timing could've been made with the libuv but we'll see if this is needed in the future

alex-courtis:
> I think the timing could've been made with the libuv

Indeed. I looked at :help but didn't look at `vim.inspect(vim)` :facepalm: 

#1110

PR #1111, opened 2022-03-27
document performance issues and add to bug report template

alex-courtis:
This should cover everything in #549 which we should be able to close.

I've not added the disclaimer about git status performance; it should be clear to the user ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ log = {
 
 Please attach `$XDG_CACHE_HOME/nvim/nvim-tree.log` if you raise an issue.
 
-### Tips
+*Performance Tips:*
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -334,10 +334,37 @@ You can toggle the help UI by pressing `g?`.
 
 ## Diagnostic Logging
 
-You may enable diagnostic logging and a file `nvim-tree-HH:MM:SS-username.log` will be created in `$XDG_CACHE_HOME/nvim`, usually `~/.cache/nvim`, containing logs from that nvim session. See `:help nvim-tree.log`.
```
alex-courtis, line 337: forgot to remove that when I changed the file name

kyazdani42 (approved):
Thanks :)

PR #1116, opened 2022-03-28
(readme typo) fix missing " and "toggle without focus" example

rakanzaidm:
fixing the second " that was missing, and changing the second parameter for the toggle to true because that's how to enable the no focus toggle (as in no_focus = true)


alex-courtis (approved):
Thank you

PR #1133, opened 2022-04-03
add open_on_setup_file option

### LINKED ISSUE #372: Auto open tree when opening a file on startup
I am trying to have nvim-tree shown as soon as vim opens. I tried the following:

`autocmd VimEnter * NvimTreeOpen`

Even though Nvim-Tree is shown, it is empty. I think it is opening in the wrong directory. If I click ENTER it does open some dir (the ".." dir").

alex-courtis:
closes #372

The name is a bit ugly however I don't want to change the existing `open_on_setup` option and break people.


kyazdani42 (approved):

brainfucksec:
Thanks for this! (I think that the instruction name is intuitive anyway)

PR #1134, opened 2022-04-03
add view.indent_markers

### LINKED ISSUE #654: [feature request] add custom indent char by user
i wanted to change the default indent marker char.'
i couldnt find something in the readme or setup.
i found this in the source code:
```lua
local function get_padding_indent_markers(depth, idx, tree, _, markers)
  local padding = ""
  if depth ~= 0 then
    local rdepth = depth/2
    markers[rdepth] = idx ~= #tree.entries
    for i=1,rdepth do
      if idx == #tree.entries and i == rdepth then
        padding = padding..'└ '
      elseif markers[i] then
        padding = padding..'│ ' -- <-----------
      else
        padding = padding..'  '
      end
    end
  end
  return padding
end
```
i guess here the indent char is located

i just want to specifiy this from a global variable or setup function.
please add this feature.

alex-courtis:
closes #654

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ let g:nvim_tree_show_icons = {
 "1 by default, notice that if 'files' is 1, it will only display
 "if nvim-web-devicons is installed and on your runtimepath.
 "if folder is 1, you can also tell folder_arrows 1 to show small arrows next to the folder icons.
-"but this will not work when you set view.indent_markers.enable (because of UI conflict)
+"but this will not work when you set renderer.indent_markers.enable (because of UI conflict)
 
 " default will show icon by default if no icon is provided
 " default shows no icon by default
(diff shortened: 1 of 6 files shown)


lua/nvim-tree/renderer/init.lua
```diff
@@ -284,4 +284,12 @@ function M.render_hl(bufnr)
   end
 end
 
+function M.setup(opts)
+  M.config = {
+    indent_markers = opts.view.indent_markers,
+  }
+
+  require("nvim-tree.renderer.padding").setup(opts)
+end
```
alex-courtis, line 293: I know renderer needs a big refactor... this is just a small step.

README.md
```diff
@@ -136,6 +135,14 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
     number = false,
     relativenumber = false,
     signcolumn = "yes",
+    indent_markers = {
```
kyazdani42, line 138: do you think it would make more sens to move the option in a new `renderer.indent_markers` ?

alex-courtis: That works. All the renderer config :g can go in there.

I'll update on the weekend.

lua/nvim-tree/legacy.lua
```diff
@@ -169,6 +169,13 @@ local migrations = {
       o.actions.change_dir.global = vim.g.nvim_tree_change_dir_global == 1
     end
   end,
+
+  nvim_tree_indent_markers = function(o)
+    utils.table_create_missing(o, "view.indent_markers")
+    if o.view.indent_markers.enable == nil then
+      o.view.indent_markers.enable = vim.g.nvim_tree_indent_markers == 1
+    end
+  end,
```
kyazdani42, line 178: looks like this hasn't been updated :)

kyazdani42 (changes requested):

lua/nvim-tree/legacy.lua
alex-courtis: Good catch

kyazdani42 (approved):

Issue #1059, opened 2022-03-09
Bug: sometimes the collapse command leads to duplicate file entries in the tree


be-ndee:
### Description

This is a bug, I sometimes experience, but at the moment I don't have certain "rules" when this happens. I hope that somebody has an idea, when this could happen or which part of the code could be a possible problem, so we could narrow down the problem.

Sometimes when I use `W` to collapse the whole tree, it results in duplicate file entries in the tree-view. Given the following directories and files:

```
dir-a/
  file-a1
  file-a2
dir-b/
  file-b1
  file-b2
```

Now my nvim-tree window looks like this:

```
dir-a/
  file-a1
  file-a2
dir-b/
```

Then I use `W` and the nvim-tree looks like this (which is correct):

```
dir-a/
dir-b/
```

But then, when I open the `dir-a`, it will look like this:

```
dir-a/
  file-a1
  file-a1
  file-a2
  file-a2
dir-b/
```

And when I open `dir-b`, it looks like this:

```
dir-a/
  file-a1
  file-a1
  file-a2
  file-a2
dir-b/
  file-b1
  file-b2
```

**KEEP IN MIND:** This is just an example, it can also be possible, that the `dir-b` also has duplicate file entries. Or everything is correct. I think it depends on, which directories are expanded and which files are open in buffers, but I could not reproduce "rules" when this happens exactly...

### Neovim version

```markdown
NVIM v0.7.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
[...]

PR #1143, opened 2022-04-09
#1059 protect against duplicates

alex-courtis:
Avoid duplicates during `populate_children`.

It is not clear how this could happen, however this is safe.


kyazdani42 (approved):
I'm a bit worried about performance here but we'll see how it behaves :)

Issue #1145, opened 2022-04-09
Preview command shouldn't open new buffers


sarmong:
When I use `tab` key to fire `preview` command, a new buffer opens for each file.
This is a bit annoying, because it clutters the buffer line and isn't really a preview anymore.

It would be nice to close the buffer of a previously _previewed_ file after opening another one, so that there can be only one _previewed_ file at a time.

As additional feature, a buffer could be made persistent if a user enters it with a cursor.

This may be a bit tricky to implement, but it kinda resembles the features of modern IDEs

kyazdani42:
I think this is a bit tricky to implement indeed, because neovim works with buffers only :smile:
What may be done is setting some options on the preview buffers to not appear in the buffer list.

lu5je0:
It maybe better to show preview in a float window.
![1](https://user-images.githubusercontent.com/17700877/162582632-8bec065e-b155-4a92-9d3a-60ca28b88876.gif)

alex-courtis:
> close the buffer of a previously _previewed_ file after opening another one

> options on the preview buffers to not appear in the buffer list

preview close will only hide the buffer as per normal `:close` rules, and I can't see any options to change that. However, we could just attempt a `:bdelete` after closing the window. We could be smart in deleting e.g. only delete if the buffer was not open before the preview.

> It maybe better to show preview in a float window.

That would be great! How did you put together that demo @lu5je0 ?

alex-courtis:
See #135 for preview floating windows.

Let's keep this issue about deleting buffers after preview.

sarmong:
I found a `bufhidden` option. 
Not sure which option would suit better for this case.

From the help:
```
	This option specifies what happens when a buffer is no longer
	displayed in a window:
	  <empty>	follow the global 'hidden' option
	  hide		hide the buffer (don't unload it), also when 'hidden'
			is not set
	  unload	unload the buffer, also when 'hidden' is set or using
			|:hide|
	  delete	delete the buffer from the buffer list, also when
			'hidden' is set or using |:hide|, like using
			|:bdelete|
	  wipe		wipe out the buffer from the buffer list, also when
			'hidden' is set or using |:hide|, like using
			|:bwipeout|

```

lu5je0:
> That would be great! How did you put together that demo @lu5je0 ?

I created a floating window by `nui.nvim`, and edit file in it.😆

alex-courtis:
[...]

PR #1149, opened 2022-04-10
feat: delete preview buffers #1145

sarmong:
This adds `bufhidden = delete` for the preview buffer, which means that the buffer will be deleted after it is no longer displayed in a window.

This value gets reset with the autocommand:
if text changes in normal or insert mode, then `bufhidden` resets and now equals to the user's `hidden` option. 

So, if you change text in preview buffer, the buffer will persist.

Also, check for `buf_loaded` is needed, so that if you press `Tab` on an already opened file, it will not become temporary with `bufhidden`

See issue #1145

### DIFF
--- a/lua/nvim-tree/actions/open-file.lua
+++ b/lua/nvim-tree/actions/open-file.lua
@@ -194,6 +194,15 @@ function M.fn(mode, filename)
   local do_split = mode == "split" or mode == "vsplit"
   local vertical = mode ~= "split"
 
+  -- Check if file is already loaded in a buffer
+  local buf_loaded = false
+  for _, bufId in ipairs(api.nvim_list_bufs()) do
+    if api.nvim_buf_is_loaded(bufId) and filename == api.nvim_buf_get_name(bufId) then
+      buf_loaded = true
+      break
+    end
+  end
+
   -- Check if filename is already open in a window
   local found = false
   for _, id in ipairs(win_ids) do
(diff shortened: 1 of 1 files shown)


sarmong:
Apparently, neovim v.0.7 introduces lua-way of defining autocommands, so this piece may be rewritten

lua/nvim-tree/actions/open-file.lua
```diff
@@ -194,6 +194,15 @@ function M.fn(mode, filename)
   local do_split = mode == "split" or mode == "vsplit"
   local vertical = mode ~= "split"
 
+  -- Check if file is already loaded in a buffer
+  local buf_loaded = false
+  for _, bufId in ipairs(api.nvim_list_bufs()) do
```
kyazdani42, line 199: bufId should be buf_id

sarmong: Thanks, changed. 
Javascript influence;)

kyazdani42 (approved):

ShiChenCong:
it's seem does not work, `is_already_loaded` always return `true`

alex-courtis:
> it's seem does not work, `is_already_loaded` always return `true`

Thanks for looking into this.

Please raise a bug and include `nvt-min.lua` so that we may replicate.

ShiChenCong:
> > it's seem does not work, `is_already_loaded` always return `true`
> 
> Thanks for looking into this.
> 
> Please raise a bug and include `nvt-min.lua` so that we may replicate.

https://github.com/kyazdani42/nvim-tree.lua/issues/1307

PR #1151, opened 2022-04-10
Feat - Add open and close tree events

HPRIOR:
Add 'open' and 'close' callbacks or 'hooks' which can be set during setup and which fire when toggling the sidebar. 

This can be used to interact with other plugins. For example, the position of a tab bar can be changed according to the width of the file tree when it is opened or closed.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -339,6 +339,16 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
       },
     },
   },
+  renderer = {
+    indent_markers = {
+      enable = false,
+      icons = {
+        corner = "└ ",
+        edge = "│ ",
+        none = "  ",
+      },
+    },
+  },
   hijack_directories = {
     enable = true,
     auto_open = true,
@@ -374,6 +384,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
     timeout = 400,
   },
   actions = {
+    use_system_clipboard = true,
     change_dir = {
       enable = true,
       global = false,
@@ -452,6 +463,7 @@ function M.setup(conf)
   require("nvim-tree.git").setup(opts)
   require("nvim-tree.view").setup(opts)
   require("nvim-tree.lib").setup(opts)
+  require("nvim-tree.renderer").setup(opts)
 
   setup_vim_commands()
   setup_autocommands(opts)
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -10,6 +10,7 @@ local utils = require "nvim-tree.utils"
 local change_dir = require "nvim-tree.actions.change-dir"
 local legacy = require "nvim-tree.legacy"
 local core = require "nvim-tree.core"
+local events = require "nvim-tree.events"
 
 local _config = {}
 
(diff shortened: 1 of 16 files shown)


alex-courtis (changes requested):
Please document in help and readme.

See CONTRIBUTING.md

lua/nvim-tree.lua
```diff
@@ -414,6 +417,8 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
       profile = false,
     },
   },
+  open_hook = function () end,
+  close_hook = function () end,
```
alex-courtis, line 421: Please default these to nil and run `update-default-opts.sh`.

alex-courtis (changes requested):

alex-courtis:
I think @HPRIOR has started something big that will grow.

Let's name configs so that we can add to it. Perhaps:
```lua
  hooks = {
    node_open = nil,
    node_close = nil,
  },
```

Maybe `actions.hooks`.

@kyazdani42 you have name ideas ;)

alex-courtis (changes requested):
These should be nvim-tree.events e.g. `nvim-tree.events.on_nvim_tree_open`

We could pass the node, like action callbacks.

HPRIOR:
Hey @alex-courtis, I'll implement the suggested changes soon. 

With regard to your previous comment (I'm not sure where it has gone) an additional section in the config would be a good idea if 'hooks' are to be added elsewhere.  Something like:

```lua
hooks = {
    tree_open = nil,
    tree_close = nil,
},
```
Would it be possible to explain how the event system works?

kyazdani42 (commented):
you don't need to setup the hooks yourself, the user will call the the event registration himself, no need to set it during setup :)

kyazdani42:
i'm not even sure you placed the event at the right location to be honest :smile: i think i would place them in view.lua. There are multiple ways the tree can be opened, it won't just work by setting it in toggle :)

HPRIOR:
@kyazdani42 thanks for your feedback. I have moved the event dispatches to `view.lua`. Let me know if there is anything else I can do. If the code is not to your liking otherwise, I would really like to see this feature at some point!

alex-courtis:
> With regard to your previous comment an additional section in the config would be a good idea

That was a mistake on my part... before I realised that we should use events ;)

alex-courtis (changes requested):
We need to document this in the help: nvim-tree.events

HPRIOR:
> We need to document this in the help: nvim-tree.events

I've added the documentation

alex-courtis (approved):

kyazdani42 (approved):

gegoune:
Another useful event would be `on_tree_leave` (or similar) for when window is not closed but no longer active.

kyazdani42:
what do you mean no longer active ? when the tree buffer is replaced by a file buffer ?

gegoune:
Apologies for not being clear. I mean event when cursor leaves nvim-tree window. Wither because file was opened or by any window movement, such as `<C-w><C-w>`.

kyazdani42:
well...
```
au WinLeave NvimTree_* ...
```
?

gegoune:
Absolutely, that's what I have at the moment, and with on open event I can replace one of autocmds, but only one. Leaving tree remains as autocmd rather than event, which feels bit odd.

kyazdani42:
Well i'm not sure how it would be different inside the plugin ^^'

gegoune:
That's absolutely alright, just thought that it could make events more complete. I am fine sticking to auto commands.

kyazdani42:
:smile: now you can define autocmd in lua with callbacks so i think it doesn't matter too much :)

gegoune:
And this is exactly what I have. Thanks though and apologies for the noise. :)

PR #1160, opened 2022-04-12
fix: fix mis-naming for system clipboard config

tshen8:
Documentation suggests that `opts.actions.use_system_clipboard` is used:
```  
actions = {
    use_system_clipboard = true,
    [...]
```
but the  actual property read is `use_sys_clipboard`.

Most likely just a typo - renaming it to match the documentation to avoid any confusion.

Related PR: https://github.com/kyazdani42/nvim-tree.lua/pull/1141


kyazdani42 (approved):
indeed, thanks :)

PR #1163, opened 2022-04-14
feat: add file completion

be-ndee:


be-ndee:
I just noticed, with this change the text after adding a file is not correct. It now says "Create file/home/ndee/foo/home/ndee/foo was properly created". Instead of just "/home/ndee/foo was properly created"... I have to fix this, before we can merge this pull request.

alex-courtis (approved):
Awesome: completes paths nicely.

kyazdani42 (approved):

Issue #1037, opened 2022-03-02
Restrict tree traversal to current working dir


steliyan:
## Is your feature request related to a problem? Please describe.
Restrict the tree traversal to the current working dir.

## Additional context
My workflow consists of opening a project dir and staying there all the time. If I need to switch between projects, I use `tmux`, so I don't really want to be able to go outside my working dir.

## Describe the solution you'd like
Integrate the following code to the plugin itself. :))

_\* Of course, handle edge-cases as I haven't tested a lot of scenarios, although I didn't hit any problems with my monkey-patching approach._ 🙈 

## Describe alternatives you've considered
Initially, I've tried passing a callback to the `dir_up` action, but this doesn't work if you have the root folder shown as you can change it, not by `-`, but by pressing `Enter` on it.

```lua
local change_dir_action = require('nvim-tree.actions.change-dir')
local original_force_dirchange = change_dir_action.force_dirchange

local function force_dirchange(foldername, ...)
  if vim.fn.getcwd() > foldername then
    return
  end

  original_force_dirchange(foldername, ...)
end

change_dir_action.force_dirchange = force_dirchange
```

## Config
```lua
require('nvim-tree').setup({
  hijack_netrw = true,
  hijack_cursor = true,
  nvim_tree_respect_buf_cwd = false,
  update_cwd = false,
  actions = {
    open_file = {
      quit_on_open = true,
    },
    change_dir = {
      enable = false,
      global = false,
    },
  },
  update_focused_file = {
    enable = true,
    update_cwd = false,
    ignore_list = {},
  },
  git = {
    enable = true,
    ignore = false,
    timeout = 500,
  }
})
```

Maybe this is something that can go in the wiki, although it's a bit hacky.

kyazdani42:
I agree it's a bit hacky and a bit too specific to be implemented in the plugin. I guess a simple wiki page would do the trick :)

alex-courtis:
I hacked this into change-dir with a restrict option: #1171 https://github.com/kyazdani42/nvim-tree.lua/pull/1171/files

It works for the situations I tried: global, change_cwd, close/open.

What do you think about this as an option @kyazdani42 ?

alex-courtis:
An alternative for you @steliyan : add a `.` [custom action](https://github.com/alex-courtis/arch/blob/master/config/nvim/lua/amc/nvim-tree.lua#L8).

alex-courtis:
[...]

PR #1171, opened 2022-04-16
add actions.change_dir.restrict_above_cwd

alex-courtis:
#1037

### DIFF
--- a/lua/nvim-tree/actions/change-dir.lua
+++ b/lua/nvim-tree/actions/change-dir.lua
@@ -8,6 +8,7 @@ local M = {
   current_tab = a.nvim_get_current_tabpage(),
   options = {
     global = false,
+    restrict = true,
     change_cwd = true,
   },
 }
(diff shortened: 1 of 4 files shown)


lua/nvim-tree/actions/change-dir.lua
```diff
@@ -19,6 +20,7 @@ function M.fn(name, with_open)
 
   local foldername = name == ".." and vim.fn.fnamemodify(utils.path_remove_trailing(core.get_cwd()), ":h") or name
   local no_cwd_change = vim.fn.expand(foldername) == core.get_cwd()
+    or not M.options.global and M.options.restrict and foldername < vim.fn.getcwd(-1, -1)
```
kyazdani42, line 23: shouldn't it be `core.get_cwd()` instead of vim.fn.getcwd ?

alex-courtis: > shouldn't it be `core.get_cwd()` instead of vim.fn.getcwd ?

I think I used that so that it works with the other options: global and change_cwd. I shall experiment further.

The question is whether you are happy for this to go in; there is an alternative for the user: https://github.com/kyazdani42/nvim-tree.lua/issues/1037#issuecomment-1100576089

alex-courtis: We definitely want to check against the global cwd, not core's, which may be underneath.

Tested with permutations of change_cwd and global.

global works exactly as documented: the user cannot change a directory upwards. We could disable restrict_above_cwd when global is set, however I prefer to let the user configure this.

kyazdani42 (approved):

PR #1177, opened 2022-04-18
#1166 validate user's options

### LINKED ISSUE #1166: Validate configuration against schema
Keys of a table passed into `setup()` function are plenty, nested and can be very easily misplaced by users. There were issues reported in the past which were simply result of wrong nesting. 

It would greatly help if plugin was validating table against schema ensuring passed in configuration is valid.

Do nothing.

**Additional context**
Edit: this is how b0o achieved it in incline https://github.com/b0o/incline.nvim/blob/main/lua/incline/config/schema.lua.

alex-courtis:
closes #1166 

Quite easy: validate user's options against DEFAULT_OPTS.

@geoboom I would be very grateful if you can test, test, test.

Edit: apologies @geoboom I meant to ask @gegoune to test this one ;)


lua/nvim-tree.lua
```diff
@@ -424,7 +430,41 @@ local function merge_options(conf)
...
+
+  walk_options(conf, DEFAULT_OPTS, "")
+
+  if #msg > 0 then
+    utils.warn(msg)
+  end
+end
+
 function M.setup(conf)
+  validate_options(conf)
```
kyazdani42, line 466: missing or `{}` in case user doesn't pass anything in there.

kyazdani42 (changes requested):
I like the implementation :)

lua/nvim-tree.lua
alex-courtis: Good catch. Guards against non-table.

gegoune:
Got `unknown option: filters.custom.1` so either this PR needs rebasing or me asking for that feature was not unreasonable after all. :)

kyazdani42:
i think the ignore list might be incorrect ?

gegoune:
Screenshot 2022-04-21 at 21 18 04

In terms of notification itself seems like there is extra empty line.

gegoune:
> i think the ignore list might be incorrect ?

```lua
  filters = {
    custom = { '.git', 'node_modules', '.cache', '.DS_Store' },
  },
```
Nothing really fancy in there I think.

kyazdani42:
not yours @gegoune , the one in the PR ^^

alex-courtis:
> Got `unknown option: filters.custom.1` so either this PR needs rebasing or me asking for that feature was not unreasonable after all. :)
> 
> Edit: according to readme `filters.custom` is a table, so it should not warn me about it, should it?

That is defiritely not working correctly. It should go in `NO_VALIDATE_OPTS`.

However, reporting an itable index is also wrong and needs to be fixed.

alex-courtis:
> In terms of notification itself seems like there is extra empty line.

That can be fixed.

gegoune:
@alex-courtis it's default config for rcarriga/nvim-notify.

alex-courtis:
Hardened and rewrote:
* removed ignore list; handle empty default tables during validate
* checks all options in each table after a failure
* use one line for message; newlines upset everything
* refactored the legacy migrations / warnings to happen before validation
* `system_open.cmd` default is now "" to avoid the nil value problem

kyazdani42 (approved):
very nice improvement :) thanks !

kyazdani42:
@gegoune does this work now for you as expected ?

gegoune:
> @gegoune does this work now for you as expected ?

Yes, it does. Looks good now and detects incorrect keys, at least handful that I tried Thanks!

alex-courtis:
> > @gegoune does this work now for you as expected ?
> 
> Yes, it does. Looks good now and detects incorrect keys, at least handful that I tried Thanks!

Thanks for the testing @gegoune 

I think we can merge now.

lua/nvim-tree.lua
```diff
@@ -324,6 +324,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
   view = {
     width = 30,
     height = 30,
+    hide_root_folder = false,
```
alex-courtis, line 326: There was an extra `hide_root_folder` in the above table. It doesn't appear to be used, so I deleted it.

kyazdani42: good catch, it wasn't in the right place !

kyazdani42 (approved):

gegoune:
Thanks for that! Hopefully it will prevent configuration-issue issues.

eeeXun:
I set up width with function. Not working anymore?
```lua
require("nvim-tree").setup({
    view = {
        width = function()
            return math.floor(vim.o.columns * 0.225)
        end,
    }
})
```

alex-courtis:
> [NvimTree] invalid option: view.width expected: number actual: function

Interesting use case: functions for configuration options is something I would never have considered.
`+` great user flexibility
`-` poor stability
`-` unpredictable behaviour when a varying value is used

Alternatives:
1. No change: do not allow functions
2. Allow a function when the type is not a table
3. Allow a function always
4. Provide a proper mechanism for this, something like `view.width_ratio`

I like flexibility; 2 is a good compromise. We would need to document that with warnings and a rule that bugs cannot be raised when such functions are used. @kyazdani42 ?

gegoune:
How about 'allow functions but require returned value of specific type and validate it'?

alex-courtis:
> How about 'allow functions but require returned value of specific type and validate it'?

That would be great, have you got a specific idea on how we would do this?

The user could do _anything_ in that function; I don't think we could validate it by executing it, as it may throw an exception when run at plugin startup.

gegoune:
Ah, that's a good point. Can't really think of a solution here.

kyazdani42:
there is only a function available for the tree width / height but the implementation for the validation is good enough :)

eeeXun:
Maybe provide an option like `relative_width`/`relative_height`. If true, then the number of `width`/`height` should be percentage of columns/lines. If false, the it should be actual columns/lines.

PR #1187, opened 2022-04-24
#1169 document event payloads

### LINKED ISSUE #1169: Document Event Payloads
`nvim-tree.events` mostly pass payloads, however we have not documented those payloads.

Add to :help

alex-courtis:
closes #1169 

We could autogenerate this from comments somehow, but that is complex and provides little value.


kyazdani42 (approved):
thanks :)

Issue #1194, opened 2022-04-24
Unknow option warning


aspeddro:
### Description

After 5bbd3a0a917122b9330dab10b55bb56cc18c9a3d, nvim-tree show warning for valid options

![image](https://user-images.githubusercontent.com/16160544/164976220-cbabf086-d92d-48ea-a675-59062879a862.png)

### Neovim version

```markdown
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

### Operating system and version

Linux arch 5.17.3-arch1-1

### nvim-tree version

5bbd3a0

### Steps to reproduce

1. Open nvim

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
      "kyazdani42/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end
```

kyazdani42:
cannot reproduce, can you send a minimal configuration please ?

gegoune:
@kyazdani42 I am sorry ;)

[...]

PR #1195, opened 2022-04-24
#1166 validate config

kyazdani42:
@alex-courtis there are a few edge cases that were not covered
somehow https://github.com/kyazdani42/nvim-tree.lua/issues/1194 but i cannot reproduce
and when an option can have different types (such a view.width/view.height)


alex-courtis:
> somehow #1194 but i cannot reproduce

That's a use case I did not account for: value table with defaults, rather than empty.

I'll think about options. Perhaps the ignore list is still a good idea.

alex-courtis:
> and when an option can have different types (such a view.width/view.height)

I'm not sure that is a [good idea](https://github.com/kyazdani42/nvim-tree.lua/issues/new/choose), however we can work with it.

alex-courtis:
Let's try this again:
* functions are allowed as values; I have added a warning to the doc
* integer indexed tables are not checked, as they are not strongly typed and have variable length contents

alex-courtis:
Tested:
* empty options
* default options
* my setup
* #1194
* function as per https://github.com/kyazdani42/nvim-tree.lua/pull/1177#issuecomment-1107832916

alex-courtis (approved):

kyazdani42:
let see if this works :smile:

PR #1198, opened 2022-04-25
feat: case sensitive sorter

vlaw:
I'm not sure if this is useful or not. But I guess having one more option is not a bad idea. 😄

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -252,7 +252,7 @@ Here is a list of the options available in the setup call:
 
 *nvim-tree.sort_by*
 - |sort_by|: changes how files within the same directory are sorted. can be
-  one of 'name' | 'modification_time'
+  one of 'name' | 'case_sensitive' | 'modification_time'.
   type: `string`
   default: `"name"`
 
--- a/lua/nvim-tree/explorer/sorters.lua
+++ b/lua/nvim-tree/explorer/sorters.lua
@@ -75,7 +75,7 @@ function M.merge_sort(t, comparator)
   split_merge(t, 1, #t, comparator)
 end
 
-function M.node_comparator_name(a, b)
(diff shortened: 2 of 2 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -252,7 +252,7 @@ Here is a list of the options available in the setup call:
 
 *nvim-tree.sort_by*
 - |sort_by|: changes how files within the same directory are sorted. can be
-  one of 'name' | 'modification_time'
+  one of 'name' | 'Name' (case sensitive) | 'modification_time'
```
alex-courtis, line 255: Great idea, but the value is too complex.

How about `name_case_sensitive`

lua/nvim-tree/explorer/sorters.lua
```diff
@@ -116,8 +128,10 @@ function M.setup(opts)
   M.sort_by = opts.sort_by
   if M.sort_by == "modification_time" then
     M.node_comparator = M.node_comparator_modification_time
+  elseif M.sort_by == "Name" then
```
alex-courtis, line 131: name should be the default; the else

alex-courtis (changes requested):

doc/nvim-tree-lua.txt
vlaw: The new option that I use is “Name”， which means case sensitive name sorter. 
'name' is still the default option, case insensitive name sorter.
I guess the doc part is misleading, need some updates.

lua/nvim-tree/explorer/sorters.lua
vlaw: nae is still the default. 'Name' is for case sensitive.

kyazdani42 (approved):

alex-courtis (approved):

PR #1201, opened 2022-04-26
NVimTreeResize make size arg optional.

ratmice:
If not specified use the default size.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -60,11 +60,16 @@ Print clipboard content for both cut and copy
 
 |:NvimTreeResize|			          *:NvimTreeResize*
 
-Resize the NvimTree window to the given size. Example: `:NvimTreeResize 50`
-resizes the window to the width of 50. If the size starts with "+" or "-" it
-adds or removes the given value to the current window width.
-Example `:NvimTreeResize -20` removes the value 20 from the current width. And
-`:NvimTreeResize +20` adds the value 20 to the current width.
+Resize the NvimTree window to the given size. Takes an optional size argument.
+
+Example `:NvimTreeResize 50` resizes the window to the width of 50.
+
+If the size starts with "+" or "-" it adds or removes the given value to the
+current window width. Example `:NvimTreeResize -20` removes the value 20 from
+the current width. And `:NvimTreeResize +20` adds the value 20 to the current
+width.
+
+If a size is not given, resizes the window back to the default size.
 
 |:NvimTreeCollapse|			          *:NvimTreeCollapse*
 
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -257,7 +257,7 @@ local function setup_vim_commands()
     command! NvimTreeClipboard lua require'nvim-tree.actions.copy-paste'.print_clipboard()
     command! NvimTreeFindFile lua require'nvim-tree'.find_file(true)
     command! NvimTreeFindFileToggle lua require'nvim-tree'.toggle(true)
-    command! -nargs=1 NvimTreeResize lua require'nvim-tree'.resize("<args>")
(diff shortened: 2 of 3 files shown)


ratmice:
The gist of this is that if you `nvim .` and then open a file, the window gets split in half (not using the default size).
I worked around this by adding something like
`autocmd BufLeave * ++nested if bufname() == 'NvimTree_' . tabpagenr() | NvimTreeResize | endif`

So in this case it helps to set the size back to the default.

ratmice:
fwiw I think this probably doesn't work as intended, but I'm super confused as to why it started behaving how I want,
Since I forgot to install the plugin from my clone even :confused:.

kyazdani42:
I think you do not need to change the existing implementation in order to make that work.

ratmice:
I merely push since I had figured it out before I saw your message,
I'm not sure how I would do it without changing the existing implementation.

I can certainly do it by hard coding the size, but have to hardcode either the size, or the reference to the config table.
Or do you mean there is another way which doesn't involve hard coding it?

kyazdani42:
what i mean is that the current implementation works already by resetting the default tree size when passing no argument. You just need to remove the mandatory argument from the function, do not change the existing implementation :)

ratmice:
Hmm, It doesn't seem to be the case for me, it believe as is it sets it to the current window size, rather than the default setup size.

ratmice:
An alternate solution occurred to me that instead of allowing a nil size, we could add a function like `require("nvim-tree").get_config()`, for retrieving the default width/height to pass into `resize`.  That might be more generally useful since you could use it for finding other configuration options?

kyazdani42:
i don't believe so. the default width/height is already saved on the view. Also the function should be working, maybe the command doesn't send a nil value in the function ?

ratmice:
Ahh perhaps, I'm really new to vim/lua scripting, I'm pretty sure my script was matching the `size == ""` test rather than the nil one.

ratmice:
So, I just added a bunch of logging, and it's both, special casing that to `if size == "" size = nil end`, It still falls through to the current size.

For the record:
```
function! LeaveNvimTree()
        NvimTreeResize
endfunction
function! EnterNvimTree()
        NvimTreeResize 30
endfunction
autocmd BufLeave * ++nested if bufname() == 'NvimTree_' . tabpagenr() | call LeaveNvimTree() | endif
autocmd BufEnter * ++nested if bufname() == 'NvimTree_' . tabpagenr() | call EnterNvimTree() | endif

lua <<EOF
require('nvim-tree').setup{
  view = {
    width = 20,
    height = 20,
  },
}
```
the logging `get_size()` calls, here always prints 30, 
https://github.com/kyazdani42/nvim-tree.lua/blob/ce2cf7131f58cef26262c1932ba543d009dc99f7/lua/nvim-tree/view.lua#L227-L231

kyazdani42:
the get_size will return what you have in your configuration and updated if you pass a value to the function resize.
If you resize to 30, of course it will be 30.

ratmice:
Alright, then I'm just going to unmark as draft, as this has the behavior I was after, and works for me.
However if you think the other mechanism I proposed `get_config()` or perhaps `get_original_config()`, for being able to get the `setup` configuration is a better way to achieve this.  That would at least allow is to avoid having store the original config in `view.lua`; then I am open to closing this and doing that instead.

kyazdani42:
but this will break the basic functionality. The resize is expected to use the get_size() that has been set after the resize call with a value. I cannot accept the PR in this state :/

ratmice:
I don't get it, this should **only** affect when its not called with an argument, or is called with an empty string, otherwise it falls through to the existing path.  In *almost all* cases the behavior you describe leads to resize with no argument setting the size to the existing size, except in the one case where the window was split in half and the size was out of sync.

That said, i'm just going to close this and try the `get_config` approach, which doesn't involve any changes that affect view, or geometry, it seems more generally useful.  That way I can just get the value I want to call `resize` with from the original configuration directly.

PR #1215, opened 2022-04-30
1210 silent migrate more options

### LINKED ISSUE #1210: bring back update_to_buf_dir
update_to_buf_dir option is unavailable

update_to_buf_dir option is back

**Additional context**
Seems to have been removed via this commit: https://github.com/kyazdani42/nvim-tree.lua/commit/3ac443207f21bed29a2a3c2402f64098842ab753

image

alex-courtis:
closes #1210

### DIFF
--- a/lua/nvim-tree/legacy.lua
+++ b/lua/nvim-tree/legacy.lua
@@ -179,19 +179,17 @@ local g_migrations = {
 }
 
 local function refactored(opts)
-  if opts.view then
-    if opts.view.mappings then
-      if opts.view.mappings.list then
-        for _, m in pairs(opts.view.mappings.list) do
-          if m.action == "toggle_ignored" then
-            m.action = "toggle_git_ignored"
-          end
-        end
+  -- mapping actions
+  if opts.view and opts.view.mappings and opts.view.mappings.list then
+    for _, m in pairs(opts.view.mappings.list) do
+      if m.action == "toggle_ignored" then
+        m.action = "toggle_git_ignored"
       end
     end
   end
 
-  if opts.update_to_buf_dir then
+  -- update_to_buf_dir -> hijack_directories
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/legacy.lua
```diff
@@ -190,6 +190,27 @@ local function refactored(opts)
       end
     end
   end
+
```
alex-courtis, line 193: These are ugly.

I started to write a generic function to do this, but it was more ugly.

kyazdani42: could you refactor the 3 line nested if statement into one in `refactored` ? Doesn't look right :smile:

lua/nvim-tree/legacy.lua
```diff
@@ -190,6 +190,27 @@ local function refactored(opts)
...
+  if opts.view then
+    if opts.view.auto_resize ~= nil then
+      utils.table_create_missing(opts, "actions.open_file")
+      if opts.actions.open_file.resize_window == nil then
+        opts.actions.open_file.resize_window = opts.view.auto_resize
+      end
+      opts.view.auto_resize = nil
+    end
+  end
```
kyazdani42, line 213: i think we can remove this though. A better strategy for the plugin stability is needed. I'll think about this and propose a solution that could work :)

lua/nvim-tree/legacy.lua
alex-courtis: Collapsed all.

lua/nvim-tree/legacy.lua
alex-courtis: Yes. This is very hacky.

However... we could just remove most of this code in a few months, as users will have migrated.

kyazdani42 (approved):

Issue #1216, opened 2022-04-30
Collected Diagnostic Sign Issues


alex-courtis:
There are some issues with diagnostic sign visibility and correctness at fd2332a

Collect any bugs discovered in here.

I have seen some sporadic problems with signs not updating, however I have not been able to reproduce.

alex-courtis:
#1157 (#1174) has been resolved

alex-courtis:
#1213 has been raised however has no reproducer

kyazdani42:
I think we could move the diagnostics update into the renderer actually, it'd be less inaccurate.

alex-courtis:
> I think we could move the diagnostics update into the renderer actually, it'd be less inaccurate.

Yes.

Having the renderer able to render indicators in the signs column would be useful: we could then put other indicators e.g. git in there.

alex-courtis:
There are timing issues with diagnostics, that we don't see with other signs e.g. git. The diagnostics are often not updated until _after_ the tree is rendered. This is a poor user experience.

We are able to handle git changes perfectly as we are reacting to all events that cause changes to git signs e.g. buffer writes. Diagnostics are lazy and late.

**Improvement:**
Render the diagnostics upon change, rather than just on tree render.
For nvim lsp we can use [vim.lsp.diagnostic.on_publish_diagnostics](https://neovim.io/doc/user/lsp.html)

alex-courtis:
#1243 updates on CocDiagnosticChange

kyazdani42:
During this week, i've experienced a few diagnostics issues, using the default lsp diagnostics provided by neovim. I had to manually refresh the tree because the diagnostics where not updated.
I'm not sure where this comes from, but i hope i can reproduce this more consistently.

alex-courtis:
We are using the `DiagnosticChanged` for lsp updates. [vim.lsp.diagnostic.on_publish_diagnostics](https://neovim.io/doc/user/lsp.html) may provide more consistent timings and perhaps even allow us to see atomic updates.

Coc has no option but the `CocDiagnosticChange` event.

kyazdani42:
ah i didn't know they added this method. I'll take a look at that.

crosscode-nl:
@kyazdani42 Is this worked on, or a separate issue created? If it is not worked on, I think I can try to work on it because it is bothering me. :)

alex-courtis:
[...]

PR #1218, opened 2022-04-30
chore(diagnostics): move updates to renderer

kyazdani42:
should address #1216

### DIFF
--- a/lua/nvim-tree/actions/change-dir.lua
+++ b/lua/nvim-tree/actions/change-dir.lua
@@ -3,7 +3,6 @@ local a = vim.api
 local log = require "nvim-tree.log"
 local utils = require "nvim-tree.utils"
 local core = require "nvim-tree.core"
-local diagnostics = require "nvim-tree.diagnostics"
 
 local M = {
   current_tab = a.nvim_get_current_tabpage(),
@@ -42,7 +41,6 @@ function M.force_dirchange(foldername, with_open)
   else
     require("nvim-tree.renderer").draw()
   end
-  diagnostics.update()
 
   log.profile_end(ps, "change dir %s", foldername)
 end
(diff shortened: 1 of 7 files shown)


alex-courtis:
This looks good so far: updates are more consistent and correct than master.

There are some issues remaining. I'll put together an nvt-min.lua and some test cases.

Investigating #1221 so I probably won't get to it until next weekend.

alex-courtis:
Test harness `nvim -nu /tmp/nvt-min.lua`
[nvt-min.lua.txt](https://github.com/kyazdani42/nvim-tree.lua/files/8600650/nvt-min.lua.txt)

Install [ccls](https://github.com/MaskRay/ccls). `pacman -S ccls`. Not sure what package manager you use; needs clang and node.

Clone https://github.com/alex-courtis/way-displays

alex-courtis:
There are many success cases; I won't list them, just the inconsistencies.

_1 - Folder Signs_
 nvim -nu /tmp/nvt-min.lua src/client.c
 create a problem in src/client.c
 :w
 :NvimTreeOpen
 src is signed
 open src
 only client.c is signed
**this is different to git behaviour**

_2 - Update Without Write_
 fix the problem in client.c
 client.c still highlighted
 close and open src
 signs are gone
**this is OK, but once again inconsistent with git, we would expect the sign only when the file is written**

_3 - Inconsistent Updates After Write_
 write client.c
 create a problem
 write client.
**depending on timing, the sign may be shown**

2 and 3 are clearly all about timing. The signs are shown correctly, however the behaviour is very inconsistent as it depends on the slow updates of lsp. I think this explains a lot of the inconsistent/unreproducable issues that people have been reporting.

alex-courtis (approved):
This works. The only small change to make is [1 - Folder Signs](https://github.com/kyazdani42/nvim-tree.lua/pull/1218#issuecomment-1114469780).

This does highlight the larger problem, which is the inconsistencies in when the signs are updated. This is not a great user experience. Leaving some ideas in #1216

Update: most of the inconsistencies occur only when using coc; when using nvim built-in lsp the response is much better.

Issue #1221, opened 2022-04-30
call NvimTreeToggle, Invalid parameter


tttoad:
### Description

When I call NvimTreeToggle, I get an error

![image](https://user-images.githubusercontent.com/14021602/166121877-5c4d9750-b14c-4019-8e0a-4545ad9ca5e3.png)

### Neovim version

```markdown
NVIM v0.6.1
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

macOS 11.2.3

### nvim-tree version

86d573d

### Steps to reproduce

:NvimTreeToggle

### Minimal config

```Lua
-- init.lua
vim.o.ignorecase=true

vim.opt.encoding="UTF-8"
vim.o.number=true
-- 设置行会导致输入指令后乱码...
-- vim.o.lines=60
vim.opt.mouse="a"
vim.o.cursorline=true
vim.opt.laststatus=2
vim.o.autoindent=true
vim.opt.tabstop=4
vim.o.smarttab=true
vim.opt.shiftwidth=4
vim.opt.softtabstop=4
vim.opt.backspace="eol,start,indent"
vim.o.showcmd=true
vim.o.whichwrap=vim.o.whichwrap.."<,>,h,l"
vim.opt.scrolloff=3
vim.opt.history=1000
vim.opt.cmdheight=2


require'nvim-tree'.setup {
}
return require('plugins')
```

gulliver-madrid:
Same issue in Windows 10, NVIM 0.6.1, default configuration.

alex-courtis:
The above minimal config does not run. Please provide a full nvt-min.lua as per the [bug report](https://github.com/kyazdani42/nvim-tree.lua/issues/new/choose) so that we can try to reproduce.

alex-courtis:
Could not replicate with minimal configuration nvim 0.6.1

I've added some logging and the following is being successfully executed:
```
setlocal winhl=EndOfBuffer:NvimTreeEndOfBuffer,Normal:NvimTreeNormal,CursorLine:NvimTreeCursorLine,WinSeparator:NvimTreeWinSeparator,StatusLine:NvimTreeStatusLine,StatusLineNC:NvimTreeStatuslineNC,SignColumn:NvimTreeSignColumn,NormalNC:NvimTreeNormalNC
```

I cannot exactly compare this to your error as you have not provided the error text, just a picture.

E474 may be ambiguous without the message. Please can you provide the english translation of the message.

[...]

Issue #1142, opened 2022-04-08
NvimTreeVertSplit highlight group is gone?


Tummetott:
### Description

I used to set a custom color for the `NvimTreeVertSplit` highlight group which is different from the `VertSplit` group. It seems like with the latest update, this highlight group is gone? Why is that the case?

### Neovim version

```markdown
NVIM v0.7.0-dev+1406-g8c25dbff4
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

macOS 10.15.7

### nvim-tree version

618ea25

### Steps to reproduce

```lua
vim.cmd('hi NvimTreeVertSplit guibg=#ff0000 guifg=#00FF00')
```

### Expected behavior

The vertical split line is colored in red/green

### Actual behavior

The vertical split line has the colors from the `VertSplit` highlight group

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
      "kyazdani42/nvim-web-devicons",
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {}
end
```

alex-courtis:
I can't reproduce this one:
```
nvim -nu /tmp/nvt-min.lua
:NvimTreeOpen
:hi NvimTreeVertSplit guibg=#ff0000 guifg=#00FF00
```

![2022-04-09-153323_974x789_scrot](https://user-images.githubusercontent.com/1505378/162558040-cd9ffa7e-2f8a-4457-9ed8-103722d0a80b.png)

[...]

PR #1223, opened 2022-05-01
Revert "fix: replace vertsplit highlights (#1142)"

alex-courtis:
Reverts kyazdani42/nvim-tree.lua#1219

See #1221

`WinSeparator` is not available as a built-in highlight group for many users.


siduck:
Oh VertSplit doesnt work for v0.7 :(

PR #1230, opened 2022-05-03
Tweak Bug Report Form

alex-courtis:
Many bug reports are not using `nvt-min.lua`, just submitting the default.

This will "encourage" reporters to use it, with `nvt-min.lua` split out into a file.

@gegoune please can you test at https://github.com/alex-courtis/nvim-tree.lua/issues/new/choose . You will need to change the URL of nvt-min.lua `kyazdani42` -> `alex-courtis`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -58,7 +58,7 @@ body:
   - type: textarea
     attributes:
       label: "Minimal config"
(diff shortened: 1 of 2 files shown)


gegoune:
@alex-courtis not sure what to test. It's the same file that was already in use, isn't it? Form looks good.

alex-courtis:
> Form looks good.

Thank you; form is all we can test ;)

kyazdani42 (approved):
let see how user behave :smile:

PR #1239, opened 2022-05-07
fix(close-node): do not close parent when current node is open

### LINKED ISSUE #1226: Feature: close_node behaviour depends on focussed node state
### Description

look at this picture
image

My cursor is at the node of vime, but if I close the node, it not close the vime rather than close the  dir lua.

I want to close the dir vime.

### Neovim version

```markdown
NVIM v0.8.0-dev+95-g07660193a
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

arch linux

### nvim-tree version

483f155

### Steps to reproduce

always exists

### Expected behavior

_No response_

### Actual behavior

_No response_

### Minimal config

```Lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {}
end
```

kyazdani42:
fixes #1226

### DIFF
--- a/lua/nvim-tree/actions/movements.lua
+++ b/lua/nvim-tree/actions/movements.lua
@@ -35,6 +35,11 @@ end
 
 function M.parent_node(should_close)
   return function(node)
+    if node.open then
+      node.open = false
+      return renderer.draw()
+    end
+
     local parent = node.parent
 
     if not parent or parent.cwd then


alex-courtis (approved):
Tested OK

I am not so comfortable making changes to behaviour that users may expect.
However, since this is the intended behaviour, we can class this as a bug, not a feature request and I am comfortable.

PR #1241, opened 2022-05-07
chore(diagnostics): clear whole group instead of keeping a table

kyazdani42:


alex-courtis (approved):
Tested OK:
* lsp
* coc

PR #1242, opened 2022-05-07
feat(renderer): set git icons in signcolumn

### LINKED ISSUE #553: Display git status in signcolumn
As I understand things right now, you leverage the signcolumn to display the lsp diagnostic status for files with potential issues.
While I like the idea of that, I'd rather use the signcolumn to display the git status information (thereby also preventing things like #72). Is that something you have thought about or would be willing to implement?

An additional cherry on top would be to have a configuraiton option whether git status signs are placed:
* to the right of file/folder names
* to the left of file/folder names
* in the sign column

Maybe you could leverage the same implementation to make the lsp diagnostic symbol location configurable in the same way? :thinking:

kyazdani42:
This is a work in progress. It works as is but there are some caveat that i'd like to discuss first.

Right, now, we will display the first icon returned by the list of icons from the node. Should we have a smarter way of doing that ?
Also, if displaying only one icons in signs, should we also always display one icon (after and before filename for instance ?)

TODO:
- [x] documentation

Fixes #553

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -32,7 +32,7 @@ local function add_sign(linenr, severity)
     return
   end
   local sign_name = sign_names[severity][1]
-  vim.fn.sign_place(1, GROUP, sign_name, buf, { lnum = linenr, priority = 2 })
+  vim.fn.sign_place(0, GROUP, sign_name, buf, { lnum = linenr, priority = 2 })
 end
 
 local function from_nvim_lsp()
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -32,7 +32,7 @@ local function add_sign(linenr, severity)
     return
   end
   local sign_name = sign_names[severity][1]
-  vim.fn.sign_place(1, GROUP, sign_name, buf, { lnum = linenr })
+  vim.fn.sign_place(1, GROUP, sign_name, buf, { lnum = linenr, priority = 2 })
 end
 
 local function from_nvim_lsp()
--- a/lua/nvim-tree/renderer/components/git.lua
+++ b/lua/nvim-tree/renderer/components/git.lua
@@ -1,7 +1,9 @@
 local _icons = require "nvim-tree.renderer.icon-config"
 local utils = require "nvim-tree.utils"
 
-local M = {}
+local M = {
+  SIGN_GROUP = "NvimTreeGitSigns"
+}
 
 local function build_icons_table()
   local i = M.icon_state.icons.git_icons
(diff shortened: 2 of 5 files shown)


lua/nvim-tree/diagnostics.lua
```diff
@@ -32,7 +32,7 @@ local function add_sign(linenr, severity)
     return
   end
   local sign_name = sign_names[severity][1]
-  vim.fn.sign_place(1, GROUP, sign_name, buf, { lnum = linenr + 1 })
+  vim.fn.sign_place(1, GROUP, sign_name, buf, { lnum = linenr + 1, priority = 2 })
```
kyazdani42, line 35: this will be higher priority than git signs

lua/nvim-tree/renderer/components/git.lua
```diff
@@ -124,6 +126,17 @@ local git_hl = {
   [" A"] = "none",
 }
 
+function M.setup_signs()
+  local i = M.icon_state.icons.git_icons
```
kyazdani42, line 130: this might not be initialized yet

lua/nvim-tree/renderer/init.lua
```diff
@@ -17,11 +17,14 @@ local M = {
...
 local namespace_id = api.nvim_create_namespace "NvimTreeHighlights"
 
-local function _draw(bufnr, lines, hl)
+local function _draw(bufnr, lines, hl, signs)
   api.nvim_buf_set_option(bufnr, "modifiable", true)
   api.nvim_buf_set_lines(bufnr, 0, -1, false, lines)
   M.render_hl(bufnr, hl)
   api.nvim_buf_set_option(bufnr, "modifiable", false)
+  for _, sign in pairs(signs) do
+    vim.fn.sign_place(1, git.SIGN_GROUP, sign.sign, bufnr, { lnum = sign.lnum, priority = 1 })
```
kyazdani42, line 26: not sure about the ID, i do not understand what it's used for. I've set it just like the one in diagnostics.

lua/nvim-tree/renderer/init.lua
```diff
@@ -111,6 +116,7 @@ function M.setup(opts)
   }
 
   _padding.setup(opts)
+  git.setup_signs()
```
kyazdani42, line 119: should we clear/setup signs on each draw ?

lua/nvim-tree/diagnostics.lua
alex-courtis: Tested OK.

We should document that behaviour.

lua/nvim-tree/renderer/init.lua
alex-courtis: It is working like this. Changing highlight groups immediately updates.

lua/nvim-tree/renderer/components/git.lua
alex-courtis: It should be.
`M.icon_state = _icons.get_config()` will always be called before `setup_signs`, as renderer init requires git.

lua/nvim-tree/renderer/init.lua
alex-courtis: sign-identifier seems to be used for mark like functionality, which we don't really need.

You could just call sign_place with 0 and have vim allocate a unique one.

alex-courtis (approved):
Tested OK:
- diagnostic fail: git_placement: after, before, signcolumn
- external change
- change some highlight groups

alex-courtis (approved):
re-tested base case - OK

lua/nvim-tree/renderer/components/git.lua
gegoune: I have got
```lua
vim.g.nvim_tree_icons = {
  git = {
    unstaged = 'M',
    staged = 'S',
    unmerged = 'U',
    renamed = 'R',
    untracked = '?',
    deleted = 'D',
    ignored = 'I',
  },
}
```
but icons in `signcolumn` are (I guess) some default (x, ★). Should I open an issue for it?

kyazdani42: are you properly setting up the nvim_tree_icons before requiring nvim-tree ?

gegoune: Ah, thanks! I was setting those `g:` before `require('nvim-tree').setup()` until recently adding a pcall-require-guard to all my plugins. User error - all works fine now, thanks!

Issue #674, opened 2021-10-01
Migration guide


kyazdani42:
After migrating to the new setup, these are the variables that have been migrated to the setup configuration table:

- `nvim_tree_disable_netrw` -> `disable_netrw`
- `nvim_tree_hijack_netrw`  -> `hijack_netrw`
- `nvim_tree_auto_open`   -> `open_on_setup`
- `nvim_tree_auto_close`   -> `auto_close`
- `nvim_tree_tab_open`    -> `open_on_tab`
- `nvim_tree_update_cwd` -> `update_cwd`
- `nvim_tree_hijack_cursor`   -> `hijack_cursor`
- `nvim_tree_system_open_command`  -> `system_open.cmd`
- `nvim_tree_system_open_command_args` -> `system_open.args`
- `nvim_tree_follow`   -> `update_focused_file.enable`
- `nvim_tree_follow_update_path` -> `update_focused_file.update_cwd`
- `nvim_tree_lsp_diagnostics` -> `lsp_diagnostics`
- `nvim_tree_auto_resize` -> `view.auto_resize`
- `nvim_tree_bindings` -> `view.mappings.list`
- `nvim_tree_hide_dotfiles` -> `filters.dotfiles`
- `nvim_tree_ignore` -> `filters.custom`
- `nvim_tree_gitignore` -> `git.ignore`

`nvim_tree_disable_keybindings` and `nvim_tree_disable_default_keybindings` have been replaced by using `view.mappings.custom_only`.
Setting `view.mappings.custom_only` to `true` and `view.mappings.list` to `{}` would do the same as `nvim_tree_disable_keybindings` and `view.mappings.custom_only` to `true` with your list of keybindings would do the same as `nvim_tree_disable_default_keybindings`.

NOTE: you need to set other `vim.g` `let g:` nvim_tree variables _BEFORE_ calling the `setup` if you want everything to work as expected :)

wookayin:
Please pin this or add a link README!

JarnoB:
Thanks, this is helpful! 
But isn't `nvim_tree_tab_open -> tab_open` supposed to be `nvim_tree_tab_open -> open_on_tab`?

kyazdani42:
good catch !

ismikhail:
The `view` config does not seem to be working for me. 
I installed the plugin with Packer like this:
```
  use {
    'kyazdani42/nvim-tree.lua',
    requires = 'kyazdani42/nvim-web-devicons'
[...]

PR #1279, opened 2022-05-21
chore/remove globals

alex-courtis:
First round. No significant refactoring. I'll do some more when I get the energy.

Second round, merge after this: #1280

Final round, merge after #1280 : #1281


lua/nvim-tree/renderer/builder.lua
```diff
@@ -209,15 +203,15 @@ function Builder:_build_file(node, padding, git_highlight, git_icons_tbl)
   local col_start = offset + #icon + git_icons_length
   local col_end = col_start + #node.name
 
-  if self.special_map[node.absolute_path] or self.special_map[node.name] then
+  if vim.tbl_contains(self.special_files, node.absolute_path) or vim.tbl_contains(self.special_files, node.name) then
```
kyazdani42, line 206: why not keeping existing implementation ?

kyazdani42 (approved):
looks good thank you !
Can you just answer this before we merge ?

lua/nvim-tree/renderer/builder.lua
alex-courtis: Do you mean the map of file names to boolean values?

I didn't see any reason for false value, so I made it a list like `window_picker.exclude.filetype`.

kyazdani42: well there is an issue then. Check out line 125 in the PR diff, special_files is used like a map, but here it is used as an array.

alex-courtis: Great catch!

`_build_folder` and `_build_file` are now the same, both tested OK.

kyazdani42 (approved):
LGTM !
Thanks a lot for this PR, we should have less issues coming up about people having issues with the setup !
Although we might have a few issues popping up in the next few days.
I'll leave you to merge this and update #674 (if you can, otherwise tell me i'll do it).

alex-courtis:
#674 updated.

I do expect a couple of issues and am watchingi for issues.

gegoune:
No issues on my side, config validator was extremely helpful! Great job Alex!

PR #1283, opened 2022-05-22
Ignore case mappings

mosheavni:
Ughh that was a nightmare!
I've been trying to restore my original <c-r> mapping and not until I dug into the code I found that the comparison is case sensitive which doesn't make sense in vim keymaps.
So hence this PR.
Thanks.

### DIFF
--- a/lua/nvim-tree/actions/init.lua
+++ b/lua/nvim-tree/actions/init.lua
@@ -193,14 +193,14 @@ local function merge_mappings(user_mappings)
     if type(map.key) == "table" then
       local filtered_keys = {}
       for _, key in pairs(map.key) do
-        if not vim.tbl_contains(user_keys, key) and not vim.tbl_contains(removed_keys, key) then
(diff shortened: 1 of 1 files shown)


alex-courtis (commented):
`{ key = "F", action = "" },` clears no mappings.

`{ key = "f", action = "" },` clears `f` and `F`.

I don't think I understand the use case for this change; case is important for mappings.

mosheavni:
it's for `<c-*`> mappings.
You're right, converting the lhs to lowercase is not very good.
Maybe I'll change the logic to look for `<c-` as the start of the lhs.
Allow me some time to fix please :)

kyazdani42:
@mosheavni please fix the luacheck issue. Although you could just write a cleaner util on your end before passing the mapping the nvim-tree ^^
@alex-courtis i'll leave you to chose what to do with this PR, i'm not sure it should be done on nvim-tree side too.

Issue #380, opened 2021-05-17
Open a directory recursively


TornaxO7:
Is it possible to open a directory recursively for example when pressing `O`? I'd like to see my project structure with nvim-tree.

kyazdani42:
not yet, it could be implemented though, but kind of slow depending on the numbers of directories under the directory you try to open

TornaxO7:
Ok, should I close this issue for you?

kyazdani42:
nope, i will forget otherwise ^^

sarahkittyy:
Please implement this, I switched from NERDTree to this and this is the one feature i miss the most. ❤️

0x00002a:
I created a function to do this based on `collapse-all.lua`:
```lua
function ExpandAll()
    local nvimlib = require('nvim-tree.lib')
    local function iter(nodes)
        for _, node in pairs(nodes) do
            if not node.open and node.nodes then
                nvimlib.expand_or_collapse(node)
                --node.open = true
            end
            if node.nodes then
                iter(node.nodes)
            end
        end
    end

    local nodes = {}
    local currnode = nvimlib.get_node_at_cursor()
    if currnode == nil then
        return
    end
    nodes = {currnode}
    iter(nodes)
end
```
it will expand all nodes under the current cursor, probably not optimal but it works :p

kyazdani42:
implemented in #1292

PR #1292, opened 2022-05-26
feat(actions): expand all under folder

### LINKED ISSUE #1289: expand_all to expand all folders to make live_filter more awesome

The new feature "live filter" is pretty awesome but it is not very usable due to two reasons:

1. It only searches expanded folders (I actually think this is desirable). 
2. There is no shortcut for expanding all folders

I think it'd make sense to add a shortcut to expand all folders. There is already one to colapse_all but not to do the opposite.

N/A

**Additional context**

N/A

kyazdani42:
fixes #1289 #380 
@alex-courtis i'd like a review on the doc and the binding to make sure its okay, because this feature will freeze neovim if used in the wrong place.


alex-courtis (changes requested):
Looks good, works.

Froze for 15 mins when expanding `~/src` but that is not surprising.

Needs a safety valve e.g.
"Expand halted after opening 500 folders"
That could be hardcoded or an option.

kyazdani42:
i think 100/200 folders would already be good enough, i'll try something out

README.md
```diff
@@ -190,6 +190,9 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
       global = false,
       restrict_above_cwd = false,
     },
+    expand_all = {
+      max_folder_discovery = 300,
```
kyazdani42, line 194: not too sure about the option name here.

lua/nvim-tree/actions/expand-all.lua
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  return iterate
+end
+
+function M.fn(base_node)
+  local node = base_node.nodes and base_node or core.get_explorer()
+  if gen_iterator()(node) then
+    utils.warn("expansion iteration was halted after " .. M.MAX_FOLDER_DISCOVERY .. " discovered folders")
```
kyazdani42, line 47: not sure if this is needed.

README.md
alex-courtis: `max_folder_expand` ?

lua/nvim-tree/actions/expand-all.lua
alex-courtis: I think it is good to tell the user what is happening.

alex-courtis:
> i think 100/200 folders would already be good enough, i'll try something out

300 is good; it stops after a few seconds.

README.md
```diff
@@ -261,6 +264,7 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
 - Double left click acts like `<CR>`
 - Double right click acts like `<C-]>`
 - `W` will collapse the whole tree
+- `E` will expand the whole tree. Be aware this might hang neovim for a while if running on a big folder (such as home dir or root dir).
```
alex-courtis, line 267: Reference `actions.expand_all.max_folder_discovery` here.

alex-courtis (approved):

README.md
kyazdani42: we'll see if this makes sense for user as it is.

PR #1293, opened 2022-05-26
feat(resize): add ability to grow and shrink the tree when drawing

### LINKED ISSUE #865: Having option to grow window width automatically
Would be awesome if the window could resize itself as more directories get opened when there is not enough space.

I thought out auto_resize would accomplish this but that seems to have another purpose.

kyazdani42:
Will adapt the view size based on the longest line.
fixes #865

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -5,6 +5,7 @@ local M = {}
 local events = require "nvim-tree.events"
 
 M.View = {
+  adaptive_size = false,
   tabpages = {},
   cursors = {},
   hide_root_folder = false,
@@ -185,6 +186,24 @@ function M.open(options)
     vim.cmd "wincmd p"
   end
   events._dispatch_on_tree_open()
+end
+
+local function grow()
+  local starts_at = M.is_root_folder_visible(require("nvim-tree.core").get_cwd()) and 1 or 0
+  local lines = vim.api.nvim_buf_get_lines(M.get_bufnr(), starts_at, -1, false)
+  local max_length = M.View.initial_width
+  for _, l in pairs(lines) do
+    if max_length < #l then
+      max_length = #l
+    end
+  end
+  M.resize(max_length)
+end
+
+function M.grow_from_content()
+  if M.View.adaptive_size and M.View.side == "left" or M.View.side == "right" then
+    grow()
+  end
 end
 
 function M.resize(size)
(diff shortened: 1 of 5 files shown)


kyazdani42:
found something, not sure if this is related, but with this option and the tree on the right side, if the window is too narrow, the live filter prompt is a bit messed up.

kyazdani42:
maybe also take into account a min_width size ?

alex-courtis:
Not related, but noticed a size change when using `view.width = 15`. The window will first display at width `17` (2 for sign column) but is then shrunk to width `15` when opening a file.

We could probably fix that in this PR.

[asciicast](https://asciinema.org/a/497739)

alex-courtis:
> maybe also take into account a min_width size ?

That is simple and reliable.

We could be lazy and just use `view.width` as the minimum.

lua/nvim-tree/view.lua
```diff
@@ -187,6 +188,23 @@ function M.open(options)
   events._dispatch_on_tree_open()
 end
 
+local function grow()
+  local lines = vim.api.nvim_buf_get_lines(M.get_bufnr(), 0, -1, false)
```
alex-courtis, line 192: This includes the root folder. This results in a very wide window when you are in a deep path.

alex-courtis (changes requested):
Tested OK in a variety of configurations e.g. git signs after.

kyazdani42:
@alex-courtis i've fixed the naming and the initial width + the shrink issue.
This should now work as expected.
The issue on your screen recording looks weird to me, the width of the window is looking bugged to me, that might be a neovim issue (or maybe something because of the signcolumn ?).
I've noticed that neovim takes the signcolumn into account when setting the width.

alex-courtis:
> The issue on your screen recording looks weird to me, the width of the window is looking bugged to me, that might be a neovim issue (or maybe something because of the signcolumn ?).

It only happens for very narrow windows. It's not worth looking at now; just thought I'd record my findings.

alex-courtis (commented):
Shrinking is not reliable / happens at the wrong time.

Defaults with `adaptive_size`:
1293-length-test

Extract [1293-length-test.tar.gz](https://github.com/kyazdani42/nvim-tree.lua/files/8792626/1293-length-test.tar.gz) to /tmp to test.

kyazdani42:
Shrinking should be reliable now !

lua/nvim-tree/view.lua
```diff
@@ -373,8 +392,10 @@ end
 
 function M.setup(opts)
   local options = opts.view or {}
+  M.View.adaptive_size = options.adaptive_size
   M.View.side = options.side
   M.View.width = options.width
+  M.View.initial_width = options.width
```
alex-courtis, line 398: `view.width = "20%"` is a problem.

alex-courtis (changes requested):
Working nicely. I look forward to using this.

lua/nvim-tree/view.lua
kyazdani42: indeed ! Completely forgot about this one

kyazdani42:
update initial width to use `get_size`

alex-courtis (approved):
👍

PR #1300, opened 2022-05-28
new option: close windows displaying removed files

emmanueltouzery:
This issue was already mentioned in the past, at least in https://github.com/kyazdani42/nvim-tree.lua/issues/1012

I have set up my neovim to not close windows when I close buffers, using [qpkorr/vim-bufkill](https://github.com/qpkorr/vim-bufkill)

The way nvim-tree explicitely closes the window after closing the buffer breaks that workflow for me (note that older versions of nvim-tree didn't do that, I believe 5d8453dfbd34ab00cb3e8ce39660f9a54cdd35f3 didn't do that).

So I added a new option, `remove_file_close_window`, defaulting to true -- the current behavior. When I change it to false, it fixes my issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ require'nvim-tree'.setup {
 require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
   auto_reload_on_write = true,
   create_in_closed_folder = false,
+  remove_file_close_window = true,
   disable_netrw = false,
   hijack_cursor = false,
   hijack_netrw = true,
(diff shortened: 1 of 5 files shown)


alex-courtis (changes requested):
Very nice; I will use this.

Tested OK: only one buffer, many buffers.

Please move the option to `actions.remove_file.close_window`.

emmanueltouzery:
yes, much better name for the setting, you're right! Changed that as requested!

alex-courtis (approved):

kyazdani42 (approved):

PR #1304, opened 2022-05-29
feat(explorer): add experimental watchers

kyazdani42:
This commit introduces watchers to update the tree.
This behavior is introduced behind an "experimental_watchers" option which should prevent instabilities.
It also adds watchers on the git module that will allow git to be updated whenever the `.git` folder changes.
It will become the default at some point.

To check:
- make sure everything is reliably connected
- let 2/3 people test this before merging
- a proper code review

Bugs encountered so far:
- not fixable apparently: git updates the index a LOT (on BufEnter < which is strange, when running git status... etcetc)

Also after some debugging, i've noticed a bug in the reloader.
New nodes where not applying git status modification on explorer reload, which caused the tree to get out of sync with git, which i've observed quite a few times these last weeks.
Also this caused an issue i've noticed when refreshing git-ignored files, which lose the ignored icon when running explorer . reload for a few milliseconds, and then reset by git reload, which was causing a weird UI behavior.

This PR fixes it too.

### DIFF
--- a/lua/nvim-tree/actions/remove-file.lua
+++ b/lua/nvim-tree/actions/remove-file.lua
@@ -93,7 +93,7 @@ function M.fn(node)
 end
 
 function M.setup(opts)
-  M.enable_reload = not opts.experimental_watchers
+  M.enable_reload = not opts.git.watcher.enable
   M.close_window = opts.actions.remove_file.close_window
 end
 
--- a/lua/nvim-tree/actions/remove-file.lua
+++ b/lua/nvim-tree/actions/remove-file.lua
@@ -93,7 +93,7 @@ function M.fn(node)
 end
 
 function M.setup(opts)
-  M.enable_reload = not opts.git.watcher.enable
+  M.enable_reload = not opts.git.watcher.enable or not opts.git.enable
   M.close_window = opts.actions.remove_file.close_window
 end
 
--- a/lua/nvim-tree/actions/remove-file.lua
+++ b/lua/nvim-tree/actions/remove-file.lua
@@ -93,7 +93,7 @@ function M.fn(node)
 end
 
 function M.setup(opts)
-  M.enable_reload = not opts.git.watcher.enable or not opts.git.enable
+  M.enable_reload = not opts.filesystem_watchers.enable
   M.close_window = opts.actions.remove_file.close_window
 end
 
--- a/lua/nvim-tree/actions/remove-file.lua
+++ b/lua/nvim-tree/actions/remove-file.lua
@@ -86,11 +86,14 @@ function M.fn(node)
       events._dispatch_file_removed(node.absolute_path)
       clear_buffer(node.absolute_path)
     end
-    require("nvim-tree.actions.reloaders").reload_explorer()
+    if M.enable_reload then
+      require("nvim-tree.actions.reloaders").reload_explorer()
+    end
   end
 end
 
 function M.setup(opts)
+  M.enable_reload = not opts.experimental_watchers
   M.close_window = opts.actions.remove_file.close_window
 end
 
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -1,5 +1,7 @@
+local utils = require "nvim-tree.utils"
 local git_utils = require "nvim-tree.git.utils"
 local Runner = require "nvim-tree.git.runner"
+local Watcher = require("nvim-tree.watcher").Watcher
 
 local M = {
   config = nil,
@@ -13,20 +15,35 @@ function M.reload()
   end
 
   for project_root in pairs(M.projects) do
-    M.projects[project_root] = {}
-    local git_status = Runner.run {
-      project_root = project_root,
-      list_untracked = git_utils.should_show_untracked(project_root),
-      list_ignored = true,
-      timeout = M.config.timeout,
-    }
-    M.projects[project_root] = {
-      files = git_status,
-      dirs = git_utils.file_status_to_dir_status(git_status, project_root),
-    }
+    M.reload_project(project_root)
   end
 
   return M.projects
+end
+
+function M.reload_project(project_root)
+  local project = M.projects[project_root]
+  if not project or not M.config.enable then
+    return
+  end
+
+  local watcher = M.projects[project_root].watcher
+  M.projects[project_root] = {}
+  local git_status = Runner.run {
+    project_root = project_root,
+    list_untracked = git_utils.should_show_untracked(project_root),
+    list_ignored = true,
+    timeout = M.config.timeout,
+  }
+  M.projects[project_root] = {
+    files = git_status,
+    dirs = git_utils.file_status_to_dir_status(git_status, project_root),
+    watcher = watcher,
+  }
+end
+
+function M.get_project(project_root)
+  return M.projects[project_root]
 end
 
 function M.get_project_root(cwd)
@@ -40,6 +57,35 @@ function M.get_project_root(cwd)
 
   local project_root = git_utils.get_toplevel(cwd)
   return project_root
+end
+
+function M.reload_tree_at(project_root)
+  local root_node = utils.get_node_from_path(project_root)
+  if not root_node then
+    return
+  end
+
+  M.reload_project(project_root)
+  local project = M.get_project(project_root)
+
+  local project_files = project.files and project.files or {}
+  local project_dirs = project.dirs and project.dirs or {}
+
+  local function iterate(n)
+    local parent_ignored = n.git_status == "!!"
+    for _, node in pairs(n.nodes) do
+      node.git_status = project_dirs[node.absolute_path] or project_files[node.absolute_path]
+      if not node.git_status and parent_ignored then
+        node.git_status = "!!"
+      end
+
+      if node.nodes and #node.nodes > 0 then
+        iterate(node)
+      end
+    end
+  end
+
+  iterate(root_node)
 end
 
 function M.load_project_status(cwd)
@@ -67,6 +113,15 @@ function M.load_project_status(cwd)
   M.projects[project_root] = {
     files = git_status,
     dirs = git_utils.file_status_to_dir_status(git_status, project_root),
+    watcher = Watcher.new {
+      absolute_path = utils.path_join { project_root, ".git" },
+      on_event = function()
+        utils.debounce(200, function()
+          M.reload_tree_at(project_root)
+          require("nvim-tree.renderer").draw()
+        end)
+      end,
+    },
   }
   return M.projects[project_root]
 end
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -111,12 +111,13 @@ function M.load_project_status(cwd)
     list_ignored = true,
     timeout = M.config.timeout,
   }
-  log.line("watcher", "git start")
-  M.projects[project_root] = {
-    files = git_status,
-    dirs = git_utils.file_status_to_dir_status(git_status, project_root),
+
+  local watcher = nil
+  if M.config.watcher.enable then
+    log.line("watcher", "git start")
     watcher = Watcher.new {
       absolute_path = utils.path_join { project_root, ".git" },
+      interval = M.config.watcher.interval,
       on_event = function()
         log.line("watcher", "git event")
         utils.debounce(200, function()
@@ -124,7 +125,13 @@ function M.load_project_status(cwd)
           require("nvim-tree.renderer").draw()
         end)
       end,
-    },
+    }
+  end
+
+  M.projects[project_root] = {
+    files = git_status,
+    dirs = git_utils.file_status_to_dir_status(git_status, project_root),
+    watcher = watcher,
   }
   return M.projects[project_root]
 end
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -120,10 +120,8 @@ function M.load_project_status(cwd)
       interval = M.config.watcher.interval,
       on_event = function()
         log.line("watcher", "git event")
-        utils.debounce(200, function()
-          M.reload_tree_at(project_root)
-          require("nvim-tree.renderer").draw()
-        end)
+        M.reload_tree_at(project_root)
+        require("nvim-tree.renderer").draw()
       end,
     }
   end
@@ -138,6 +136,7 @@ end
 
 function M.setup(opts)
   M.config = opts.git
+  M.config.watcher = opts.filesystem_watchers
 end
 
 return M
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -111,13 +111,14 @@ function M.load_project_status(cwd)
     list_ignored = true,
     timeout = M.config.timeout,
   }
+  log.line("watcher", "git start")
   M.projects[project_root] = {
     files = git_status,
     dirs = git_utils.file_status_to_dir_status(git_status, project_root),
     watcher = Watcher.new {
       absolute_path = utils.path_join { project_root, ".git" },
       on_event = function()
-        log.line("watcher", "git")
+        log.line("watcher", "git event")
         utils.debounce(200, function()
           M.reload_tree_at(project_root)
           require("nvim-tree.renderer").draw()
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -1,3 +1,4 @@
+local log = require "nvim-tree.log"
 local utils = require "nvim-tree.utils"
 local git_utils = require "nvim-tree.git.utils"
 local Runner = require "nvim-tree.git.runner"
@@ -116,6 +117,7 @@ function M.load_project_status(cwd)
     watcher = Watcher.new {
       absolute_path = utils.path_join { project_root, ".git" },
       on_event = function()
+        log.line("watcher", "git")
         utils.debounce(200, function()
           M.reload_tree_at(project_root)
           require("nvim-tree.renderer").draw()
--- a/lua/nvim-tree/watcher.lua
+++ b/lua/nvim-tree/watcher.lua
@@ -14,11 +14,6 @@ function Watcher.new(opts)
   end
   log.line("watcher", "Watcher:new   '%s'", opts.absolute_path)
 
-  local stat, _ = uv.fs_stat(opts.absolute_path)
-  if not stat or stat.type ~= "directory" then
-    return nil
-  end
-
   local watcher = setmetatable({
     _opts = opts,
   }, Watcher)
@@ -33,13 +28,14 @@ function Watcher:start()
 
   self._p, _, name = uv.new_fs_poll()
   if not self._p then
+    self._p = nil
     utils.warn(
       string.format("Could not initialize an fs_poll watcher for path %s : %s", self._opts.absolute_path, name)
     )
     return nil
   end
 
-  local poll_cb = vim.schedule_wrap(function(err, _, _)
+  local poll_cb = vim.schedule_wrap(function(err)
     if err then
       log.line("watcher", "poll_cb for %s fail : %s", self._opts.absolute_path, err)
     else
@@ -47,8 +43,7 @@ function Watcher:start()
     end
   end)
 
-  -- TODO option for interval ms
-  rc, _, name = uv.fs_poll_start(self._p, self._opts.absolute_path, 1, poll_cb)
+  rc, _, name = uv.fs_poll_start(self._p, self._opts.absolute_path, M.interval, poll_cb)
   if rc ~= 0 then
     utils.warn(string.format("Could not start the fs_poll watcher for path %s : %s", self._opts.absolute_path, name))
     return nil
@@ -74,7 +69,8 @@ function Watcher:restart()
 end
 
 function M.setup(opts)
-  M.enabled = opts.experimental_watchers
+  M.enabled = opts.git.watcher.enable
+  M.interval = opts.git.watcher.interval
 end
 
 M.Watcher = Watcher
--- a/lua/nvim-tree/watcher.lua
+++ b/lua/nvim-tree/watcher.lua
@@ -29,11 +29,13 @@ end
 function Watcher:start()
   log.line("watcher", "Watcher:start '%s'", self._opts.absolute_path)
 
-  local rc, name
+  local rc, _, name
 
   self._p, _, name = uv.new_fs_poll()
   if not self._p then
-    utils.warn(string.format("Could not initialize an fs_poll watcher for path %s : %s", self._opts.absolute_path, name))
+    utils.warn(
+      string.format("Could not initialize an fs_poll watcher for path %s : %s", self._opts.absolute_path, name)
+    )
     return nil
   end
 
--- a/lua/nvim-tree/watcher.lua
+++ b/lua/nvim-tree/watcher.lua
@@ -1,5 +1,6 @@
 local uv = vim.loop
 
+local log = require "nvim-tree.log"
 local utils = require "nvim-tree.utils"
 
 local M = {}
@@ -10,6 +11,7 @@ function Watcher.new(opts)
   if not M.enabled then
     return nil
   end
+  log.line("watcher", "Watcher:new   '%s'", opts.absolute_path)
 
   local ok, fs_event = pcall(uv.new_fs_event)
   if not ok then
@@ -26,6 +28,7 @@ function Watcher.new(opts)
 end
 
 function Watcher:start(opts)
+  log.line("watcher", "Watcher:start '%s'", self._path)
   local ok = pcall(
     uv.fs_event_start,
     self._w,
@@ -44,6 +47,7 @@ function Watcher:start(opts)
 end
 
 function Watcher:stop()
+  log.line("watcher", "Watcher:stop  '%s'", self._path)
   if self._w then
     uv.fs_event_stop(self._w)
   end
--- a/lua/nvim-tree/watcher.lua
+++ b/lua/nvim-tree/watcher.lua
@@ -72,11 +72,6 @@ function Watcher:stop()
   end
 end
 
-function Watcher:restart()
-  self:stop()
-  return self:start()
-end
-
 M.Watcher = Watcher
 
 return M
--- a/lua/nvim-tree/watcher.lua
+++ b/lua/nvim-tree/watcher.lua
@@ -1,0 +1,63 @@
+local uv = vim.loop
+
+local utils = require "nvim-tree.utils"
+
+local M = {}
+local Watcher = {}
+Watcher.__index = Watcher
+
+function Watcher.new(opts)
+  if not M.enabled then
+    return nil
+  end
+
+  local ok, fs_event = pcall(uv.new_fs_event)
+  if not ok then
+    utils.warn(string.format("Could not initialize a watcher for path %s", opts.absolute_path))
+    return nil
+  end
+
+  local watcher = setmetatable({
+    _path = opts.absolute_path,
+    _w = fs_event,
+  }, Watcher)
+
+  return watcher:start(opts)
+end
+
+function Watcher:start(opts)
+  local ok = pcall(
+    uv.fs_event_start,
+    self._w,
+    self._path,
+    {},
+    vim.schedule_wrap(function()
+      opts.on_event(self._path)
+    end)
+  )
+  if not ok then
+    utils.warn(string.format("Could not start the watcher for path %s", self._path))
+    return nil
+  end
+
+  return self
+end
+
+function Watcher:stop()
+  if self._w then
+    uv.fs_event_stop(self._w)
+  end
+end
+
+function Watcher:restart()
+  self:stop()
+  return self:start()
+end
+
+function M.setup(opts)
+  M.enabled = opts.experimental_watchers
+end
+
+M.Watcher = Watcher
+
+return M
--- a/lua/nvim-tree/watcher.lua
+++ b/lua/nvim-tree/watcher.lua
@@ -7,39 +7,48 @@ local M = {}
 local Watcher = {}
 Watcher.__index = Watcher
 
+-- TODO check for a Watcher with the same opts.absolute_path and return that
 function Watcher.new(opts)
   if not M.enabled then
     return nil
   end
   log.line("watcher", "Watcher:new   '%s'", opts.absolute_path)
 
-  local ok, fs_event = pcall(uv.new_fs_event)
-  if not ok then
-    utils.warn(string.format("Could not initialize a watcher for path %s", opts.absolute_path))
+  local stat, _ = uv.fs_stat(opts.absolute_path)
+  if not stat or stat.type ~= "directory" then
     return nil
   end
 
   local watcher = setmetatable({
-    _path = opts.absolute_path,
-    _w = fs_event,
+    _opts = opts,
   }, Watcher)
 
-  return watcher:start(opts)
+  return watcher:start()
 end
 
-function Watcher:start(opts)
-  log.line("watcher", "Watcher:start '%s'", self._path)
-  local ok = pcall(
-    uv.fs_event_start,
-    self._w,
-    self._path,
-    {},
-    vim.schedule_wrap(function()
-      opts.on_event(self._path)
-    end)
-  )
-  if not ok then
-    utils.warn(string.format("Could not start the watcher for path %s", self._path))
+function Watcher:start()
+  log.line("watcher", "Watcher:start '%s'", self._opts.absolute_path)
+
+  local rc, name
+
+  self._p, _, name = uv.new_fs_poll()
+  if not self._p then
+    utils.warn(string.format("Could not initialize an fs_poll watcher for path %s : %s", self._opts.absolute_path, name))
+    return nil
+  end
+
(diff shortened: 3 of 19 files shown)


kyazdani42:
@gegoune also tagging you here in case you wish to test this :)

alex-courtis (changes requested):
Have not reviewed all code yet.

It is functioning for changes made to the buffer in vim and, more importantly, updates when files are changed when vim is not focussed. This is incredible!

alex-courtis:
> * not fixable apparently: git updates the index a LOT

I think I see a problem: the node watcher is seeing the changes in `.git` as well as the git watcher.

We are also seeing too many events for the node. We have multple watchers for the same path; we only need one.

I've pushed up some logging:

Modify a buffer under `src` and write:
```
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:14] [watcher] node '/home/alex/src/way-displays/src'
```

`git checkout .` from another terminal:
```
[2022-05-30 15:19:27] [watcher] git
[2022-05-30 15:19:27] [watcher] node '/home/alex/src/way-displays/.git'
[2022-05-30 15:19:27] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:27] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:27] [watcher] node '/home/alex/src/way-displays/src'
[2022-05-30 15:19:27] [watcher] git
[2022-05-30 15:19:27] [watcher] node '/home/alex/src/way-displays/.git'
[2022-05-30 15:19:27] [watcher] git
[2022-05-30 15:19:27] [watcher] node '/home/alex/src/way-displays/.git'
[2022-05-30 15:19:27] [watcher] git
[2022-05-30 15:19:27] [watcher] node '/home/alex/src/way-displays/.git'
```

There don't seem to be too many events for the git watcher. There may also be side-effects. Taking out the node watcher for `.git` may solve a lot of issues.

alex-courtis:
Turning on profile logging shows that all the git jobs are running concurrently. The debounce timing is going to need some tuning.

alex-courtis:
Open nvim-tree:
```
[2022-05-30 16:00:03] [watcher] node start    '/home/alex/src/way-displays'
[2022-05-30 16:00:03] [watcher] Watcher:new   '/home/alex/src/way-displays'
[2022-05-30 16:00:03] [watcher] Watcher:start '/home/alex/src/way-displays'
[2022-05-30 16:00:03] [git] running job with timeout 400ms
[2022-05-30 16:00:03] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:03] [git] done
[2022-05-30 16:00:03] [git] job success
[2022-05-30 16:00:03] [watcher] git start
[2022-05-30 16:00:03] [watcher] Watcher:new   '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:03] [watcher] Watcher:start '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:03] [watcher] node start    '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:03] [watcher] Watcher:new   '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:03] [watcher] Watcher:start '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:03] [watcher] node start    '/home/alex/src/way-displays/doc'
[2022-05-30 16:00:03] [watcher] Watcher:new   '/home/alex/src/way-displays/doc'
[2022-05-30 16:00:03] [watcher] Watcher:start '/home/alex/src/way-displays/doc'
[2022-05-30 16:00:03] [watcher] node start    '/home/alex/src/way-displays/inc'
[2022-05-30 16:00:03] [watcher] Watcher:new   '/home/alex/src/way-displays/inc'
[2022-05-30 16:00:03] [watcher] Watcher:start '/home/alex/src/way-displays/inc'
[2022-05-30 16:00:03] [watcher] node start    '/home/alex/src/way-displays/pro'
[2022-05-30 16:00:03] [watcher] Watcher:new   '/home/alex/src/way-displays/pro'
[2022-05-30 16:00:03] [watcher] Watcher:start '/home/alex/src/way-displays/pro'
[2022-05-30 16:00:03] [watcher] node start    '/home/alex/src/way-displays/src'
[2022-05-30 16:00:03] [watcher] Watcher:new   '/home/alex/src/way-displays/src'
[2022-05-30 16:00:03] [watcher] Watcher:start '/home/alex/src/way-displays/src'
```
Extra `.git` watcher, otherwise OK.

Modify and write buf under `way-displays/src`:
```
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:13] [git] running job with timeout 400ms
[2022-05-30 16:00:13] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] done
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
[2022-05-30 16:00:13] [git] job success
```
Too many events. Less than the number of files the directory.

`git checkout .` from another terminal:
```
[2022-05-30 16:00:21] [watcher] git event
[2022-05-30 16:00:21] [watcher] node event '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:21] [watcher] node start    '/home/alex/src/way-displays/.git/branches'
[2022-05-30 16:00:21] [watcher] Watcher:new   '/home/alex/src/way-displays/.git/branches'
[2022-05-30 16:00:21] [watcher] Watcher:start '/home/alex/src/way-displays/.git/branches'
[2022-05-30 16:00:21] [watcher] node start    '/home/alex/src/way-displays/.git/hooks'
[2022-05-30 16:00:21] [watcher] Watcher:new   '/home/alex/src/way-displays/.git/hooks'
[2022-05-30 16:00:21] [watcher] Watcher:start '/home/alex/src/way-displays/.git/hooks'
[2022-05-30 16:00:21] [watcher] node start    '/home/alex/src/way-displays/.git/info'
[2022-05-30 16:00:21] [watcher] Watcher:new   '/home/alex/src/way-displays/.git/info'
[2022-05-30 16:00:21] [watcher] Watcher:start '/home/alex/src/way-displays/.git/info'
[2022-05-30 16:00:21] [watcher] node start    '/home/alex/src/way-displays/.git/logs'
[2022-05-30 16:00:21] [watcher] Watcher:new   '/home/alex/src/way-displays/.git/logs'
[2022-05-30 16:00:21] [watcher] Watcher:start '/home/alex/src/way-displays/.git/logs'
[2022-05-30 16:00:21] [watcher] node start    '/home/alex/src/way-displays/.git/objects'
[2022-05-30 16:00:21] [watcher] Watcher:new   '/home/alex/src/way-displays/.git/objects'
[2022-05-30 16:00:21] [watcher] Watcher:start '/home/alex/src/way-displays/.git/objects'
[2022-05-30 16:00:21] [watcher] node start    '/home/alex/src/way-displays/.git/refs'
[2022-05-30 16:00:21] [watcher] Watcher:new   '/home/alex/src/way-displays/.git/refs'
[2022-05-30 16:00:21] [watcher] Watcher:start '/home/alex/src/way-displays/.git/refs'
[2022-05-30 16:00:21] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:21] [git] running job with timeout 400ms
[2022-05-30 16:00:21] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:21] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:21] [git] running job with timeout 400ms
[2022-05-30 16:00:21] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:21] [watcher] node event '/home/alex/src/way-displays/src'
[2022-05-30 16:00:21] [git] running job with timeout 400ms
[2022-05-30 16:00:21] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:21] [watcher] git event
[2022-05-30 16:00:21] [watcher] node event '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:21] [watcher] git event
[2022-05-30 16:00:21] [watcher] node event '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:21] [watcher] git event
[2022-05-30 16:00:21] [watcher] node event '/home/alex/src/way-displays/.git'
[2022-05-30 16:00:21] [git] done
[2022-05-30 16:00:21] [git] done
[2022-05-30 16:00:21] [git] done
[2022-05-30 16:00:21] [git] job success
[2022-05-30 16:00:21] [git] job success
[2022-05-30 16:00:21] [git] job success
[2022-05-30 16:00:22] [git] running job with timeout 400ms
[2022-05-30 16:00:22] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-05-30 16:00:22] [git] done
[2022-05-30 16:00:22] [git] job success
```
Watching all the nodes under `.git` is a problem.
There are only 4 git events, however some may be side-effects. Removing the node watchers on `.git` may Fix Things.

alex-courtis:
Here's a dirty little c program to show you all the events that are happening in the directory, using inotify: https://gist.github.com/alex-courtis/17c0ab4b9fefaf8ffe2e9b99e46d23b2

Write process.c buffer. This is interesting as it shows the temporary file that vim writes. We don't want to watch that.
```
./dw -n /home/alex/src/way-displays/src
IN_CLOSE_WRITE:   file: 4913
IN_CLOSE_WRITE:   file: process.c
```

Turn on all close events and write the file. There are a horrible number of close events that we do not want to listen to:
```
./dw -a /home/alex/src/way-displays/src
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_WRITE:   file: 4913
IN_CLOSE_WRITE:   file: process.c
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: log.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: cfg.cpp
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: server.c
IN_CLOSE_NOWRITE: dir
IN_CLOSE_NOWRITE: file: fds.c
IN_CLOSE_NOWRITE: file: process.c
IN_CLOSE_NOWRITE: file: process.c
```

When one of my apps is polling for a config file change in a directory, it only listens to close write events, ignoring the others: https://github.com/alex-courtis/way-displays/blob/f8b12563ccbcc248aa0121b6590f730994cd784f/src/fds.c#L48

I think we will need to somehow configure `fs_event_start` to act in the same manner, or filter out the unwanted events when we receive them.

alex-courtis:
One more thing: symlinks

Watching symlinks (at least via inotify) does not behave as you would expect.

You need to watch the directory containing the symlink's target i.e. the link_to's directory.

lua/nvim-tree/actions/remove-file.lua
```diff
@@ -86,11 +86,14 @@ function M.fn(node)
       events._dispatch_file_removed(node.absolute_path)
       clear_buffer(node.absolute_path)
     end
-    require("nvim-tree.actions.reloaders").reload_explorer()
+    if M.enable_reload then
```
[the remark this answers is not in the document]
kyazdani42: Well i cannot do that, because reload explorer is used for when using filter change.

kyazdani42:
Nice review as always :) Thanks for the debugging, i'll take a look when i have some time this week

kyazdani42:
@alex-courtis i've ignored the .git folder for the explorer to avoid watching it twice.

alex-courtis:
The 4913 files are a test by vim to see if it can write to the directory: https://groups.google.com/g/vim_dev/c/sppdpElxY44

We might be able to hardcode ignore them, as they are basically hardcoded in vim.

alex-courtis:
Pushed a change to use [fs_poll](https://github.com/luvit/luv/blob/master/docs.md#uv_fs_poll_t--fs-poll-handle) instead of fs_event.

fs_poll is beatiful because it is throttled. Even at 1ms it only results in single git events. This also sidesteps the 4913 file problem.

TODO:
* add option for poll interval
* add more `.git` watchers e.g. `.git/info` OR remove the `.git` watchers (they may not be useful any more)
* `watcher.lua` maintain a list of active Watchers and prevent multiple for the same path
* think about how to handle removed directories - removing their Watcher

alex-courtis (approved):

lua/nvim-tree/watcher.lua
```diff
@@ -7,49 +7,62 @@ local M = {}
...
-    end)
-  )
-  if not ok then
-    utils.warn(string.format("Could not start the watcher for path %s", self._path))
+function Watcher:start()
+  log.line("watcher", "Watcher:start '%s'", self._opts.absolute_path)
+
+  local rc, name
+
+  self._p, _, name = uv.new_fs_poll()
```
alex-courtis, line 34: The [fail](https://github.com/luvit/luv/blob/master/docs.md#error-handling) pseudo-type makes me sad.

lua/nvim-tree/watcher.lua
```diff
@@ -0,0 +1,82 @@
...
+function Watcher.new(opts)
+  if not M.enabled then
+    return nil
+  end
+  log.line("watcher", "Watcher:new   '%s'", opts.absolute_path)
+
+  local stat, _ = uv.fs_stat(opts.absolute_path)
+  if not stat or stat.type ~= "directory" then
+    return nil
+  end
```
kyazdani42, line 20: not sure about this check, it is already done in the caller i believe

lua/nvim-tree/watcher.lua
```diff
@@ -0,0 +1,82 @@
...
+
+  self._p, _, name = uv.new_fs_poll()
+  if not self._p then
+    utils.warn(
+      string.format("Could not initialize an fs_poll watcher for path %s : %s", self._opts.absolute_path, name)
+    )
+    return nil
+  end
+
+  local poll_cb = vim.schedule_wrap(function(err, _, _)
```
kyazdani42, line 42: uneeded ignored params

lua/nvim-tree/watcher.lua
```diff
@@ -0,0 +1,82 @@
...
+    if err then
+      log.line("watcher", "poll_cb for %s fail : %s", self._opts.absolute_path, err)
+    else
+      self._opts.on_event(self._opts.absolute_path)
+    end
+  end)
+
+  -- TODO option for interval ms
+  rc, _, name = uv.fs_poll_start(self._p, self._opts.absolute_path, 1, poll_cb)
```
kyazdani42, line 51: i think we should just make experimental watchers a table with enable and interval options

kyazdani42 (commented):
This fs_poll refactoring is good. 
Let me know what you think about the comments i made.
Once we fix these (+ the existing watcher check), i'd be in favor of merging so users can start experimenting with it and we get some feedback.

lua/nvim-tree/watcher.lua
alex-courtis: done

lua/nvim-tree/watcher.lua
alex-courtis: done

lua/nvim-tree/watcher.lua
alex-courtis: Added `git.watcher`

alex-courtis:
> There is also the git watcher that triggers on bufenter that bothers me quite a bit. Could we somehow prevent the poll event for git whenever trigger by a bufenter ?

I'm not seeing that.

The only paths that can trigger a git job are:
* watcher
* explorer new / expand

These are all skipped:
* various reload_explorer
* `NvimTreeRefresh` (I want to put that back; will discuss later)
* `BufEnter` `NvimTree_*`

alex-courtis:
> (+ the existing watcher check)

done

alex-courtis:
I've put `:NvimTreeRefresh` back as we still have `R`. It's useful for non-git related external file system changes.

alex-courtis:
We are seeing single git jobs for most cases of editing buffers.

In the case of an external event e.g. a git checkout, we are seeing many:
```
[2022-06-05 14:48:53] [watcher] node event '/home/alex/.local/share/nvim/vundle/nvim-tree.lua.lu5je0/scripts'
[2022-06-05 14:48:53] [git] running job with timeout 400ms
[2022-06-05 14:48:53] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-06-05 14:48:53] [watcher] node event '/home/alex/.local/share/nvim/vundle/nvim-tree.lua.lu5je0/lua'
[2022-06-05 14:48:53] [git] running job with timeout 400ms
[2022-06-05 14:48:53] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-06-05 14:48:53] [watcher] git event
[2022-06-05 14:48:53] [watcher] node event '/home/alex/.local/share/nvim/vundle/nvim-tree.lua.lu5je0/doc'
[2022-06-05 14:48:53] [git] running job with timeout 400ms
[2022-06-05 14:48:53] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-06-05 14:48:53] [git] done
[2022-06-05 14:48:53] [git] done
[2022-06-05 14:48:53] [git] done
[2022-06-05 14:48:53] [git] job success
[2022-06-05 14:48:53] [git] job success
[2022-06-05 14:48:53] [git] job success
[2022-06-05 14:48:53] [watcher] node event '/home/alex/.local/share/nvim/vundle/nvim-tree.lua.lu5je0'
[2022-06-05 14:48:53] [git] running job with timeout 400ms
[2022-06-05 14:48:53] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-06-05 14:48:53] [git] done
[2022-06-05 14:48:53] [git] job success
[2022-06-05 14:48:53] [git] running job with timeout 400ms
[2022-06-05 14:48:53] [git] git --no-optional-locks status --porcelain=v1 --ignored=matching -u
[2022-06-05 14:48:53] [git] done
[2022-06-05 14:48:53] [git] job success
```

That doesn't bother me too much. I'd like to address that in a separate change.
I'm thinking of preventing concurrent git jobs and having a queue of 1 outstanding job, which will likely result in only 2 git jobs.

alex-courtis:
I attempted to create a watcher to the containing directory of the symlink target:
`watcher = watch.create_watcher(vim.fn.fnamemodify(link_to, ":p:h"))`
Writes are not seen.

In fact, many external events are not seen, such as appending to the file via `>>`. `fs_poll` on a directory may not be sufficient and polling on files may be necessary.

It is working for the great majority of cases right now. I think we can leave it as it is and address problems if they arise.

alex-courtis:
TODO later or never:

* think about how to handle removed directories i.e. removing their Watcher
* remove debounce as it should no longer be necessary
* consider recursive directory search, enabling watchers for all; use case is not clear to me right now
* symlink owning directories

Let's launch this.

As usual, testing would be gratefully appreciated.

lua/nvim-tree/git/init.lua
```diff
@@ -13,22 +16,37 @@ function M.reload()
...
 
   return M.projects
 end
 
+function M.reload_project(project_root)
+  local project = M.projects[project_root]
+  if not project or not M.config.enable then
+    return
+  end
+
```
alex-courtis, line 30: We could move the debounce in here or maybe Runner::run 

That would debounce all the jobs, not just the .git ones.

That can come later.

kyazdani42:
I think the changes are invalid now. The watcher option should not be under 'git'.
There should be 1 watcher for git changes whenever git is enabled and 1 watcher for filesystem changes when experimental_watchers is enabled.
I'm changing these in order to keep semantic meaning correct.

kyazdani42:
I've fixed this. I'm going to implement the watcher removal for whenever the directory is removed.

kyazdani42:
Also for the issue with the bufenter i've mentioned:

https://user-images.githubusercontent.com/23156099/172046152-6169549d-791c-4c5a-ae99-16df7e85ae6e.mp4

kyazdani42:
I think we can leave clearing the existing watchers for another PR since this seems to be a bit of a tricky job ^^

alex-courtis:
> I think the changes are invalid now. The watcher option should not be under 'git'.
> There should be ~1 watcher for git changes whenever git is enabled and~ 1 watcher for filesystem changes when experimental_watchers is enabled.

Yes. I abstracted anything git out of watchers so that we could create many watchers for different event types. That would not actually be useful or performant. Watchers with different intervals on the same path is also not useful.

It would be useful is to add multiple on_event callbacks to a watcher. That will be trivial to implement if we actually need it.

`filesystem_watchers` makes much more sense, thanks for moving.

Future possibilities:
```
  filesystem_watchers = {
    enable = false,
    interval = 100,
    types = {
      git = true,
      something = true,
      something = true,
    },
  },
```

I cannot imagine what that something might be though.

alex-courtis:
> I've also removed the debounce since it's not needed anymore.

I've got some ideas I'd like to implement very separately: https://github.com/kyazdani42/nvim-tree.lua/issues/1317#issuecomment-1146919908

I think debounce would still be useful. We do get a lot of FS events in a very short period of time and debounce could likely batch them all into one, which would avoid the extra trailing job as above.

alex-courtis:
> I dont think it would add value. I don't see any use case for enabling watchers like this.

It creates a lot of problems.

It does solve edge cases such as external checkout or, say, lsp refactors changing subdirectory files, however those would likely result in other changes that the wathchers would see.

alex-courtis:
> > symlink owning directories
> 
> I think we should that.

That one is tricky. Individual file watchers are the only solution I can see right now, which would be (unnecessarily) complex.

alex-courtis:
> Also for the issue with the bufenter i've mentioned:

I see. That is not happening with an nvt-min.lua setup ( how I develop / test ). It is happening with my full vim setup.

Maybe autowrites?

kyazdani42:
- debounce is not needed since we are polling already and i haven't observed multiple calls that would need to be debounced.
- symlink owned dirs could be implemented quite easily i believe, instead of watching the "absolute_path", we should just watch the "link_to" and match based on that value also. That would just add a few lines of code and work just as expected.
- i think this might be caused by a plugin, i'll try to pin down which one it is.

kylo252:
isn't the poll-based watcher very expensive for this kind of operation? I actually only just found out that gitsigns seems to be using the same thing :(

Instead, an event-based approach should be the way to go imo.

w/ [inotify](https://man7.org/linux/man-pages/man7/inotify.7.html):
- https://github.com/gitwatch/gitwatch
- https://github.com/facebook/watchman

w/ libuv's [fs_event](https://github.com/luvit/luv/blob/master/docs.md#uv_fs_event_t--fs-event-handle):
- https://github.com/neovim/neovim/pull/12593
- https://github.com/kevinhwang91/nvim-bqf/blob/0cc539c52e51d32d8febf1f04c5e7ed5353fead2/lua/bqf/filter/fzf.lua#L319

kyazdani42:
inotify is unix based, won't work on windows systems.
This PR was implemented first using fs_event, but we switched to fs_poll because it looked like better (polling wasn't doing too many refresh compared to fs_event).
Could you explain why you think fs_poll would be more expensive than fs_event ?

kylo252:
> Could you explain why you think fs_poll would be more expensive than fs_event ?

it's just the nature of using a polling model, it's basically calling stat periodically, so uses CPU even when nothing changes.

Example from nodejs: `fs.watch()` uses `fs_event` while `fs.watchfile()` uses `fs_poll`, and here's what they have to say about it in their API docs

> Using [fs.watch()](https://nodejs.org/docs/latest/api/fs.html#fswatchfilename-options-listener) is more efficient than `fs.watchFile` and `fs.unwatchFile`. `fs.watch` should be used instead of `fs.watchFile` and `fs.unwatchFile` when possible.

kyazdani42:
I think it makes more sense indeed, that might've been why i've used fs_event the first time i implemented a watcher :)
@alex-courtis did you have a particular case for using the polling method instead ?

alex-courtis:
> * symlink owned dirs could be implemented quite easily i believe, instead of watching the "absolute_path", we should just watch the "link_to" and match based on that value also. That would just add a few lines of code and work just as expected.

I thought so too. Not the case; fs_poll is not seeing the "to directory" changes. More experimentation will need to be done, so I put it on the "later" pile.

alex-courtis:
> * debounce is not needed since we are polling already and i haven't observed multiple calls that would need to be debounced.

Yes. That would only be necessary as a last resort if we still see to many events and a concurrent git job guard is not sufficient.

The situation could be that we would always run 2 git jobs: the first for the first event and the second for all the trailing events. `fs_poll` avoids this as it is throttled.

alex-courtis:
> did you have a particular case for using the polling method instead ?

The throttled nature suits our current implementation, spawning only single jobs. We can change.

alex-courtis:
The context of this reply is for watching a directory, rather than individual files.

TL;DR libuv doesn't give us the information that inotify does in the callback.

## Options

### inotify

That is an ideal solution. It sees all the changes and can be very granular, providing a lot of information including the file name that the change applies to.

We can restrict it to see only `IN_CLOEXEC` events and then only respond to `IN_CLOSE_WRITE` for files, which provides exactly what we require. See [dw.c](https://gist.github.com/alex-courtis/17c0ab4b9fefaf8ffe2e9b99e46d23b2) for a hacky inotify poller to tell us everything that is going on in a directory.

Not usable in vim as it's OS specific and there is no API.

### libuv

This abstraction layer is fit for purpose as it is OS independent and has first class vim API support.

Note that all callback information from libuv refers to the directory being watched, not the affected file.

#### [fs_event](https://github.com/luvit/luv/blob/master/docs.md#uv_fs_event_t--fs-event-handle)

This should provide the same functionality as inotify. It sees almost all the events that inotify sees. I _imagine_ that the linux implementation uses poll/inotify, which is optimised at the OS level.

fs_event watching a directory results in an incredible number of events, including many reads.

Unfortunately, it doesn't provide the actual nature of the event, just "change" or "rename", with almost every event being a change. It doesn't provide the file that relates to the event, just the directory.

> w/ libuv's [fs_event](https://github.com/luvit/luv/blob/master/docs.md#uv_fs_event_t--fs-event-handle):
> 
> * [fswatch autoread neovim/neovim#12593](https://github.com/neovim/neovim/pull/12593)
> * https://github.com/kevinhwang91/nvim-bqf/blob/0cc539c52e51d32d8febf1f04c5e7ed5353fead2/lua/bqf/filter/fzf.lua#L319

Those both appear to respond to changes to files rather than directories. They will see much more specific and relevant events.

#### [fs_poll](https://github.com/luvit/luv/blob/master/docs.md#uv_fs_poll_t--fs-poll-handle)

This calls back for the same events that fs_event does, except that it is throttled to the polling interval. That basically fits this use case, although still with many false positives.

I did take a look at a few other plugins like gitsigns and they follow this kludgey approach.

> @alex-courtis did you have a particular case for using the polling method instead ?

It is a practical solution that works with our implementation right now. It's definitely not perfect and we can change back to fs_event as we improve things.

## Next Steps

#1320 outlines the planned improvements.

Handling concurrent git jobs opens up our options, allowing us to use more granular watchers producing more events. This would allow us to move back to fs_event.

Handling directory removal / deletion will require introspection of events and possibly reading the directory to discover changes.

Monitoring individual files as well as directories may be needed to achieve "perfect" results however would add a lot of complexity and many extra filesystem reads.

∴ fs_poll or fs_event restricted to directories is a practical solution that covers the vast majority of cases. Any outlying events that are missed will be picked up by a subsequent event which will usually occur shortly afterwards.

gegoune:
@sindrets just merged similar thing to [diffview.nvim](https://github.com/sindrets/diffview.nvim) - perhaps it can add some value here as well?
https://github.com/sindrets/diffview.nvim/commit/7d0d9ecc9ddee85d11a8a52cd30432646cdebf71


Which makes me think - if each git related plugin (gitsigns I assume, nvim-tree, diffview, maybe neogit?) will add their own watchers it surely must not be the most optimal approach. Perhaps some plugin or maybe even core functionality should exist and add single watchers triggering (autocmd?) relevant `refresh` functions of each of those plugins?

cc @lewis6991, @TimUntersberger

TimUntersberger:
@gegoune 

Neogit doesn't listen to filesystem events. I guess in the future there might be a need for that.

Extracting the watcher code and moving it to plenary would definitely help a lot of plugin devs that need something similar.

Issue #1297, opened 2022-05-28
[Feature request] Display node whose name length is wider than the width of nvim-tree window in floating window.


lu5je0:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Display node whose name length is wider than the width of nvim-tree window in floating window just like IntelliJ IEAD.
![2](https://user-images.githubusercontent.com/17700877/170813946-b0289158-d47b-44fb-89b3-f16809ae988f.gif)

**Additional context**
I can submit a PR if you guys find it useful.

kyazdani42:
go for it, looks like a good feature to me :)

alex-courtis:
I like that. Users could use this or #1293 to show the file names.

PR #1305, opened 2022-05-29
feat: display node whose name length is wider than the width of nvim-tree window in floating window.

lu5je0:
Referencing #1297

TODO:
- ~~add an option in `require'nvim-tree'.setup`~~
- ~~show floating window when `winsaveview().leftcol` is not zero~~

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -124,6 +124,7 @@ Values may be functions. Warning: this may result in unexpected behaviour.
         add_trailing = false,
         group_empty = false,
         highlight_git = false,
+        full_name = false,
         highlight_opened_files = "none",
         root_folder_modifier = ":~",
         indent_markers = {
@@ -514,6 +515,10 @@ UI rendering setup
 
     *nvim-tree.renderer.group_empty*
     Compact folders that only contain a single folder into one node in the file tree.
+      Type: `boolean`, Default: `false`
+
+    *nvim-tree.renderer.full_name*
+    Display node whose name length is wider than the width of nvim-tree window in floating window.
       Type: `boolean`, Default: `false`
 
     *nvim-tree.renderer.highlight_git*
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -385,6 +385,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
     add_trailing = false,
     group_empty = false,
     highlight_git = false,
+    full_name = false,
     highlight_opened_files = "none",
     root_folder_modifier = ":~",
     indent_markers = {
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
     add_trailing = false,
     group_empty = false,
     highlight_git = false,
+    full_name = false,
     highlight_opened_files = "none",
     root_folder_modifier = ":~",
     indent_markers = {
--- a/lua/nvim-tree/renderer/components/full-name.lua
+++ b/lua/nvim-tree/renderer/components/full-name.lua
@@ -1,0 +1,83 @@
+local M = {}
+
+local api = vim.api
+local fn = vim.fn
+
+local function hide(win)
+  if win then
+    if api.nvim_win_is_valid(win) then
+      api.nvim_win_close(win, true)
+    end
+  end
+end
+
+local function show()
+  local line_nr = api.nvim_win_get_cursor(0)[1]
+  if line_nr == 1 and require("nvim-tree.view").is_root_folder_visible() then
+    return
+  end
+  if vim.wo.wrap then
+    return
+  end
+  -- only work for left tree
+  if vim.api.nvim_win_get_position(0)[2] ~= 0 then
+    return
+  end
+
+  local line = fn.getline "."
+  local leftcol = fn.winsaveview().leftcol
+  -- hide full name if left column of node in nvim-tree win is not zero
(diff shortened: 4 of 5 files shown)


alex-courtis (changes requested):
Doesn't look good for a right tree.

[1293-length-test.tar.gz](https://github.com/kyazdani42/nvim-tree.lua/files/8795212/1293-length-test.tar.gz)


1305-floating-node

The sign column is added to the left of the float. It should be removed, for left and right.

lua/nvim-tree.lua
```diff
@@ -377,6 +377,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
     add_trailing = false,
     group_empty = false,
     highlight_git = false,
+    floating_node = false,
```
alex-courtis, line 380: This option name is not intuitive; perhaps `renderer.full_name`.

lua/nvim-tree/renderer/components/floating-node.lua
```diff
@@ -0,0 +1,82 @@
...
+
+local function hide(win)
+  if win then
+    if api.nvim_win_is_valid(win) then
+      api.nvim_win_close(win, true)
+    end
+  end
+end
+
+local popup_win = nil
```
alex-courtis, line 14: This should be a member.

lua/nvim-tree/renderer/components/floating-node.lua
```diff
@@ -0,0 +1,82 @@
...
+local function show()
+  hide(popup_win)
```
alex-courtis, line 16: Is this necessary?
The first autocommand should do the job.

lua/nvim-tree/renderer/components/floating-node.lua
```diff
@@ -0,0 +1,82 @@
...
+  local line_nr = api.nvim_win_get_cursor(0)[1]
+  if line_nr == 1 then
```
alex-courtis, line 18: We need to call `is_root_folder_visible` to decide whether to display line 1.

alex-courtis (changes requested):

alex-courtis:
I'm not sure that this feature is necessary; #1293 achieves the goal of showing the full name and it is a lot simpler.

I do appreciate all your efforts @lu5je0

kyazdani42:
for the people that don't want to grow / reduce the size, maybe this would be helpful anyway :)

lua/nvim-tree/renderer/components/floating-node.lua
lu5je0: I forgeted this😄

lua/nvim-tree/renderer/components/floating-node.lua
lu5je0: Yes, this is not necessary, I will remove it.

lua/nvim-tree.lua
lu5je0: looks good

lu5je0:
> 

The popup should be hide for right tree.

lu5je0:
> Doesn't look good for a right tree.
> 
> [1293-length-test.tar.gz](https://github.com/kyazdani42/nvim-tree.lua/files/8795212/1293-length-test.tar.gz)
> 
> 1305-floating-node [ ](https://user-images.githubusercontent.com/1505378/170912119-ed147d59-84db-41b4-8d2d-cd5bf7408931.gif) 1305-floating-node [ ](https://user-images.githubusercontent.com/1505378/170912119-ed147d59-84db-41b4-8d2d-cd5bf7408931.gif)
> 
> Edit: ~The sign column is added to the left of the float. It should be removed, for left and right.~
> 
> The popup should be left trimmed up to the icon/name.

Yes, it should be hidden for the right tree

lua/nvim-tree/renderer/components/floating-node.lua
lu5je0: Change this to `M.popup_win`?

doc/nvim-tree-lua.txt
```diff
@@ -475,6 +476,10 @@ UI rendering setup
     Compact folders that only contain a single folder into one node in the file tree.
       Type: `boolean`, Default: `false`
 
+    *nvim-tree.renderer.full_name*
+    Display node whose name length is wider than the width of nvim-tree window in floating window.
```
alex-courtis, line 480: Need to specify that this only applies to left trees. Maybe add:

    Only works when |nvim-tree.view.side| is `left`.

alex-courtis:
> Yes, it should be hidden for the right tree

That works for me. Tested OK.

alex-courtis (changes requested):
Looking good. Still to do:

* Solution / answer for "show floating window when winsaveview().leftcol is not zero"
* Rebase
* stylua/luacheck
*

lu5je0:
> Solution / answer for "show floating window when winsaveview().leftcol is not zero"
There are quite complex to do this. Maybe we can just disable full name in this situation simply.

alex-courtis:
> > Solution / answer for "show floating window when winsaveview().leftcol is not zero"
> 
> There are quite complex to do this. Maybe we can just disable full name in this situation simply.

That works for me. I'm not sure when this could happen, however it is not catastrophic if we don't show the full name.

alex-courtis (approved):

kyazdani42 (approved):

kyazdani42:
@RaafatTurki this is unexpected behavior, although this might be caused by neovim. I've noticed some issues when playing with winbar lately. Although it works nicely most of the time.
If you can reproduce consistently, please open a new issue.

PR #1308, opened 2022-05-30
feat: guard against multiple setup calls

### LINKED ISSUE #1306: Handle Mulitple setup Calls
Running setup multiple times has been an issue discovered when resolving some user issues.

Warn when setup multiple times.

Exit early when calling setup multiple times.

**Additional context**

alex-courtis:
closes #1306

We have had a few cases where users have done this.

We could make nvim-tree capable of "resetting itself" and applying new config, however that doesn't seem worth the effort. We would have to tear down the autocommands etc. etc.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -16,7 +16,9 @@ local collapse_all = require "nvim-tree.actions.collapse-all"
 
 local _config = {}
 
-local M = {}
+local M = {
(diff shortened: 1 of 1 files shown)


kyazdani42 (approved):
nice thanks

icholy:
I don't think this plays nice with packer. See #1313

PR #1310, opened 2022-06-01
feat(view): ability to centralize node initially

kylo252:
This is only relevant for large nested folders, but I find it easier when the first selection is centered.

```help
    *nvim-tree.view.centralize_selection*
    Reposition the view so that the current node is initially centralized, see |zz|.
```

Hopefully this is useful for others :smile:

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -105,6 +105,7 @@ Values may be functions. Warning: this may result in unexpected behaviour.
       respect_buf_cwd = false,
       view = {
         adaptive_size = false,
+        centralize_selection = false,
         width = 30,
         height = 30,
         hide_root_folder = false,
@@ -452,6 +453,11 @@ Window / buffer setup.
     *nvim-tree.view.adaptive_size*
     Resize the window on each draw based on the longest line.
     Only works when |nvim-tree.view.side| is `left` or `right`.
+      Type: `boolean`, Default: `false`
+
+    *nvim-tree.view.centralize_selection*
+    When entering nvim-tree, reposition the view so that the current node is
+    initially centralized, see |zz|.
       Type: `boolean`, Default: `false`
 
     *nvim-tree.view.hide_root_folder*
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -347,6 +347,18 @@ local function setup_autocommands(opts)
   if opts.reload_on_bufenter and not has_watchers then
     create_nvim_tree_autocmd("BufEnter", { pattern = "NvimTree_*", callback = reloaders.reload_explorer })
   end
+
+  if opts.view.centralize_selection then
+    create_nvim_tree_autocmd("BufEnter", {
+      pattern = "NvimTree_*",
+      callback = function()
+        vim.schedule(function()
+          local keys = api.nvim_replace_termcodes("zz", true, false, true)
+          api.nvim_feedkeys(keys, "n", true)
+        end)
+      end,
+    })
+  end
 end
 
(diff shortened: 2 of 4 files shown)


lua/nvim-tree.lua
```diff
@@ -342,6 +342,18 @@ local function setup_autocommands(opts)
...
     create_nvim_tree_autocmd("BufEnter", { pattern = "NvimTree_*", callback = reloaders.reload_explorer })
   end
+
+  if opts.view.centralize_selection then
+    create_nvim_tree_autocmd("BufEnter", {
+      pattern = "NvimTree_*",
+      callback = function()
+        vim.schedule(function()
+          local keys = api.nvim_replace_termcodes("zz", true, false, true)
+          api.nvim_feedkeys(keys, "n", true)
```
alex-courtis, line 352: Does not obey `hijack_cursor = true`.

Putting this autocmd before `CursorMoved` does not fix this.

Please try changing this autocmd to be nested to see if you can make hijack_cursor work. If it's not possible I am happy to let it go.

lua/nvim-tree/view.lua
```diff
@@ -392,17 +395,10 @@ function M.is_root_folder_visible(cwd)
...
-  M.View.side = options.side
-  M.View.width = options.width
-  M.View.initial_width = get_size()
-  M.View.height = options.height
-  M.View.hide_root_folder = options.hide_root_folder
-  M.View.preserve_window_proportions = options.preserve_window_proportions
-  M.View.winopts.number = options.number
-  M.View.winopts.relativenumber = options.relativenumber
-  M.View.winopts.signcolumn = options.signcolumn
+  opts = opts or {}
```
alex-courtis, line 398: Let's not do this refactor. It does not appear to be necessary.

doc/nvim-tree-lua.txt
```diff
@@ -424,6 +425,10 @@ Window / buffer setup.
     Only works when |nvim-tree.view.side| is `left` or `right`.
       Type: `boolean`, Default: `false`
 
+    *nvim-tree.view.centralize_selection*
+    Reposition the view so that the current node is initially centralized, see |zz|.
```
alex-courtis, line 429: Please clarify that this is on BufEnter.

Perhaps: "When entering nvim-tree, reposition the view so that the current node is moved to the center, see |zz|.

alex-courtis (changes requested):
Looks good, works.

lua/nvim-tree/view.lua
kylo252: > Let's not do this refactor. It does not appear to be necessary.

I mean it's not necessary in the sense that you can keep copying all the parameters by hand and hope you didn't forget any of them, or you can make it both explicit and obvious that this is the intended behavior. 

I'll put it back regardless, but I just wanted to point that out.

---

_P.s. notice how calling this merge function notified me that `height`/`weight` are initially missing.._

kyazdani42: please rebase, i've fixed this issue when height is top/bottom

lua/nvim-tree.lua
kylo252: I fixed it for now, but I do think that the problem originates from event used for `hijack_cursor`, example: try `:NvimTreeToggle` while having `hijack_cursor` to be set.  I don't think this would've been a problem, if the initial "legal" position is set when entering.

That said, I'm not sure it's worth the effort or the complexity to try to cover all the edge-cases that would originate from such a change.

kyazdani42 (approved):
looks good to me, waiting for alex's approval :)

alex-courtis (changes requested):
Remove `place_cursor_on_node` and we are OK to merge.

alex-courtis (approved):

Issue #1318, opened 2022-06-04
nvim-tree has problem with directories using decomposed UTF-8 (UTF-8-MAC)


ronisbr:
### Description

When trying to create a file inside a directory in which the path has accented characters using decomposed UTF-8 (UTF-8-MAC), nvim-tree create a file in another path in which all the accents are removed.

For example, if we are in `/tmp/seção` and try to create a file, nvim-tree creates in `/tmp/secao`.

The problem here is that every time in macOS you create a directory using Finder with accents, they will use this decomposed UTF-8. Hence, this issue leads to lot of problems.

### Neovim version

```text
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
Compilado por brew@Monterey

Features: +acl +iconv +tui
See ":help feature-compile"

    arquivo vimrc de sistema: "$VIM/sysinit.vim"
            padrão para $VIM: "/usr/local/Cellar/neovim/0.7.0/share/nvim"

Run :checkhealth for more info
```

### Operating system and version

macOS 12.4

### nvim-tree version

25921aa

### Minimal config

```lua
The bug happens with the default nat-min.lua.
```

### Steps to reproduce

Execute:

```
cd /tmp
mkdir tmp
mv tmp `echo seção | iconv -f UTF-8 -t UTF-8-MAC`
cd seção
```

Open neovim and nvim-tree.
Type `a` to create a new file.

### Expected behavior

The input line should be:

```
Create file /tmp/seção/
```

in which a new file will be created in the current directory.

### Actual behavior

The input line is:

```
Create file /tmp/secao/
```

in which a new file will be created in the wrong directory.

alex-courtis:
Working on linux:

`Create file /tmp/1318/seção/`

This might be tricky.

alex-courtis:
I don't have an environment (mac) on which I can test this.

@ronisbr could you attempt to fix this yourself?

In the meantime I'll take a look and see if anything obvious becomes apparent.

alex-courtis:
Are there any other nvim-tree situations in which such problems occur? Is it only for create-file?

ronisbr:
Hi @alex-courtis !

Thanks for the quick response.

> Unfortunately that proves nothing as I cannot perform the iconv.

Yes, without iconv it will use composed UTF-8, which works fine.

> @ronisbr could you attempt to fix this yourself?

I have almost no knowledge in Lua, but I will try to at least debug the problem!

[...]

PR #1319, opened 2022-06-05
doc: add help wanted and contributing

alex-courtis:
Dealing with issues like #1318 is becoming very difficult.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,6 +395,22 @@ Please attach `$XDG_CACHE_HOME/nvim/nvim-tree.log` if you raise an issue.
 
 * Try temporarily disabling git integration by setting `git.enable = false` in your setup.
 
+## Help Wanted
+
+Developers with the following environments:
+
+* Apple macOS
+* Windows
+  * WSL
+  * msys
+  * powershell
+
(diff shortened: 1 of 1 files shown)


gegoune:
Auto labeling issues with respective OS labels based on value specified in issue might not be the worst idea. Then some volunteers could be auto mentioned. I am on macOS and willing to help according to my abilities.

kyazdani42 (approved):
let's do that first, we'll improve the bug triage as a 2nd step :)

alex-courtis:
> Auto labeling issues with respective OS labels based on value specified in issue might not be the worst idea.

Nice! Could we do that with, say, a dropdown on the bug report form?

gegoune:
I think so. Then https://github.com/actions/labeler could be used on issue: created event.

alex-courtis:
Could you PR that?

The way to test a new form is to make the change on master of a fork or change the default branch of a fork.

gegoune:
I am thinking whether auto labelling is a way to go here, as it not always is relevant. Most issues are platform independent and those that are platform specific should be labeled as such manually once it's established. I can come up with PR but after short reflection on the subject I no longer think it's a way forward - let me know your thoughts.

alex-courtis:
Manual labelling works. Added 3 more "OS .*" labels.

Only a few issues actually need those labels.

PR #1321, opened 2022-06-05
feat(renderer): indent_markers add a item icon

rammiah:
I add a item icon for indent_markers, make it possible to make nvim-tree view like tree command output.

When we config indent_markers to this:

```lua
render.indent_markers = {
    enable = true,
    icons = {
        corner = "└─",
        item = "├─",
        edge = "│ ",
        none = "  ",
    },
},
```

nvim-tree view:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
       icons = {
         corner = "└ ",
         edge = "│ ",
+        item = "│ ",
         none = "  ",
       },
     },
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -113,6 +113,7 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
       icons = {
         corner = "└ ",
         edge = "│ ",
+        item = "│ ",
```
kyazdani42, line 116: should be empty string by default :)

kyazdani42 (changes requested):

README.md
rammiah: it is used to replace the innerest file's edge icon,  shouldn't be empty

kyazdani42: i see sorry

kyazdani42 (approved):

alex-courtis:
nice

PR #1323, opened 2022-06-06
feat(trash): default command 'trash' -> 'gio trash'

RaafatTurki:
This makes sense since most linux distros already ship with glib2 instead of requiring the user to tweak or install 3rd party dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
     },
   },
   trash = {
-    cmd = "trash",
+    cmd = "gio trash",
     require_confirm = true,
   },
   live_filter = {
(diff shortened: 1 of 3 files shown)


alex-courtis (approved):
Not quite sure what gio is (man page is not enlightening) but it is definitely part of glib2 and should be present for all.

macOS and windows users will need to supply their own :shrug:

kyazdani42 (approved):

hermitmaster:
FYI - this was a change that broke config on MacOS for anyone using `trash`

kyazdani42:
is trash a default command on macos ?

hermitmaster:
No, but it is a trivial and fairly standard `brew` install. Looks like `brew install glib` is a decent alternative to make it work with default config. Maybe document it as an optional requirement on MacOS? `glib` provides `gio trash` on MacOS.

gegoune:
> Maybe an improvment on the README/docs or a check to only enable gio trash for linux based systems ?

Perhaps `checkhealth` would be a better place to report things like that to the user. It would stop over polluting README. I a sure bunch of other things could be detected and reported there as well.

PR #1325, opened 2022-06-07
prevent open buffer on NvimTree window when quite_on_open is true

btstream:
As title said.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -385,6 +385,10 @@ function M._prevent_buffer_override()
     require("nvim-tree.renderer").draw()
     a.nvim_win_close(curwin, { force = true })
     require("nvim-tree.actions.open-file").fn("edit", bufname)
+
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/view.lua
```diff
@@ -385,6 +385,10 @@ function M._prevent_buffer_override()
     require("nvim-tree.renderer").draw()
     a.nvim_win_close(curwin, { force = true })
     require("nvim-tree.actions.open-file").fn("edit", bufname)
+
+    if require("nvim-tree").get_config().actions.open_file.quit_on_open then
```
alex-courtis, line 389: I like that this is a direct call instead of an au.

I will test over the weekend.

alex-courtis: Please follow the existing pattern and add to `M.View.quit_on_open`. It needs a cleanup, but Not Today.

alex-courtis (changes requested):
This works reliably.

alex-courtis (approved):

Issue #1322, opened 2022-06-05
NvimTreeFocus on a file outside cwd does not work


lixvbnet:
Hi,

First of all, thanks for creating this wonderful plugin!

**My config**
```lua
update_cwd = true,
update_focused_file = {
    enable = true,
    update_cwd = false,
},
```

**Description**
NvimTreeFocus on a file outside cwd does not work.

**Expected behavior**:
Allow NvimTreeFocus to update cwd to locate and focus the file. (Or make its behavior configurable ? 🤔 )

I've read issue #986 and got the following config.

```lua
  update_focused_file = {
    update_cwd  = true
  },
```

This would also solve this issue, but I'd like to avoid setting this. 

Suppose I open nvim in `~/MyProject/`, with the above config, when I **first open a file outside the initial directory** such as `/usr/local/test.out`, and **then open a file inside its subfolder**, say `~/MyProject/folderA/a.txt`, the working directory will be changed to that subfolder, then I would only see files under folderA.

But I wish to have cwd remain to be my initial folder `~/MyProject/` unless I choose to change it by running NvimTreeFocus on a file outside it.  Is this possible at all?

**Screenshot**
![image](https://user-images.githubusercontent.com/75612664/172055289-827d21ba-c9a9-47de-ae1c-5878bdbdbd48.png)

Here I'm unable to locate `encode.go` , which is outside my initial working directory.

kyazdani42:
This should be the case though. If the node is found under the current cwd, it should not update the cwd.
Are you using macos ?

lixvbnet:
> This should be the case though. If the node is found under the current cwd, it should not update the cwd. Are you using macos ?

Yes I use macos.    What I am suggesting is allow NvimTreeFocus **update the cwd** if the node is **not found** in current cwd (or, in other words, outside current cwd).  What do you think ?

kyazdani42:
We could add parameters indeed, the focus function is quite small and could be extended.
Feel free to open a PR.
Sorry i misunderstood your request :)

lixvbnet:
Sure, I’ll take a try!

PR #1328, opened 2022-06-09
extend NvimTreeFindFile: allow change root

lixvbnet:
For #1322 

This adds `NvimTreeFindFile!` command.  (`NvimTreeFindFile` without bang works same as before)

With the bang, it locates current file in the tree, expands the paths if necessary.
And the optional `root_dirs` config allows setting preferred root directories such as source code or library roots.

example:
```lua
root_dirs = { "/usr/local/go/src", "~/go/pkg/mod" },
update_cwd = false,
update_focused_file = {
  enable = false,
  update_cwd = false,
},
```

image

go to definition (`gd`), then run `NvimTreeFindFile!` (Edit: changed to `NvimTreeFindFile!` ) again:

image

Instead of setting tree root to the `json` folder, it **prefers** root directories specified in `root_dirs`.

~~I cannot run the sh script for it reports errors, so those texts are manually added. please help double check~~

Updated the PR, and now it is possible to have this behavior inserted into existing `update_focused_file` by following config:
```lua
root_dirs = { "/usr/local/go/src", "~/go/pkg/mod" },
update_focused_file = {
  enable = true,
  update_root = true,
},
```


alex-courtis (changes requested):
Reviewed functionality but not code. It works.

The most useful piece here is `root_dirs`, however it is optional and this functionality works well with and without it. A sensible default would be `root_dirs = { "/usr/include", },`

- [x] `NvimTreeFocus!` does not make sense; it should be `NvimTreeFindFile!`.

- [x] `update_focused_file` should also be enhanced, as it mirrors `NvimTreeFindFile`. Add an `update_focused_file.update_root` option to mirror the bang.

I would prefer that we do not talk about cwd; this change is about the root. cwd is ambiguous as it usually refers to the vim cwd.(Yes, I'm aware that `actions.change_dir.restrict_above_cwd` probably should have been named `actions.change_dir.restrict_above_root`)

alex-courtis:
> I cannot run the sh script for it reports errors, so those texts are manually added. please help double check

What was the error?

lixvbnet:
> > I cannot run the sh script for it reports errors, so those texts are manually added. please help double check
> 
> What was the error?

```shell
sed: 2: "/BEGIN_DEFAULT_OPTS/,/E ...": extra characters at the end of d command
sed: 2: "/BEGIN_DEFAULT_OPTS/,/E ...": extra characters at the end of d command
scripts/update-default-opts.sh: line 31: syntax error near unexpected token `('
scripts/update-default-opts.sh: line 31: `lua <(cat /tmp/DEFAULT_MAPPINGS.M.lua scripts/generate_default_mappings.lua)'
```

macos issues.  but I managed to run the script by replacing `sed -i` with `sed -i ''` and changing the `<` in L31 with pipe.

I just accidently deleted my remote dev branch, now it's back..

---

> `NvimTreeFocus!` does not make sense; it should be `NvimTreeFindFile!`
Yeah I agree.  I didn't notice this command.  will update this PR later

kyazdani42 (approved):
looks good to me

lixvbnet:
PR updated , please double check

lua/nvim-tree.lua
```diff
@@ -125,10 +153,15 @@ function M.find_file(with_open, bufnr)
...
   end
 
+  -- if we don't schedule, it will search for NvimTree
   vim.schedule(function()
-    -- if we don't schedule, it will search for NvimTree
-    if _config.update_focused_file.update_cwd then
+    -- bang forces it to run M.change_root
+    if bang then
+      M.change_root(filepath)
+    elseif _config.update_focused_file.update_cwd then
```
alex-courtis, line 161: `update_focused_file.update_cwd` and `update_focused_file.update_root` are actually the same except for `root_dirs`.

We can remove `update_focused_file.update_root` and just have `update_focused_file.update_cwd` use `root_dirs`.

`update_base_dir_with_filepath` can be replaced with `change_root` as they are doing the same thing.

 `change_root` needs to be modified to honor `update_focused_file.ignore_list`.

Are you OK with that @kyazdani42 ?

alex-courtis (changes requested):
`update_focused_file.update_root` is working.

alex-courtis:
Having `update_focused_file.update_cwd` and `update_cwd` is confusing. The first updates the root and the second update's vim's cwd.

After or during this PR I'd like to change `update_focused_file.update_cwd` to `update_focused_file.update_root` for better clarity. It can be a silent `legacy.refactored`.

alex-courtis:
> macos issues. but I managed to run the script by replacing `sed -i` with `sed -i ''`, add an `;` at the end of `d`, and changing the `<` in L31 with pipe.

Thanks. Update in another PR https://github.com/kyazdani42/nvim-tree.lua/pull/1327/commits/e46b7571de4935fa06aca3e6bc487fe17837f5c1

lixvbnet:
> Having `update_focused_file.update_cwd` and `update_cwd` is confusing. The first updates the root and the second update's vim's cwd.

That's still quite confusing. For `vim's cwd`, is it the starting directory from command line, i.e. `M.init_root` in my PR?
As per my understanding, setting `update_cwd` to true means changing `M.init_root` when the tree root directory is changed, right?

lua/nvim-tree.lua
lixvbnet: I think `update_focused_file.update_root` might be a better name here. Why not keep it and mark `update_focused_file.update_cwd` as deprecated?

kyazdani42:
a better name would be `sync_root_with_cwd` or something like that for the main option

kyazdani42:
update_cwd is useful whenever you `:cd some-folder`, which changes neovim base cwd, thus nvim-tree catches DirChanged event and updates. It is also useful i believe when multiple tabs have different cwd, changing between tabs will dispatch a DirChanged event (i think).

lixvbnet:
> update_cwd is useful whenever you `:cd some-folder`, which changes neovim base cwd, thus nvim-tree catches DirChanged event and updates. It is also useful i believe when multiple tabs have different cwd, changing between tabs will dispatch a DirChanged event (i think).

thanks, that make sense.
I just added one more commit to unify the behavior. Note that I chose to keep `update_focused_file.update_root`, and `update_focused_file.update_cwd` has the same behavior but deprecated. Please double check, and feel free to make changes.

lua/nvim-tree.lua
```diff
@@ -373,6 +402,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
   open_on_setup_file = false,
   open_on_tab = false,
   sort_by = "name",
+  root_dirs = { "/usr/include" },
```
kyazdani42, line 405: i must ask though, why adding /usr/include by default ? I'm on nixos and this directory does not even exist, people on windows will also not have it. I'd prefer keeping this table empty by default.

lixvbnet: yes empty table is better. Updated.

alex-courtis: Fair enough. I didn't think of other OSs ;)

lua/nvim-tree.lua
alex-courtis: Thank you.

lua/nvim-tree.lua
alex-courtis: This is very useful:

```lua
		root_dirs = { "/usr/include/yaml-cpp", "/usr/include/linux", "/usr/include", },
```

alex-courtis (changes requested):
This tested OK for many cases.

- [x] Remove `init_root`

When `opts.update_cwd = true,` the user has indicated that they wish to use vim's new cwd, not vim's initial cwd.
When `opts.update_cwd = false,` init_root will be vim's cwd, so it is redundant.

alex-courtis:
> a better name would be `sync_root_with_cwd` or something like that for the main option

This is good. We are resolving any possible confusions with these renames.

I'll make a change after this one is resolved to deprecate and tidy.

alex-courtis:
> > update_cwd is useful whenever you `:cd some-folder`, which changes neovim base cwd, thus nvim-tree catches DirChanged event and updates. It is also useful i believe when multiple tabs have different cwd, changing between tabs will dispatch a DirChanged event (i think).
> 
> thanks, that make sense. I just added one more commit to unify the behavior. Note that I chose to keep `update_focused_file.update_root`, and `update_focused_file.update_cwd` has the same behavior but deprecated. Please double check, and feel free to make changes.

I often `:cd src` or `:cd lua` so that I can focus on code and not see tests / doc etc.

lixvbnet:
> This tested OK for many cases.
> 
> * [ ]  Remove `init_root`
> 
> When `opts.update_cwd = true,` the user has indicated that they wish to use vim's new cwd, not vim's initial cwd. When `opts.update_cwd = false,` init_root will be vim's cwd, so it is redundant.

There's a small use case though (which I use a lot..), that I often navigate up and down using `<C-]` and `-` in nvim-tree to look around source/library code.  So vim's cwd would change back and force.  When I come back to my previously open file in initial working directory, I'd like it to set root to my initial root rather than the folder containing the file. This way I never get lost.

alex-courtis:
> > This tested OK for many cases.
> > 
> > * [ ]   Remove `init_root`
> > 
> > When `opts.update_cwd = true,` the user has indicated that they wish to use vim's new cwd, not vim's initial cwd. When `opts.update_cwd = false,` init_root will be vim's cwd, so it is redundant.
> 
> There's a small use case though (which I use a lot..), that I often navigate up and down using `<C-]` and `-` in nvim-tree to look around source/library code. (although library root is set in `root_dirs`, but it's a big tree and sometimes you may want to cd into one package and focus on that part and change back to root afterwards). So vim's cwd would change back and force. When I come back to my previously open file in initial working directory, I'd like it to set root to my initial root rather than the folder containing the file. This way I never get lost.

That is a valid use case and would be useful. It's outside the scope of this change and we shouldn't be changing existing behavior to do this.

We should make a separate change to do this, with this functionality controlled by a (default false) option, say `opts.prefer_startup_root`.

lixvbnet:
> We should make a separate change to do this, with this functionality controlled by a (default false) option, say `opts.prefer_startup_root`.

yeah that's better. Added a commit for this option.

alex-courtis (approved):

PR #1333, opened 2022-06-10
feat: wrapped the setup recall guard in an option

dominicpalmer:
Following points raised in https://github.com/kyazdani42/nvim-tree.lua/issues/1313, this change wraps the setup call guard (from https://github.com/kyazdani42/nvim-tree.lua/commit/b1ecb75a6cf0bc83ec64d987c7f4128ada3c254e) in an option named `suppress_setup_recall_warning`.

While it makes sense to inform the user that they shouldn't recall `setup`, forcibly warning them every time via the UI is intrusive.

Note that this PR doesn't alter the default behaviour of always warning. The user must explicitly set `suppress_setup_recall_warning` to `true` to suppress the messages. 

The option also prevents the early return. If it's preferable to keep the early return in place, the block ought to be changed to:

```
  if M.setup_called then
    if not opts.suppress_setup_recall_warning then
      utils.warn "nvim-tree.lua setup called multiple times"
    end
    return
  end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Setup should be run in a lua file or in a lua heredoc (`:help lua-heredoc`) if u
 
 Legacy `g:` options have been migrated to the setup function. See [this issue](https://github.com/kyazdani42/nvim-tree.lua/issues/674) for information on migrating your configuration.
 
-Setup may only be run once; subsequent calls will result in a warning. Do not invoke from, say, a packer config function and then call it again later. You can suppress the warning using the `suppress_setup_recall_warning` option.
+Setup may only be run once; subsequent calls will be a NOP and will result a warning. Do not invoke from, say, a packer config function and then call it again later. You can suppress the warning using the `suppress_setup_recall_warning` option.
 
 ```vim
 " vimrc
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Setup should be run in a lua file or in a lua heredoc (`:help lua-heredoc`) if u
 
 Legacy `g:` options have been migrated to the setup function. See [this issue](https://github.com/kyazdani42/nvim-tree.lua/issues/674) for information on migrating your configuration.
 
-Setup may only be run once; subsequent calls will result in a warning. Do not invoke from, say, a packer config function and then call it again later.
+Setup may only be run once; subsequent calls will result in a warning. Do not invoke from, say, a packer config function and then call it again later. You can suppress the warning using the `suppress_setup_recall_warning` option.
 
 ```vim
 " vimrc
@@ -77,6 +77,7 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
   hijack_cursor = false,
   hijack_netrw = true,
   hijack_unnamed_buffer_when_opening = false,
+  suppress_setup_recall_warning = false,
   ignore_buffer_on_setup = false,
   open_on_setup = false,
   open_on_setup_file = false,
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -83,8 +83,10 @@ SETUP                                                *nvim-tree.setup*
 To configure the tree (and make it runnable), you should call the setup
 function.
 
-Setup may only be run once; subsequent calls will result in a warning. Do not
-invoke from, say, a packer config function and then call it again later.
+Setup may only be run once; subsequent calls will be a NOP and will result a
+warning. Do not invoke from, say, a packer config function and then call it
+again later. You can suppress the warning using the `suppress_setup_recall_warning`
+option.
 
 Values may be functions. Warning: this may result in unexpected behaviour.
 >
@@ -95,6 +97,7 @@ Values may be functions. Warning: this may result in unexpected behaviour.
       hijack_cursor = false,
       hijack_netrw = true,
       hijack_unnamed_buffer_when_opening = false,
+      suppress_setup_recall_warning = false,
       ignore_buffer_on_setup = false,
       open_on_setup = false,
       open_on_setup_file = false,
@@ -314,6 +317,11 @@ Can be one of 'name', 'case_sensitive', 'modification_time' or 'extension'.
 
 *nvim-tree.hijack_unnamed_buffer_when_opening*
 Opens in place of the unnamed buffer if it's empty.
+  Type: `boolean`, Default: `false`
+
+*nvim-tree.suppress_setup_recall_warning*
+Suppresses the warning that's displayed when setup is called multiple times.
+When suppressed, subsequent setup calls will still result in a NOP.
   Type: `boolean`, Default: `false`
 
 *nvim-tree.hijack_cursor*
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -36,7 +36,7 @@ Setup should be run in a lua file or in a lua heredoc (`:help lua-heredoc`) if u
 
 Legacy `g:` options have been migrated to the setup function. See [this issue](https://github.com/kyazdani42/nvim-tree.lua/issues/674) for information on migrating your configuration.
 
-Setup may only be run once; subsequent calls will result in a warning. Do not invoke from, say, a packer config function and then call it again later.
+Setup may only be run once; subsequent calls will result in a warning. Do not invoke from, say, a packer config function and then call it again later. You can suppress the warning using the `suppress_setup_recall_warning` option.
```
alex-courtis, line 39: Let's be explicit:

"Setup may only be run once; subsequent calls will be a NOP will result in a warning."

lua/nvim-tree.lua
```diff
@@ -589,17 +590,18 @@ function M.setup(conf)
...
-
   legacy.migrate_legacy_options(conf or {})
 
   validate_options(conf)
 
   local opts = merge_options(conf)
+
+  if not opts.suppress_setup_recall_warning and M.setup_called then
+    utils.warn "nvim-tree.lua setup called multiple times"
+    return
```
alex-courtis, line 601: We should return even when `suppress_setup_recall_warning` is set.

nvim-tree will behave unpredictably (duplicate autocommands etc) when setup is run twice.

alex-courtis (changes requested):
Please update :help as per readme.

Please run update-default-opts as per https://github.com/kyazdani42/nvim-tree.lua/blob/master/CONTRIBUTING.md

dominicpalmer:
Ran update-default-opts, updated help, updated readme wording, and switched to always returning early even if the warning is suppressed.

README.md
```diff
@@ -77,6 +77,7 @@ require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
   hijack_cursor = false,
   hijack_netrw = true,
   hijack_unnamed_buffer_when_opening = false,
+  suppress_setup_recall_warning = false,
```
kylo252, line 80:
```suggestion
```

doc/nvim-tree-lua.txt
```diff
@@ -95,6 +97,7 @@ Values may be functions. Warning: this may result in unexpected behaviour.
       hijack_cursor = false,
       hijack_netrw = true,
       hijack_unnamed_buffer_when_opening = false,
+      suppress_setup_recall_warning = false,
```
kylo252, line 100:
```suggestion
```

doc/nvim-tree-lua.txt
```diff
@@ -316,6 +319,11 @@ Can be one of 'name', 'case_sensitive', 'modification_time' or 'extension'.
 Opens in place of the unnamed buffer if it's empty.
   Type: `boolean`, Default: `false`
 
+*nvim-tree.suppress_setup_recall_warning*
+Suppresses the warning that's displayed when setup is called multiple times.
+When suppressed, subsequent setup calls will still result in a NOP.
+  Type: `boolean`, Default: `false`
+
```
kylo252, line 326:
```suggestion
```

lua/nvim-tree.lua
```diff
@@ -368,6 +368,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
   hijack_cursor = false,
   hijack_netrw = true,
   hijack_unnamed_buffer_when_opening = false,
+  suppress_setup_recall_warning = false,
```
kylo252, line 371:
```suggestion
```

lua/nvim-tree.lua
```diff
@@ -589,17 +590,20 @@ function M.setup(conf)
...
   legacy.migrate_legacy_options(conf or {})
 
   validate_options(conf)
 
   local opts = merge_options(conf)
+
+  if M.setup_called then
+    if not opts.suppress_setup_recall_warning then
+      utils.warn "nvim-tree.lua setup called multiple times"
+    end
```
kylo252, line 602:
```suggestion
utils.warn "nvim-tree.lua setup called multiple times"
```

kylo252 (commented):
Please switch to using the builtin logger, where the user can set the level once for all errors/warnings/info/debug/trace.

are you going to create a suppress flag for each new warning you add?

every instance of `util.warn` should be replaced with `log[LEVEL](...)` instead

--- 

Furthermore, this is just an implementation detail about allowing to call `setup()` multiple times or not, so it should be a debug entry and not a warning (since there's nothing to do about it, unless you meant it as a breaking change, which you should've clarified in https://github.com/kyazdani42/nvim-tree.lua/pull/1308 and marked it as such)

dominicpalmer:
> are you going to create a suppress flag for each new warning you add?

I didn't add the warning. The addition of a suppression flag was to honour @alex-courtis's original change, while giving the user the option to no longer see the warning. My original argument was to remove the guard altogether and instead focus on allowing config change without restarting, but I know that's a bigger undertaking.

@kylo252 - I've replaced the call to `utils.warn` with a log write and removed the option

alex-courtis:
> Please switch to using the builtin logger, where the user can set the level once for all errors/warnings/info/debug/trace.
> ...
> every instance of `util.warn` should be replaced with `log[LEVEL](...)` instead

The logger is for diagnostic purposes only, when reporting an issue. It is not enabled by default and we do not expect the user to look at the log under normal operations.

> are you going to create a suppress flag for each new warning you add?

Warnings are explicit and the user is expected to resolve them. This suppression is unique and applies for a specific and exceptional use case.

> Furthermore, this is just an implementation detail about allowing to call `setup()` multiple times or not, so it should be a debug entry and not a warning (since there's nothing to do about it, unless you meant it as a breaking change, which you should've clarified in #1308 and marked it as such)

Calling setup multiple times resulted in nvim-tree behaving unexpectedly and incorrectly. Many users raised issues following multiple setups, resolved by removing the extra calls to setup.

#1308 was not a breaking change. It was a bug fix that prevented the incorrect / unexpected behaviour. The warning was added so that the (many) users experiencing such problems were alerted to them and could resolve them without needing to raise an issue.

The most common problem case: user calls setup with no options via packer config. User calls setup again with their options. nvim-tree runs with mostly defaults and many of the user's options not applied.

alex-courtis (changes requested):
Change approved after reverting bd7a941

alex-courtis:
Raised #1338 

It is not a trivial amount of work, however would satisfy all users.

dominicpalmer:
Alright. If the suppression will get approved as-is then I'm for it. I'm representing the user here, and frankly, any short term fix that stops the message getting displayed everytime I hot reload my config, I'm all for it. Now that #1338 is raised there's an incentive for a full fix to be worked on regardless.

kyazdani42:
i've opened #1340 which should make setup work as expected, thus not needing this PR. 
Can you guys check it and report any inconsistent behavior you might observe ?

kyazdani42:
closing in favor of merged #1340

PR #1340, opened 2022-06-12
chore(setup): make setup idempotent

### LINKED ISSUE #1338: idempotent setup capable of multiple invocations
setup may only be called once, with a suppressable warning raised on subsequent calls, which are a NOP.

- Users do not like this warning #1333
- User may not invoke setup again with a different configuration
- Valid use cases such as hot-reloading the vim config exist

setup may be called multiple times, with all new options being applied.

Continue warning the user with subsequent setup calls being a NOP

**Additional context**
Changes needed, but not limited to:
- Audit each module's setup to ensure idemoptency
- Ensure all computed fields such as `View.winopts` are reset
- Teardown autocommands. Using an augroup is not sufficient as some autocommands are dependent on options.
- Teardown watchers

kyazdani42:
I'm not a 100% sure this works as expected for every user configuration, but tested on mine and didn't notice weird behavior.
Autocmd clear properly and set up properly on 2nd setup call
git cleared whenever tree is loaded
view is closed and reopened if open on setup is true in 2nd setup call.

fixes #1338 
superseeds #1333

### DIFF
--- a/lua/nvim-tree/actions/init.lua
+++ b/lua/nvim-tree/actions/init.lua
@@ -6,7 +6,8 @@ local view = require "nvim-tree.view"
 local util = require "nvim-tree.utils"
 local nvim_tree_callback = require("nvim-tree.config").nvim_tree_callback
 
-local DEFAULT_MAPPINGS = { -- BEGIN_DEFAULT_MAPPINGS
+-- BEGIN_DEFAULT_MAPPINGS
+local DEFAULT_MAPPINGS = {
   {
     key = { "<CR>", "o", "<2-LeftMouse>" },
     action = "edit",
@@ -217,7 +218,8 @@ local DEFAULT_MAPPINGS = { -- BEGIN_DEFAULT_MAPPINGS
     action = "toggle_help",
     desc = "toggle help",
   },
-} -- END_DEFAULT_MAPPINGS
+}
+-- END_DEFAULT_MAPPINGS
 
 local M = {
   mappings = {},
--- a/lua/nvim-tree/actions/init.lua
+++ b/lua/nvim-tree/actions/init.lua
@@ -6,223 +6,223 @@ local view = require "nvim-tree.view"
 local util = require "nvim-tree.utils"
 local nvim_tree_callback = require("nvim-tree.config").nvim_tree_callback
 
--- BEGIN_DEFAULT_MAPPINGS
-local M = {
-  mappings = {
-    {
-      key = { "<CR>", "o", "<2-LeftMouse>" },
-      action = "edit",
-      desc = "open a file or folder; root will cd to the above directory",
-    },
-    {
-      key = "<C-e>",
-      action = "edit_in_place",
-      desc = "edit the file in place, effectively replacing the tree explorer",
-    },
-    {
-      key = "O",
-      action = "edit_no_picker",
-      desc = "same as (edit) with no window picker",
-    },
-    {
-      key = { "<C-]>", "<2-RightMouse>" },
-      action = "cd",
-      desc = "cd in the directory under the cursor",
-    },
-    {
-      key = "<C-v>",
-      action = "vsplit",
-      desc = "open the file in a vertical split",
-    },
-    {
-      key = "<C-x>",
-      action = "split",
-      desc = "open the file in a horizontal split",
-    },
-    {
-      key = "<C-t>",
-      action = "tabnew",
-      desc = "open the file in a new tab",
-    },
-    {
-      key = "<",
-      action = "prev_sibling",
-      desc = "navigate to the previous sibling of current file/directory",
-    },
-    {
-      key = ">",
-      action = "next_sibling",
-      desc = "navigate to the next sibling of current file/directory",
-    },
-    {
-      key = "P",
-      action = "parent_node",
-      desc = "move cursor to the parent directory",
-    },
-    {
-      key = "<BS>",
-      action = "close_node",
-      desc = "close current opened directory or parent",
-    },
-    {
-      key = "<Tab>",
-      action = "preview",
-      desc = "open the file as a preview (keeps the cursor in the tree)",
-    },
-    {
-      key = "K",
-      action = "first_sibling",
-      desc = "navigate to the first sibling of current file/directory",
-    },
-    {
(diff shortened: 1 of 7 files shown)


alex-courtis:
Checking:
- [ ] watchers
- [ ] autocommands
- [ ] actions/change-dir.lua
- [ ] actions/copy-paste.lua
- [ ] actions/create-file.lua
- [ ] actions/open-file.lua
- [ ] actions/init.lua
- [ ] actions/system-open.lua
- [ ] actions/remove-file.lua
- [ ] actions/trash.lua
- [ ] colors.lua
- [ ] actions/expand-all.lua
- [ ] actions/rename-file.lua
- [ ] explorer/explore.lua
- [ ] explorer/init.lua
- [ ] explorer/reload.lua
- [ ] explorer/filters.lua
- [ ] explorer/sorters.lua
- [ ] explorer/watch.lua
- [ ] git/init.lua
- [ ] lib.lua
- [ ] live-filter.lua
- [ ] log.lua
- [ ] renderer/components/git.lua
- [ ] renderer/components/icons.lua
- [ ] renderer/components/padding.lua
- [ ] renderer/init.lua
- [ ] diagnostics.lua
- [ ] view.lua

lua/nvim-tree/git/init.lua
```diff
@@ -134,6 +134,16 @@ function M.load_project_status(cwd)
   return M.projects[project_root]
 end
 
+function M.purge_state()
+  for _, p in pairs(M.projects) do
+    if p.watcher then
+      p.watcher:stop()
```
alex-courtis, line 140: It might be easier (and more reliable) to just create a `Watcher.purge` that goes through all `_watchers`.

We won't need to clear node and git watchers explicitly, and any future watchers will be purged with no effort required.

alex-courtis (changes requested):
- [ ] log.lua - need to clear M.path
- [ ] filters.lua - M.ignore_list needs to be cleared
- [ ] diagnostics.lua - need to clear autocommands
- [ ] delete keybindings and set them back; anything like `*@:lua require'nvim-tree.actions'.on_keypress` I guess

alex-courtis:
> * we might want to log the setup calls ?

Not sure that is worth the effort. We are confident that this works :shrug:

alex-courtis:
> view is closed and reopened if open on setup is true in 2nd setup call.

nice

lua/nvim-tree/git/init.lua
kyazdani42: for log.lua, i don't think we should clear M.path, since it could help detecting subsequent calls. Also the setup function of log.lua already removes the file if the truncate option is true.

kyazdani42:
@kylo252 this is something i've thought about for managing state and events better, but i think it's way outside the scope of this PR.
I think we should have an internal event api to manage event and logic flows, and an external document event api for user to subscribe to, more flexible than the existing one (which is already good i believe, but could be greater)

lua/nvim-tree/diagnostics.lua
```diff
@@ -38,29 +28,13 @@ end
...
-  else
-    -- nvim version < 0.7
-    for buf, diagnostics in pairs(vim.lsp.diagnostic.get_all()) do
-      if a.nvim_buf_is_valid(buf) then
-        local bufname = a.nvim_buf_get_name(buf)
-        if not buffer_severity[bufname] then
-          local severity = get_lowest_severity(diagnostics)
-          buffer_severity[bufname] = severity
-        end
+  for _, diagnostic in ipairs(vim.diagnostic.get()) do
```
alex-courtis, line 31: clean

lua/nvim-tree/diagnostics.lua
```diff
@@ -166,17 +145,6 @@ function M.setup(opts)
   for lhs, rhs in pairs(links) do
     vim.cmd("hi def link " .. lhs .. " " .. rhs)
   end
-
-  if M.enable then
```
alex-courtis, line 170: Let's see how it goes without the autocommands...

kyazdani42: the autocmds were just moved in the nvim-tree setup for autocmds

kyazdani42:
@alex-courtis i think this should be ready now.
I've cleaned up the keymaps, but i'm not sure wether we should refactor the mappings table to reset it to its default.

lua/nvim-tree/actions/init.lua
```diff
@@ -397,6 +397,22 @@ local function copy_mappings(user_mappings)
   return user_mappings
 end
 
+local function cleanup_existing_mappings()
```
alex-courtis, line 400: That will remove any mappings the user manually hacked in, but their setup should add them back.

alex-courtis:
> @alex-courtis i think this should be ready now. I've cleaned up the keymaps, but i'm not sure wether we should refactor the mappings table to reset it to its default.

We can make it easy for ourselves as we do with options:
- change M.mappings to `local DEFAULT_MAPPINGS = { -- BEGIN_DEFAULT_MAPPINGS` which will be immutable
- after cleanup_existing_mappings `M.mappings = vim.deepcopy(DEFAULT_MAPPINGS)`

lua/nvim-tree/watcher.lua
```diff
@@ -74,4 +74,11 @@ end
 
 M.Watcher = Watcher
 
+function M.purge_watchers()
+  for _, watcher in pairs(M._watchers) do
+    watcher:stop()
+  end
+  M._watchers = {}
```
alex-courtis, line 81: Pushed this up.

We were keeping the old watchers and so the new ones would not start.

alex-courtis (changes requested):
Working well.

- [ ] https://github.com/kyazdani42/nvim-tree.lua/pull/1340#issuecomment-1159840095

kyazdani42:
I've added the final changes, i think it is good to go :)

alex-courtis (approved):
Pushed a fix for the update script

alex-courtis:
> > Setup may only be run once; subsequent calls will result in a warning.
> 
> Does this still apply now? I don't get a warning anymore. smile

Thank you. Doc updated.

PR #1343, opened 2022-06-13
draft / ideas for git job concurrency protection

alex-courtis:
I toyed with protecting at the git runner level, however that resulted in a lot of unwanted drawing etc, with outdated information that may "flash" incorrect information.

Exiting the node watcher early results in only the successful task reloading/drawing etc. with the latest git data.

At most 2 git jobs will be run when queueing occurs. That can be debounced away if we want to optimise further.

TODO: refactor git / node watchers to use one common, guarded function.

### DIFF
--- a/lua/nvim-tree/explorer/watch.lua
+++ b/lua/nvim-tree/explorer/watch.lua
@@ -3,13 +3,10 @@ local utils = require "nvim-tree.utils"
 local git = require "nvim-tree.git"
 local Watcher = require("nvim-tree.watcher").Watcher
 
-local M = {}
-
-local function reload_and_get_git_project(path)
-  local project_root = git.get_project_root(path)
-  git.reload_project(project_root)
-  return project_root, git.get_project(project_root) or {}
-end
+local M = {
+  running_project = {},
+  pending_project = {},
+}
 
 local function update_parent_statuses(node, project, root)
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/explorer/watch.lua
```diff
@@ -3,13 +3,10 @@ local utils = require "nvim-tree.utils"
...
-local M = {}
-
-local function reload_and_get_git_project(path)
-  local project_root = git.get_project_root(path)
-  git.reload_project(project_root)
-  return project_root, git.get_project(project_root) or {}
-end
+local M = {
+  running_project = {},
+  pending_project = {},
```
alex-courtis, line 8: This pattern could probably be extracted for use by LSP and COC.

lua/nvim-tree/explorer/watch.lua
```diff
@@ -35,18 +32,38 @@ function M.create_watcher(absolute_path)
...
+      log.line("watcher", "node event '%s'", path)
       local n = utils.get_node_from_path(absolute_path)
       if not n then
         return
       end
-      log.line("watcher", "node event '%s'", path)
 
+      local project_root = git.get_project_root(path)
+
+      if M.running_project[project_root] then
```
alex-courtis, line 43: Thread safety is not a concern here.

PR #1348, opened 2022-06-15
feat: add cwd arg and fallback to open_replacing_current_buffer

javiertury:
Make `open_replacing_current_buffer` reach feature parity with `open`

- Accept `cwd` argument
- Fallback to `vim.loop.cwd()` instead of `return`

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -53,18 +53,22 @@ function M.open(cwd)
   end
 end
 
-function M.open_replacing_current_buffer()
+function M.open_replacing_current_buffer(cwd)
   if view.is_visible() then
     return
   end
 
   local buf = api.nvim_get_current_buf()
   local bufname = api.nvim_buf_get_name(buf)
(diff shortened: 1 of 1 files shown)


lua/nvim-tree.lua
```diff
@@ -53,18 +53,22 @@ function M.open(cwd)
...
     return
   end
 
   local buf = api.nvim_get_current_buf()
   local bufname = api.nvim_buf_get_name(buf)
-  if bufname == "" or vim.loop.fs_stat(bufname) == nil then
-    return
+
+  if cwd == "" or cwd == nil then
+    if bufname ~= "" and vim.loop.fs_stat(bufname) ~= nil then
```
alex-courtis, line 65: Let's retain the existing behaviour of returning early when this is false.

alex-courtis (changes requested):
Looks good, works.

kyazdani42 (approved):
LGTM

alex-courtis (approved):

PR #1357, opened 2022-06-18
fix(#1356): view.close switch_buf_if_last_buf prefers alt buf

### LINKED ISSUE #1356: Closing nvim-tree in vinegar mode will switch to first buffer
### Description

When opening nvim-tree in vinegar mode (via `open_replacing_current_buffer()`), closing the tree (via `q`) will switch the buffer to the first buffer.

You can simply open three buffers, open the third buffer (when inspecting via `:buffers`, you should be in the third), and then you should go into the tree with the config below.

Currently, I simply mapping a closing command like this:

```lua
local function close_tree()
    vim.cmd [[ buffer # ]]
end

local list = {
    ...
    { key = 'q', action = 'close_tree', action_cb = close_tree },
}

nvim_tree.setup {
    ...
    view = {
        mappings = {
            list = list,
        },
    },
}
```

### Neovim version

```text
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@Monterey

Features: +acl +iconv +tui
```

macOS 12.4

### nvim-tree version

bdb6d4a

### Minimal config

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {}
  function toggle_replace()
      local view = require 'nvim-tree.view'
      if view.is_visible() then
          view.close()
      else
          require('nvim-tree').open_replacing_current_buffer()
      end
  end

  vim.keymap.set('n', '-', toggle_replace)
end
```

### Steps to reproduce

1. nvim -nu /tmp/nvt-min.lua ./fileA
2. Press <kbd>-</kbd>, and open ./fileB
3. Press <kbd>-</kbd>, and then <kbd>q</kbd>

### Expected behavior

fileB is open

### Actual behavior

fileA is open

alex-courtis:
closes #1356 

Use `#` if available when closing last buf, falling back to next as per previous behaviour.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -130,16 +130,29 @@ local function open_window()
   set_window_options_and_buffer()
 end
 
-local function get_existing_buffers()
-  return vim.tbl_filter(function(buf)
-    return a.nvim_buf_is_valid(buf) and vim.fn.buflisted(buf) == 1
-  end, a.nvim_list_bufs())
+local function get_alt_or_next_buf()
+  local alt_bufnr = vim.fn.bufnr "#"
+
+  local alt_buf = nil
+  local next_buf = nil
+
+  for _, buf in ipairs(a.nvim_list_bufs()) do
+    if a.nvim_buf_is_valid(buf) and vim.fn.buflisted(buf) == 1 then
+      next_buf = next_buf or buf
+      if buf == alt_bufnr then
+        alt_buf = buf
+      end
+    end
+  end
+
+  return alt_buf or next_buf
 end
 
 local function switch_buf_if_last_buf()
   if #a.nvim_list_wins() == 1 then
-    if #get_existing_buffers() > 0 then
-      vim.cmd "sbnext"
+    local buf = get_alt_or_next_buf()
+    if buf then
+      vim.cmd("sb" .. buf)
     else
       vim.cmd "new"
     end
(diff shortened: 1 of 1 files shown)


hgiesel:
This works for me

lua/nvim-tree/view.lua
```diff
@@ -130,16 +130,29 @@ local function open_window()
...
+  for _, buf in ipairs(a.nvim_list_bufs()) do
+    if a.nvim_buf_is_valid(buf) and vim.fn.buflisted(buf) == 1 then
+      next_buf = next_buf or buf
+      if buf == alt_bufnr then
+        alt_buf = buf
+      end
+    end
+  end
+
+  return alt_buf or next_buf
```
kyazdani42, line 148: The logic could be greatly simplified:
- check at the start that alt_bufnr is valid and listed, and return it in that case (create a small function like is_buf_displayed or something like that).
- then do the loop and return whenever you find a valid buffer

alex-courtis: nice

kyazdani42 (approved):

PR #1373, opened 2022-06-26
feat(watcher): partial git refresh

alex-courtis:
First PR of many. These can be incrementally merged.

Optimising watchers has two parts:
1. git refresh
2. file system refresh

Preventing concurrency for both is a problem as they cover a different set of files.

Avoid this problem for directory watchers by only git refreshing the sub-tree and making the sets of files the same.

Experimentation has shown that this is a substantial time save: ~1/20th for my large [test repo](https://github.com/qmk/qmk_firmware)

### DIFF
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -22,25 +22,36 @@ function M.reload()
   return M.projects
 end
 
-function M.reload_project(project_root)
+function M.reload_project(project_root, path)
   local project = M.projects[project_root]
   if not project or not M.config.enable then
     return
   end
 
-  local watcher = M.projects[project_root].watcher
-  M.projects[project_root] = {}
+  if path and not path:match("^" .. project_root) then
+    path = nil
+  end
+
   local git_status = Runner.run {
     project_root = project_root,
+    path = path,
     list_untracked = git_utils.should_show_untracked(project_root),
     list_ignored = true,
     timeout = M.config.timeout,
   }
-  M.projects[project_root] = {
-    files = git_status,
-    dirs = git_utils.file_status_to_dir_status(git_status, project_root),
-    watcher = watcher,
-  }
+
+  if path then
+    for p in pairs(project.files) do
+      if p:match("^" .. path) then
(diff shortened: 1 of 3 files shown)


lua/nvim-tree/git/init.lua
```diff
@@ -22,25 +22,36 @@ function M.reload()
...
-function M.reload_project(project_root)
+function M.reload_project(project_root, path)
   local project = M.projects[project_root]
   if not project or not M.config.enable then
     return
   end
 
-  local watcher = M.projects[project_root].watcher
-  M.projects[project_root] = {}
+  if path and not path:match("^" .. project_root) then
```
kyazdani42, line 31: how is this even possible ?

lua/nvim-tree/git/runner.lua
```diff
@@ -45,7 +45,7 @@ function Runner:_getopts(stdout_handle, stderr_handle)
   local untracked = self.list_untracked and "-u" or nil
   local ignored = (self.list_untracked and self.list_ignored) and "--ignored=matching" or "--ignored=no"
   return {
-    args = { "--no-optional-locks", "status", "--porcelain=v1", ignored, untracked },
+    args = { "--no-optional-locks", "status", "--porcelain=v1", ignored, untracked, self.path },
```
kyazdani42, line 48: this is great

kyazdani42 (approved):
that's a great change :) Please just answer my question and feel free to merge if you feel it's not valid

lua/nvim-tree/git/init.lua
alex-courtis: How is "no match" possible? Not sure...

I'd prefer to have this safety measure in place, as falling back to a full refresh is better than having the git invocation fail.

kyazdani42: what i meant is that if the path does not match, it will not load a project and thus return early.

alex-courtis: Ah I see... if there is no match, the update for that directory would be meaningless.

I'll change in the next PR, to keep things simple.

PR #1374, opened 2022-06-26
feat(watcher): debounce FS watchers

alex-courtis:
requires #1373 

Not very useful until we move to fs_event as fs_poll is aready throttled.

I have played around a few options but was not happy:
1. Abandon refresh when there is a refresh in progress for that directory - can miss actual changes
2. Queue one refresh when there is one in progress - can result in two refreshes, which will be a big problem when we move to real time FS events

Debounce refreshes by absolute path seems the best option.
Having multiple refresh events running for different paths is OK, as we are only looking at the FS and git for that path.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -273,6 +273,7 @@ Setup may only be run once; subsequent calls will result in a warning.
       filesystem_watchers = {
         enable = false,
         interval = 100,
+        debounce = 50,
       },
       git = {
         enable = true,
@@ -523,6 +524,10 @@ This will be experimental for a few weeks and will become the default.
     Increase to at least 1000ms if changes are not visible. See
     https://github.com/luvit/luv/blob/master/docs.md#uvfs_poll_startfs_poll-path-interval-callback
       Type: `number`, Default: `100` (ms)
+
+    *nvim-tree.filesystem_watchers.debounce*
+    Idle milliseconds between filesystem change and action.
+      Type: `number`, Default: `50` (ms)
 
 *nvim-tree.view*
 Window / buffer setup.
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -273,7 +273,7 @@ Setup may only be run once; subsequent calls will result in a warning.
       filesystem_watchers = {
         enable = false,
         interval = 100,
-        debounce = 50,
+        debounce_delay = 50,
       },
       git = {
         enable = true,
@@ -525,7 +525,7 @@ This will be experimental for a few weeks and will become the default.
     https://github.com/luvit/luv/blob/master/docs.md#uvfs_poll_startfs_poll-path-interval-callback
       Type: `number`, Default: `100` (ms)
 
-    *nvim-tree.filesystem_watchers.debounce*
+    *nvim-tree.filesystem_watchers.debounce_delay*
     Idle milliseconds between filesystem change and action.
       Type: `number`, Default: `50` (ms)
 
--- a/lua/nvim-tree/explorer/watch.lua
+++ b/lua/nvim-tree/explorer/watch.lua
@@ -3,9 +3,7 @@ local utils = require "nvim-tree.utils"
 local git = require "nvim-tree.git"
 local Watcher = require("nvim-tree.watcher").Watcher
 
-local M = {
-  refreshing_paths = {},
-}
+local M = {}
 
 local function reload_and_get_git_project(path)
   local project_root = git.get_project_root(path)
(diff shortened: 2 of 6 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -273,6 +273,7 @@ Setup may only be run once; subsequent calls will result in a warning.
       filesystem_watchers = {
         enable = false,
         interval = 100,
+        debounce = 50,
```
alex-courtis, line 276: This doesn't really matter for now; we can tune it when we see how fast fs_event fires.

lua/nvim-tree/utils.lua
```diff
@@ -328,4 +330,25 @@ function M.key_by(tbl, key)
...
+function M.debounce(context, timeout, fn)
+  if M.debouncers[context] then
+    M.debouncers[context]:close()
+  end
+
+  M.debouncers[context] = uv.new_timer()
+  M.debouncers[context]:start(
+    timeout,
+    0,
+    vim.schedule_wrap(function()
```
alex-courtis, line 346: Should be thread safe.

lua/nvim-tree/explorer/watch.lua
```diff
@@ -35,25 +60,17 @@ function M.create_watcher(absolute_path)
...
+        refresh_path(path)
+      end)
     end,
   }
 end
 
 function M.setup(opts)
   M.enabled = opts.filesystem_watchers.enable
   M.interval = opts.filesystem_watchers.interval
+  M.debounce = opts.filesystem_watchers.debounce
```
kyazdani42, line 73: i'd call this debounce_timeout or something like that :)

lua/nvim-tree/utils.lua
```diff
@@ -328,4 +330,25 @@ function M.key_by(tbl, key)
   return keyed
 end
 
+---Execute fn timeout ms after the last invocation with context
+---@param context string identifies the callback to debounce
+---@param timeout number ms to wait
+---@param fn function callback to execute on completion
+function M.debounce(context, timeout, fn)
+  if M.debouncers[context] then
+    M.debouncers[context]:close()
```
kyazdani42, line 339: i'd pcall this in case this is already closing somehow.

lua/nvim-tree/explorer/watch.lua
```diff
@@ -22,6 +24,29 @@ local function is_git(path)
...
+  if not n then
+    return
+  end
+  log.line("watcher", "node event '%s'", path)
+
+  if M.refreshing_paths[path] then
+    log.line("watcher", "node event already running")
+    return
+  end
+  M.refreshing_paths[path] = true
```
kyazdani42, line 38: this is already managed by the debounce function right ? as this is synchronous it should not run 2 events in parallel anyway.

alex-courtis: Oh bugger... that should not have been committed. It was from "2. Queue one refresh when there is one in progress".

Removing, thanks for spotting.

lua/nvim-tree/utils.lua
alex-courtis: Done.

`pcall(uv.close, M.debouncers[context])` looks strange; `uv.close` is actually `uv_handle_t` close.

It's OK if the call fails, the timer will be released after this.

lua/nvim-tree/explorer/watch.lua
alex-courtis: `debounce_delay`

lua/nvim-tree/watcher.lua
```diff
@@ -48,7 +48,7 @@ function Watcher:start()
     if err then
       log.line("watcher", "poll_cb for %s fail : %s", self._opts.absolute_path, err)
     else
-      self._opts.on_event(self._opts.absolute_path)
+      self._opts.on_event(self._opts)
```
alex-courtis, line 51: Allows users to add their own data to opts.

lua/nvim-tree/git/init.lua
```diff
@@ -120,19 +123,25 @@ function M.load_project_status(cwd)
...
     log.line("watcher", "git start")
     watcher = Watcher.new {
       absolute_path = utils.path_join { project_root, ".git" },
-      interval = M.config.watcher.interval,
-      on_event = function()
+      project_root = project_root,
+      interval = M.config.filesystem_watchers.interval,
+      on_event = function(opts)
+        log.line("watcher", "git event scheduled '%s'", opts.project_root)
+        utils.debounce("git:watcher:" .. opts.project_root, M.config.filesystem_watchers.debounce_delay, function()
```
alex-courtis, line 138: Forgot to debounce the git watcher.

kyazdani42 (approved):

alex-courtis:
Whoops... merged to feat/optimise-watchers-1-partial-git-refresh

Manually rebased / merged to master.

PR #1385, opened 2022-07-01
fix(actions): close_node doesn't close for grouped node

rammiah:
When `group_empty` option is enabled, parent_node and close_node dosen't close correctly.

example:
image

after call close_node on node d.go expect: 
image

actual:
image

So I look up until the first group node, and close it will be expected.


kyazdani42 (approved):
thanks for the fix :)

PR #1388, opened 2022-07-02
feat(actions): expand_all 'exclude' option

KrasimirZahariev:
Hello, guys !

A small proposal from me - an 'exclude' option to the expand_all action.

I find it useful when recursively expanding child nodes. Say you have nested '.git', or you don't want to expand some kind of 'build' directories.
This also helps if you 'expand_all' while you have selected a file, which has siblings you don't want to expand.
Dunno is it a common use case, that is why I defaulted to an empty table value.


kyazdani42 (approved):
LGTM :)

PR #1392, opened 2022-07-03
chore(iterators): create Iterator module and migrate iterators to use it

kyazdani42:
Missing one iterator (see TODO comment), because i'm not sure about the interface. This is a first attempt to clean up the repetitive code for iterators. 
Not sure this is the best implementation i could go with, but it works for now.

### DIFF
--- a/lua/nvim-tree/actions/expand-all.lua
+++ b/lua/nvim-tree/actions/expand-all.lua
@@ -1,6 +1,7 @@
 local core = require "nvim-tree.core"
 local renderer = require "nvim-tree.renderer"
 local utils = require "nvim-tree.utils"
+local Iterator = require "nvim-tree.iterators.node-iterator"
 
 local M = {}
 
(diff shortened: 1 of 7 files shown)


lua/nvim-tree/actions/expand-all.lua
```diff
@@ -20,34 +21,38 @@ local function expand(node)
...
-    for _, node in pairs(parent.nodes) do
-      if node.nodes and not node.open and not M.EXCLUDE[node.name] then
-        expansion_count = expansion_count + 1
-        expand(node)
-      end
-
-      if node.open then
-        if iterate(node) then
-          return true
+    Iterator.new(parent.nodes)
```
alex-courtis, line 39: I love the fluid nature. This is classic builder pattern. Perhaps we could make the naming more concise e.g.

```lua
NodeIterator.builder(parent.nodes)
  :hidden()
  :matcher(...)
  :applier(...)
  :recursor(...)
  :build()()
```

The `:build()()` is a bit redundant and implies that we have an intermediate `NodeIteratorBuilder` which doesn't add much value, so `:consume()` or `iterate()` is appropriate.

lua/nvim-tree/iterator.lua
```diff
@@ -0,0 +1,67 @@
...
+    _apply_fn_on_node = function(_) end,
+    _match = function(_) end,
+    _recurse_with = function(node)
+      return node.nodes
+    end,
+  }, Iterator)
+end
+
+function Iterator:allow_hidden()
+  self._filter_hidden = function(_)
```
alex-courtis, line 19: /nit this could just be a boolean member.

alex-courtis (changes requested):
Fantastic.

- NOP case `Iterator.new(nodes):consume()` tested ok.
- `Iterator.new(nodes)` fails but I don't think that matters.
- Tested expand-all as it is a big change
- Reviewed other usages but they are all trivial refactors and did not test

PR #1394, opened 2022-07-04
Add instructions to disable icon display

herrbischoff:
I had to dig quite a bit, including in the code itself to figure out how to do this. This shouldn't be needed for a a straight-forward configuration change. So, in order to save (new) users some time, I amended the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,26 @@ For complete list of available configuration options see [:help nvim-tree.setup]
 
 Each option is documented in `:help nvim-tree.OPTION_NAME`. Nested options can be accessed by appending `.`, for example [:help nvim-tree.view.mappings](doc/nvim-tree-lua.txt)
 
+Using the icons requires a [patched font](https://www.nerdfonts.com/). If you
+don't have one installed, you're probably seeing undesirable "?" box glyphs.
+To disable the display of icons altogether, you can use the following minimal
+configuration:
+
+```lua
+require("nvim-tree").setup({
+  renderer = {
+    icons = {
+      show = {
+        file = false,
+        folder = false,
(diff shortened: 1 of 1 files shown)


gegoune:
I disagree with assumption that it is not straight forward to disable icons therefore don't see this change as warranted. It's all documented.

README.md
```diff
@@ -81,6 +81,26 @@ For complete list of available configuration options see [:help nvim-tree.setup]
 
 Each option is documented in `:help nvim-tree.OPTION_NAME`. Nested options can be accessed by appending `.`, for example [:help nvim-tree.view.mappings](doc/nvim-tree-lua.txt)
 
+Using the icons requires a [patched font](https://www.nerdfonts.com/). If you
+don't have one installed, you're probably seeing undesirable "?" box glyphs.
+To disable the display of icons altogether, you can use the following minimal
+configuration:
```
alex-courtis, line 87: This is mostly covered in `:help nvim-tree-introduction` but clarity is nice.

We can just add "To disable the display of icons altogether set each of |renderer.icons.show| to false." in that section.

alex-courtis (changes requested):

README.md
herrbischoff: Good idea, thanks for the suggestion.

alex-courtis: Agreed

herrbischoff: Sounds good. I will make the changes shortly.

alex-courtis:
Pushed https://github.com/kyazdani42/nvim-tree.lua/commit/08db5a576dbdc44e3996e643978c8e6a5a460d56

PR #1396, opened 2022-07-04
Optionally suppress the symlink destination

tobywan:
Setting `nvim-tree.renderer.symlink_destination=false` will prevent the arrow and symlink `link_to` file
appearing in the tree.
The original symlink name, icons and colour are not affected.

This is useful when `nvim-tree.view.adaptive_size = true`, because the full path of the `node.link_to` is used which can mean that the tree window is excessively large for files deep in the file system.
Compare:
symlink_destination-true

symlink_destination-false


kyazdani42:
@tobywan looks good to me, just fix the style :)

tobywan:
Thanks, I pushed a 3rd commit that _should_ have fixed it. Is your comment after that push?

kyazdani42:
yes, check the CI :) there is a missing space before the `..`

tobywan:
> yes, check the CI :) there is a missing space before the `..`

D'Oh!
My bad, spotted the extra leading space but missed that one.

alex-courtis:
Not able to review right now, however this is a fantastic option that I will definitely use.

kyazdani42 (approved):

PR #1398, opened 2022-07-05
fix(actions): change_dir to wrong directory

rammiah:
vim.cmd only accept 1 arg, path will be discarded, see `:h vim.cmd`，so we should concate cmd and args


rammiah:
by default cd will change to user home directory

kyazdani42 (approved):

kyazdani42:
thanks :)

PR #1399, opened 2022-07-05
fix(iterators): the index does not increase

edwarmv:


kyazdani42 (approved):

kyazdani42:
well my bad, thanks for the patch :)

PR #1410, opened 2022-07-09
refactor(actions): move actions into semantic modules

kyazdani42:
The purpose of this PR is to cleanup the actions module which got a bit overwhelming when searching for a particular module.


lua/nvim-tree/actions/dispatch.lua
```diff
@@ -5,43 +5,51 @@ local M = {}
 
 local Actions = {
```
alex-courtis, line 6: My first instinct is was to move each of these to an dispatch.lua in each directory however that adds no value.

alex-courtis (approved):
Very nice. Did not test this refactor.

Groupings make sense. Maybe some small name tweaks:
change-dir -> root
fs-operations -> fs
open-file/open-file.lua -> node/open.lua

lua/nvim-tree/actions/dispatch.lua
kyazdani42: i guess so :) i still want to improve this module since it's a bit clunky imo.

Issue #1413, opened 2022-07-11
[Bookmarks] followup features


kyazdani42:
Since the new marks api has been merged into master, we should now add some integrations with it. This issue is to track related feature requests:
- [x] bulk move (#1419)
- [ ] bulk delete #1804
- [x] navigation (through vim.ui.select) (#1415)
- [x] navigation (prev / next mark) (#1415)
- [ ] command runner (in vim)
- [ ] command runner (system)
- [ ] marks persistence
- [ ] bulk open

RaafatTurki:
Awesome PR, one issue I've found so far was icons not aligning with the correct folder when `hide_root_folder = true`
[preview](https://asciinema.org/a/FNt42JubdBzoYH6opEcbmTcyR)

kyazdani42:
yep @RaafatTurki i've fixed it in the PR add navigation

kyazdani42:
unrelated @RaafatTurki, please open a new issue, i believe this is related to https://github.com/kyazdani42/nvim-tree.lua/pull/1408

tuber84:
explain to a beginner: as a result, bookmarks work or not, in the help in paragraph 8 about bookmarks, the default key "m" sets the bookmark icon against the folder, but after restarting the editor, the bookmarks are not saved. And without saving bookmarks when their program exits, they are useless. Or, most likely, did I not understand something?

alex-courtis:
> And without saving bookmarks when their program exits, they are useless.

Bookmarks are not persisted. The documentation does not state that they are persisted. Persistence is possible but not implemented.

Your opinion on their usefulness is noted, however such comments from you are not welcome in the future.

serhez:
What about bulk open?

martin-braun:
Bulk move and bulk open are actions designed for temporary marks (not bookmarks), while navigating marks and marks persistence are more designed for permanent marks (like bookmarks).

I think it would make sense to have two types of marks. Bookmarks (blue) and temporary marks (red). Bookmarks could be toggled with capital `M` instead of `m`. Bookmarks then should be persisted in the `.local` directory somewhere.

What do you think?

Also, could we add a bullet point for a proper documentation of all the features?

Finally, I'd like to say I love the direction this plugin is heading towards to.

alex-courtis:
See #1795

petobens:
@alex-courtis until someone tackles a bulk delete  PR, do you happen to have some snippet to achieve such functionality (using current API methods)?

[...]

PR #1415, opened 2022-07-11
feat(marks): add navigation (next, previous, select)

kyazdani42:
This feature will edit the file if the node is a file, and focus the folder in the tree if the node is a folder.

Added: 
- next mark
- previous mark
- select mark (vim.ui.select)

implements part of #1413

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1264,4 +1264,16 @@ default.
 To get the list of marked paths, you can call
 `require("nvim-tree.marks").get_marks()`. This will return `{node}`.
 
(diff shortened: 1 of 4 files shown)


alex-courtis (approved):
Tested OK:
- file and folder
- from nvim-tree
- from buffer
- update_focused_file.enabled
- 0, 1, n marks

`Type number and <Enter> or click with the mouse (q or empty cancels):` is odd. I cannot make mouse selection work in a clean setup or my full setup. Not really related.

An option might be to use the location list to do the navigation, however that would probably result in a more (complicated) implementation with little benefit.

PR #1416, opened 2022-07-11
feat(movement): allow circular movement for sibling next and prev

kyazdani42:


alex-courtis (approved):
Tested OK:

- root
- 0, 1, n siblings
- variety of folder structures

Not sure about always circular movement. `nowrapscan` people like me might want an option to disable, however we can add that later if requested.

PR #1424, opened 2022-07-14
feat(mapping): deprecate user mappings and add on_attach

kyazdani42:
See the help doc. I think this is better than providing our custom binding methods
The idea was taken from gitsigns.nvim.
We have a new keymap module that uses the API directly. I've taken the same map from the actions/init.lua file and migrated the action keys to callback keys. We will need to update the script for the default mappings too once this lands.


alex-courtis:
To confirm: we are completely removing actions e.g. `edit_in_place` and the user's mappings will call the API directly.

This is the right way forward. Users invoking `nvim_buf_set_keymap` directly is obvious and clear and will be more future proof.
The mapping / action code is messy and can be difficult / unintuitive for the user. This makes the problem Go Away.

> Its missing a way for users to reuse the existing actions.

This shouldn't be a problem if they are mapping directly to the API.

We will need to provide all the default mappings in the help, so that users can copy-paste to create their own e.g.
```lua
vim.api.nvim_buf_set_keymap(bufnr, 'n', 'hs', '<cmd>lua require"gitsigns".stage_hunk()<CR>', {})
vim.api.nvim_buf_set_keymap(bufnr, 'n', 'Y', '<cmd>lua require"nvim-tree.actions.fs.copy-paste".copy_path()<CR>', {})
```

Users could not use regular vim keymaps, however that is acceptable.

> users can easily migrate to this

We will need to create a page to explain the migration process with examples.

We will unfortunately need to provide a legacy migration path, pointing them to the migration doc. Users will be upset if we don't do this.

> We could start thinking on how to handle a public API for the plugin.

We can do that at any time before this. It's on my list ;)

doc/nvim-tree-lua.txt
```diff
@@ -539,6 +539,32 @@ This will be experimental for a few weeks and will become the default.
     Idle milliseconds between filesystem change and action.
       Type: `number`, Default: `50` (ms)
 
+*nvim-tree.on_attach*
```
alex-courtis, line 542: This should be more obvious, something like `nvim-tree.apply_mappings` or `setup_mappings`.

kyazdani42: I believe on_attach is good enough though, it respects the neovim api ecosystem.
Also this could be used for other things, such as setting buffer options or window options.

kyazdani42:
let's leave it open for now, i'd like to have the API ready before making this change. Also i agree a migration tool would be useful for people since it might be a bit hard to convert from action name to code location.

kyazdani42:
@alex-courtis I can start working on an API module for nvim-tree today, it shouldn't take too long.
Let me know if you are already working on this.

doc/nvim-tree-lua.txt
alex-courtis: > setting buffer options or window options

That works.

alex-courtis:
> @alex-courtis I can start working on an API module for nvim-tree today, it shouldn't take too long. Let me know if you are already working on this.

It is only an idea right now. I had some basic thoughts: https://github.com/kyazdani42/nvim-tree.lua/pull/1412#pullrequestreview-1033695743

We could "namespace" the api for brevity. Example usage:
```lua
local nvt = require "nvim-tree-api"

nvt.marks.get()

nvt.tree.toggle(...)
```

kyazdani42:
@alex-courtis i've migrated the keymap code to a new module. We will do a 2 step deprecation, 1st switching the default to the on_attach (in the next week or two) with a warning message, and in a month or two we'll remove the whole code and options related to this.
For now we can leave it like this, so users that wish to try this can, and we make sure the UX is all right, then we'll do the migration.
We could also write a script that will generate the new configuration for the user. Might not be too complicated.
We might need to take time and notice maintainers of things like lunarvim because they have internal stuff to change and it might take some time for them to implement.

alex-courtis:
This is definitely working:
```lua
        vim.keymap.set("n", "z", inject_node(function(node)
			nt_api.node.open.edit(node)
        end), { buffer = bufnr, noremap = true })
```

We could provide more convenience, something like:
```lua
        -- public api
	function nt_api_add_node_mapping(bufnr, key, fn)
		vim.keymap.set("n", key, function(node)
			fn(node)
		end, { buffer = bufnr, noremap = true })
	end

        -- user in on_attach
	nt_api_add_node_mapping(bufnr, "Z", nt_api.node.open.edit)
```

I'd like to take a look at the conversion script. This will produce new mappings for the old, which we can then put into help. This exercise will clarify the simplest way to do this.

The next step from there is to generate these mappings from `keymap.lua` and add to `update-help.sh`.

alex-courtis:
> We might need to take time and notice maintainers of things like lunarvim because they have internal stuff to change and it might take some time for them to implement.

Yes. We can give them some time; there is no rush to remove the legacy mappings.

alex-courtis:
Pushed a draft mechanism to write current mappings -> on_attach

Feel free to revert / change etc.

Run `:NvimTreeCreateOnAttach`

alex-courtis:
Not sure how we are going to do the DEFAULT_KEYMAPS for update-help.sh

Might have to do something horrible like:
```lua
  {
    key = { "<CR>", "o", "<2-LeftMouse>" },
    callback = { "node", "open", "edit", },
    desc = "open a file or folder; root will cd to the above directory",
  },

...

function M.set_keymaps(bufnr)
  local opts = { noremap = true, silent = true, nowait = true, buffer = bufnr }
  for _, km in ipairs(M.keymaps) do
    local keys = type(km.key) == "table" and km.key or { km.key }
    for _, key in ipairs(keys) do
      local fn = Api
      for _, t in ipairs(km.callback) do
        fn = fn[t]
      end
      vim.keymap.set("n", key, fn, opts)
    end
  end
end

...
```

kyazdani42:
i'm moving your commit in a separate PR so we can discuss it more. I think we can probably have a better strategy for migrating user configs under the hood too.
I think the default keymap is not too problematic, we could simply stringify the lua code for the callback.
I'm merging this since it's non breaking. let's iterate :)

alex-courtis:
> i'm moving your commit in a separate PR so we can discuss it more. I think we can probably have a better strategy for migrating user configs under the hood too. I think the default keymap is not too problematic, we could simply stringify the lua code for the callback. I'm merging this since it's non breaking. let's iterate :)

This was the right desicion. Early feedback is useful. https://github.com/kyazdani42/nvim-tree.lua/issues/1461

PR #1427, opened 2022-07-16
fix(explorer): reload `executable` stat

klesh:
The `NvimTreeRefresh` command didn't refresh `executable` field of the `node`.

## Step to reproduce:
1. Open `nvim-tree` by `:NvimTreeOpen`
2. Change the executable flag of any file by `chmod` command
3. Execute `:NvimTreeRefresh`
4. The style of the file node doesn't change at all while it should be

### DIFF
--- a/lua/nvim-tree/explorer/reload.lua
+++ b/lua/nvim-tree/explorer/reload.lua
@@ -64,7 +64,9 @@ function M.reload(node, status)
         end
       end
       local n = nodes_by_path[abs]
-      n.executable = builders.is_executable(abs, n.extension)
+      if n then
+        n.executable = builders.is_executable(abs, n.extension)
+      end
     end
   end
 
--- a/lua/nvim-tree/explorer/reload.lua
+++ b/lua/nvim-tree/explorer/reload.lua
@@ -62,7 +62,6 @@ function M.reload(node, status)
             table.insert(node.nodes, link)
           end
         end
-
       end
       local n = nodes_by_path[abs]
       n.executable = builders.is_executable(abs, n.extension)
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/explorer/reload.lua
```diff
@@ -62,7 +62,10 @@ function M.reload(node, status)
             table.insert(node.nodes, link)
           end
         end
+
       end
+      local n = nodes_by_path[abs]
+      n.executable = builders.is_executable(abs, n.extension)
```
kyazdani42, line 68: check if n exists first please to make sure that this won't crash.

lua/nvim-tree/explorer/reload.lua
```diff
@@ -62,7 +62,10 @@ function M.reload(node, status)
             table.insert(node.nodes, link)
           end
         end
+
```
kyazdani42, line 65: empty line to remove

kyazdani42 (commented):
Thanks for fixing this, i've noticed it many times and it was bothering me too

alex-courtis (approved):
Tested OK after refreshing. Consistent with other operations e.g. rename.

Will work automatically after #1431

lua/nvim-tree/explorer/reload.lua
klesh: I did. But I removed it before I created PR because I thought line 49 already covered the case.
What do you think?

lua/nvim-tree/explorer/reload.lua
klesh: fixed

lua/nvim-tree/explorer/reload.lua
kyazdani42: indeed but it still might not go into the conditionals, so make the check again :)

klesh: Ok, done

PR #1433, opened 2022-07-17
fix(git): prevent execution of git commands when git integration is disabled

yehy4:
This PR contains one commit to prevent the execution of Git commands even when Git integration is explicitly disabled.

To replicate the issue prior to this commit:
1. Set `git.enable` to `false`.
2. Add `print(cmd)` [here](https://github.com/kyazdani42/nvim-tree.lua/blob/06e48c29c4543331eee41753e69e3bd2235bf430/lua/nvim-tree/git/utils.lua#L8).
3. Expand multiple directories in the tree.
4. Run `:NvimTreeRefresh`.
5. Observe that multiple Git commands are printed/executed (one for each expanded directory).

Explanation: Without this commit, when `get_project_root` is called (for example as a result of executing `:NvimTreeRefresh`), `get_toplevel` (which [unconditionally executes](https://github.com/kyazdani42/nvim-tree.lua/blob/06e48c29c4543331eee41753e69e3bd2235bf430/lua/nvim-tree/git/utils.lua#L6-L7) Git commands) is called.

### DIFF
--- a/lua/nvim-tree/git/init.lua
+++ b/lua/nvim-tree/git/init.lua
@@ -68,11 +68,18 @@ function M.get_project_root(cwd)
     return nil
   end
 
-  local project_root = git_utils.get_toplevel(cwd)
-  return project_root
+  if M.config.git.enable then
+    return git_utils.get_toplevel(cwd)
+  end
+
+  return nil
 end
 
 local function reload_tree_at(project_root)
+  if M.config.git.enable then
+    return nil
+  end
+
   log.line("watcher", "git event executing '%s'", project_root)
   local root_node = utils.get_node_from_path(project_root)
   if not root_node then


alex-courtis (changes requested):
Good catch, thanks for finding.

I've only been observing "no git" with git logging on, which doesn't show `get_toplevel`. I might add a log and profile around it and `should_show_untracked`.

We can be more proactive about the checks. Please perform a "git enabled check" for all the git/utils.lua and git/init.lua functions as per:

```lua
function M.reload()
  if not M.config.git.enable then
    return {}
  end
  ...
```

All callers check for nil/empty table so this should be safe.

alex-courtis:
Added log/profile here: #1435 

Merge the above after this PR.

yehy4:
Done. I added a check to `reload_tree_at` which was the only function in `git/init.lua` that did not already perform the check. 

I did modify the functions in `git/utils.lua` since these functions only get called from `git/init.lua`, so adding the checks there seems unnecessary. If you still think we should add checks in  `git/utils.lua`, please let me know.

lua/nvim-tree/git/init.lua
```diff
@@ -68,11 +68,18 @@ function M.get_project_root(cwd)
...
+    return git_utils.get_toplevel(cwd)
+  end
+
+  return nil
 end
 
 local function reload_tree_at(project_root)
+  if M.config.git.enable then
+    return nil
+  end
```
kyazdani42, line 81: @alex-courtis @yehy4 is this change correct ? It looks like it prevents git to reload when enabled.

yehy4: Sorry, that was a mistake. Thank you for catching it. Fixed in #1443.

PR #1439, opened 2022-07-18
fix: correct 'opt' typo

Crystalix007:
Fixes a typo which incorrectly tries to index `vim.op` instead of `vim.opt`.


kyazdani42:
haven't seen the notification, already pushed to master but thanks :)

PR #1446, opened 2022-07-21
feat(live_filter): filter nodes from loaded buffers

kyazdani42:
Breaking: changing `f` binding to `ff` for the input type filter
Adding `fb` for buffer type filter.

Note: if folders are not shown (`:help nvim-tree.live_filter.always_show_folders`), the filter will not find the nodes since these will be filtered out. 
This option will only work well with find_files enabled and always_show_folders true.

We need to improve the filter algorithm to be able to search recursively. The UI can also be improved for the buffer filtering, something like displaying the buffer number. But this will come in other PRs

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -850,8 +850,9 @@ Configuration for various actions.
 *nvim-tree.live_filter*
 Configurations for the live_filtering feature.
 The live filter allows you to filter the tree nodes dynamically, based on
-regex matching (see |vim.regex|).
-This feature is bound to the `f` key by default.
+regex matching (see |vim.regex|) or loaded buffers.
+This feature is bound to the `ff` key by default for the input filter, and
+bound to `fb` for the buffer filter.
 The filter can be cleared with the `F` key by default.
 
     *nvim-tree.live_filter.prefix*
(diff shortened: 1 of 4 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -1062,7 +1066,8 @@ DEFAULT MAPPINGS                                     *nvim-tree-default-mappings
     { key = "]c",                             action = "next_git_item" }
     { key = "-",                              action = "dir_up" }
     { key = "s",                              action = "system_open" }
-    { key = "f",                              action = "live_filter" }
+    { key = "ff",                             action = "live_filter" }
```
alex-courtis, line 1069: On pressing ff when fb is active, the filter changes text to `/Buffer/` which is not useful.

alex-courtis:
> We need to improve the filter algorithm to be able to search recursively.

I'd like that for regular filter. Out of scope here.

alex-courtis (changes requested):
Functionality is useful, works.

The UX is not great. We do not need to present this functionality to the user as a filter, even if we do it that way internally.

Proposal:
- leave existing `f` and `F` unchanged
  - avoids breaking change
- use `B` to toggle buffer filtering, like `H` `I` etc
  - `B` will not show `[FILTER]: /Buffer/` as this is confusing
- `f` and `B` may be run together, like `f` and `H`

There is no indication that `B` is on, but that is OK as there is no indication that `H` is on. Users appear happy with that.

kyazdani42:
After your comments, i wonder if this feature should even be inside the live-filter :smile: I think we could just implement this filter on the renderer since we already check for opened buffers there.

kyazdani42:
Also this might be a good idea to do like neo-tree and display the buf numbers next to the filenames.
I'm closing in favor of a better implementation.

PR #1450, opened 2022-07-22
refactor(actions): remove linefeed on info messages.

### LINKED ISSUE #1448: Blocking pop-up notifications when creating a folder/file
### Description

I press "a" to create a file/folder, type name, and then press enter.
After that i get a message at the bottom which says "[NimvTree] path/to/file was properly created. Press enter to continue". I have to press enter again to continue using nvim-tree. 

When i delete file/folder i get the same notification but i don't need to press enter to confirm it (this works correctly). 

### Neovim version

```text
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
```

Linux 5.15.55-1-lts

### nvim-tree version

08ab346

### Minimal config

```lua
Behavior is reproduced with a standard config
```

### Steps to reproduce

1. open any file/folder in nvim
2. open nvim-tree
3. create file/directory with "a" hotkey

### Expected behavior

Don't wait for enter after notification pop-up

### Actual behavior

Have to press enter to confirm that i read notification

cevhyruz:
This will fix blocking messages in simple fs operations. 

will fix #1448

### DIFF
--- a/lua/nvim-tree/actions/fs/copy-paste.lua
+++ b/lua/nvim-tree/actions/fs/copy-paste.lua
@@ -120,11 +120,11 @@ local function add_to_clipboard(node, clip)
   for idx, _node in ipairs(clip) do
     if _node.absolute_path == node.absolute_path then
       table.remove(clip, idx)
-      return utils.notify.info(node.absolute_path .. " removed to clipboard.\n")
+      return utils.notify.info(node.absolute_path .. " removed to clipboard.")
     end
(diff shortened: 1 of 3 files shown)


kyazdani42 (approved):

kyazdani42:
Thanks for the fix, the newline was used before because it was needed when doing vim.api.nvim_out_write, vim.notify must handle the formatting properly :)

PR #1453, opened 2022-07-24
chore(watchers): Watcher shares single fs_event from Event, node watchers use unique path prefixed debounce context

alex-courtis:
Split Watcher into Watcher and Event.

Event owns the fs_event for an absolute path. `Watcher:new` always creates a new Watcher which may share an existing Event.

Node watchers use a unique context: absolute path + incremennting index

Example:
```
real
real/file
link1 -> real
sub/link2 -> real
```

- One Event created for `/path/to/real`. Two node Watchers created: `/path/to/real:1`, `/path/to/real:2`.
- Modify `real/file`: one event fires, two watchers fire.
- Open sub: node watcher created `/path/to/real:3`
- Modify `real/file`: one Event fires, three watchers fire.
- Remove `link1`: node Watcher destroyed: `/path/to/real:2`
- Modify `real/file`: one Event fires, two watchers fire.

All the refresh/draw events will now correctly occur for each affected link/directory, along with the associated git event. This extra git event could be optimised away, however the added complexity is not worth it for the rare cases in which this will actually occur.

### DIFF
--- a/lua/nvim-tree/explorer/watch.lua
+++ b/lua/nvim-tree/explorer/watch.lua
@@ -45,21 +45,23 @@ function M.create_watcher(absolute_path)
     return nil
   end
 
-  log.line("watcher", "node start '%s'", absolute_path)
-  return Watcher.new {
-    absolute_path = absolute_path,
-    on_event = function(opts)
-      log.line("watcher", "node event scheduled '%s'", opts.absolute_path)
-      utils.debounce("explorer:watch:" .. opts.absolute_path, M.debounce_delay, function()
-        refresh_path(opts.absolute_path)
-      end)
-    end,
-  }
+  local function callback(watcher)
+    log.line("watcher", "node event scheduled %s", watcher.context)
+    utils.debounce(watcher.context, M.debounce_delay, function()
+      refresh_path(watcher._path)
+    end)
+  end
+
+  M.uid = M.uid + 1
+  return Watcher:new(absolute_path, callback, {
+    context = "explorer:watch:" .. absolute_path .. ":" .. M.uid,
+  })
 end
 
 function M.setup(opts)
   M.enabled = opts.filesystem_watchers.enable
   M.debounce_delay = opts.filesystem_watchers.debounce_delay
+  M.uid = 0
 end
 
 return M
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -169,15 +169,17 @@ end
 ---Matching executable files in Windows.
 ---@param ext string
 ---@return boolean
-local PATHEXT = vim.env.PATHEXT or ""
-local wexe = vim.split(PATHEXT:gsub("%.", ""), ";")
-local pathexts = {}
-for _, v in pairs(wexe) do
-  pathexts[v] = true
-end
-
 function M.is_windows_exe(ext)
-  return pathexts[ext:upper()]
+  if not M.pathexts then
+    local PATHEXT = vim.env.PATHEXT or ""
+    local wexe = vim.split(PATHEXT:gsub("%.", ""), ";")
+    M.pathexts = {}
+    for _, v in pairs(wexe) do
+      M.pathexts[v] = true
+    end
+  end
+
+  return M.pathexts[ext:upper()]
 end
 
 function M.rename_loaded_buffers(old_path, new_path)
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -379,4 +379,23 @@ function M.clear_prompt()
   end
 end
 
+-- return a new table with values from array
+function M.array_shallow_clone(array)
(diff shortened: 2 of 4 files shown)


lua/nvim-tree/utils.lua
```diff
@@ -169,15 +169,17 @@ end
...
+  if not M.pathexts then
+    local PATHEXT = vim.env.PATHEXT or ""
+    local wexe = vim.split(PATHEXT:gsub("%.", ""), ";")
+    M.pathexts = {}
+    for _, v in pairs(wexe) do
+      M.pathexts[v] = true
+    end
+  end
+
+  return M.pathexts[ext:upper()]
```
alex-courtis, line 182: Not related to the change, however this hurt my eyes and I could not help tidying it.

lua/nvim-tree/explorer/watch.lua
```diff
@@ -45,21 +45,23 @@ function M.create_watcher(absolute_path)
...
+  local function callback(watcher)
+    log.line("watcher", "node event scheduled %s", watcher.context)
+    utils.debounce(watcher.context, M.debounce_delay, function()
+      refresh_path(watcher._path)
+    end)
+  end
+
+  M.uid = M.uid + 1
+  return Watcher:new(absolute_path, callback, {
+    context = "explorer:watch:" .. absolute_path .. ":" .. M.uid,
```
alex-courtis, line 57: Unique debouncer context added so that the real/link jobs do not clobber each other.

lua/nvim-tree/watcher.lua
```diff
@@ -16,87 +22,129 @@ local FS_EVENT_FLAGS = {
...
 function M.purge_watchers()
-  for _, watcher in pairs(M._watchers) do
-    watcher:destroy()
+  log.line("watcher", "purge_watchers")
+
+  for _, w in ipairs(utils.array_shallow_clone(Watcher._watchers)) do
+    w:destroy()
+  end
+
+  for _, e in pairs(Event._events) do
```
alex-courtis, line 145: This should not be necessary however it does not hurt.

lua/nvim-tree/watcher.lua
```diff
@@ -3,10 +3,16 @@ local uv = vim.loop
 local log = require "nvim-tree.log"
 local utils = require "nvim-tree.utils"
 
-local M = {
+local M = {}
+
+local Event = {
+  _events = {},
```
alex-courtis, line 9: Static class member.

lua/nvim-tree/watcher.lua
```diff
@@ -16,87 +22,129 @@ local FS_EVENT_FLAGS = {
...
+
+  w:start()
+
+  table.insert(Watcher._watchers, w)
+
+  return w
+end
+
+function Watcher:start()
+  self._listener = function()
```
alex-courtis, line 121: I wanted to call back to a member function like `Watcher:on_event` however it doesn't seem possible.

lua/nvim-tree/watcher.lua
```diff
@@ -16,87 +22,129 @@ local FS_EVENT_FLAGS = {
...
-  for i, w in ipairs(M._watchers) do
-    if w == self then
-      table.remove(M._watchers, i)
-      break
-    end
+
+  Event._events[self._path] = nil
+end
+
+function Watcher:new(path, callback, data)
```
alex-courtis, line 99: _opts flattened and parameters added for clarity

lua/nvim-tree/watcher.lua
```diff
@@ -16,87 +22,129 @@ local FS_EVENT_FLAGS = {
...
 end
 
 M.Watcher = Watcher
 
 function M.purge_watchers()
-  for _, watcher in pairs(M._watchers) do
-    watcher:destroy()
+  log.line("watcher", "purge_watchers")
+
+  for _, w in ipairs(utils.array_shallow_clone(Watcher._watchers)) do
```
kyazdani42, line 141: not sure i understand why a shallow clone is needed here ?

kyazdani42 (approved):

kyazdani42:
LGTM, i wonder if we should move the debouncing inside the watcher event listener to avoid repeating the logic everytime we want to introduce a new watcher. Maybe for a followup PR. Looks good to me, let's merge this.

lua/nvim-tree/watcher.lua
alex-courtis: `Watcher:destroy` will remove self from `Watcher._watchers`. That confuses the iterator.

alex-courtis:
> LGTM, i wonder if we should move the debouncing inside the watcher event listener to avoid repeating the logic everytime we want to introduce a new watcher.

That makes sense: `Watcher:new` takes a debounce timeout argument and applies it to `_listener`.

Debounce would still be used "standalone" by diagnostics and anything else.

Issue #1368, opened 2022-06-21
quit as last buffer 


YanzhaoW:
### Description

Without any configuration, neovim simply doesn't quit as nvim-tree is the last buffer. Many people suggest, as is also mentioned in the README.md (#1115 ), that I should add following command:
```
vim.api.nvim_create_autocmd("BufEnter", {
  nested = true,
  callback = function()
    if #vim.api.nvim_list_wins() == 1 and vim.api.nvim_buf_get_name(0):match("NvimTree_") ~= nil then
      vim.cmd "quit"
    end
  end
})
```
This is very bad solution because it can cause neovim to crash if you accidentally quit without saving. Following errors will be repeated if you press any key in neovim:
```
Press ENTER or type command to continue
Error executing vim.schedule lua callback: ...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:378: Vim(append):Error executing lua callback: /home/ywang/.config/nvim/lua/plugs/n
vimtree_config.lua:34: Vim(quit):E37: No write since last change
stack traceback:
        [C]: in function 'cmd'
        /home/ywang/.config/nvim/lua/plugs/nvimtree_config.lua:34: in function </home/ywang/.config/nvim/lua/plugs/nvimtree_config.lua:32>
        [C]: in function 'nvim_win_close'
        ...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:378: in function <...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:356>
stack traceback:
        [C]: in function 'nvim_win_close'
        ...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:378: in function <...e/pack/packer/start/nvim-tree.lua/lua/nvim-tree/view.lua:356>
Press ENTER or type command to continue
```
I would suggest remove this tip in the README.md.

Is there any other doable solution to quit neovim when the nvim-tree is the last buffer?

### Neovim version

```text
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
```

### Minimal config

```lua
require'nvim-tree'.setup{
    view = {
        side = 'right'
    },
    update_focused_file = {
        enable = true,
        update_cwd = true,
        ignore_list = {}
    }
}
-- vim.cmd([[autocmd BufEnter * ++nested if winnr('$') == 1 && bufname() == 'NvimTree_' . tabpagenr() | quit | endif]])
[...]

PR #1466, opened 2022-07-26
Reliably determine if nvim-tree is last window and exit gracefully

beauwilliams:
This is a PR aims to solve #1368 by parsing the window layout of the current window to reliably determine whether NvimTree if the last window and if so quits.

Also solving an apparent issue highlighted in snippet provided by readme this PR adds `confirm quit` so that we can also get a graceful exit in the case of unsaved buffers in the background.

I was sorely missing this feature when it was deprecated so I am hoping that this is good enough for it to be added back in :)

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -420,7 +420,7 @@ local function setup_autocommands(opts)
 	group = vim.api.nvim_create_augroup("NvimTreeClose", {clear = true}),
 	callback = function()
 		local layout = vim.api.nvim_call_function("winlayout", {})
-		if layout[1] == "leaf" and vim.api.nvim_buf_get_option(vim.api.nvim_win_get_buf(layout[2]), "filetype") == "NvimTree" and layout[3] == nil then vim.cmd("quit") end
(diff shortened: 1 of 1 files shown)


kyazdani42:
To be honest, we removed this because there were too many people asking for different implementations of this feature. We spent a lot of time answering these issues.
We don't want to maintain this, that's why we left it to the users. I believe a wiki page + a note in the readme would be good enough.

beauwilliams:
I totally understand, there were quite a few different ones flying around. 

This solution should be quite a bit more robust in its approach compared to what was floating around, so at least the community can use it as they please

kyazdani42:
what i mean is that we won't merge that again, because even if it works better for some people, some other people will want different implementations.
You can write a new wiki page once you are done with the implementation :)

beauwilliams:
No problem :)

alex-courtis:
Wiki / readme: yes

nvim-tree functionality: no

This is hugely problematic due to:
* unpredictable vim event ordering, especially factoring in nested autocommands
* vim plugins / automation modifying buffers and windows during `BufEnter`

PR #1473, opened 2022-07-28
Fixed broken remove_keymaps option.

adam-bates:
New here, just trying to set up nvim-tree. `remove_keymaps` option wasn't working for me and it looks like this is why.

Feel free to change as you see fit.


adam-bates:
Closing this as there's a good chance its wrong the more I look at the code. I can see now I was mixing the old style with the new style of mappings which was throwing me off.

alex-courtis:
You were right: https://github.com/kyazdani42/nvim-tree.lua/commit/9bbf95e61691feb6696880701cdb1b6cafa97d41

PR #1476, opened 2022-07-31
chore(mappings): migrate legacy mappings under the hood

kyazdani42:
NOTE: this should not be merged before we tested it thoroughly.
NOTE2: this should not be merged before the help menu is rewritten
NOTE3: this can be used to print the on_attach migration to the user.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -666,8 +666,9 @@ function M.setup(conf)
   validate_options(conf)
 
   local opts = merge_options(conf)
-  local netrw_disabled = opts.disable_netrw or opts.hijack_netrw
+  legacy.move_mappings_to_keymap(opts)
 
+  local netrw_disabled = opts.disable_netrw or opts.hijack_netrw
   _config.root_dirs = opts.root_dirs
   _config.prefer_startup_root = opts.prefer_startup_root
   _config.update_focused_file = opts.update_focused_file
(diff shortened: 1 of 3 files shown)


lua/nvim-tree.lua
```diff
@@ -666,8 +666,9 @@ function M.setup(conf)
   validate_options(conf)
 
   local opts = merge_options(conf)
-  local netrw_disabled = opts.disable_netrw or opts.hijack_netrw
+  legacy.move_mappings_to_keymap(opts)
```
alex-courtis, line 669: I like this approach: moving the mappings after merge means we only have to do this once.

lua/nvim-tree/legacy.lua
```diff
@@ -297,6 +298,99 @@ local function removed(opts)
   end
 end
 
+local OLD_MAPPING_TABLE = {
```
alex-courtis, line 301: We could be lazy and move this to `DEFAULT_KEYMAPS` with a `legacy_action` field.

alex-courtis (commented):
> NOTE2: this should not be merged before the help menu is rewritten

We could be lazy: retain the view.mappings.list for help to look up, if we want to merge this before.

> NOTE3: this can be used to print the on_attach migration to the user.

Yes. `vim.keymap.set` is ready to print.

lua/nvim-tree/legacy.lua
kyazdani42: Yes, also i've noticed from some PRs on nvim-treesitter that vim.keymap.set takes an optional "desc" argument which we should use so the nvim-tree will work in which-key and allow us to use that in the help menu

alex-courtis:
See #1579

PR #1478, opened 2022-08-01
chore(mappings): POC for help and :help on_attach keymaps

alex-courtis:
1. generate help view by mapping keymap functions back to `DEFAULT_KEYMAPS`
2. generate :help via update script
a. user readable table `/tmp/DEFAULT_KEYMAPS.help`
b. on_attach function `/tmp/DEFAULT_KEYMAPS.on_attach.lua`

I did mess around with `callback` as a string to loadstring. It works, however I lost the ability to map back, as the function from loadstring is a new wrapper function.

We don't need the complication of loadstring in code, so text processing in the update script is enough. If necessary, we can apply the same processing to generate lua to perform the legacy conversion.

### DIFF
--- a/lua/nvim-tree/renderer/help.lua
+++ b/lua/nvim-tree/renderer/help.lua
@@ -1,83 +1,96 @@
-local log = require "nvim-tree.log"
-
 local M = {}
 
-local function shorten_lhs(lhs)
-  lhs = lhs:gsub("LeftMouse", "LM")
-  lhs = lhs:gsub("RightMouse", "RM")
-  lhs = lhs:gsub("MiddleMouse", "MM")
-  lhs = lhs:gsub("ScrollWheelDown", "SWD")
-  lhs = lhs:gsub("ScrollWheelUp", "SWU")
-  lhs = lhs:gsub("ScrollWheelLeft", "SWL")
-  lhs = lhs:gsub("ScrollWheelRight", "SWR")
+local function tidy_lhs(lhs)
+  -- nvim_buf_get_keymap replaces leading "<" with "<lt>" e.g. "<lt>CTRL-v>"
+  lhs = lhs:gsub("^<lt>", "<")
+
+  -- shorten ctrls
+  if lhs:lower():match "^<ctrl%-" then
+    lhs = lhs:lower():gsub("^<ctrl%-", "<C%-")
+  end
+
+  -- uppercase ctrls
+  if lhs:match "^<C%-" then
+    lhs = lhs:upper()
+  end
+
   return lhs
+end
+
+local function shorten_lhs(lhs)
+  return lhs
+    :gsub("LeftMouse>$", "LM>")
+    :gsub("RightMouse>$", "RM>")
+    :gsub("MiddleMouse>$", "MM>")
+    :gsub("ScrollWheelDown>$", "SD>")
+    :gsub("ScrollWheelUp>$", "SU>")
+    :gsub("ScrollWheelLeft>$", "SL>")
+    :gsub("ScrollWheelRight>$", "SR>")
+end
+
+-- sort lhs roughly as per :help index
+local PAT_MOUSE = "^<.*Mouse"
+local PAT_CTRL = "^<C\\-"
+local PAT_SPECIAL = "^<.+"
+local function sort_lhs(a, b)
+  -- mouse last
+  if a:match(PAT_MOUSE) and not b:match(PAT_MOUSE) then
+    return false
+  elseif not a:match(PAT_MOUSE) and b:match(PAT_MOUSE) then
+    return true
+  end
+
+  -- ctrl first
+  if a:match(PAT_CTRL) and not b:match(PAT_CTRL) then
+    return true
+  elseif not a:match(PAT_CTRL) and b:match(PAT_CTRL) then
+    return false
+  end
+
+  -- special next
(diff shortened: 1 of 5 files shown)


lua/nvim-tree/renderer/help.lua
```diff
@@ -1,27 +1,62 @@
...
-    local name
-    if cb and cb:sub(1, 35) == require("nvim-tree.config").nvim_tree_callback("test"):sub(1, 35) then
-      name = cb:match "'[^']+'[^']*$"
-      name = name:match "'[^']+'"
-    elseif b.action then
-      name = b.action
+  for _, bkm in ipairs(buf_keymaps) do
+    local default_keymap = nil
+    for _, dkm in ipairs(require("nvim-tree.keymap").DEFAULT_KEYMAPS) do
+      if bkm.callback == dkm.callback then
```
alex-courtis, line 28: Very happy with this. It successfully maps user's on_attach mappings as well as defaults.

`inject_node` was not a problem as it returns an anonymous function that mapped well.

lua/nvim-tree/renderer/help.lua
```diff
@@ -1,27 +1,62 @@
...
+  --   elseif b.action then
+  --     name = b.action
+  --   else
+  --     name = (b.name ~= nil) and b.name or cb
+  --     name = '"' .. name .. '"'
+  --   end
+  --   table.insert(processed, { key, name, true })
+  -- end
+
   table.sort(processed, function(a, b)
```
alex-courtis, line 60: This sort order looks a bit strange now; maybe we switch to ordering by key.

kyazdani42 (commented):
I think we should probably adapt this to attach short descriptions on the keymap directly so we don't need to do some custom logic :)
see `:help vim.api.nvim_set_keymap`, these applies to vim.keymap.set as well.
This way we could just set the same description from the old mapping table, and the new mapping table, so we can keep retro compatibility while moving to the new ones.

lua/nvim-tree/renderer/help.lua
kyazdani42: indeed, simplify this please !

scripts/generate_default_keymaps.lua
```diff
@@ -0,0 +1,54 @@
...
+    io.write(string.format(fmt, m.key, m.short, m.long))
+  end
+end
+io.write "\n"
+io.close(file)
+
+-- lua on_attach
+file = io.open("/tmp/DEFAULT_KEYMAPS.on_attach.lua", "w")
+io.output(file)
+io.write "local function on_attach(bufnr, mode, opts)\n"
```
kyazdani42, line 45: why are mode and opts passed to on_attach ? They should be unique for each mapping i believe.

alex-courtis:
> I think we should probably adapt this to attach short descriptions on the keymap directly so we don't need to do some custom logic :) see `:help vim.api.nvim_set_keymap`, these applies to vim.keymap.set as well. This way we could just set the same description from the old mapping table, and the new mapping table, so we can keep retro compatibility while moving to the new ones.

Oh wow... that makes things easy.

scripts/generate_default_keymaps.lua
alex-courtis: I'm not sure that the user should care about mode and opts... we should pass in sensible defaults and they can change if they want to.

RE mode: I don't think anything but "n" makes sense.

lua/nvim-tree/renderer/help.lua
alex-courtis: It's now ordering something like :help index

scripts/generate_default_keymaps.lua
alex-courtis: Updated to include desc, removed mode and opts.

alex-courtis:
> This way we could just set the same description from the old mapping table, and the new mapping table, so we can keep retro compatibility while moving to the new ones.

Working nicely. desc is set to old mapping action and new keymap desc.short.

scripts/generate_default_keymaps.lua
kyazdani42: yes 'n' by default, but the old bindings could have mode associated with them (if the user wishes to do stuff with the visual selection).

kyazdani42 (approved):
this is much better. Since i won't be too much active until end of august, i'll leave you deciding if you want to merge :)

lua/nvim-tree/renderer/help.lua
```diff
@@ -1,48 +1,96 @@
...
+local function shorten_lhs(lhs)
+  return lhs
+    :gsub("LeftMouse>$", "LM>")
+    :gsub("RightMouse>$", "RM>")
+    :gsub("MiddleMouse>$", "MM>")
+    :gsub("ScrollWheelDown>$", "SD>")
+    :gsub("ScrollWheelUp>$", "SU>")
+    :gsub("ScrollWheelLeft>$", "SL>")
+    :gsub("ScrollWheelRight>$", "SR>")
+end
```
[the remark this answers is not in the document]
alex-courtis: I can revert. Now that mouse is at the end, the indenting won't look so strange.

alex-courtis:
> this is much better. Since i won't be too much active until end of august, i'll leave you deciding if you want to merge :)

I will not merge, it's not urgent. I might collect all our work in a single branch that we can review as one when you return.

alex-courtis:
See #1579

PR #1481, opened 2022-08-02
doc: reinstate mapping doc, mark on_attach as experimental

alex-courtis:
We don't have a complete story for on_attach yet, so restore the documentation for old style mappings.

As a new user, I would find it difficult to set up mappings: I don't have any examples or indication of how to create on_attach mappings for non-custom functions.


kyazdani42 (approved):
agreed :)

PR #1482, opened 2022-08-02
fix(#1480): break symlink cycle on find-file, search-node

### LINKED ISSUE #1480: nvim gets stuck when finding a file in a directory with a looping symlink
### Description

When navigating up to a large directory (e.g. from `C:\large-directory\subdirectory` to `C:\large-directory`) that contains a looping symlink (e.g. `C:\large-directory\loop -> C:\large-directory`), nvim will try to find the previously selected file (e.g. `C:\large-directory\subdirectory\neat.file`).
But that looping symlink will make nvim loop through the directory again and again, causing it to lock up while its RAM usage gradually grows larger and larger (>1GB).

### Neovim version

```text
NVIM v0.8.0-dev-739-ge59bc078d
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compiled by runneradmin@fv-az158-506

Features: -acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM\sysinit.vim"
  fall-back for $VIM: "C:/Program Files (x86)/nvim/share/nvim"

```

Microsoft Windows [Version 10.0.19043.1826]

### nvim-tree version

451901c

### Minimal config

```lua
-- Yes, this bug occurs with this minimal config as it is
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

1. `mkdir /tmp/crasher && cd /tmp/crasher`
2. `mkdir ./a_{0..500}`
3. `ln -s ./ ./loop`
4. `mkdir ./z && touch ./z/cool.file`
5. Open nvim and toggle nvim-tree
6. Navigate into the `z` folder with `Ctrl+]`
7. Navigate back up using `-`

I also followed the suggestion in https://github.com/kyazdani42/nvim-tree.lua/issues/549#issuecomment-1200552247 and what happens in the logs is that after `START find file C:\Users\foo\AppData\Local\Everything\Everything.db`, `[watcher] Watcher:new` events are spawned on throughout all folders in `C:\Users\foo\AppData\Local` over and over again, looping back and forth hundreds of times. The log file weighs tens of megabytes when I crashed nvim.

### Expected behavior

nvim doesn't lock up

### Actual behavior

nvim locks up, while its RAM usage gradually grows larger and larger (>1GB)

alex-courtis:
closes #1480 

* detect cycles on search-node `S`
* search link paths, not target paths on find-file

expand-all is OK as it stops after `max_folder_discovery`, however it does follow cycles. We should look at that if time permits.

### DIFF
--- a/lua/nvim-tree/actions/finders/find-file.lua
+++ b/lua/nvim-tree/actions/finders/find-file.lua
@@ -33,10 +33,7 @@ function M.fn(fname)
     end)
     :applier(function(node)
       line = line + 1
-      local abs_match = vim.startswith(fname_real, node.absolute_path .. utils.path_separator)
-      local link_match = node.link_to and vim.startswith(fname_real, node.link_to .. utils.path_separator)
-
-      if abs_match or link_match then
+      if vim.startswith(fname_real, node.absolute_path .. utils.path_separator) then
         node.open = true
         if #node.nodes == 0 then
           core.get_explorer():expand(node)
--- a/lua/nvim-tree/actions/finders/find-file.lua
+++ b/lua/nvim-tree/actions/finders/find-file.lua
@@ -27,13 +27,24 @@ function M.fn(fname)
 
   local line = core.get_nodes_starting_line()
 
+  local absolute_paths_searched = {}
+
   local found = Iterator.builder(core.get_explorer().nodes)
     :matcher(function(node)
       return node.absolute_path == fname_real or node.link_to == fname_real
     end)
     :applier(function(node)
       line = line + 1
-      if vim.startswith(fname_real, node.absolute_path .. utils.path_separator) then
+
+      if vim.tbl_contains(absolute_paths_searched, node.absolute_path) then
+        return
+      end
+      table.insert(absolute_paths_searched, node.absolute_path)
+
+      local abs_match = vim.startswith(fname_real, node.absolute_path .. utils.path_separator)
+      local link_match = node.link_to and vim.startswith(fname_real, node.link_to .. utils.path_separator)
+
+      if abs_match or link_match then
         node.open = true
         if #node.nodes == 0 then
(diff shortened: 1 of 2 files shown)


alex-courtis:
@archiif your testing would be gratefully appreciated.

kyazdani42 (commented):
does this change means that nvim-tree will not find file in symlink directories ? I think we should only prevent looping indefinitely right ?

alex-courtis:
> does this change means that nvim-tree will not find file in symlink directories ? I think we should only prevent looping indefinitely right ?

I recall that it did. I'll re-test on the weekend.

archiif:
Thanks a lot for this PR! I did a quick test and it fixed my issue.
nvim-tree doesn't seem to track the current directory, so when you `C-]` into a symlinked directory and go back up again, you end up in the new target directory. So I'm not sure how to test what @kyazdani42 is saying.

alex-courtis:
> nvim-tree doesn't seem to track the current directory, so when you `C-]` into a symlinked directory and go back up again, you end up in the new target directory.

That's working as intended: nvim-tree simply finds the file, and find always picks the real directory. That could be changed in the future, but would need some thought.

alex-courtis:
Still testing...

alex-courtis:
> does this change means that nvim-tree will not find file in symlink directories ? I think we should only prevent looping indefinitely right ?

It is not finding files in symlink directories outside cwd. More work needed.

PR #1492, opened 2022-08-09
docs: Bring in line with standard documentation

anuvyklack:


gegoune:
Amazing!

doc/nvim-tree-lua.txt
```diff
@@ -1425,10 +1653,10 @@ Navigation for marks is not bound by default in nvim-tree because we don't
...
-vim.keymap.set("n", "<leader>ms", require("nvim-tree.api").marks.navigate.select)
-
- vim:tw=78:ts=4:sw=4:et:ft=help:norl:
+>
+ vim.keymap.set("n", "<leader>mn", require("nvim-tree.api").marks.navigate.next)
+ vim.keymap.set("n", "<leader>mp", require("nvim-tree.api").marks.navigate.prev)
+ vim.keymap.set("n", "<leader>ms", require("nvim-tree.api").marks.navigate.select)
+<
+------------------------------------------------------------------------------
+ vim:ft=help:et:tw=78:ts=4:sw=4:norl:
```
alex-courtis, line 1662: This appears unchanged; is there anything we can add to assist with standardisation?

alex-courtis (changes requested):
Looks good. Where are the standards documented? I'm never sure what to follow; even vim's help is inconsistent.

- [ ] Default mappings and options are updated via a script, see [CONTRIBUTING.md](https://github.com/kyazdani42/nvim-tree.lua/blob/master/CONTRIBUTING.md). Scripts will need to be updated.

doc/nvim-tree-lua.txt
anuvyklack: I don't get, what you mean. I place `ft=help` first because I found out that all options before `ft` don't recognize, and their default values are used.

anuvyklack:
By standard documentation here, I mean the style in which the Vim options descriptions are formatted.

I decided to rewrite the documentation when I installed the nvim-tree myself and found its documentations unusable. All default values are in very huge table, which is marked up as "code" and all options descriptions underneath and options names are vim-help links without destination. So I made the default values also be the table of content, to allow user immediately jump to particular option description while scanning with eyes the table with defaults. The default value and Vim help tag separated with Lua comment sign: `--`, sou you can copy any line into your Neovim config file, and it doesn't cause an error.

To create the script to make such formatting with sed is a lot of work that not worth it. And the result of the script will never be like manually tailored. I have already done most of the work, so to maintain the existing docs won't be much work, since I doubt, that you will frequently change all options names and descriptions.

doc/nvim-tree-lua.txt
alex-courtis: > To create the script to make such formatting with sed is a lot of work that not worth it. And the result of the script will never be like manually tailored. I have already done most of the work, so to maintain the existing docs won't be much work, since I doubt, that you will frequently change all options names and descriptions.

The script is already present. It updates `nvim-tree-setup` and `nvim-tree-default-mappings`. Fixing this script is not optional.

alex-courtis:
> To create the script to make such formatting with sed is a lot of work that not worth it. And the result of the script will never be like manually tailored. I have already done most of the work, so to maintain the existing docs won't be much work, since I doubt, that you will frequently change all options names and descriptions.

The script is already present. It updates `nvim-tree-setup` and `nvim-tree-default-mappings`. Fixing this script is not optional.

doc/nvim-tree-lua.txt
alex-courtis: > I found out that all options before `ft` don't recognize, and their default values are used.

Thank you.

RE additional options: any other formatting options such as `copyindent`

alex-courtis:
@anuvyklack have you made any progress? The script should be simple to fix; it just requires updating the markers in the help file that the script uses.

alex-courtis:
@gegoune @kyazdani42 how do you feel about this one? The author will not fix `update-help.sh` however it should be simple for us.

I am happy to reject this one; the current doc is fine and these "standards" are arbitrary.

gegoune:
I would be happy to make it its way in. Maybe someone else will pick it up? I could see if I could amend the script, maybe with some guidance?

alex-courtis:
> I would be happy to make it its way in. Maybe someone else will pick it up? I could see if I could amend the script, maybe with some guidance?

It should be simple; `update-help.sh` expects some markers which were deleted.
Default mappings text:
https://github.com/kyazdani42/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L1202
https://github.com/kyazdani42/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L1254
Default mappings lua:
https://github.com/kyazdani42/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L1255
https://github.com/kyazdani42/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L1255
Default options lua:
https://github.com/kyazdani42/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L167
https://github.com/kyazdani42/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L373 

The author did not make any attempt to understand and instead chose to argue about it.

alex-courtis:
RE links after options `auto_reload_on_write = true, -- |nvim-tree.auto_reload_on_write|` they should go in nvim-tree.lua to be scraped.

Once again, author made no attempt to understand.

alex-courtis:
These changes do require some attention e.g. the removal of the links such as `:NvimTreeRefresh`.

I did not fully review these changes as I found the author to be very difficult to communicate with.

gegoune:
@anuvyklack Would you be willing to continue your work on this PR given @alex-courtis' effort describing remaining required changes?

If not, I should perhaps have some free time to look into it in a week or two.

PR #1097, opened 2022-03-21
feat: Use vim.ui.input for rename and create

baahrens:

PR #1499, opened 2022-08-11
Use select instead of input for overwriting file

### LINKED ISSUE #1483: Use `vim.ui.select` for selecting an option instead of `vim.ui.input`

https://github.com/kyazdani42/nvim-tree.lua/commit/8dc2144e878921111c75a942af6abed685e15c01 switched from `vim.ui.select` to `vim.ui.input`, which causes a problem for me: when the filename is too long, the title is truncated and I couldn't see the choices (y/n/r).

I think using `vim.ui.select` is a better fit since in this case we're selecting an option rather than inputting something. If user doesn't happy with the UI/UX, they can override `vim.ui.select` instead.

**Additional context**
image

In this image, I'm copying and pasting a file into the same folder, `nvim-tree` prompts for overriding or renaming the file, but the title is truncated and I couldn't see the possible choices/what to enter.

mosheavni:
Since the input is for 1 of 3 options (y = yes, n = no, r = rename), it only makes sense to use a select instead of input, which also fixes the truncated input box when using https://github.com/stevearc/dressing.nvim
Closes #1483

### DIFF
--- a/lua/nvim-tree/actions/fs/copy-paste.lua
+++ b/lua/nvim-tree/actions/fs/copy-paste.lua
@@ -94,11 +94,11 @@ local function do_single_paste(source, dest, action_type, action_fn)
   end
 
   if dest_stats then
-    vim.ui.input({ prompt = dest .. " already exists. Overwrite? y/n/r(ename): " }, function(choice)
+    vim.ui.select({ "Yes", "No", "Rename" }, { prompt = dest .. " already exists. Overwrite? y/n/r(ename): " }, function(choice)
       utils.clear_prompt()
-      if choice == "y" then
+      if choice == "Yes" then
         on_process()
-      elseif choice == "r" then
+      elseif choice == "Rename" then
         vim.ui.input({ prompt = "New name: ", default = dest, completion = "dir" }, function(new_dest)
           utils.clear_prompt()
           if new_dest then


lua/nvim-tree/actions/fs/copy-paste.lua
```diff
@@ -94,11 +94,11 @@ local function do_single_paste(source, dest, action_type, action_fn)
   end
 
   if dest_stats then
-    vim.ui.input({ prompt = dest .. " already exists. Overwrite? y/n/r(ename): " }, function(choice)
+    vim.ui.select({ "Yes", "No", "Rename" }, { prompt = dest .. " already exists. Overwrite? y/n/r(ename): " }, function(choice)
```
alex-courtis, line 97: This works when using dessing.

However, the user experience for vanilla vim is unacceptable:
```
/tmp/1499/aaa already exists. Overwrite? y/n/r(ename):
1: Yes
2: No
3: Rename
Type number and <Enter> or click with the mouse (q or empty cancels):
```

alex-courtis (changes requested):
We have tried and then reverted something similar: #1097

A solution that provides a good user experience for vanilla vim (ex command) and ui plugins is necessary.

lua/nvim-tree/actions/fs/copy-paste.lua
mosheavni: so without dressing, the `input` is a disaster.
Both options are bad for some and good for others. can we let users decide?

alex-courtis: That is a reasonable idea. A single interaction that covers both cases does not seem feasible.

Add an option something like `select_popup` and provide alternate vim.ui.input/select for each of the invocations.

Before we add an option, is there a mechanism by which we can detect whether the user is using dressing of some sort? I believe there are several out there...

mosheavni: I believe something like this could work:

```
local hasdressing, _ = pcall(require,"dressing")
if hasdressing then
  ...
end
```

alex-courtis: > pcall(require,"dressing")

This will work for dressing, however this is vim and so there will be many other vim.ui implementations. Adding a condition for each of these is not feasible.

Some digging has indicated that dressing and others function by overriding `vim.ui.select` and it does not appear possible to determine whether this override is in effect.

Therefore two implementations with an option is the solution.

alex-courtis (changes requested):
- [ ] add option `vim_ui_popup = false`
- [ ] implement two copy-paste selects, removing `Overwrite? y/n/r(ename)` from the popup version
- [ ] test all nvim-tree `vim.ui.select` and `vim.ui.input` with dressing, adding alternative if required
- [ ] address stylua violations

mosheavni:
> * [ ]  add option `vim_ui_popup = false`
> * [ ]  implement two copy-paste selects, removing `Overwrite? y/n/r(ename)` from the popup version
> * [ ]  test all nvim-tree `vim.ui.select` and `vim.ui.input` with dressing, adding alternative if required
> * [ ]  address stylua violations

Let me know if I can help, I don't really know the codebase but would love to contribute

alex-courtis:
> > * [ ]   add option `vim_ui_popup = false`
> > * [ ]   implement two copy-paste selects, removing `Overwrite? y/n/r(ename)` from the popup version
> > * [ ]   test all nvim-tree `vim.ui.select` and `vim.ui.input` with dressing, adding alternative if required
> > * [ ]   address stylua violations
> 
> Let me know if I can help, I don't really know the codebase but would love to contribute

> > add option `vim_ui_popup = false`

Add this to `DEFAULT_OPTS`. Retain the value in `copy-paste.lua`. Document it like, say, `reload_on_bufenter`.

> > implement two copy-paste selects, removing Overwrite? y/n/r(ename) from the popup version

Put your new select behind the option, tidying the text.

> > test all nvim-tree vim.ui.select and vim.ui.input with dressing, adding alternative if required

Please try all the vim.ui. codepaths with dressing and ensure that they look reasonable.

> > address stylua violations

https://github.com/kyazdani42/nvim-tree.lua/runs/7824109954?check_suite_focus=true

alex-courtis:
Are you still working on this one @mosheavni ?

mosheavni:
> Are you still working on this one @mosheavni ?

nope

alex-courtis:
Resolved: https://github.com/kyazdani42/nvim-tree.lua/commit/ac8d259badf915cd9aaad406503d116230296c44

mosheavni:
> Resolved: [ac8d259](https://github.com/kyazdani42/nvim-tree.lua/commit/ac8d259badf915cd9aaad406503d116230296c44)

i appreciate it very much. thanks.

PR #1501, opened 2022-08-11
Fix focusing directories with a trailing slash in their path doesn't work

carlosecp:
The issue here was that the directory path generated when a new directory was created through the `fs.create-file` function was not formatted correctly as `utils.focus_file` required, so it wasn't focused after it's creation. The solution was to remove the trailing slash on this path, which was solved using the `utils.path_remove_trailing` utility when attempting to focus the newly created dir in the above mentioned functions.

Solves https://github.com/kyazdani42/nvim-tree.lua/issues/1500


carlosecp:
The commit's message probably needs a rebase, I didn't know how to format it.

alex-courtis (approved):
Tested OK nvim v0.7.2

PR #1505, opened 2022-08-13
feat: add config option for a tree indent width

mivort:
This PR adds 'indent_width' option to configure visible indent for tree nesting levels (default is 2). Both enabled and disabled indent markers are supported.

A little context: 9a02dedd92fad67b04b2a3fee2de20555956b089 seemingly removed the ability to have a compact 1-symbol width indentations (and also without indent markers it was only possible to have a fixed 2-space indents, no more, no less). This PR adds ability to have a both smaller and larger (3+) indentations.

### DIFF
--- a/lua/nvim-tree/renderer/components/padding.lua
+++ b/lua/nvim-tree/renderer/components/padding.lua
@@ -100,7 +100,7 @@ function M.setup(opts)
     if #symbol == 0 then
       return " "
     end
-    return symbol:match("[%z\1-\127\194-\244][\128-\191]*")
+    return symbol:match "[%z\1-\127\194-\244][\128-\191]*"
   end
 
   for k, v in pairs(M.config.indent_markers.icons) do
--- a/lua/nvim-tree/renderer/components/padding.lua
+++ b/lua/nvim-tree/renderer/components/padding.lua
@@ -30,25 +30,27 @@ local function get_padding_indent_markers(depth, idx, nodes_number, markers, wit
     for i = 1, depth do
       local glyph
       if idx == nodes_number and i == depth then
-        glyph = M.config.indent_markers.icons.corner
+        local bottom_width = M.config.indent_width - 2 +
+          (with_arrows and not inline_arrows and has_folder_sibling and 2 or 0)
+        glyph = M.config.indent_markers.icons.corner ..
+          string.rep(M.config.indent_markers.icons.bottom, bottom_width) ..
+          (M.config.indent_width > 1 and " " or "")
       elseif markers[i] and i == depth then
-        glyph = M.config.indent_markers.icons.item
+        glyph = M.config.indent_markers.icons.item .. indent
       elseif markers[i] then
-        glyph = M.config.indent_markers.icons.edge
+        glyph = M.config.indent_markers.icons.edge .. indent
       else
-        glyph = M.config.indent_markers.icons.none
+        glyph = M.config.indent_markers.icons.none .. indent
       end
 
       if not with_arrows or (inline_arrows and (depth ~= i or not node.nodes)) then
-        padding = padding .. glyph .. indent
+        padding = padding .. glyph
       elseif inline_arrows then
         padding = padding
-      elseif idx == nodes_number and i == depth and has_folder_sibling then
-        padding = padding .. base_padding .. glyph .. "──" .. indent
-      elseif depth == i and not node.nodes and has_folder_sibling then
-        padding = padding .. base_padding .. glyph .. indent .. base_padding
+      elseif idx ~= nodes_number and depth == i and not node.nodes and has_folder_sibling then
+        padding = padding .. base_padding .. glyph .. base_padding
       else
-        padding = padding .. base_padding .. glyph .. indent
+        padding = padding .. base_padding .. glyph
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lua/nvim-tree/renderer/components/padding.lua
```diff
@@ -25,30 +25,30 @@ local function get_padding_indent_markers(depth, idx, nodes_number, markers, wit
...
+      if not with_arrows or (inline_arrows and (depth ~= i or not node.nodes)) then
+        padding = padding .. glyph .. indent
       elseif inline_arrows then
         padding = padding
-      elseif idx == nodes_number and i == rdepth and has_folder_sibling then
-        padding = padding .. base_padding .. glyph .. "── "
-      elseif rdepth == i and not node.nodes and has_folder_sibling then
-        padding = padding .. base_padding .. glyph .. " " .. base_padding
+      elseif idx == nodes_number and i == depth and has_folder_sibling then
+        padding = padding .. base_padding .. glyph .. "──" .. indent
```
alex-courtis, line 47: We should extend this and the corner:

20220814_155144 497x927

A `renderer.indent_markers.icons.bottom = "─"` should be added.

alex-courtis (changes requested):

lua/nvim-tree/renderer/components/padding.lua
mivort: Done, I've added `icons.bottom` char and added support for corner extension in all three possible option combinations (inline/non-inline arrows, no arrows). Without icons it looks like this:

| `inline_arrows = true` | `inline_arrows = false` | `folder_arrow = false` |
|---|---|---|
|image|image|image|

alex-courtis:
This is looking good; everything appears functional.

This is a complex piece of functionality that has experienced difficulties in the past.

I will:
* exhaustively test manually
* soak it for a week or so in a variety of configurations

alex-courtis:
When we have two character icons the bottom is indented too far. That is a result of the assumption that glyphs are a single character, which is reasonable.

20220816_104244 565x871

This is an edge case however we should deal with it.

Please trim icons to the first character; set to a single space if they are length 0.
Add to the documentation:
`Icons shown before the file/directory. Length 1.`

alex-courtis:
`indent_width = 0` is "working" when indent markers enabled and disabled. What should we do when 0 and markers?

This is a questionable use case and adds complication e.g. `local indent = string.rep(" ", M.config.indent_width - 1)`

Let's remove it:
* set to 1 when less than 1 in padding setup
* add documentation `Number of spaces for an each tree nesting level. Minimum 1.`

alex-courtis:
Test Cases

`indent_markers.enable = false`
- [ ] 1
- [ ] 2
- [ ] 3
- [ ] 3 folder
- [ ] 3 folder folder_arrow
- [ ] 3 folder_arrow

`indent_markers.enable = true`
- [ ] 3 folder
- [ ] 3 folder folder_arrow
- [ ] 3 folder_arrow
- [ ] 3 inline_arrows
- [ ] 3 inline_arrows folder
- [ ] 3 inline_arrows folder folder_arrow
- [ ] 3 inline_arrows folder_arrow

alex-courtis (changes requested):
- [ ] minimum width 1
- [ ] single character icons

alex-courtis:
This has worked well for a week in a variety of situations.

Discovered existing bug #1524 which can be fixed afterwards.

Will merge following changes https://github.com/kyazdani42/nvim-tree.lua/pull/1505#pullrequestreview-1073398428

alex-courtis:
> * [ ]  minimum width 1
> * [ ]  single character icons

Nudge @mivort 

https://github.com/kyazdani42/nvim-tree.lua/pull/1505#issuecomment-1216025335

mivort:
@alex-courtis Sorry for the late update. I've updated doc and added checks for minimal indent width and marker symbol width.
Please take note I've used [this solution](https://stackoverflow.com/a/13238257/3313557) to implement retrieval of the first UTF-8 character (since `:sub(1, 1)` doesn't properly handle multi-byte characters).

alex-courtis:
Retest da6a766

`indent_markers.enable = false`
- [ ] 1
- [ ] 2
- [ ] 3
- [ ] 3 folder
- [ ] 3 folder folder_arrow
- [ ] 3 folder_arrow

`indent_markers.enable = true`
- [ ] 1
- [ ] 1 folder
- [ ] 1 folder folder_arrow
- [ ] 1 folder_arrow
- [ ] 1 inline_arrows
- [ ] 1 inline_arrows folder
- [ ] 1 inline_arrows folder folder_arrow
- [ ] 1 inline_arrows folder_arrow
- [ ] 3
- [ ] 3 folder
- [ ] 3 folder folder_arrow
- [ ] 3 folder_arrow
- [ ] 3 inline_arrows
- [ ] 3 inline_arrows folder
- [ ] 3 inline_arrows folder folder_arrow
- [ ] 3 inline_arrows folder_arrow

alex-courtis:
> @alex-courtis Sorry for the late update. I've updated doc and added checks for minimal indent width and marker symbol width. Please take note I've used [this solution](https://stackoverflow.com/a/13238257/3313557) to implement retrieval of the first UTF-8 character (since `:sub(1, 1)` doesn't properly handle multi-byte characters).

Nice work.

Please add a comment, something like:
`  -- return the first character from the UTF-8 encoded string; we may use utf8.codes from Lua 5.3 when available`

Testing now...

alex-courtis:
The recent update to nerd hack font has change the glyphs and they don't line up nicely any more:

20220830_094137 521x766

Fortunately that's unrelated to this change and everything here is working nicely.

alex-courtis:
Many thanks @mivort 

We can now customise the tree to look exactly as we want.

alex-courtis:
> The recent update to nerd hack font has change the glyphs and they don't line up nicely any more:

Resolution: #1561

PR #1509, opened 2022-08-14
fix(#1503): focus last win before close

### LINKED ISSUE #1503: Update_focused_file.update_root acting incorrectly for window working directory
### Description

Background - I have the following options enabled:

- `sync_root_with_cwd = true`
- `update_focused_file = {
    enable = true,
    update_root = true,
    ignore_list = {'help'}
  }`

Scenario: 
1. I have a single tab with 1 file open, where the window `lcd` is set to a certain directory.
2. I open another window with `:new` and open a file in a different directory.
3. I use :NvimTreeToggle to open nvim-tree and see that the root is set to the correct directory of the file that I just opened.
4. Within the window of the new file, I `:lcd ..` to one level up. Again, nvim-tree is updated properly.
5. If I then close and reopen nvim-tree two times with :NvimTreeToggle, then the root is updated to the direct parent directory of the file, which seems to contradict the intended behavior of `Update_focused_file.update_root`.

This does not happen when changing focus between two windows. The root of nvim-tree is properly updated without resetting directly to the parent directory of the second file. Also, it seems like it has to be the bottom window in the layout for it to reset. It doesn't happen with the top one.

### Neovim version

```text
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.7.2_1/share/nvim"

```

macOS 12.5

### nvim-tree version

1685484

### Minimal config

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {
    sync_root_with_cwd = true,
    update_focused_file = {
      enable = true,
      update_root = true,
      ignore_list = {'help'}
    }
  }
end
```

### Steps to reproduce

1. Open a file in Nvim.
2. Do `:new`
3. Open a file in a different directory in the second window.
4. Do `:lcd ..`
5. Execute `:NvimTreeToggle` twice.
6. See the root of nvim-tree reset to the parent of the lower file.

### Expected behavior

The root should not reset to the parent directory because according to the description of `update_focused_file.update_root`, It should only `Update the root directory of the tree if the file is not under current
    root directory.`

### Actual behavior

After toggling nvim-tree a few times, the root is updated to the containing directory of the file.

alex-courtis:
closes #1503 

@alexisgeoffrey please test this fix for ~1 week.

```sh
cd <your-package-directory>/nvim-tree.lua
git pull
git checkout 1503-focus-last-win-before-close
```

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -188,11 +188,11 @@ function M.close()
   local current_win = a.nvim_get_current_win()
   for _, win in pairs(a.nvim_list_wins()) do
     if tree_win ~= win and a.nvim_win_get_config(win).relative == "" then
-      a.nvim_win_close(tree_win, true)
       local prev_win = vim.fn.winnr "#" -- this tab only
       if tree_win == current_win and prev_win > 0 then
         a.nvim_set_current_win(vim.fn.win_getid(prev_win))
       end
+      a.nvim_win_close(tree_win, true)
       events._dispatch_on_tree_close()
       return
     end


alexisgeoffrey:
I can confirm so far that this fixes my issue! Thank you! I will continue using it this week to make sure I don't notice any problems.

Issue #1518, opened 2022-08-19
sort_by = 'modification_time' does not update sort order on refresh


sebcode:
### Description

With `sort_by = 'modification_time'` set, I'd expect the sort order of tree nodes to change whenever the modification time of a file changes after the tree refreshes (e.g. using :NvimTreeRefresh). But that's not the case.

### Neovim version

```text
NVIM v0.8.0-dev-878-g02e9b5a8e
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux 5.15.0-46-generic

### nvim-tree version

09a51266bca28dd87febd63c66bdbd74f7764a63

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
      "kyazdani42/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {
    sort_by = 'modification_time'
  }
end
```

### Steps to reproduce

1. `touch aaa; sleep 2; touch bbb`
1. nvim -nu nvt-min.lua
1. :NvimTreeOpen
1. Open the file `aaa` and save it `:w`
1. :NvimTreeRefresh

### Expected behavior

[...]

PR #1519, opened 2022-08-19
fix(#1518): sort_by=modification_time not reordering on refresh.

sebcode:
Potential fix for #1518.

### DIFF
--- a/lua/nvim-tree/explorer/reload.lua
+++ b/lua/nvim-tree/explorer/reload.lua
@@ -41,8 +41,18 @@ function M.reload(node, status)
       break
     end
 
+    local stat
+    local function fs_stat_cached(path)
+      if stat ~= nil then
+        return stat
+      end
+
+      stat = uv.fs_stat(path)
+      return stat
+    end
+
     local abs = utils.path_join { cwd, name }
-    t = t or (uv.fs_stat(abs) or {}).type
+    t = t or (fs_stat_cached(abs) or {}).type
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/explorer/reload.lua
```diff
@@ -65,6 +65,7 @@ function M.reload(node, status)
       local n = nodes_by_path[abs]
       if n then
         n.executable = builders.is_executable(abs, n.extension or "")
+        n.fs_stat = uv.fs_stat(abs)
```
alex-courtis, line 68: This extra stat is expensive and unnecessary.

Please retain a value from the start of the loop:
```lua
    local stat = uv.fs_stat(abs)
    t = t or (stat or {}).type
```

alex-courtis (changes requested):
- [x] remove extra fs_stat

lua/nvim-tree/explorer/reload.lua
sebcode: Looks like the first stat is rarely necessary since `uv.fs_scandir_next` returns the type for most file systems I tested with (ext3, zfs, vfat) _except_ reiserfs 💀.

Refactored the code to call `stat` only when necessary.

alex-courtis: Nice work.

alex-courtis (approved):
Re-tested OK.

/tmp functioning correctly.

Issue #1520, opened 2022-08-19
File type changes are not detected


sebcode:
### Description

If I remove a file and create a folder with the same name, the file is still displayed with a file icon instead of a folder icon.

### Neovim version

```text
NVIM v0.8.0-dev-878-g02e9b5a8e
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux 5.15.0-46-generic

### nvim-tree version

09a5126

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
      "kyazdani42/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

1. cd /tmp/; nvim -nu nvt-min.lua
1. :NvimTreeOpen
1. Run in separate terminal: `cd tmp; touch aaa; sleep 1; rm aaa; mkdir aaa`

### Expected behavior

The file node for `aaa` should change into a folder node.

### Actual behavior

The folder `aaa` is displayed as a file node instead of a folder node.

PR #1521, opened 2022-08-19
fix(#1520) file type changes are not detected.

sebcode:
Potential fix for #1520.

### DIFF
--- a/lua/nvim-tree/explorer/reload.lua
+++ b/lua/nvim-tree/explorer/reload.lua
@@ -45,6 +45,18 @@ function M.reload(node, status)
     t = t or (uv.fs_stat(abs) or {}).type
     if not filters.should_ignore(abs) and not filters.should_ignore_git(abs, status.files) then
       child_names[abs] = true
+
+      -- Recreate node if type changes.
+      if nodes_by_path[abs] then
+        local n = nodes_by_path[abs]
+
+        if n.type ~= t then
+          utils.array_remove(node.nodes, n)
+          common.node_destroy(n)
+          nodes_by_path[abs] = nil
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/explorer/reload.lua
```diff
@@ -45,6 +54,20 @@ function M.reload(node, status)
...
       child_names[abs] = true
+
+      -- Recreate node if type changes.
+      if nodes_by_path[abs] then
+        local n = nodes_by_path[abs]
+
+        if n.type ~= t then
+          local idx = table_find(node.nodes, n)
+          if idx then
+            table.remove(node.nodes, idx)
```
alex-courtis, line 65: We need to destroy the node here: `common.node_destroy(n)`

lua/nvim-tree/explorer/reload.lua
```diff
@@ -18,6 +18,15 @@ local function update_status(nodes_by_path, node_ignored, status)
   end
 end
 
+local function table_find(table, value)
```
alex-courtis, line 21: This is useful. Please move to utils:

```lua
function M.array_index_of(array, item)
```

alex-courtis (changes requested):
Tested OK:
file -> link
file -> dir
link -> file
link -> dir
dir -> file
dir -> link

This will fit nicely into some ongoing work to change nodes to classes.

- [ ] utils function
- [ ] destory node

lua/nvim-tree/explorer/reload.lua
sebcode: Actually `utils.array_remove` is exactly what I need

alex-courtis (approved):
Re-tested OK

Issue #1529, opened 2022-08-21
View float: Invalid window id


solipfuct:
### Description

I set the floating window open/close to `<leader>e`. But if I close the window with this key combination, I get an error message.

### Neovim version

```text
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux arch 5.19.1-arch2-1

### nvim-tree version

81eb718

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
      "kyazdani42/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  local opts = { noremap = true, silent = true }
  vim.g.mapleader = " "
  vim.keymap.set("", "<Space>", "<Nop>", opts)
  vim.keymap.set("n", "<leader>e", ":NvimTreeRefresh<CR>:NvimTreeToggle<CR>", opts)
  
  require("nvim-tree").setup {
    sort_by = "case_sensitive",
    sync_root_with_cwd = true,
    respect_buf_cwd = true,
  
    update_focused_file = {
      enable = true,
      update_root = true
    },
    view = {
      float = {
        enable = true
      }
    }
  }
end
```

### Steps to reproduce

1. nvim -nu ~/nvt-min.lua
[...]

PR #1535, opened 2022-08-21
fix: attempt to focus non-existent win after close

MrFixThis:
* Removed: error when attempting to focus a non-existent window when closing nvim-tree floating instance given any event


alex-courtis:
Whoops... I didn't see this before posting #1529 

I went with a naive "does this window still exist" check, relying on the focus shift to close it. Thoughts?

MrFixThis:
No problem buddy, the key here is just to solve the broken closing behavior of the floating nvim-tree instance. I'm gonna close the PR to allow you merge the solution you gave, which looks concrete.

PR #1546, opened 2022-08-23
fix(view): Fix closing nvim-tree float when file is removed

### LINKED ISSUE #1539: View float: Cannot close last window at remove open file
### Description

When deleting an open file with floating view enabled, I get an error.

### Neovim version

```text
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
```

Linux arch 5.19.1-arch2-1

### nvim-tree version

e3353c4

### Minimal config

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  local opts = { noremap = true, silent = true }
  vim.g.mapleader = " "
  vim.keymap.set("", "<Space>", "<Nop>", opts)
  vim.keymap.set("n", "<leader>e", ":NvimTreeRefresh<CR>:NvimTreeToggle<CR>", opts)
  
  require("nvim-tree").setup {
    sort_by = "case_sensitive",
    sync_root_with_cwd = true,
    respect_buf_cwd = true,
  
    update_focused_file = {
      enable = true,
      update_root = true
    },
    view = {
      float = {
        enable = true
      }
    }
  }
end
```

### Steps to reproduce

1. nvim -nu ~/nvt-min.lua
2. Open tree `<space>e`
3. Create file `a`
4. Open file
5. Open tree `<space>e`
6. Delete `d` current file

### Expected behavior

Deletes a file that is opened without an error

### Actual behavior

```
E5108: Error executing lua: ...t/nvim-tree.lua/lua/nvim-tree/actions/fs/remove-file.lua:12: Vim:E444: Cannot close last window
stack traceback:
        [C]: in function 'nvim_win_close'
        ...t/nvim-tree.lua/lua/nvim-tree/actions/fs/remove-file.lua:12: in function 'close_windows'
        ...t/nvim-tree.lua/lua/nvim-tree/actions/fs/remove-file.lua:29: in function 'clear_buffer'
        ...t/nvim-tree.lua/lua/nvim-tree/actions/fs/remove-file.lua:86: in function 'on_confirm'
        /usr/share/nvim/runtime/lua/vim/ui.lua:93: in function 'input'
        ...t/nvim-tree.lua/lua/nvim-tree/actions/fs/remove-file.lua:71: in function 'handle_tree_actions'
        ...r/start/nvim-tree.lua/lua/nvim-tree/actions/dispatch.lua:120: in function 'dispatch'
        ...acker/start/nvim-tree.lua/lua/nvim-tree/actions/init.lua:260: in function <...acker/start/nvim-tree.lua/lua/nvim-tree/actions/init.lua:259>
```

Khrees:
Resolves #1539

### DIFF
--- a/lua/nvim-tree/actions/fs/remove-file.lua
+++ b/lua/nvim-tree/actions/fs/remove-file.lua
@@ -23,7 +23,7 @@ local function clear_buffer(absolute_path)
   local bufs = vim.fn.getbufinfo { bufloaded = 1, buflisted = 1 }
   for _, buf in pairs(bufs) do
     if buf.name == absolute_path then
-      if buf.hidden == 0 then
+      if buf.hidden == 0 and (#bufs > 1 or view.View.float.enable) then
         local winnr = a.nvim_get_current_win()
         a.nvim_set_current_win(buf.windows[1])
         vim.cmd ":bn"
--- a/lua/nvim-tree/actions/fs/remove-file.lua
+++ b/lua/nvim-tree/actions/fs/remove-file.lua
@@ -3,10 +3,15 @@ local luv = vim.loop
 
 local utils = require "nvim-tree.utils"
 local events = require "nvim-tree.events"
+local view = require "nvim-tree.view"
 
 local M = {}
 
 local function close_windows(windows)
+  if view.View.float.enable and #a.nvim_list_wins() == 1 then
+    return
+  end
+
   for _, window in ipairs(windows) do
     if a.nvim_win_is_valid(window) then
       a.nvim_win_close(window, true)
@@ -18,11 +23,13 @@ local function clear_buffer(absolute_path)
   local bufs = vim.fn.getbufinfo { bufloaded = 1, buflisted = 1 }
   for _, buf in pairs(bufs) do
     if buf.name == absolute_path then
-      if buf.hidden == 0 and #bufs > 1 then
+      if buf.hidden == 0 then
         local winnr = a.nvim_get_current_win()
         a.nvim_set_current_win(buf.windows[1])
         vim.cmd ":bn"
-        a.nvim_set_current_win(winnr)
+        if not view.View.float.enable then
+          a.nvim_set_current_win(winnr)
+        end
       end
       a.nvim_buf_delete(buf.bufnr, { force = true })
       if M.close_window then


alex-courtis (approved):

lua/nvim-tree/actions/fs/remove-file.lua
```diff
@@ -18,11 +23,13 @@ local function clear_buffer(absolute_path)
   local bufs = vim.fn.getbufinfo { bufloaded = 1, buflisted = 1 }
   for _, buf in pairs(bufs) do
     if buf.name == absolute_path then
-      if buf.hidden == 0 and #bufs > 1 then
+      if buf.hidden == 0 then
```
alex-courtis, line 26: What's the reason for this change?

It is a change in behaviour for non-floating:  the window will be closed even if there are other buffers present.

alex-courtis (changes requested):
Love your work Khrees.

Tested:
* delete only open buffer
* delete current buffer
* delete other buffer
* delete unopened buffer

Last case leaves nvim-tree open which is reasonable behaviour.

- [x] https://github.com/kyazdani42/nvim-tree.lua/pull/1546/files#r956528621

lua/nvim-tree/actions/fs/remove-file.lua
Khrees: I am not able to reproduce such behaviour. With non-floating window, it seems to behave the same regardless of this change. How can I reproduce this?

alex-courtis: > I am not able to reproduce such behaviour. With non-floating window, it seems to behave the same regardless of this change. How can I reproduce this?

I should have noted how to reproduce this as I now cannot... the next buffer is correctly displayed.

The window switch and `:bn` are now invoked when there are no other buffers in the window, however that seems fine.

**Real question:** do we need this change? If it's necessary for floating windows it should be done only for the case of floating windows.

Khrees: I reverted this change so non-floating windows should not be affected.

When I was trying to fix the original bug from the referenced issue, the deleted buffer was still displayed for some reason. The important bit here seems to be switching focus to window with soon-to-be-deleted buffer.

alex-courtis: > When I was trying to fix the original bug from the referenced issue, the deleted buffer was still displayed for some reason. The important bit here seems to be switching focus to window with soon-to-be-deleted buffer.

:+1:

PR #1556, opened 2022-08-26
fix(view): Fix view.width overriding view.float.open_win_config

Khrees:
Resolves #1543

PR #1559, opened 2022-08-28
feat(adaptive_size): floating window's optional adaptive size specification

MrFixThis:
Given the last merged PR (#1556) suggested by @Khrees, the floating window's adaptive size has been fully blocked, making impossible to allow it for those who actually use it.

```lua
if M.View.float.enable and not M.View.adaptive_size then
  -- if the floating windows's adaptive size is not desired, then the
  -- float size should be defined in view.float.open_win_config
  return
end
```

So i propose to check both the floating window and adaptive size configuration to avoid the resizing behavior in floating windows only when the adaptive size is disabled.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -246,7 +246,8 @@ function M.grow_from_content()
 end
 
 function M.resize(size)
-  if M.View.float.enable then
+  if M.View.float.enable and not M.View.adaptive_size then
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Tested OK with a variety of `open_win_config` and `float` false.

alex-courtis:
Please review @Khrees

Khrees:
Looks good to me.

PR #1571, opened 2022-09-02
fix(view): Fix filter and file info popups with float enabled

Khrees:
This PR fixes floating nvim-tree behaviour involving pop-ups:

- [x] live-filter: focusing on filter window caused nvim-tree window to close,
- [x] file-popup: file info pop-up was displayed below nvim-tree window.

### DIFF
--- a/lua/nvim-tree/actions/node/file-popup.lua
+++ b/lua/nvim-tree/actions/node/file-popup.lua
@@ -32,6 +32,7 @@ local function setup_window(node)
     width = max_width + 1,
     height = #lines,
     noautocmd = true,
+    zindex = 60,
   })
   local winnr = a.nvim_open_win(0, false, open_win_config)
   current_popup = {
--- a/lua/nvim-tree/live-filter.lua
+++ b/lua/nvim-tree/live-filter.lua
@@ -25,6 +25,15 @@ local overlay_bufnr = nil
 local overlay_winnr = nil
 
 local function remove_overlay()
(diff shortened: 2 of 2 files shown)


lua/nvim-tree/actions/node/file-popup.lua
```diff
@@ -32,6 +32,7 @@ local function setup_window(node)
     width = max_width + 1,
     height = #lines,
     noautocmd = true,
+    zindex = 60,
```
alex-courtis, line 35: :+1: 

The default value for floats are 50. In general, values below 100 are recommended, unless there is a good reason to overshadow builtin elements.

lua/nvim-tree/live-filter.lua
```diff
@@ -92,12 +101,24 @@ local function configure_buffer_overlay()
...
+      event = "WinLeave",
+      pattern = "NvimTree_*",
+      group = a.nvim_create_augroup("NvimTree", { clear = false }),
+    }
+
+    min_width = min_width - 2
+  end
+
   configure_buffer_overlay()
   overlay_winnr = a.nvim_open_win(overlay_bufnr, true, {
```
alex-courtis, line 117: Not sure what the zindex is here, but it's working :shrug:

alex-courtis (approved):
Works beautifully, thanks for identifying and fixing.

Forcibly setting zindex is the right thing to do; there's not a good use case for the user specifying their own.

PR #1573, opened 2022-09-03
fix(#1564): reset &bufhidden when opening an existing (maybe previewed) buffer

### LINKED ISSUE #1564: If a preview buffer is opened, we should set bufhidden=""
### Description

If we opened a preview buffer, we should not expect it to be removed even we don't edit anything

see screenshot below, the right behavior should be keep the opened buffer
Kapture 2022-08-30 at 14 13 42

### Neovim version

```text
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
```

macOS 12.5.1

### nvim-tree version

08ab346

### Minimal config

```lua
no special config needed
```

### Steps to reproduce

1. <Tab> to preview a file
2. <CR> to open the same file
3. switch to another buffer
4. the opened file is been deleted

### Expected behavior

expect the opened preview file wouldn't be deleted

### Actual behavior

_No response_

alex-courtis:
fixes #1564 

@clatisus please test.

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -264,6 +264,7 @@ function M.fn(mode, filename)
     open_in_new_window(filename, mode, win_ids)
   else
     api.nvim_set_current_win(found_win)
+    vim.bo.bufhidden = ""
   end
 
   if M.resize_window then


clatisus:
thanks for the fix! it works now

PR #1594, opened 2022-09-20
fix(#1553): set correct side on vim open directory

### LINKED ISSUE #1553: NvimTree aligns itself on the right the first time I open a file, but closing and opening the file again fixes it.
### Description

Pretty much the title. I have configured NvimTree to start when opening a directory. When started this way, the first time I open a file NvimTree will align itself on the right side of the screen for some reason. Closing the file and opening it again will correctly re-align NvimTree on the left side.

### Neovim version

```text
NVIM v0.7.2
Build type: Release
LuaJIT 2.1.0-beta3
```

5.19.2-arch1-1

### nvim-tree version

e3353c4

### Minimal config

```lua
-- PLUGINS
local Plug = vim.fn['plug#']
vim.call('plug#begin', '~/.config/nvim/plugged')

vim.call('plug#end')



-- NVIM TREE CONFIG
require("nvim-tree").setup({
  sort_by = "name",
  open_on_setup = true,
  hijack_cursor = true,
  view = {
    side = "left",
    width = 20, 
    mappings = {
      list = {
        { key = "-", action = "dir_up" },
	{key = "+", action = "cd"}
      },
    },
  },
  renderer = {
    group_empty = true,
  },
  filters = {
    dotfiles = true,
  },
})
```

### Steps to reproduce

1) Open Nvim with a directory as a target.
2) NvimTree will open in fullscreen mode as normal
3) Choose a file and open it by pressing enter
4) NvimTree will be positioned on the right side of the screen while the file will be opened on the left side
5) Close the file with :q
6) Now open it back again in the same way
7) Now the file is on the right side while NvimTree is on the left, as it should

### Expected behavior

NvimTree should be aligned to the left side the first time it is opened.

### Actual behavior

_No response_

alex-courtis:
resolves #1553

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -7,7 +7,7 @@ local view = require "nvim-tree.view"
 
 local M = {}
 
-local function get_split_cmd()
+local function get_split_side()
   local side = view.View.side
   if side == "right" then
     return "aboveleft"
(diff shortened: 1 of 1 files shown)


alex-courtis:
Tested:
* side = "left"
* side = "right"
* split action
* vsplit action

side = "top" and side = "bottom" are not completely functional right now, however this change does not make anything any worse. Raised #1595

PR #1609, opened 2022-09-27
fix(view): window options are set even if it is not a drawer

uga-rosa:
Because they are window-local, options such as winfixwidth will remain after the file is opened.


uga-rosa:
There may be side effects, so if you have a better fix, feel free to close it.

kyazdani42:
it's needed though, since we need to set the buffer into the window and make the window a tree window. Closing since it doesn't seem to be right.

Issue #1270, opened 2022-05-15
Race Condition With open_on_setup


alex-courtis:
### Description

Inconsistent behaviour is occurring with the following options:
```
open_on_setup
open_on_setup_file
ignore_buffer_on_setup
disable_netrw
hijack_netrw
```

### Neovim version

```text
NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux 5.17.5-arch1-1

### nvim-tree version

7293f8d

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
      "kyazdani42/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {
	  open_on_setup = true,
	  open_on_setup_file = false,
	  ignore_buffer_on_setup = true,
	  disable_netrw = false,
	  hijack_netrw = true,
  }
end
```

### Steps to reproduce

`vim -nu /tmp/nvt-min.lua <file>` sometimes opens the tree
`nvim -nu /tmp/nvt-min.lua` sometimes opens the tree
`nvim -nu /tmp/nvt-min.lua .` sometimes opens netrw

alex-courtis:
Alternatives like:
```lua
[...]

PR #1618, opened 2022-10-02
fix(#1270): open_on_setup_file does not override open_on_setup, hijack_directories does not override startup behaviour

alex-courtis:
Investigating #1270 and found some bugs and inconsistencies:

Bugs: `open_on_setup_file` overrides `open_on_setup`

Inconsistencies: `hijack_directories` overrides `open_on_setup`. Not necessarily a bug, however this behaviour has been updated to match the documentation.

Documentation has been clarified.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -390,29 +390,33 @@ Hijack netrw windows (overridden if |disable_netrw| is `true`)
   Type: `boolean`, Default: `true`
 
 *nvim-tree.open_on_setup*
-Will automatically open the tree when running setup if startup buffer is
-a directory, is empty or is unnamed. nvim-tree window will be focused.
+Will automatically open the tree when running setup if startup buffer is a
+directory, empty or unnamed.
+nvim-tree window will be focused.
 See |nvim-tree.focus_empty_on_setup|
   Type: `boolean`, Default: `false`
 
 *nvim-tree.open_on_setup_file*
-Will automatically open the tree when running setup if startup buffer is a file.
+Will automatically open the tree when running setup if startup buffer is a
+file.
 File window will be focused.
-File will be found if update_focused_file is enabled.
+File will be found if |nvim-tree.update_focused_file| is enabled.
   Type: `boolean`, Default: `false`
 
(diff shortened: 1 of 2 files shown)


alex-courtis:
Test cases:

defaults
- [x] file
- [x] dir
- [x] empty
- [x] none

open_on_setup
- [x] file
- [x] dir
- [x] empty
- [x] none

open_on_setup_file
- [x] file
- [x] dir
- [x] empty
- [x] none

open_on_setup, open_on_setup_file
- [x] file
- [x] dir
- [x] empty
- [x] none

open_on_setup, ignore_buffer_on_setup
- [x] file
- [x] dir
- [x] empty
- [x] none

open_on_setup_file, ignore_ft_on_setup
- [x] file ft match
- [x] file
- [x] dir
- [x] empty
none

open_on_setup, ignore_ft_on_setup
- [x] file ft match
- [x] file
- [x] dir
- [x] empty
- [x] none

alex-courtis:
If we were building from scratch today, I'd define options like these, however that is way too disruptive a change to make:

```lua
  startup = {
    open = {
      always = false,
      dir = false,
      file = false,
      empty = false,
    },
    focus_tree = {
      always = false,
      dir = false,
      file = false,
      empty = false,
    },
    ignore_ft = {
    }
  }
```

kyazdani42:
i have edge case testing for open on setup also:
- `man man` with `export MANPAGER="nvim +Man!"`
- lazy loading, but this shouldn't be affected.

alex-courtis:
> * `man man` with `export MANPAGER="nvim +Man!"`

This doesn't open the tree unless:
```lua
    ignore_buffer_on_setup = true,
    open_on_setup = true,
```

> git operations with nvim

The tree will be opened when `open_on_setup_file = true`. I've never actually noticed this, but now you point it out I do...

This can be stopped via:
```lua
		ignore_ft_on_setup = {
			"gitcommit"
		},
```

Perhaps I'm missing something?

kyazdani42:
thats good then, the behavior might be the same as before.

kyazdani42 (approved):
lets try this, if users complain we can just revert.

alex-courtis:

PR #1620, opened 2022-10-02
ADD: auto_close_when_last 

NEX-S:
`auto_close_when_last` can autoclose the nvim-tree when it's last remaining window

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -344,6 +344,16 @@ local function setup_autocommands(opts)
 
   local has_watchers = opts.filesystem_watchers.enable
 
+  if opts.auto_close_when_last then
+    create_nvim_tree_autocmd("BufEnter", {
+      callback = function ()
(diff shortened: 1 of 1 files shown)


gegoune:
This PR is likely to be rejected. For explanation why please search through closed issue/PRs. In the nutshell there are too many edge cases and issues with that functionality. You can always set it up in your configuration.

gegoune:
@alex-courtis Would IT be worth having a look if that implementation is actually worth adding?

alex-courtis:
We specifically removed this functionality and [do not intend to implement](https://github.com/kyazdani42/nvim-tree.lua#tips--tricks).

> vim events are unpredictably ordered, especially when other plugins and automation is involved

> BufEnter is the last event that can be acted upon and that event can have side effects

> This is a difficult vim problem with no good solution. It's not specific to nvim-tree.

> BufEnter / WinClosed can be problematic: the state of other buffers/tabs/windows is not always their correct state at the time of the event. Some plugins etc. do not use ++nested on their autocommands, which adds more confusion as events are "lost", if they perform actions inside their autocmmand that would raise events.

PR #1621, opened 2022-10-03
feat: add float autoclose option

emmanueltouzery:
I like the float mode of the treeview, however I'd like it not to autoclose when the popup loses focus.

So I've added an option to that effect.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -698,8 +698,12 @@ Window / buffer setup.
     Configuration options for floating window
 
         *nvim-tree.view.float.enable*
-        Display nvim-tree window as float (enforces |nvim-tree.actions.open_file.quit_on_open| if set).
+        Display nvim-tree window as float (enforces if set).
           Type: `boolean`, Default: `false`
+
+        *nvim-tree.view.float.quit_on_focus_loss
+        When in floating mode, autoclose the popup when the popup loses the focus.
+          Type: `boolean`, Default: `true`
 
         *nvim-tree.view.float.open_win_config*
         Floating window config. See |nvim_open_win| for more details.
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -418,7 +418,7 @@ local function setup_autocommands(opts)
     })
   end
 
-  if opts.view.float.enable then
+  if opts.view.float.enable and opts.view.float.quit_on_focus_loss then
     create_nvim_tree_autocmd("WinLeave", { pattern = "NvimTree_*", callback = view.close })
   end
 end
@@ -463,6 +463,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
     },
     float = {
       enable = false,
+      quit_on_focus_loss = true,
       open_win_config = {
         relative = "editor",
         border = "rounded",
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -267,7 +267,7 @@ function M.fn(mode, filename)
 end
 
 function M.setup(opts)
-  M.quit_on_open = opts.actions.open_file.quit_on_open or opts.view.float.enable
+  M.quit_on_open = opts.actions.open_file.quit_on_open
   M.resize_window = opts.actions.open_file.resize_window
   if opts.actions.open_file.window_picker.chars then
     opts.actions.open_file.window_picker.chars = tostring(opts.actions.open_file.window_picker.chars):upper()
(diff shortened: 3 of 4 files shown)


alex-courtis:
I asked for that to be [unconditionally added](https://github.com/kyazdani42/nvim-tree.lua/pull/1462#pullrequestreview-1056510957) as I didn't consider the use case that the user desires this functionality.

alex-courtis (changes requested):
`view.float.autoclose = true`

Tested OK:
* window focus is changed

Needs work:
* `actions.open_file.quit_on_open = false`: float closes when opening file

emmanueltouzery:
Yes, the `close_on_open` was hardcoded as per the original PR https://github.com/kyazdani42/nvim-tree.lua/pull/1462 -- the comment was that it didn't work properly otherwise.

This is also documented:
```
*nvim-tree.view.float.enable*
        Display nvim-tree window as float (enforces |nvim-tree.actions.open_file.quit_on_open| if set).
```
I didn't have that much testing time with the feature yet, so I'm not sure what I'd prefer yet. Right now I think that I'd like the setup of this PR (don't close on focus loss, close on open).

I understand that the settings as set up by this PR are confusing.

What about if we rename this setting to `auto_close_when_unfocused`? Or maybe `auto_close_on_focus_loss`? And we reserve the option to potentially one day respect `quit_on_open` also for the floating configuration, which I view as an independent thing that also wasn't supported so far.

alex-courtis:
> This is also documented:

Good catch! That can be changed.

> I didn't have that much testing time with the feature yet, so I'm not sure what I'd prefer yet. Right now I think that I'd like the setup of this PR (don't close on focus loss, close on open).
> 
> I understand that the settings as set up by this PR are confusing.
> 
> What about if we rename this setting to auto_close_when_unfocused? Or maybe auto_close_on_focus_loss? And we reserve the option to potentially one day respect quit_on_open also for the floating configuration, which I view as an independent thing that also wasn't supported so far.

We can keep it simple and do it now: floating window respects both `actions.open_file.quit_on_open` and `view.float.autoclose` now. No need for another option.

`autoclose` -> `quit_on_focus_loss` might be the clearest, consistent option.

alex-courtis:
RE defaults: users will experience disruption if they relied on the implicit `quit_on_open`, however I think that is a reasonable change for a newly released feature.

emmanueltouzery:
I don't understand the disruption? `quit_on_open` didn't work for the floating window setup, and still doesn't? So it seems to me there is no change?

I'll update the PR with the setting rename.

alex-courtis:
> I don't understand the disruption? `quit_on_open` didn't work for the floating window setup, and still doesn't? So it seems to me there is no change?

The implicit `quit_on_open` is always firing, regardless of `autoclose`. My proposal is that we always respect the user's `quit_on_open` setting for floating windows. This will be a more predictable user experience and offers the full range of behaviour choices.

Making both changes now will be less disruptive and I would be very greatful if you could make this change in this PR.

emmanueltouzery:
ok, i'll research what is the breakage that the quit_on_open locking was supposed to workaround. i think fixing that shouldn't be too hard.

emmanueltouzery:
OK, I've updated the PR. I do not understand what was the issue with quit_on_open. I removed the setting blocking it and.. it worked right away for me. I couldn't find any issue... It seems to work just fine for me.

It does turn out there can be a conflict between both options. If you say `quit_on_focus_loss=true` but `quit_on_open=false`... i mean when we open a file, we focus it. Therefore it's a focus loss. But the three other cases are clear.

quit_on_focus_loss=true
quit_on_open=true
=> no problem, clear what should happen (and it's the current behavior)

quit_on_focus_loss=false
quit_on_open=true
=> ok, clear what should happen

quit_on_focus_loss=false
quit_on_open=false
=> clear

quit_on_focus_loss=true
quit_on_open=false
=> ambiguous, quit_on_focus_loss wins and effectively we have quit_on_open=true enforced

quit_on_focus_loss=false
quit_on_open=false
=> ok, clear what should happen

I've tested the four cases, everything looks good to me. I intend to use true+true for now, maybe i'll switch to quit_on_focus_loss=false+quit_on_open=true, I'll see. But so far I can't see any issues.

emmanueltouzery:
Ok, I found an issue. The tree contents are not updated when we switch buffers. I'll check it out.

emmanueltouzery:
false alert: I needed update_focused_file.update_root=true. I could have sworn that I didn't need that in the past. Anyway, nevermind. I know of no bugs right now.

alex-courtis:
Thanks for updating. Unfortunately I'm not able to look at this until the weekend.

emmanueltouzery:
no problem, thank you! In the meantime I'll keep using it through the week and be careful for glitches. As I said, I intend to use true+true for now, maybe i'll switch to quit_on_focus_loss=false+quit_on_open=true, I'll see.

alex-courtis (approved):
Tested OK.

Defaults match current behaviour.

Thanks for going the extra mile.

PR #1634, opened 2022-10-09
fix(#1629): nvim start with file named *NvimTree* opens tree instead of buffer

### LINKED ISSUE #1629: `nvim NvimTree.lua` does not open
### Description

I have some custom mappings stored in `nvim/after/ftplugin/NvimTree.lua` to run on filetype NvimTree. Opening this file via `nvim after/ftplugin/NvimTree.lua` actually opens NvimTree on that file and not the file itself.

### Neovim version

```text
NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by builduser

Features: +acl +iconv +tui
See ":help feature-compile"

  fall-back for $VIM: "/usr/share/nvim"

```

arch

### nvim-tree version

7282f7d -n 1

### Minimal config

```lua
no config needed
```

### Steps to reproduce

1. NvimTreeOpen after/ftplugin/NvimTree.lua

### Expected behavior

nvim-tree actually opens the file NvimTree.lua

### Actual behavior

the plugin NvimTree itself opens, not the file

alex-courtis:
fixes #1629

[find_existing_windows](https://github.com/nvim-tree/nvim-tree.lua/blob/54afa503a978164204716a98fb67fb21cbd10017/lua/nvim-tree.lua#L136) misidentifies a buffer named `*NvimTree*` as the tree's buffer.

* `setup` -> `on_enter` -> `find_existing_windows` is the only codepath
* view is always closed in `setup`

∴ this may be removed

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -135,8 +135,7 @@ end
 
 local function find_existing_windows()
   return vim.tbl_filter(function(win)
-    local buf = api.nvim_win_get_buf(win)
-    return api.nvim_buf_get_name(buf):match "NvimTree" ~= nil
+    return utils.is_nvim_tree_buf(api.nvim_win_get_buf(win))
   end, api.nvim_list_wins())
 end
 
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -133,6 +133,13 @@ function M.tab_change()
   end
 end
 
+local function find_existing_windows()
+  return vim.tbl_filter(function(win)
+    local buf = api.nvim_win_get_buf(win)
+    return api.nvim_buf_get_name(buf):match "NvimTree" ~= nil
+  end, api.nvim_list_wins())
+end
+
 local function is_file_readable(fname)
   local stat = luv.fs_stat(fname)
   return stat and stat.type == "file" and luv.fs_access(fname, "R")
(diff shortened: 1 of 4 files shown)


lua/nvim-tree.lua
```diff
@@ -258,13 +251,7 @@ function M.on_enter(netrw_disabled)
     end
   end
 
-  -- Session that left a NvimTree Buffer opened, reopen with it
-  local existing_tree_wins = find_existing_windows()
-  if existing_tree_wins[1] then
-    api.nvim_set_current_win(existing_tree_wins[1])
-  end
-
-  if should_open or existing_tree_wins[1] ~= nil then
```
kyazdani42, line 267: i think this is needed for user using sessions

kyazdani42: maybe a better fix would be to change the matching string to `NvimTree_`

alex-courtis: Good catch; I didn't think of nvim sessions, just user calling setup multiple times.

Added a util to check that the buffer name matches and that it is a real file. We can't test file type as it's not imported via default `sessionoptions`.

@barrett-ruth I'd be grateful for a re-test.

alex-courtis:
Raised #1639 to be completed in a separate branch.

barrett-ruth:
@alex-courtis works great for me.

kyazdani42 (approved):

alex-courtis:

PR #1638, opened 2022-10-10
doc update

princejoogie:
update_root -> update_cwd


alex-courtis:
`update_root` is correct. See `:help nvim-tree.update_focused_file.update_root`

PR #1637, opened 2022-10-09
fix: prevent quit_on_open from opening file in the same window

Almo7aya:
fixes #1628

 call `view._prevent_buffer_override` on `BufWipeout` instead of `view.abandon_current_window` when `quit_on_open = true`

PR #1640, opened 2022-10-10
fix(#1639): ensure tree autocommands match filetype as well as name

### LINKED ISSUE #1639: Strict Matching Of NvimTree Buffers
#1629 demonstated incorrect handling of buffers named `NvimTree*` on startup.

There are further cases, mostly autocommands, in which similar tests will need to be applied. Something like:

```lua
  if opts.view.centralize_selection then
      pattern = "NvimTree_*",
      callback = function()
        if utils.is_current_buf_nvim_tree() then
          vim.schedule(function()
            local keys = api.nvim_replace_termcodes("zz", true, false, true)
            api.nvim_feedkeys(keys, "n", true)
          end)
        end
      end,
    })
  end
```

It's probably safe to just test the filetype as it should be set at most of these points.

alex-courtis:
fixes #1639 

requires #1629 

requires #1632

requires #1637

Autocommands are firing for buffers named `*NvimTree_*` which are real files.


alex-courtis:
Test cases:
- [ ] hijack_cursor
- [ ] wipe tree, no quit_on_open
- [ ] enter tree, no watchers and reload_on_bufenter
- [ ] leave window, float.enable and float.quit_on_focus_loss
- [ ] leave window with filter, float.enable and float.quit_on_focus_loss
- [ ] buf leave, full_name
- [ ] cursor move, full_name

lua/nvim-tree.lua
```diff
@@ -370,8 +377,16 @@ local function setup_autocommands(opts)
   end
 
   if not opts.actions.open_file.quit_on_open then
-    create_nvim_tree_autocmd("BufWipeout", { pattern = "NvimTree_*", callback = view._prevent_buffer_override })
+    create_nvim_tree_autocmd("BufWipeout", {
+      pattern = "NvimTree_*",
+      callback = function()
+        if api.nvim_buf_get_option(0, "filetype") == "NvimTree" then
```
kyazdani42, line 383: vim.bo.filetype is more luaish

alex-courtis: Today I Learned

gegoune: `vim.bo.filetype` is also a shorthand for `vim.bo[0].filetype`.

alex-courtis: Fantastic. I might merge these checks into `is_nvim_tree_buf`.

kyazdani42 (approved):

PR #1648, opened 2022-10-14
fix: preview on floating window

### LINKED ISSUE #1643: Error when calling preview when on floating mode
### Description

Clicking `<Tab>` to preview a file when floating mode is enabled results to `Invalid window id: 1002` error 

image

The issue occurs in this line https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/view.lua#L339 since nvim-tree on floating mode is auto_closed by default, the `winnr` is always invalid since the window is already gone, I suggest using `a.nvim_win_is_valid(wnr)` before trying to focus the window.

I'm not sure what is the expected behavior here, should it reopens the floating window again when previewing a file 

### Neovim version

```text
NVIM v0.9.0-dev-1214-ge54541f7f-dirty
Build type: Release
LuaJIT 2.1.0-beta3
```

MacOS 12.6

### nvim-tree version

main - b01e7be

### Minimal config

```lua
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			"kyazdani42/nvim-tree.lua",
		},
		config = {
			package_root = package_root,
		},
	})
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
	require("nvim-tree").setup({
		view = {
			float = {
				enable = true,
			},
		},
	})
end

vim.api.nvim_set_keymap("n", "<Space>e", "<cmd>:NvimTreeToggle<CR>", {})
```

### Steps to reproduce

1- nvim -nu /tmp/nvt-min.lua
2- :NvimTreeToggle
3- press `<Tab>` on any file

### Expected behavior

_No response_

### Actual behavior

_No response_

Almo7aya:
fixes #1643

My changes prevent nvim-tree from using already closed window, usually when using floating window mode.

More details in the issue #1643

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -209,7 +209,7 @@ local function open_in_new_window(filename, mode, win_ids)
     cmd = string.format("edit %s", fname)
   end
 
-  if mode == "preview" then
+  if mode == "preview" and view.View.float.enable then
     -- ignore "WinLeave" autocmd on preview
     -- because the registered "WinLeave"
     -- will kill the floating window immediately
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -212,7 +212,7 @@ local function open_in_new_window(filename, mode, win_ids)
 
   if mode == "preview" then
     -- ignore "WinLeave" autocmd on preview
-    -- because the requested "WinLeave"
+    -- because the registered "WinLeave"
     -- will kill the floating window immediately
     set_current_win_no_autocmd(api.nvim_get_current_win(), "WinLeave")
   end
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -167,9 +167,10 @@ end
 
 -- This is only to avoid the BufEnter for nvim-tree to trigger
 -- which would cause find-file to run on an invalid file.
-local function set_current_win_no_autocmd(winid)
+local function set_current_win_no_autocmd(winid, autocmd)
+  autocmd = autocmd or "BufEnter"
   local eventignore = vim.opt.eventignore:get()
-  vim.opt.eventignore:append "BufEnter"
+  vim.opt.eventignore:append(autocmd)
   api.nvim_set_current_win(winid)
   vim.opt.eventignore = eventignore
 end
@@ -207,6 +208,13 @@ local function open_in_new_window(filename, mode, win_ids)
     cmd = string.format("%s %s %s", split_side, split_cmd, fname)
   else
     cmd = string.format("edit %s", fname)
+  end
+
+  if mode == "preview" then
+    -- ignore "WinLeave" autocmd on preview
+    -- because the requested "WinLeave"
+    -- will kill the floating window immediately
+    set_current_win_no_autocmd(api.nvim_get_current_win(), "WinLeave")
   end
 
   set_current_win_no_autocmd(target_winid)
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -168,7 +168,6 @@ end
 -- This is only to avoid the BufEnter for nvim-tree to trigger
 -- which would cause find-file to run on an invalid file.
 local function set_current_win_no_autocmd(winid, autocmd)
-  autocmd = autocmd or "BufEnter"
   local eventignore = vim.opt.eventignore:get()
   vim.opt.eventignore:append(autocmd)
   api.nvim_set_current_win(winid)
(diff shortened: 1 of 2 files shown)


alex-courtis (changes requested):
Testing with nvt-min.lua from the issue resulted in an empty nvim-tree window. Upon `R` the contents were correctly repopulated.

[asciicast](https://asciinema.org/a/528519)

Almo7aya:
Ummmm, after some digging I found out that I have 
```lua
update_focused_file = {
	enable = true,
},
```
in my config file, when removing it the issue you faced occurs, I'm checking...

You can try this config with my branch and `update_focused_file = true`

```lua
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			{
				"almo7aya/nvim-tree.lua",
				branch = "fix/preview_action_on_floating_windows",
			},
		},
		config = {
			package_root = package_root,
		},
	})
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
	require("nvim-tree").setup({
		update_focused_file = {
			enable = true,
		},
		view = {
			float = {
				enable = true,
			},
		},
	})
end

vim.api.nvim_set_keymap("n", "<Space>e", "<cmd>:NvimTreeToggle<CR>", {})

```

alex-courtis:
> You can try this config with my branch and `update_focused_file = true`

That works as expected.

Almo7aya:
seems like 
```lua
  if opts.update_focused_file.enable then
      callback = function()
        M.find_file(false)
      end,
    })
  end
```
has something todo with this, removing `M.find_file(false)` results in the same issue you mentioned

alex-courtis:
Cursory investigation:

It seems that the tree window is being closed during the preview action: https://github.com/Almo7aya/nvim-tree.lua/blob/79f631bc1d52b387f4ae59fad1291d894afa97f5/lua/nvim-tree/actions/node/open-file.lua#L134

Rather than being reactive to the closed window in `focus`, we could keep the floating tree window open.

Almo7aya:
After a deep debugging sessions, I discovered that we have an option called `view.float.quit_on_focus_loss` which is enabled by default to kill the window when losing focus 
```lua
  if opts.view.float.enable and opts.view.float.quit_on_focus_loss then
    create_nvim_tree_autocmd("WinLeave", { pattern = "NvimTree_*", callback = view.close })
  end
``` 
so ignoring this autocmd on preview fixes the issue

alex-courtis:
Still blank after preview. The floating window is definitely still being closed. This can be validated via `:echo win_getid()` before and after the preview.

The next call to `set_current_win_no_autocmd` is the one actually causing the WinLeave event to be fired. This could be moved into an else, something like
```lua
  if mode == "preview" then
    -- ignore "WinLeave" autocmd on preview
    -- because the registered "WinLeave"
    -- will kill the floating window immediately
    set_current_win_no_autocmd(api.nvim_get_current_win(), "WinLeave")
  else
    set_current_win_no_autocmd(target_winid)
  end
```

Unfortunately there is another WinLeave event fired after `open_in_new_window`.

Almo7aya:
I was calling `set_current_win_no_autocmd ` with the wrong window id, I adjust it now, I can't replicate the blank window after preview

alex-courtis:
> I was calling `set_current_win_no_autocmd ` with the wrong window id, I adjust it now, I can't replicate the blank window after preview

Nice work. I'm out of time today, I'll take it for a drive tomorrow.

lua/nvim-tree/view.lua
```diff
@@ -335,8 +335,10 @@ end
 
 function M.focus(winnr, open_if_closed)
   local wnr = winnr or M.get_winnr()
-
-  if a.nvim_win_get_tabpage(wnr or 0) ~= a.nvim_win_get_tabpage(0) then
+  if not a.nvim_win_is_valid(wnr) then
```
alex-courtis, line 338: I don't think we need this any more.

I tested successfully without these changes.

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -209,7 +209,15 @@ local function open_in_new_window(filename, mode, win_ids)
     cmd = string.format("edit %s", fname)
   end
 
-  set_current_win_no_autocmd(target_winid)
+  if mode == "preview" then
```
alex-courtis, line 212: This is correct, however is only desirable for floating windows. There may be (future) WinLeave events in the non-floating case that we still wish to receive.

- [ ] Please change to `if mode == "preview" and view float is set then`

alex-courtis (changes requested):
Tested OK with open-file changes only:

- [ ] float enabled
- [ ] float disabled
- [ ] filter visible
- [ ] quit_on_focus_lost disabled

alex-courtis:
Thank you so much for this fix!

PR #1651, opened 2022-10-15
chore(#1606): always use file system watchers

### LINKED ISSUE #1606: Finalise filesystem_watchers
Remove any codepaths that are conditional on either of these options.

Silently remove these options.

Update: there are genuine use cases for disabled watchers. Resolve remaining issues such as large directories.

Tasks:
- [x] add `filesystem_watchers.ignore`
- [x] tear down watcher on failure, notifying the user once

alex-courtis:
resolves #1606 

Watchers are stable and always used.

Any remaining reloading is user invoked or necessary for actions such as filtering.

I would greatly appreciate some testing on this one. I'll soak for a week.


lua/nvim-tree/actions/init.lua
```diff
@@ -393,7 +393,6 @@ function M.setup(opts)
   require("nvim-tree.actions.node.open-file").setup(opts)
   require("nvim-tree.actions.root.change-dir").setup(opts)
   require("nvim-tree.actions.fs.create-file").setup(opts)
-  require("nvim-tree.actions.fs.rename-file").setup(opts)
```
gegoune, line 396: Was that removed accidentally?

lua/nvim-tree/marks/init.lua
```diff
@@ -70,7 +70,6 @@ end
 
 function M.setup(opts)
   vim.fn.sign_define(SIGN_NAME, { text = opts.renderer.icons.glyphs.bookmark, texthl = "NvimTreeBookmark" })
-  require("nvim-tree.marks.bulk-move").setup(opts)
```
gegoune, line 73: Was that removed accidentally?

gegoune (commented):
LGTM except for two questions. Quick testing revealed not problems.

lua/nvim-tree/marks/init.lua
alex-courtis: That setup function was removed as it was no longer needed.

lua/nvim-tree/actions/init.lua
alex-courtis: That setup function was removed as it was no longer needed.

lua/nvim-tree/legacy.lua
```diff
@@ -313,6 +321,9 @@ function M.migrate_legacy_options(opts)
   -- silently move
   refactored(opts)
 
+  -- silently clean unnecessary options that are now default behaviour
+  unnecessary(opts)
```
alex-courtis, line 325: Contemplating a mechanism of informing users, however users do get upset when messages are displayed on startup.

Perhaps some sort of one-shot mechanism or a way to disable messages.

gegoune: Perhaps something like `vim.notify_once()` could be used? I really dont mind those messages either, would rather know about changes I need to make than have broken setup.

alex-courtis: Today I Learned `notify_once`.

I was hopingi it persists across sessions, however sadly not.

#1667

alex-courtis:
Watchers disabled is desirable.

Issue #1005, opened 2022-02-19
Auto-close not work


tin-pham:
OS: PopOS
Nvim Version: v0.7.0-dev
Nvim-tree-version: I just update this plugin, but i don't know how to check version of this
Bug: auto-close not work
Step to reproduce: Close the buffer while the tree open
![image](https://user-images.githubusercontent.com/20679090/154821320-5bb5890e-3399-441b-9101-61b3fae7a303.png)

![image](https://user-images.githubusercontent.com/20679090/154821325-f99e34a5-fe2a-48d1-a188-85541ceed0f5.png)

vlstv:
doesn't work for me as well

bryant-the-coder:
> OS: PopOS
> Nvim Version: v0.7.0-dev
> Nvim-tree-version: I just update this plugin, but i don't know how to check version of this
> Bug: auto-close not work
> Step to reproduce: Close the buffer while the tree open
> ![image](https://user-images.githubusercontent.com/20679090/154821320-5bb5890e-3399-441b-9101-61b3fae7a303.png)
> 
> 
> ![image](https://user-images.githubusercontent.com/20679090/154821325-f99e34a5-fe2a-48d1-a188-85541ceed0f5.png)
> 

font? it doesn't work for me as well

kyazdani42:
cannot reproduce, can you provide your configuration and the exact steps to reproduce ?

tin-pham:
Open like this
![image](https://user-images.githubusercontent.com/20679090/154866176-398af319-e6de-4ffc-aa4f-2849d34c74a1.png)
While the cursor in the buffer, press :wq or :q to quit 

The tree still live
![image](https://user-images.githubusercontent.com/20679090/154866195-98acb33b-e971-4811-b4b4-c9df3179f825.png)


I don't know but other people have the same issue as me, I think that the new update problem, when I update this issue start occur

kyazdani42:
cannot reproduce, can you share you complete nvim-tree configuration ?
Also if you are using packer, you can navigate to `~/.local/share/nvim/site/pack/packer/start/nvim-tree.lua` and do a `git log`

tin-pham:
> cannot reproduce, can you share you complete nvim-tree configuration ? Also if you are using packer, you can navigate to `~/.local/share/nvim/site/pack/packer/start/nvim-tree.lua` and do a `git log`

https://github.com/tin-pham/NeoVimLua

[...]

PR #1662, opened 2022-10-16
feat: Sync nvim-tree across tabs

rwblokzijl:
This PR syncs nvim tree across tabs if `open_on_tab` is set.

Previously: 
 - On tab change this would open nvim-tree
 - When having closed nvim-tree it would still open nvim tree on tab change

Now:

 - when having closed nvim-tree it will sync this on tab change
 - if nvim-tree is closed (using `:quit`) this is also synced across tabs
 - if nvim-tree is the final window in a tab, the tab is closed (if this is the last tab, vim is closed)

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -192,6 +192,7 @@ end
 
 function M.reset_highlight()
   colors.setup()
+  view.reset_winhl()
   renderer.render_hl(view.get_bufnr())
 end
 
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -436,6 +436,13 @@ function M.is_root_folder_visible(cwd)
   return cwd ~= "/" and not M.View.hide_root_folder
 end
 
+-- used on ColorScheme event
+function M.reset_winhl()
+  if M.get_winnr() then
+    vim.wo[M.get_winnr()].winhl = M.View.winopts.winhl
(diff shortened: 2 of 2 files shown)


alex-courtis:
> * if nvim-tree is the final window in a tab, the tab is closed (if this is the last tab, vim is closed)

This was present in the past however was removed as it is unavoidably problematic. https://github.com/nvim-tree/nvim-tree.lua/issues/1005#issuecomment-1079824801

You may implement this functionality yourself: https://github.com/nvim-tree/nvim-tree.lua#tips--tricks

lua/nvim-tree.lua
```diff
@@ -88,6 +88,7 @@ end
 
 function M.open(cwd)
   cwd = cwd ~= "" and cwd or nil
+  vim.g.nvim_tree_tab_open = 1
```
alex-courtis, line 91: We do not use global variables to track nvim-tree state.

Please test via something like:
* iterate windows on tab
* test `utils.is_nvim_tree_buf`

alex-courtis (changes requested):
I've not tested nor fully reviewed yet.

Please:

- [ ] remove auto close
- [ ] avoid global variable

rwblokzijl:
Closed by force push i think. See https://github.com/nvim-tree/nvim-tree.lua/pull/1698

Issue #1607, opened 2022-09-26
Remove and Clarify netrw Options


alex-courtis:
Eager disable of netrw is working: https://github.com/kyazdani42/nvim-tree.lua/commit/fb8735e96cecf004fbefb086ce85371d003c5129

Remove the netrw options (with a message pointing to a wiki page) and associated codepaths.

See #1270

alex-courtis:
https://github.com/nvim-tree/nvim-tree.lua/wiki/Tips#netrw-keeps-popping-up and :help should be enough.

Don't remove netrw functionality, as users do sometimes use other netrw functionality  #1751

PR #1665, opened 2022-10-17
fix(#1664): respect hijack_directories.enable on startup when not open_on_setup

### LINKED ISSUE #1664: hijack_directories Not Respected When Not `open_on_setup`
c5536db #1618 broke this behaviour.

See https://github.com/nvim-tree/nvim-tree.lua/discussions/1660#discussioncomment-3889215

alex-courtis:
resolves #1664 

#1618 removed this functionality; this PR restores it.

Behaviour is unreliable when netrw is enabled, usually opening netrw. See #1607


alex-courtis:

Issue #1679, opened 2022-10-23
Show node names longer than window width


przepompownia:
**Is this a question?**
No

**Is your feature request related to a problem? Please describe.**
Node names are shown truncated if they exceed window.

**Describe the solution you'd like**
Add an autocommand that displays the full name of node under cursor. 
Looking at the solution for `fern.vim` I wrote a module in lua: https://github.com/przepompownia/vim-arctgx/blob/master/lua/arctgx/lineHover.lua It can be simply used without any modifications in nvim-tree:
```vim
lua vim.api.nvim_create_autocmd({'FileType'}, {pattern = 'NvimTree', callback = require('arctgx.lineHover').enableForWindow})
```
but IMHO this or similar function could be a basic option in nvim-tree. 

**Describe alternatives you've considered**
See also the example from [nvim-dap-ui](https://github.com/rcarriga/nvim-dap-ui/blob/master/lua/dapui/render/line_hover.lua) where the author of the plugin added copying extmarks from the original line. 

**Additional context**
On the below images the hover window shows only raw text so highlights are lost:
![image](https://user-images.githubusercontent.com/11404453/197401679-6a28c536-f883-4c97-9632-ae45db2791a6.png)
![image](https://user-images.githubusercontent.com/11404453/197401686-12bb784c-fcc3-4794-a094-23c9b2044122.png)

Edit: I see that `file-popup.lua` does similar job but for file details, not the full name.

ssiyad:
Couldn't you just use `view = { adaptive_size = true }`?

przepompownia:
> Couldn't you just use `view = { adaptive_size = true }`?

It can be usable for me in some cases, especially with closing NvimTree window right after opening some buffer from a node, but I don't like consuming too much space by the window if it stays open.

jorgsk:
I just came here to make this feature request! I also sometimes have limited screen real estate and it would be great to view full file names.

alex-courtis:
Does `:help nvim-tree.renderer.full_name` provide the functionality you desire?

przepompownia:
It's exactly that! Thanks @alex-courtis

pleshevskiy:
@alex-courtis
It doesn't work if only one character goes outside the width of the tree. 
![image](https://user-images.githubusercontent.com/7839004/197862671-8eb31601-07c2-4f49-ac53-6c3137845ca5.png)

but in other cases it works great!
[...]

PR #1688, opened 2022-10-26
fix: reduce signcolumn/foldcolumn from window width

ssiyad:
https://github.com/nvim-tree/nvim-tree.lua/issues/1679#issuecomment-1291030546

### DIFF
--- a/lua/nvim-tree/renderer/components/full-name.lua
+++ b/lua/nvim-tree/renderer/components/full-name.lua
@@ -12,6 +12,28 @@ local function hide(win)
   end
 end
 
+-- reduce signcolumn/foldcolumn from window width
+local function effective_win_width()
+  local win_width = fn.winwidth(0)
+
+  -- return zero if the window cannot be found
+  local win_id = fn.win_getid()
+
+  if win_id == 0 then
+    return win_width
+  end
+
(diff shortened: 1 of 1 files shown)


pleshevskiy:
I've checked and it works! Thanks!

alex-courtis (changes requested):
Looking good:

20221029_135419 441x477

Please be more defensive:

- [ ] `win_getid` may return 0 which will result in an incorrect result from `getwininfo`
- [ ] `getwininfo` may return an empty list

`textoff` is OK as it is "guaranteed" to be a number for a successful result.

lua/nvim-tree/renderer/components/full-name.lua
```diff
@@ -32,14 +54,17 @@ local function show()
     return
   end
 
-  local width = fn.strdisplaywidth(fn.substitute(line, "[^[:print:]]*$", "", "g"))
-  if width < fn.winwidth(0) then
+  local text_width = fn.strdisplaywidth(fn.substitute(line, "[^[:print:]]*$", "", "g"))
+  local win_width = effective_win_width()
```
alex-courtis, line 58: nice

alex-courtis (approved):
Works beautifully, thank you for the contribution.

PR #1698, opened 2022-10-28
Sync closing of nvim-tree across tabs

rwblokzijl:

PR #1701, opened 2022-10-30
chore: remove vim.* "requires"

alex-courtis:
See #1698

Inconsistencies between vim.api and nvim-tree api are becoming a problem. Explicitly specify `vim.api` instead of "require"ing it.

Also remove `vim.fn` and `vim.loop` "require"s.


alex-courtis:
I'm going to soak this branch for a few days of regular usage.

gegoune (approved):

alex-courtis:
Soaking more following master merge.

PR #1705, opened 2022-10-31
feat(watcher): add filesystem_watchers.ignore_dirs

alex-courtis:
#1606

Add `filesystem_watchers.ignore_dirs`.

Tested:
* exact path
* trailing directory
* project directory
* escaped specials like `\\.`

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -308,6 +308,7 @@ Subsequent calls to setup will replace the previous configuration.
       filesystem_watchers = {
         enable = true,
         debounce_delay = 50,
+        ignore_dirs = {},
       },
       git = {
         enable = true,
@@ -600,6 +601,12 @@ performance.
     *nvim-tree.filesystem_watchers.debounce_delay*
     Idle milliseconds between filesystem change and action.
       Type: `number`, Default: `50` (ms)
+
+    *nvim-tree.filesystem_watchers.ignore_dirs*
+    List of vim regex for absolute directory paths that will not be watched.
+    Backslashes must be escaped e.g. `"my-project/\\.build$"`. See |string-match|.
+    Useful when path is not in `.gitignore` or git integration is disabled.
+      Type: {string}, Default: `{}`
 
 *nvim-tree.on_attach*
 Function ran when creating the nvim-tree buffer.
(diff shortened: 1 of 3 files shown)


alex-courtis:
@yehuohan

I would be very grateful if you could test:

```
cd /path/to/nvim-tree.lua
git pull
git checkout 1606-watchers-ignore-dir
```

* disable git integration
* no ignore (high CPU)
* ignore_dirs = { "CMakeFiles" }` (normal)

yehuohan:
@alex-courtis Sorry for late test until this weekend.

With `ignore_dirs` parameter, there is no large memory consumption anymore to build with cmake.
While there is still permission fail log like below. Is it works expectedly?

```
[2022-11-05 16:24:51] [watcher] event_cb for E:\nvim\neovim\build\CMakeFiles\CMakeTmp\CMakeFiles fail : EPERM
```

Here is the complete nvim-tree log: 

[nvim-tree-ignore_dirs.zip](https://github.com/nvim-tree/nvim-tree.lua/files/9943378/nvim-tree-ignore_dirs.zip)

And the test configuration for nvim:

```vim
let $DotVimDir=resolve(expand('<sfile>:p:h'))
set rtp^=$DotVimDir
call plug#begin($DotVimDir.'/bundle')
call plug#end()

lua << EOF
    local tcb = require('nvim-tree.config').nvim_tree_callback
    require('nvim-tree').setup{
        auto_reload_on_write = false,
        log = {
          enable = true,
          truncate = true,
          types = {
            all = false,
            config = false,
            copy_paste = false,
            dev = false,
            diagnostics = false,
            git = true,
            profile = true,
            watcher = true,
          },
        },
        filesystem_watchers = { enable = true },
        diagnostics = { enable = false },
        git = { enable = false },
        ignore_dirs = { "CMakeFiles" },
    }
EOF
```

alex-courtis:
Many thanks for all your testing @yehuohan

> By the way, does this PR has the feature `tear down watcher on failure, notifying the user once`?
> It seemed the watcher wasn't torn down and nvim got into non-response at the final from large memory consumption when testing without the `ignore_dirs` parameter.

There was a bug with the message on teardown: #1726 which prevented the actual teardown.

Perhaps that was the cause of the unnecessary memory usage. I would be very grateful if you pull latest and try again.

yehuohan:
@alex-courtis 

I tested with the latest nvim-tree.lua without `ignore_dirs`.  There is no large memory consumption and the permission message is displayed. Great!

worked

alex-courtis:
> I tested with the latest nvim-tree.lua without `ignore_dirs`. There is no large memory consumption and the permission message is displayed. Great!

Fantastic! Many thanks.

PR #1706, opened 2022-10-31
feat: command to clear the clipboard

wyrid:
There does not appear to be any way to empty the clipboard other than going back to each file and removing it. This will add the command to empty it all at once (currently left unbound by default).


alex-courtis (approved):
Works, many thanks for your contribution.

PR #1726, opened 2022-11-05
utils may not have a notify field

europeanplaice:
Hi. `utils` may not have a field `notify` and this PR removes it.


alex-courtis:
Many thanks for finding and fixing this @europeanplaice

europeanplaice:
I'm happy to help.

Issue #591, opened 2021-08-26
Session conflict when setting buffer name


ahmedkhalf:
Reproduction:

1. Save the current session when nvim-tree is open. `:mksession!`
2. Load session using `nvim -S Session.vim`

You now have the following error.
```
"NvimTree" [Not edited] --No lines in buffer--                                                                                                                                                
Error detected while processing CursorMoved Autocommands for "NvimTree":                                                                                                                      
E5108: Error executing lua ...te/pack/packer/start/nvim-tree.lua/lua/nvim-tree/lib.lua:102: Expected 1 argument                                                                               
Press ENTER or type command to continue
```
Furthermore, nvim_tree is not open, but it should be open since session was saved while open.
Furthermore, entries are doubled as in: #590

kyazdani42:
I've addressed the first issue in 0a178d17bc7a1c56cc. The second one is quite tricky because the session tries to load a buffer named NvimTree although it's already loaded on startup.
The setup refactoring should be able to fix this issue by deferring the setup call `after` the session file has loaded. I haven't found a way to detect if vim is opening through a vim session though.

ahmedkhalf:
> haven't found a way to detect if vim is opening through a vim session though.

perhaps a hackish way would be to check for `-S` in `v:argv`

kyazdani42:
should be fixed in 1.1.5 (4cc2abe)

ahmedkhalf:
I believe this was wrongly closed @kyazdani42 .

As this issue is still present:
https://github.com/kyazdani42/nvim-tree.lua/issues/591#issuecomment-906984527

kyazdani42:
indeed, sorry though this was fixed :smile:

ahmedkhalf:
> indeed, sorry thought this was fixed smile

After the recent updates, the second issue is now fixed, but now a third issue is born :(. When I save the session with nvim-tree open, it does not restore my vim with nvim-tree, it is instead closed.

kyazdani42:
because the setup/entry is not working with sessions at all. I have to use 
```
						*SessionLoad-variable*
While the session file is loading the SessionLoad global variable is set to 1.
Plugins can use this to postpone some work until the SessionLoadPost event is
triggered.
```
I will probably take a look into it soon

ahmedkhalf:
thank you! but sorry at the same time :(, issue #2 has been reintroduced when having nvim-tree open when dashboard.nvim is open.

[...]

PR #1734, opened 2022-11-09
proper session restoration

zwhitchcox:
Currently, if you open a session after vim is already opened (not with `vim -S ./path/to/session.vim`), the old `nvim-tree` buffers are not restored. This ensures that `nvim-tree` is reloaded whenever a session is loaded.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -78,6 +78,11 @@ function M.wipe_rogue_buffer()
   local was_wiped = false
   for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
     if not matches_bufnr(bufnr) and utils.is_nvim_tree_buf(bufnr) then
+
+      -- if restored from session, values might be incorrect
+      for option, value in pairs(BUFFER_OPTIONS) do
+        vim.bo[bufnr][option] = value
+      end
       pcall(vim.api.nvim_buf_delete, bufnr, { force = true })
       was_wiped = true
     end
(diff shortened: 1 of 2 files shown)


zwhitchcox:
So, I know this code is wrong, but I'm not sure how, so I was hoping I could get some feedback on it 😅

zwhitchcox:
cc @kyazdani42  (from https://github.com/nvim-tree/nvim-tree.lua/issues/591#issuecomment-912728592)

kyazdani42 (commented):
looks good to me. cc @alex-courtis

zwhitchcox:
Just made another commit...I was getting an error to save the `NvimTree_*` file before exiting before this fix.

lua/nvim-tree/view.lua
```diff
@@ -74,16 +74,24 @@ local function matches_bufnr(bufnr)
...
-local function wipe_rogue_buffer()
+function M.wipe_rogue_buffer()
+  local was_wiped = false
   for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
     if not matches_bufnr(bufnr) and utils.is_nvim_tree_buf(bufnr) then
+
+      -- if restored from session, values might be incorrect
+      for option, value in pairs(BUFFER_OPTIONS) do
+        vim.bo[bufnr][option] = value
+      end
```
kyazdani42, line 85: this is outside the scope of this function though.
Also if the buffer is deleted, i'm not sure what's the point of resetting its options.

zwhitchcox: I was having to do this due to a misconfiguration in my setup. I'm having another issue now. Going to close this PR for now until I fix it.

Issue #1740, opened 2022-11-10
Error while running `:NvimTreeCollapseKeepBuffers`


alfa07:
### Description

I am getting error below while trying execute: `NvimTreeCollapseKeepBuffers`

```
Error executing Lua callback: ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:272: attempt to get length of local 'nodes' (a nil value)
stack traceback:
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:272: in function '_get_nodes_number'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:285: in function 'build'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:265: in function '_build_line'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:289: in function 'build'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:265: in function '_build_line'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:289: in function 'build'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:265: in function '_build_line'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:289: in function 'build'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:265: in function '_build_line'
        ...r/start/nvim-tree.lua/lua/nvim-tree/renderer/builder.lua:289: in function 'build'
        ...cker/start/nvim-tree.lua/lua/nvim-tree/renderer/init.lua:78: in function 'draw'
        ...ua/lua/nvim-tree/actions/tree-modifiers/collapse-all.lua:41: in function 'fn'
        ...m/site/pack/packer/start/nvim-tree.lua/lua/nvim-tree.lua:320: in function <...m/site/pack/packer/start/nvim-tree.lua/lua/nvim-tree.lua:319>
Press ENTER or type command to continue
```

### Neovim version

```text
❯ nvim --version
NVIM v0.8.0
Build type: Release
LuaJIT 2.1.0-beta3
Compiled by brew@HMBRW-A-001-M1-004.local

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
[...]

PR #1741, opened 2022-11-11
fix(#1740): set open state only on folders

alfa07:
Fixes [issue 1740](https://github.com/nvim-tree/nvim-tree.lua/issues/1740)

If we set `node.open` indiscriminately (e.g. for files) then we fail later on this line when we try to draw the tree.

First we will recurse here on a file:
https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/renderer/builder.lua#L263-L267

And then `tree.node` will be `nil` here:
https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/renderer/builder.lua#L285

### DIFF
--- a/lua/nvim-tree/actions/tree-modifiers/collapse-all.lua
+++ b/lua/nvim-tree/actions/tree-modifiers/collapse-all.lua
@@ -31,7 +31,9 @@ function M.fn(keep_buffers)
   Iterator.builder(core.get_explorer().nodes)
     :hidden()
     :applier(function(node)
-      node.open = keep_buffers == true and matches(node.absolute_path)
+      if node.nodes ~= nil then
+        node.open = keep_buffers == true and matches(node.absolute_path)
+      end
     end)
     :recursor(function(n)
       return n.nodes


alfa07:
@alex-courtis @kyazdani42 do you think this is the right way to fix the issue? Another option is to fix the check of `node.open` in [here](https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/renderer/builder.lua#L263) to `is_folder and node.open`.

alex-courtis:
> @alex-courtis @kyazdani42 do you think this is the right way to fix the issue? Another option is to fix the check of `node.open` in [here](https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/renderer/builder.lua#L263) to `is_folder and node.open`.

This fixes the root cause, which is the incorrect setting of `node.open`.

Let's not change the builder itself as there may be other assumptions around open and nodes.

alex-courtis:
Fix tested OK however it did reveal an issue with the matcher: no-keep folders that begin with the path of a keep folder are left open:

```sh
:; mkdir 1740
:; cd 1740
:; mkdir dir dir1 dir2 dir3
:; echo 0 > dir/file0
:; echo 1 > dir1/file1
:; echo 2 > dir2/file2
```

[asciicast](https://asciinema.org/a/537046)

If you could als resolve that in this PR I would be most grateful @alfa07

alex-courtis (approved):
Tested OK

Issue #1743, opened 2022-11-12
Remove “/..” from root directory if “restrict_above_cwd” is enabled


alex-courtis:
### Discussed in https://github.com/nvim-tree/nvim-tree.lua/discussions/1735

<div type='discussions-op-text'>

<sup>Originally posted by **davilera** November  9, 2022</sup>
Currently, nvim-tree’s default behavior is to append `/..` to the root directory. This makes sense because the user can't browse one directory up. However, if `actions.change_dir.restrict_above_cwd` is enabled, this is no longer possible.

Therefore, I'd recommend appending `/..` in `local function format_root_name` if, and only if, `actions.change_dir.restrict_above_cwd` is **not** enabled. IMO, this would result in a better UX.</div>

lmartinpierrat:
I would argue that you should be able to remove the /.. entirely even if you don't have restrict_above_cwd

I have no interest in having this added to the header, which I would rather have just the name of the directory I'm in, which  is always my project name.

I spent quite a while trying to find the right filename-modifiers to get rid of the /.. which lead me to look into the code and saw it was hardcoded unfortunately.

alex-courtis:
> I would argue that you should be able to remove the /.. entirely even if you don't have restrict_above_cwd

I'm inclined to agree. It's not adding much value and removing it should not break anything.

The `cd` action will be a bit misleading as it will change to the root's parent rather than the root, however displaying the `/..` is also misleading as that is not the root's actual directory.

davilera:
> I would argue that you should be able to remove the /.. entirely even if you don't have restrict_above_cwd

Should it have its own setting at least?

alex-courtis:
> > I would argue that you should be able to remove the /.. entirely even if you don't have restrict_above_cwd
> 
> Should it have its own setting at least?

Usually yes, as this change would be a break to existing functionality.

However, as this is a more cosmetic change we could do it without a setting.

Issue #876, opened 2021-12-25
Only show tail directory in current working directory marker


petobens:
**OS**
Arch 5.15.11-arch2-1

**Neovim version**
NVIM v0.7.0-dev+764-g3e81c1f9b5

**Nvim-tree version**
master

**Describe the bug/feature**
I would like to only display the  current (tail) folder in the marker at the top of window indicating the current directory. The rationale for this is that I use a narrow window width (40cols) and when moving around directories the innermost folder gets truncated. For instance in the GIF you can see that only `~/git-repos/private/dotfiles/nvim` gets displayed even when the inner directory changes (to `plugin-config` or `lua`)

![Peek 2021-12-25 15-04-root](https://user-images.githubusercontent.com/2583971/147390991-02d06c8c-0c0c-4da5-97ef-6374deccad08.gif)

kyazdani42:
`:help g:nvim_tree_root_folder_modifier`

petobens:
Missed that one. A follow-up (two part) question/feature request then:
1. How I can display both the parent and current dir? I tried playing with `:h:t` without success
2. Is is it possible to play with the root folder format? I would like to display it with a leading symbol/icon as in: `-> root_folder_info`

kyazdani42:
1. not sure how, play with the modifiers a bit `:help filename-modifiers`
2. i don't think so, it uses fnamemodify under the hood

PR #1746, opened 2022-11-15
Fix/1743

davilera:
This PR addresses issue #1743. It adds a new renderer setting named `add_root_updir` to let the user customize whether `/..` should be appended (or not) to the root directory.


gegoune:
Perhaps instead of adding another narrow setting we could allow users to set `header` (or similar) to either string or function? Some data could be passed to callback. This would allow greater flexibility without adding various toggles.

@alex-courtis What do you think? There are most likely other parameters that could benefit from more generalised solution.

petobens:
> Perhaps instead of adding another narrow setting we could allow users to set header (or similar) to either string or function? Some data could be passed to callback. This would allow greater flexibility without adding various toggles.

Big plus 1 to this. I asked for something similar here https://github.com/nvim-tree/nvim-tree.lua/issues/876#issuecomment-1030631944

alex-courtis:
> Perhaps instead of adding another narrow setting we could allow users to set `header` (or similar) to either string or function? Some data could be passed to callback. This would allow greater flexibility without adding various toggles.
> 
> @alex-courtis What do you think? There are most likely other parameters that could benefit from more generalised solution.

Yes. Such flexibility via functions desirable. We do this with `sort_by` `view.width` `view.float.open_win_config` `on_attach`

Let's continue with this pattern.

Proposed help:

```
    *nvim-tree.renderer.root_folder_modifier*
    In what format to show root folder. `"/.."` will be appended.
    See |filename-modifiers| for available options.
      Type: `string`, Default: `":~"`

    *nvim-tree.renderer.root_folder_name*
    Function to set the root folder's displayed name.
    When not set, the root folder title is |nvim-tree.renderer.root_folder_modifier|
      Type: `function`, Default: `nil`

      Function is passed the absolute path of the root folder and should return a string.
      e.g. >
        root_folder_title = function(path)
          return "root directory: " .. vim.fn.fnamemodify(path)
        end
<
```

alex-courtis:
Are you comfortable implementing this @davilera ? It looks like most of the parts are there.

gegoune:
Would we need both? Can't `nvim-tree.renderer.root_folder` take `function` or `string`? If `string` we can run it through `fnamemodify`, if `function` it will be up to user to return what they want. Can still default to `":~"`

alex-courtis:
> Would we need both? Can't `nvim-tree.renderer.root_folder` take `function` or `string`? If `string` we can run it through `fnamemodify`, if `function` it will be up to user to return what they want. Can still default to `":~"`

Of course! That is a much better experience.

We could then "silently move": `root_folder_modifier` -> `root_folder`.

gegoune:
Although maybe `root_folder_label` is a better name?

davilera:
> Are you comfortable implementing this @davilera ?

I'll be happy to give it a shot! I'll work on the PR later on today :-)

> Although maybe `root_folder_label` is a better name?

Ok! Just to be clear, in my PR I'll get rid of `root_folder_modifier` (as well as the one I proposed: `add_root_updir`) and I'll add `root_folder`.

davilera:
There you go, guys. By the way, I renamed the setting to `root_folder_label` as proposed by @gegoune... that's what I wanted to say in my previous message, but I made a mistake.

If someone could please review the commit, I'd really appreciate it.

gegoune:
LGTM! Thank you so much!

@alex-courtis I have added warning for this breaking change, also moved `opts.focus_empty_on_setup = nil` to conditional statement, I am not sure if it wasn't in there by mistake or not. Do you mind having a look?

alex-courtis:
> @alex-courtis I have added warning for this breaking change, also moved `opts.focus_empty_on_setup = nil` to conditional statement, I am not sure if it wasn't in there by mistake or not. Do you mind having a look?

Good catch.

I moved the change from removed to refactored, so that there's no warning, if that's OK with you.

lua/nvim-tree/renderer/builder.lua
```diff
@@ -294,14 +294,17 @@ function Builder:build(tree)
   return self
 end
 
-local function format_root_name(root_cwd, modifier)
-  local base_root = utils.path_remove_trailing(vim.fn.fnamemodify(root_cwd, modifier))
-  return utils.path_join { base_root, ".." }
```
alex-courtis, line 299: We have lost the `/..` at the end of the default which might upset some users.

Proposal: append it to the default: `root_folder_label = ":~:s?$?/..?"`

That adds some complexity. I'd appreciate some test/feedback.

gegoune:
That makes more sense, thanks.

alex-courtis (changes requested):
Tested function OK:
* empty string
* populated string

Fail:
* return number
* return table

alex-courtis:
> That makes more sense, thanks.

I'm not sure what to do about those ones... they'll hang around forever...

lua/nvim-tree/renderer/builder.lua
gegoune: Doesn't matter how complex default is. Most users won't ever change it.

gegoune:
@davilera Can you enforce return type of that function to string? It should be done after that I think.

davilera:
@gegoune, I just pushed a couple of commits to address this:

> enforce return type of that function to string

and this:

> We have lost the `/..` at the end of the default which might upset some users. Proposal: append it to the default

I hope it works as expected!

lua/nvim-tree/renderer/builder.lua
```diff
@@ -294,14 +296,21 @@ function Builder:build(tree)
...
+local function format_root_name(root_cwd, root_label)
+  if type(root_label) == "function" then
+    label = root_label(root_cwd)
+    if type(label) == "string" then
```
gegoune, line 302: Thanks, looks good, I am only thinking whether we should notify user that returned value must be string and until it is default value will be used instead. `vim.notify_once()`?

gegoune:
Looks good to me, thank you! I have left one more comment above and will leave it to @alex-courtis to give it one last look before merging it. Awesome work!

lua/nvim-tree/renderer/builder.lua
davilera: I just realized I forgot to add make the `label` variable `local` in line 301. Should I commit the fix?

alex-courtis:
> An idea for most likely follow up PR. How often will that function run? It could be used to display other things such as summary of git or diagnostic statuses. Eg: `myroot/proj [E6,W3,H9]`.

Yes. We'll need to experiment.

I did cowboy an API addition `get_nodes` https://github.com/nvim-tree/nvim-tree.lua/commit/e38e061710c593d9b88c8ebb2c48f2842c89dc31 however that's minimal and missing things like diagnostics.

lua/nvim-tree/renderer/builder.lua
```diff
@@ -294,14 +296,21 @@ function Builder:build(tree)
...
-local function format_root_name(root_cwd, modifier)
-  local base_root = utils.path_remove_trailing(vim.fn.fnamemodify(root_cwd, modifier))
-  return utils.path_join { base_root, ".." }
+local function format_root_name(root_cwd, root_label)
+  if type(root_label) == "function" then
+    local label = root_label(root_cwd)
+    if type(label) == "string" then
+      return label
+    else
+      root_label = DEFAULT_ROOT_FOLDER_LABEL
```
alex-courtis, line 305: nice

alex-courtis:
Thank you all... I'm using this now.

PR #1778, opened 2022-11-27
feat(diagnostics): only show diagnostic and git icon on closed folder

### LINKED ISSUE #1781: show_on_dirs option for lowest visible only
The git and diagnostics views both have a setting `show_on_dirs` which makes parent folders display the same icon as the file. This is extremely helpful when they're collapsed I.e. if there is an error in `/first/second/third.lua` and the tree is collapsed, there will be an error icon next to `first`, indicating that there's a problem in one of its children. However, if both `first` and `second` are expanded, the icon is displayed on both parents as well as the file. As a visual cue, it exaggerates the number of "things", particularly when the tree expands far enough that they're not all visible at the same time.

It would be really nice to have an option to display the icon only on the lowest visible leaf. E.g.
* if `first` is collapsed, display it on `first`
* if first is expanded but `second` is collapsed, display it on `second`
* if all are expanded, display it on `third.lua`

chomosuke:
fixes #1781 

On certain projects there might be deeply nested folders (e.g. java projects). Showing diagnostics on all the parent folders feels very cluttered. It also makes it harder to get a idea of how many files have diagnostics at a glance.

This PR adds the `show_on_open_dirs` which when set to false will only show diagnostics on closed directory.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -290,6 +290,7 @@ Subsequent calls to setup will replace the previous configuration.
       diagnostics = {
         enable = false,
         show_on_dirs = false,
+        show_on_open_dirs = true,
         debounce_delay = 50,
         severity = {
           min = vim.diagnostic.severity.HINT,
@@ -554,6 +555,11 @@ Show LSP and COC diagnostics in the signcolumn
     *nvim-tree.diagnostics.show_on_dirs*
     Show diagnostic icons on parent directories.
       Type: `boolean`, Default: `false`
+
+    *nvim-tree.diagnostics.show_on_open_dirs*
+    Show diagnostics icons on directories that are open.
+    Only relevant when `diagnostics.show_on_dirs` is `true`.
+      Type: `boolean`, Default: `true`
 
     *nvim-tree.diagnostics.icons*
     Icons for diagnostic severity.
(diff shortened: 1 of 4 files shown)


alex-courtis (changes requested):
This is valuable functionality, thank you for raising. I will greatly appreciate this functionality.

We currently have feature parity between git and diagnostics via `show_on_dirs` albeit with different defaults.

Let's continue parity: please add `show_on_open_dirs` for git as well.

Update: see #1781 which requested this functionality.

chomosuke:
Thanks! I'm glad you liked it.

I actually attempted to replicate the functionality for git too before opening this PR but had a little bit of trouble getting it to update without constantly hitting refresh. Hence why I didn't replicate it. 

After some digging I've managed to get it to update without refreshing now. This PR should be ready to review again @alex-courtis. :)

gegoune:
This is indeed very valuable addition. Thank you!

alex-courtis (approved):
Tested combinations `enable` `show_on_dirs` `show_on_open_dirs`:
* git
* lsp

alex-courtis:
It seems that my testing was insufficient: #1786 
I'd be grateful if you looked at that one @chomosuke 

I'm using this functionality for git and diagnostics... my view is so much clearer... many thanks!

PR #1782, opened 2022-11-28
[WIP] allow passing a custom function as a window picker

davidsierradz:
Hi! According to [[WIP] allow using the winbar instead of statusbar as a window picker by davidsierradz · Pull Request #1767 · nvim-tree/nvim-tree.lua](https://github.com/nvim-tree/nvim-tree.lua/pull/1767), this replaces that PR.

With something like:

```lua
...
    open_file = {
      window_picker = {
        enable = true,
        custom_function = function()
          return require('window-picker').pick_window({
            selection_chars = '12345',
            use_winbar = 'always',
            filter_rules = {
              bo = {
                filetype = { 'notify', 'packer', 'qf', 'diff', 'fugitive', 'fugitiveblame' },
                buftype = { 'nofile', 'terminal', 'help' },
              },
            },
          })
        end,
      },
  },
...
```

Let me know what you think, thanks!


alex-courtis (changes requested):
Wow! That is clean!

Please add :help and update as per https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md I'm happy to assist with the :help

Please provide an example so that I may test this out and we can add to the :help

davidsierradz:
> Please add :help and update as per https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md I'm happy to assist with the :help
>
> Please provide an example so that I may test this out and we can add to the :help

Thanks, will do!

davidsierradz:
This function uses `vim.ui.select` to pick a window (not the best example, but works):

```lua
function()
  local window_id
  local window_list = {}
  for _, win in ipairs(vim.api.nvim_list_wins()) do
    if vim.api.nvim_win_get_config(win).relative == '' then
      table.insert(window_list, win)
    end
  end
  vim.ui.select(window_list, {
    prompt = 'Select window:',
  }, function(choice)
    window_id = choice
  end)
  return window_id
end
```

This one uses [s1n7ax/nvim-window-picker](https://github.com/s1n7ax/nvim-window-picker):

```lua
function()
  return require('window-picker').pick_window()
end
```

alex-courtis:
> This function uses `vim.ui.select` to pick a window (not the best example, but works):

Fantastic. I'm out of time and will review next weekend.

lua/nvim-tree.lua
```diff
@@ -621,6 +621,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
       resize_window = true,
       window_picker = {
         enable = true,
+        picker = "default",
```
alex-courtis, line 624: I've changed this to a string/function to leave the door open for other methods in the future.

alex-courtis (changes requested):
This works beautifully, and allows full control for the user i.e. ability to pick the nvim-tree window itself.

Please:

- [x] add to :help (I'll help out with this)
- [x] use the `s1n7ax/nvim-window-picker` example in the help

alex-courtis (approved):

Your complete example https://github.com/nvim-tree/nvim-tree.lua/pull/1782#issuecomment-1330039322 would be a fantastic addition to the [wiki Recipes](https://github.com/nvim-tree/nvim-tree.lua/wiki/Recipes)

PR #1783, opened 2022-11-28
fixed: git rename not showing up for the renamed file

chomosuke:
After renaming a file, rename icon only shows up on parent directory. No icon shows up on the file itself. This PR fixes that.

Edit: It now also deal with escape characters in quoted filenames (e.g. "a \\"quoted\\" file name") correctly

### DIFF
--- a/lua/nvim-tree/git/runner.lua
+++ b/lua/nvim-tree/git/runner.lua
@@ -10,13 +10,23 @@ function Runner:_parse_status_output(line)
   if utils.str_find(status, "R") then
     -- rename have format of "from -> to"
     -- we just what the "to" part
-    local _, j = path:find " -> "
-    -- Note: The edge case where " -> " is a part of a file name is not addressed
-    path = path:sub(j + 1, -1)
+    if path:sub(1, 1) == '"' then
+      -- incase " -> " is a part of the file name
+      local _, j = path:find '" %-> "'
+      path = path:sub(j, -1)
+    else
+      local _, j = path:find " %-> "
+      path = path:sub(j + 1, -1)
+    end
   end
 
   -- removing `"` when git is returning special file status containing spaces
-  path = path:gsub('^"', ""):gsub('"$', "")
+  if path:sub(1, 1) == '"' then
+    path = path:sub(2, -2)
+    path = path:gsub('\\"', '"')
+    path = path:gsub("\\\\", "\\")
+  end
+
   -- replacing slashes if on windows
   if vim.fn.has "win32" == 1 then
     path = path:gsub("/", "\\")
@@ -131,6 +141,7 @@ function Runner:_wait()
   local function is_done()
     return self.rc ~= nil
   end
+
   while not vim.wait(30, is_done) do
   end
 end
--- a/lua/nvim-tree/git/runner.lua
+++ b/lua/nvim-tree/git/runner.lua
@@ -4,29 +4,7 @@ local utils = require "nvim-tree.utils"
 local Runner = {}
 Runner.__index = Runner
 
-function Runner:_parse_status_output(line)
-  local status = line:sub(1, 2)
-  local path = line:sub(4, -2)
-  if utils.str_find(status, "R") then
-    -- rename have format of "from -> to"
-    -- we just what the "to" part
-    if path:sub(1, 1) == '"' then
-      -- incase " -> " is a part of the file name
-      local _, j = path:find('" -> "', nil, true)
-      path = path:sub(j, -1)
-    else
-      local _, j = path:find(" -> ", nil, true)
-      path = path:sub(j + 1, -1)
-    end
-  end
-
-  -- removing `"` when git is returning special file status containing spaces
-  if path:sub(1, 1) == '"' then
-    path = path:sub(2, -2)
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/git/runner.lua
```diff
@@ -6,8 +6,27 @@ Runner.__index = Runner
 
 function Runner:_parse_status_output(line)
   local status = line:sub(1, 2)
+  local path = line:sub(4, -2)
+  if utils.str_find(status, "R") then
+    -- rename have format of "from -> to"
+    -- we just what the "to" part
+    if path:sub(1, 1) == '"' then
+      -- incase " -> " is a part of the file name
```
gegoune, line 14: That feels awfully user case specific.

chomosuke: What do you mean by use case specific? If we see the first character as `"` then the file names are definitely enclosed in quotes no? And if that's the case there might be ` -> ` present in the file name so the only way to reliably tell which one is ` -> `  is to search for `" -> "`?

gegoune: Apologies, I took it out of context completely. Of course this is a rename marker here. Sorry to have wasted your time.

alex-courtis (changes requested):
We can actually avoid `->` detection and quoting/escaping issues completely by using the `-z` machine readable option. See `git help status`.

The only complication is that we must split the results on NUL instead of LF in `Runner:_run_git_job`. `Runner:_log_raw_output` will likely need to be changed to log the split results instead of raw results.

Running with `log.types.git` is useful.

lua/nvim-tree/git/runner.lua
```diff
@@ -6,8 +6,27 @@ Runner.__index = Runner
...
+      local _, j = path:find '" %-> "'
+      path = path:sub(j, -1)
```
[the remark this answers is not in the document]
chomosuke: Hmmm, how would you do that? I couldn't find anything on Google that let me use `string.sub` to match substring.

chomosuke:
@alex-courtis `-z` was the first thing that I tried, but I couldn't get it to work because `utils.str_find` would throw something like `using binary blob as string`. I then tried to replace `utils.str_find` with `string.find` but got `string expected, got no value`. I also tried to replace all `\0` with `\n` with `gsub("\0", "\n")` but got identical result as `gsub("", "\n")` i.e. every character was on its own line. Do you know how can I split on `\0`?

lua/nvim-tree/git/runner.lua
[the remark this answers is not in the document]
alex-courtis: Oh bugger... I got that wrong. It should be string.find

I've recently used it for a path here: https://github.com/nvim-tree/nvim-tree.lua/blob/02b9376cdf3a3abc51bab9e274ccd7caa40ab94e/lua/nvim-tree/explorer/filters.lua#L30

alex-courtis:
> @alex-courtis `-z` was the first thing that I tried, but I couldn't get it to work because `utils.str_find` would throw something like `using binary blob as string`. I then tried to replace `utils.str_find` with `string.find` but got `string expected, got no value`. I also tried to replace all `\0` with `\n` with `gsub("\0", "\n")` but got identical result as `gsub("", "\n")` i.e. every character was on its own line. Do you know how can I split on `\0`?

Ouch, that is tricky.

It looks like we can use the magic `"+%z"` to handle zero byte string pattern matching: https://www.lua.org/pil/21.2.2.html

Time box that at 20 mins: if you can't get that working, I'm happy to go with the `" -> "` implementation  using `string.find`s.

I'm out of time; I'll review again next weekend.

chomosuke:
@alex-courtis Done, this is much cleaner!

lua/nvim-tree/git/runner.lua
```diff
@@ -44,7 +52,7 @@ function Runner:_getopts(stdout_handle, stderr_handle)
   local untracked = self.list_untracked and "-u" or nil
   local ignored = (self.list_untracked and self.list_ignored) and "--ignored=matching" or "--ignored=no"
   return {
-    args = { "--no-optional-locks", "status", "--porcelain=v1", ignored, untracked, self.path },
+    args = { "--no-optional-locks", "status", "--porcelain=v1", "-z", ignored, untracked, self.path },
```
alex-courtis, line 55: Fantastic! Thank you.

lua/nvim-tree/git/runner.lua
```diff
@@ -4,26 +4,34 @@ local utils = require "nvim-tree.utils"
...
     for line in prev:gmatch "[^\n]*\n" do
-      i = i + self:_parse_status_output(line)
+      if skip_next_line then
+        skip_next_line = false
+      else
+        local status = line:sub(1, 2)
+        local path = line:sub(4, -2)
+        if utils.str_find(status, "R") then
+          -- skip next line if it is a rename entry
+          skip_next_line = true
```
alex-courtis, line 30: Interesting that -z swaps from and to... It works.

alex-courtis (approved):
Tested OK.

Many thanks for the change and going the extra mile!

Issue #1155, opened 2022-04-11
Rename before extension


sarmong:
Most of the time, when you rename a file, you don't want to change the extension, you just want to change the name of a file.

Therefore, it will be rather useful to add a new action, that will put the cursor directly before the file extension.

File browsers such as ranger or lf can be configured to have this feature (they also have the command prompt at the bottom as vim)

kyazdani42:
go for it :)

sarmong:
Any tips? I don't really know where to start, that's why opened an issue;)

kyazdani42:
go to the rename-file.lua and make a bit of an abstraction to rename either from filename or something else i suppose :)

sarmong:
Hm, I don't think it is possible with the current implementation of `vim.ui.input` as theit allows to provide default value and callback function, and there is really nowhere to move the cursor

kyazdani42:
hum this look weird indeed, could you maybe just remove the whole name and saving the extension, call input without the name, then pasting the extension ? Not sure how to achieve that exactly but i believe it should be possible.

ianhomer:
I've created a PR for this, feedback welcome. I've created new action "relative_rename" that changes the base name of the file - without changing directory or extension. This is what I tend to want to do from nvimtree most of the time.  I use another extension to move files.

Issue #1753, opened 2022-11-19
Improve API Documentation


alex-courtis:
A description of each API function with its parameters and returns will be useful.

Events are a good template.

Move nvim-tree-api section to before nvim-tree-events.

Segregate access / mutate API functions.

alex-courtis:
All done.

PR #1791, opened 2022-11-30
relative rename action (#1155)

ianhomer:
https://github.com/nvim-tree/nvim-tree.lua/issues/1155

### DIFF
--- a/lua/nvim-tree/actions/dispatch.lua
+++ b/lua/nvim-tree/actions/dispatch.lua
@@ -25,6 +25,7 @@ local Actions = {
   trash = require("nvim-tree.actions.fs.trash").fn,
   remove = require("nvim-tree.actions.fs.remove-file").fn,
   rename = require("nvim-tree.actions.fs.rename-file").fn(false),
+  relative_rename = require("nvim-tree.actions.fs.rename-file").fn(false, true),
 
   -- Movements in tree
   close_node = require("nvim-tree.actions.moves.parent").fn(true),
--- a/lua/nvim-tree/actions/fs/rename-file.lua
+++ b/lua/nvim-tree/actions/fs/rename-file.lua
@@ -24,7 +24,7 @@ function M.rename(node, to)
   events._dispatch_node_renamed(node.absolute_path, to)
 end
 
-function M.fn(with_sub)
+function M.fn(with_sub, relative_rename)
   return function(node)
     node = lib.get_last_group_node(node)
     if node.name == ".." then
@@ -32,9 +32,31 @@ function M.fn(with_sub)
     end
 
     local namelen = node.name:len()
-    local abs_path = with_sub and node.absolute_path:sub(0, namelen * -1 - 1) or node.absolute_path
+    local abs_directory = node.absolute_path:sub(0, namelen * -1 - 1)
+    local default_path, prepend, append
+    if relative_rename then
+      local filename = node.absolute_path:sub(abs_directory:len() + 1)
+      local extension_index = filename:find("%.") or -1
+      if extension_index > -1 then
+        default_path = filename:sub(0, extension_index -1)
+        append = filename:sub(extension_index)
+      else
+        default_path = filename
+        append = ""
+      end
+      prepend = abs_directory
+      print(filename .. "|" .. (extension_index or "na") .. "|" .. prepend .. "|" ..append)
+    else
+      prepend = ""
+      append = ""
+      if with_sub then
+        default_path = abs_directory
+      else
+        default_path = node.absolute_path
+      end
+    end
 
-    local input_opts = { prompt = "Rename to ", default = abs_path, completion = "file" }
+    local input_opts = { prompt = "Rename to ", default = default_path, completion = "file" }
 
     vim.ui.input(input_opts, function(new_file_path)
       utils.clear_prompt()
@@ -42,7 +64,7 @@ function M.fn(with_sub)
         return
       end
 
-      M.rename(node, new_file_path)
+      M.rename(node, prepend .. new_file_path .. append)
       if M.enable_reload then
         require("nvim-tree.actions.reloaders.reloaders").reload_explorer()
       end


alex-courtis:
> Not sure whether this should be bound to key by default. Would like to add to documentation somewhere, although other actions such as full_rename are only mentioned in docs via the default key mapping.

That's OK... proper descriptions are incoming #1579

lua/nvim-tree/actions/dispatch.lua
```diff
@@ -25,6 +25,7 @@ local Actions = {
   trash = require("nvim-tree.actions.fs.trash").fn,
   remove = require("nvim-tree.actions.fs.remove-file").fn,
   rename = require("nvim-tree.actions.fs.rename-file").fn(false),
+  relative_rename = require("nvim-tree.actions.fs.rename-file").fn(false, true),
```
alex-courtis, line 28: Please add a default action to `lua/nvim-tree/actions/init.lua` and run `scripts/update-help.sh`

`<c-r>` is taken; what do you think about `"e"` ?

That will address the documentation. "in-vim" documentation is coming #1579

alex-courtis, line 28: Please add API: `lua/nvim-tree/api.lua`

alex-courtis (changes requested):
Works beautifully. I will enjoy using this feature.

You will be the owner of this feature, responsible for resolving any immediate issues that arise.

alex-courtis:
@sarmong your testing of this feature would be gratefully appreciated.

lua/nvim-tree/actions/dispatch.lua
ianhomer: Thanks - and yes, I'm good with e - it's good to have the different renames available at different key strokes.

I had to change `gsed` instead of `sed` in the `update-help.sh` for it to run (I'm currently on a mac, and mac sed behaves different to gnu sed), but I didn't commit that script change.

ianhomer: ok - I've renamed action to rename_relative and used that name in the API since that naming matches the "rename_sub" pattern.

tested OK with `:lua require"nvim-tree.api".fs.rename_relative()`

alex-courtis:
Tested OK:
* API
* mapping

alex-courtis (changes requested):
I'm thinking again about the name `relative_rename`, which isn't quite a full explanation and base name isn't quite right.

Looking at vimlike naming for inspiration it maps to `fnamemodify(bufname, ":t:r")`. That's "tail root". We have two changes going on here. The root is arguably the most significant.

If we were building from scratch today I imagine something like this:
"r" rename  `":t"` - using tail is different, but arguably better UX
"e" rename_root `":t:r"`
"<c-r>" rename_full `":p"`

Thinking about fnamemodify now makes me realise that we could use that instead of manually calculating the name. Apologies, I should have realised this earlier.

Proposal

- [x] change rename_relative to rename_root - Alex
- [ ] use fnamemodify in `rename-file.lua`.fn - Ian

Maybe later

Change rename to use the tail; this is a break of existing functionality and needs further thought. Raised #1803

ianhomer:
Thx for the feedback, I'm traveling for the next few days, but will action this when I get back. No urgency to get it in, would prefer to get it right than rush it in.

sarmong:
Hey, thanks for making this feature!

1. If you have directory with dots in its name then it will trim out part after the last dot.
2. Imho. `rename_root` is also a bit confusing, because root oftentimes refers to the root directory of the project. [I found that](https://stackoverflow.com/questions/13362022/what-do-you-call-a-file-name-without-the-extension) in most programing languages file name without extension is called either `stem` or `basename`. I've never heard or used `stem` in this context, but it is a shorter word, whereas `basename` is much more common and understandable.

gegoune (changes requested):
Thanks for doing it.

Once again we are adding specialised functions rather than providing building blocks.

Imagine having a function that receives path for buffer under cursor - whatever is returned will be a candidate for renaming.

Input: `/my/awesom.file.c`, returning: `awesome`, would set that as a rename candidate.

We could then easily provide some builtins built on that, and adding more would be as simple as that:

```lua
action = actions.rename(path) return vim.fn.fnamemodify(path, ':t') end,
```

We could also make it accept string which would be a modifier for `fnamemodify` like so:
```lua
action = ':t',
```

Function would allow users greater flexibility. Like renaming files to include part of directory name, or some other stuff.


`full_rename` could then be refactored to just return the same `path` for backwards compatibility.

Are you willing to go this direction, @ianhomer? What do you think @alex-courtis? I think this would go a long way.

ianhomer:
I've done a minor change to use fnamemodify instead of custom string searching ... however I've not gone as far a few of the other suggestions here. 

Regarding the `/my/|.file.c` suggestion,  for me I don't find that behaviour as convenient as just seeing the basename, since my primary use case for renaming a file is a slight tweak of the current name, perhaps correcting a spelling mistake. If the original name has gone from the default value in the input box, then my rename is now harder.   I also found the full absolute path in the input box a mild annoyance since I had more I could change than I wanted to.  Perhaps that's another optional configuration - some people may prefer one vs the other.

I do like the generic function suggestion, and using vim fnamemodify modifiers. Not sure if there is a convenient vim function to rename a part of a filename based on the modifiers provided, but it wouldn't be hard to write that logic if needs be.

I'll be offline for the next few days, but happy to pick up any loose ends towards the end of the week if needs be.

ianhomer:
> 2\. I've never heard or used `stem` in this context, but it is a shorter word, whereas `basename` is much more common and understandable.

"stem" is the unambiguous term for this (in C/Python world at least) - agree that it's not commonly known and in that sense perhaps not great semantics,  "basename" normally includes the extension.  Having read that stackoverflow post I'd personally vote for "stem" since it's short, sweet and well defined in C and Python.

gegoune:
> Regarding the `/my/|.file.c` suggestion, for me I don't find that behaviour as convenient as just seeing the basename, since my primary use case for renaming a file is a slight tweak of the current name, perhaps correcting a spelling mistake. If the original name has gone from the default value in the input box, then my rename is now harder. I also found the full absolute path in the input box a mild annoyance since I had more I could change than I wanted to. Perhaps that's another optional configuration - some people may prefer one vs the other.

That was just an example how new functionality could be used.

Your explanation of intentions and expectations only solidifies my earlier assumptions - there are so many use cases that there is no way nvim-tree can satisfy them all. Therefore providing an easy framework in form of `actions.rename()` seems like a way to go.

alex-courtis:
> I've updated so that we have a rename API that takes arguments. Only argument it currently supports are the filename modifiers ":p", ":t" and ":t:r". Couldn't see a way to generically support all modifiers, however getting the API contract to where we want is key. Now this first argument is the file modifier, not the path to replace.

This works now and can be customised via `action_cb`:
```lua
          { 
            key = "<C-P>",
            action = "tr",
            action_cb = function()
              require("nvim-tree.actions.fs.rename-file").fn()(":t")
            end
          },
```
We just need to add API...

It will be a _lot_ more usable when we get to `on_attach` https://github.com/nvim-tree/nvim-tree.lua/pull/1579

alex-courtis:
> Users will see this difference and for some it might not be desirable. I wonder if we need to support the previous "full_rename" behaviour with another action with users could configure in their mapping if they wished.

I'm happy; I was being to cautious. It's a break, but a reasonable one. #1803

alex-courtis:
Summary (please correct me):

Done:
* rename basename
* rename full path -> filename
* rename modifiers

Outstanding:
* input dialogue hacking etc.
* full user control of renaming

It seems that we can merge this then deal with outstanding. Continuing discussion in #1803

@gegoune are you happy with the addition of `rename_basename` as a default action?

doc/nvim-tree-lua.txt
```diff
@@ -1196,8 +1196,10 @@ exists.
     - create
     - remove
     - trash
+    - rename_node `(node: table, modifier?: string vim.fn.fnamemodify argument)`
```
alex-courtis, line 1199: I took the liberty of adding this new pattern. Will be improved later: #1753

lua/nvim-tree/actions/fs/rename-file.lua
```diff
@@ -24,25 +30,54 @@ function M.rename(node, to)
   events._dispatch_node_renamed(node.absolute_path, to)
 end
 
-function M.fn(with_sub)
-  return function(node)
+function M.fn(default_modifier)
+  default_modifier = default_modifier or ":t"
+
+  return function(node, modifier)
```
alex-courtis, line 36: I removed capability for the first argument to be a modifier after adding the API, as I couldn't see a use case.

I may be missing something...

alex-courtis (approved):

alex-courtis:
You work here is greatly appreciated @ianhomer 

This raised all sorts of good things well outside of the scope of your change. Thank you for pushing this one all this way.

alex-courtis (approved):
Tested OK: `r`, `e`, `<c-r>` custom:
* file with no dot
* file with many dots
* rename remove / add trailing dot
 
`e` (`file.`) -> `file.` fails to rename, however that is quite reasonable as it is not clear what should happen.

Thanks for resolving the bugs I introduced @ianhomer

Vinni-Cedraz:
Can I write my issue here or is it better to open an issue? instead of renaming the file it is opening it on a new tab. Here is my setup:

```
local status_ok, nvim_tree = pcall(require, "nvim-tree")
if not status_ok then
	return
end

local config_status_ok, nvim_tree_config = pcall(require, "nvim-tree.config")
if not config_status_ok then
	return
end

local tree_cb = nvim_tree_config.nvim_tree_callback

nvim_tree.setup({
	update_focused_file = {
		enable = true,
		update_cwd = true,
	},

	renderer = {
		root_folder_modifier = ":t",
		icons = {
			glyphs = {
				default = "",
				symlink = "",
				folder = {
					arrow_open = "",
					arrow_closed = "",
					default = "",
					open = "",
					empty = "",
					empty_open = "",
					symlink = "",
					symlink_open = "",
				},
				git = {
					unstaged = "",
					staged = "S",
					unmerged = " ",
					renamed = "➜ ",
					untracked = "U",
					deleted = " ",
					ignored = "◌",
				},
			},
		},
	},
	diagnostics = {
		enable = true,
		show_on_dirs = true,
		icons = {
			hint = " ",
			info = " ",
			warning = " ",
			error = " ",
		},
	},
	view = {
		width = 30,
		side = "left",
		mappings = {
			list = {
				{ key = { "l", "<CR>", "o" }, cb = tree_cb("edit") },
				{ key = "h", cb = tree_cb("close_node") },
				{ key = "v", cb = tree_cb("vsplit") },
				{ key = "r", action = "rename_relative" },
			},
		},
	},
})
```

gegoune:
New issue please.

PR #1792, opened 2022-11-30
fix: Implicit current buf on centralize selection

thedaviddelta:
The recent implementation change to center the selected file produces a very odd error on my setup.
If I have `nvim-tree` open (for example, when opening an empty project) and then open a file using any other mean (e.g. Telescope)  a `Invalid buffer id: 1` error pops into my neovim.
Tweaking the code I've discovered the cause: the `centralize_selection` autocmd that runs when closing the tree tries to reads the `bufnr` of Telescope and then schedules the `zz` call, running after some milliseconds when the Telescope buffer is already killed.
By implicitly using the current buffer (by using `0` as a param) instead of explicitly retrieving it solves this problem. It'd probably also work if the retrieving is moved inside the `schedule` function, but haven't tried as I don't think it's better to do so.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -407,9 +407,8 @@ local function setup_autocommands(opts)
     create_nvim_tree_autocmd("BufEnter", {
       pattern = "NvimTree_*",
       callback = function()
-        local bufnr = vim.api.nvim_get_current_buf()
         vim.schedule(function()
-          vim.api.nvim_buf_call(bufnr, function()
+          vim.api.nvim_buf_call(0, function()
             vim.cmd [[norm! zz]]
           end)
         end)


gegoune:
`0` is probably good here but maybe using buffer number from event object will be more robust?

thedaviddelta:
Changed. I think it's also working correctly this way.

lua/nvim-tree.lua
```diff
@@ -407,8 +407,8 @@ local function setup_autocommands(opts)
     create_nvim_tree_autocmd("BufEnter", {
       pattern = "NvimTree_*",
       callback = function()
-        local bufnr = vim.api.nvim_get_current_buf()
```
gegoune, line 410: That can be removed entirely.

lua/nvim-tree.lua
```diff
@@ -407,8 +407,8 @@ local function setup_autocommands(opts)
...
         vim.schedule(function()
+          local bufnr = vim.api.nvim_get_current_buf()
```
gegoune, line 411: No need to fetch that. It's already available in callback by accepting `event` argument.

lua/nvim-tree.lua
```diff
@@ -407,8 +407,8 @@ local function setup_autocommands(opts)
...
           vim.api.nvim_buf_call(bufnr, function()
```
gegoune, line 412:
```suggestion
          vim.api.nvim_buf_call(event.buf, function()
```

lua/nvim-tree.lua
```diff
@@ -407,8 +407,8 @@ local function setup_autocommands(opts)
     create_nvim_tree_autocmd("BufEnter", {
       pattern = "NvimTree_*",
       callback = function()
```
gegoune, line 409:
```suggestion
      callback = function(event)
```

gegoune (changes requested):

gegoune:
Thanks, that's not exactly what I had in mind. Apologies for not being clear. I have requested changes with more detail.

thedaviddelta:
imagen
This produces the same problem, as the event is populated before the schedule starts.
We should just use `0`, or simply query the id **inside** the schedule callback.

gegoune:
Didn't test it properly then. Schedule it then. Maybe Alex will have better solution.

thedaviddelta:
Should I revert to the explicit or the implicit implementation?

gegoune:
If `0` in scheduled context works then it seems like the easiest solution avoiding extra call to api.

alex-courtis:
Gentle nudge for re-review @gegoune

gegoune (approved):

PR #1784, opened 2022-11-28
feat(view): add filters.git_clean, filters.no_buffer

alex-courtis:
fixes #1692

```help
    *nvim-tree.filters.git_clean*
    Do not show files with no git status. This will show ignored files when
    |nvim-tree.git.ignore| is set, as they are effectively dirty.
    Toggle via the `toggle_git_clean` action, default mapping `C`.
      Type: `boolean`, Default: `false`

    *nvim-tree.filters.no_buffer*
    Do not show files that have no listed buffer.
    Toggle via the `toggle_no_buffer` action, default mapping `B`.
      Type: `boolean`, Default: `false`
```

no_buffer's matching is a bit fuzzy and may miss some updates on buffer delete/wipe however it is not worth the performance cost of making it perfectly accurate.

- [ ] ~~`filters.diagnostics_ok`~~

Raised #1785 which may be completed separately.

PR #1809, opened 2022-12-06
Improve git icons on directories

chomosuke:
After adding option git.show_on_open_dirs, I've discovered some undesirable behavior:
- ignored directory's ignored status going away after opening it.
- having no way of knowing a file a deleted once the directory it's in is opened.

This PR fixes the above two short coming and also made parent directories show all the git status icons from their children.

The ordering of the git icons in `lua/nvim-tree/renderer/components/git.lua:10` are made to preserve the original ordering of some of the multi-icons status. I'd be happy to change it if anyone have different preferences :).


alex-courtis:
> The ordering of the git icons in `lua/nvim-tree/renderer/components/git.lua:10` are made to preserve the original ordering of some of the multi-icons status. I'd be happy to change it if anyone have different preferences :).

I think we can remove that. The order of statuses to icons retrieved by `git.get_icons` is Good Enough and easy to change following its normalisation.

alex-courtis (changes requested):
I've tested a variety of cases, however it's not practical to test all cases.

Please:

- [ ] use natural order for statuses: https://github.com/nvim-tree/nvim-tree.lua/pull/1809#issuecomment-1345438281
- [ ] `common.lua` -> `node.lua`
- [ ] use `git_statuses` `git_status` consistently in all of `node.lua`
- [ ] merge master and retest: #1784 introduced some changes to git statuses

Thank you so much: this refactor and consolidation is a great improvement to the codebase.

chomosuke:
> > The ordering of the git icons in `lua/nvim-tree/renderer/components/git.lua:10` are made to preserve the original ordering of some of the multi-icons status. I'd be happy to change it if anyone have different preferences :).
> 
> I think we can remove that. The order of statuses to icons retrieved by `git.get_icons` is Good Enough and easy to change following its normalisation.

Hmmm are you sure? Imo it looks quite weird having folders with the same combination of changes being different, for example:

image

Here both marks and renderer have some files changed and some new files. But it looks like they have different git status if that make sense.

Though tbh even with their icons sorted they'd still have different highlight. Maybe what we need is to sort the statuses?

alex-courtis:
> Here both marks and renderer have some files changed and some new files, but it looks like they have different git status.
> 
> Though tbh even with their icons sorted they'd still have different highlight. Maybe what we need is to sort the statuses?

Whoops... I didn't notice that it was a maplike table. Options:

1. revert 23f6276ef76ccb12e8c9e02550d33525c469154a
2. change it to an array-like with `staged` etc. as a table member and use `utils.key_by`

I like 1.

lua/nvim-tree/renderer/components/git.lua
```diff
@@ -60,19 +60,19 @@ local function warn_status(git_status)
 end
 
 local function get_icons_(node)
-  local git_statuses = explorer_common.get_git_status(node)
-  if git_statuses == nil then
+  local git_status = explorer_common.get_git_status(node)
```
alex-courtis: I was actually thinking we could use `git_statuses` everwhere as it is a much better variable name, similar to `iconss`.

However... looking through the whole codebase we are using `git_status` _everywhere_ so I guess that change is outside of the scope of this PR.

alex-courtis (changes requested):
Please undo my bad idea: https://github.com/nvim-tree/nvim-tree.lua/pull/1809#issuecomment-1354132930

lua/nvim-tree/renderer/components/git.lua
chomosuke: Ahhh sorry, I misunderstood. Perhaps I'll open a PR right after this one specifically for git_statuses?

alex-courtis: If you're keen I would be most grateful! There are a lot of usages...

alex-courtis (approved):

alex-courtis:

PR #1814, opened 2022-12-08
Enable cursor line

rishabhjain9191:
nvim tree relies on user's cursorline option to enable/disble the cursor line within nvim-tree buffer. So, In case, user has not set the cursorline, same won't be applied to nvim-tree which will give bad experience while navigating the tree.
Hence, Have set cursorline to true for nvim buffer.


alex-courtis:

PR #1817, opened 2022-12-09
feat: dispatch Event.NodeRenamed on cut-paste.

antosha417:
Hey! 
Really small pr on event dispatching. 
I expect `NodeRenamed` to be triggered when I do `cut` and then `paste` some nodes.


alex-courtis (approved):
Tested OK:
* paste in other directory
* paste in same directory and rename

Issue #1824, opened 2022-12-12
Don't modify jump list when `edit_in_place`


antosha417:
### Description

I'm trying out vinegar mode and see that jump list behaves not as expect it to. I think we add a position there which becomes invalid after we do `edit_in_place`. And it causes weird behavior. 

### Neovim version

```text
NVIM v0.8.1
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

macOS 12.3

### nvim-tree version

0cd8ac4

### Minimal config

```lua
Default `nvt-min.lua` will do.
```

### Steps to reproduce

1. `touch file{1..3}`
2. `nvim -nu /tmp/nvt-min.lua`
3. `:clearjumps`
4. `:e file1`
5. `:e file2`
6. `:e file3`
7.`:lua require"nvim-tree".open_replacing_current_buffer()`
8. go to file1 and press `<c-e>` to open it.
9. verify that `:jumps` shows 
```
:jumps
 jump line  col file/text
   4     1    0
   3     1    0 file2
   2     1    0 file3
>
```
10. Press `<c-o>` to go previous position in the jumplist. 
11.  I expect to jump to file3. But it jumps to file2, skipping file3.
12. Now if you press `<c-i>` and you will go to the `file3`. Press it again and you will stay in `file3`. Press it again and you will go to `file1`.

### Expected behavior

Jump list behaves like a stack. Without skipping elements.

### Actual behavior

Jump list looks curated.

PR #1825, opened 2022-12-12
fix: Don't modify jumplist when `edit_in_place`.

antosha417:
closes https://github.com/nvim-tree/nvim-tree.lua/issues/1824


alex-courtis (approved):
Thank you for the fix

PR #1827, opened 2022-12-14
Fixed highlight-opened-files doesn't automatically refresh

chomosuke:
This PR refresh opened files highlight automatically whenever a buffer is read or unloaded.


alex-courtis (approved):
Works nicely, thanks for finding this one.

PR #1830, opened 2022-12-16
fix: avoid some unnecessary explorer reloads

D0ot:
it seems that the reload is unnecessary when *renderer.config.highlight_opened_files* is "none". 

The PR is made because I find a problem when using Telescope.
When in a large codebase, Telescope causes a lot of BufUnload events when it close, which makes my nvim stuck for seconds....

### DIFF
--- a/lua/nvim-tree/actions/finders/find-file.lua
+++ b/lua/nvim-tree/actions/finders/find-file.lua
@@ -13,7 +13,7 @@ local running = {}
 ---Find a path in the tree, expand it and focus it
 ---@param fname string full path
 function M.fn(fname)
-  if not core.get_explorer() then
+  if not core.get_explorer() or not view.is_visible() then
     return
   end
 


chomosuke:
Good catch! Tested and works :)

chomosuke (approved):
Good catch! Tested and works as expected :)

lua/nvim-tree.lua
```diff
@@ -363,15 +363,15 @@ local function setup_autocommands(opts)
 
   create_nvim_tree_autocmd("BufReadPost", {
     callback = function()
-      if filters.config.filter_no_buffer or renderer.config.highlight_opened_files then
+      if filters.config.filter_no_buffer or renderer.config.highlight_opened_files ~= "none" then
         reloaders.reload_explorer()
```
chomosuke, line 367:
```suggestion
        utils.debounce("BufReadPost:filter_buffer", 10, function()
          reloaders.reload_explorer()
        end)
```
Actually please change them to something like this. This would make reload only happen once if a bunch of BufReadPost happened in quick succession.

lua/nvim-tree.lua
```diff
@@ -363,15 +363,15 @@ local function setup_autocommands(opts)
...
       end
     end,
   })
 
   create_nvim_tree_autocmd("BufUnload", {
     callback = function(data)
-      if filters.config.filter_no_buffer or renderer.config.highlight_opened_files then
+      if filters.config.filter_no_buffer or renderer.config.highlight_opened_files ~= "none" then
```
chomosuke, line 374:
```suggestion
        utils.debounce("BufUnload:filter_buffer", 10, function()
          reloaders.reload_explorer(nil, data.buf)
        end)
```
Same as above

chomosuke (changes requested):

lua/nvim-tree.lua
alex-courtis: I like this.

`update_focused_file.debounce_delay` default `15` has recently been added, to cope with telescope's extra buffer switching.

Please use that value for now, even thought it is only for focused. We should do a full audit of such instances but Not Today.

lua/nvim-tree.lua
D0ot: changes made, tested and works as expected. please review again :smile:

lua/nvim-tree.lua
```diff
@@ -363,17 +363,17 @@ local function setup_autocommands(opts)
 
   create_nvim_tree_autocmd("BufReadPost", {
     callback = function()
-      if filters.config.filter_no_buffer or renderer.config.highlight_opened_files then
```
chomosuke, line 366: Ohhh? I thought the if statement can still remain though, it'll still bring (possibly negligible) performance improvement ;)

chomosuke (changes requested):

chomosuke:
On a side note, you can run `scripts/setup-hooks.sh` to setup a commit hook and prevent accidentally committing changes without passing `stylua` and `luacheck`. :)

lua/nvim-tree.lua
```diff
@@ -364,15 +364,19 @@ local function setup_autocommands(opts)
   create_nvim_tree_autocmd("BufReadPost", {
     callback = function()
       if filters.config.filter_no_buffer or renderer.config.highlight_opened_files then
```
alex-courtis, line 366: `highlight_opened_files` has a default of "none".

We should test for that too.

alex-courtis (changes requested):

lua/nvim-tree.lua
alex-courtis: Raised #1836

lua/nvim-tree.lua
alex-courtis: 4fc74ca resolves this

alex-courtis (approved):

alex-courtis:
#1843 provides some more avoidance of reloads.

chomosuke (approved):

PR #1838, opened 2022-12-18
fix: clear clipboard error

wilsonross:
Calling `api.fs.clear_clipboard()` produces the following error:

```
E5108: Error executing lua: ...ed/nvim-tree.lua/lua/nvim-tree/actions/fs/copy-paste.lua:135: attempt to index field 'notify' (a nil value)
stack traceback:
        ...ed/nvim-tree.lua/lua/nvim-tree/actions/fs/copy-paste.lua:135: in function 'clear_clipboard'
        /Users/rosswilson/Home/nvim/plugin/tree.lua:34: in function </Users/rosswilson/Home/nvim/plugin/tree.lua:33>
```

Occurs when the clipboard is cleared and attempts to notify. Updated to remove `utils` and instead use `notify` directly

### DIFF
--- a/lua/nvim-tree/actions/fs/copy-paste.lua
+++ b/lua/nvim-tree/actions/fs/copy-paste.lua
@@ -132,7 +132,7 @@ end
 function M.clear_clipboard()
   clipboard.move = {}
   clipboard.copy = {}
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Many thanks for finding and fixing

PR #1843, opened 2022-12-19
fix(#1841): do not refresh on buffer events when not a file buffer

### LINKED ISSUE #1841: Error when generating CMake files.
### Description

I get an error when generating the CMake files through [this plugin](https://github.com/Shatur/neovim-tasks):

image

I have started to use nvim-tree very recently so I don't know much about it, but I think this happens because after generating the CMake files, nvim-tree tries to reload itself and something goes wrong.

I've also opened an issue in the CMake plugin github page and the author told me: "Not sure why, but looks like its nvim-tree tries to reload after adding things to quickfix."

Although I've included my config and the steps to reproduce, I think developers of this plugin might just be able to tell what the problem is by just looking at the error message. The only thing I can think of is maybe set "auto_reload_on_write" to false, but that would also mean that git changes would not show in the tree when I save my files, am I wrong?

### Neovim version

```text
NVIM v0.8.1
```

Windows 11

### nvim-tree version

e0cfbbb

### Minimal config

```lua
require('packer').startup(function(use)
	use 'wbthomason/packer.nvim'
        use {
            'nvim-tree/nvim-tree.lua',
            requires = {
                'nvim-tree/nvim-web-devicons',
            },
            tag = 'nightly'
        }
        use 'Shatur/neovim-tasks'
end)

require('nvim-tree').setup()

local Path = require('plenary.path')
require('tasks').setup({
    default_params = {
    cmake = {
        cmd = 'cmake',
        build_dir = tostring(Path:new('{cwd}', 'build', '{os}-{build_type}')),
        build_type = 'Debug',
        dap_name = 'lldb',
        args = {
        configure = { '-D', 'CMAKE_EXPORT_COMPILE_COMMANDS=1', '-G', 'Ninja' },
    },
},
  },
  save_before_run = true,
  params_file = 'neovim.json',
  quickfix = {
      pos = 'botright',
      height = 12,
  },
  dap_open_command = function() return require('dap').repl.open() end,
})
```

### Steps to reproduce

1. On the root folder (where the `CMakeLists.txt` file is located) of a CMake project, run `:Task start cmake configure` to generate the CMake files.
2. Select the build target (e.i., the executable) by running `:Task set_module_param cmake target`.
3. Finally, build and run the project by running `:Task start cmake run`.
4. You should see the same error you saw on the image posted above.

### Expected behavior

_No response_

### Actual behavior

_No response_

alex-courtis:
fixes #1841

`vim.fn.filereadable` is expensive. Empty `buftype` should be good enough to detect whether it's a buffer that we can show/highlight.

Follows hotfix 4fc74ca32157ecb275e62647fbe9cff0b8e9b9c8

Follows on from #1830

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -362,8 +362,12 @@ local function setup_autocommands(opts)
   end
 
   create_nvim_tree_autocmd("BufReadPost", {
-    callback = function()
-      if filters.config.filter_no_buffer or renderer.config.highlight_opened_files ~= "none" then
+    callback = function(data)
+      -- update opened file buffers
+      if
+        (filters.config.filter_no_buffer or renderer.config.highlight_opened_files ~= "none")
+        and vim.bo[data.buf].buftype == ""
+      then
         reloaders.reload_explorer()
       end
     end,
@@ -371,7 +375,11 @@ local function setup_autocommands(opts)
 
   create_nvim_tree_autocmd("BufUnload", {
     callback = function(data)
-      if filters.config.filter_no_buffer or renderer.config.highlight_opened_files ~= "none" then
+      -- update opened file buffers
+      if
+        (filters.config.filter_no_buffer or renderer.config.highlight_opened_files ~= "none")
+        and vim.bo[data.buf].buftype == ""
+      then
         reloaders.reload_explorer(nil, data.buf)
       end
     end,


chomosuke (approved):
Looks good! Haven't had time to dig into #1841 but this does make sense and looks like something that would fix it.

alex-courtis:
> Haven't had time to dig into https://github.com/nvim-tree/nvim-tree.lua/issues/1841

No worries at all; there's no hurry.

chomosuke:
1 min too late but update: I just tested this PR and it does work!

alex-courtis:
> 1 min too late but update: I just tested this PR and it does work!

Thanks @chomosuke

PR #1854, opened 2022-12-24
Issue 1848 - escaping of '(' on windows

### LINKED ISSUE #1848: Problems with Braces in Pathnames under Windows
### Description

I have a SvelteKit project that I'm developing on a Windows machine.

In SvelteKit, you can define routing groups (invisible parent folders, for lag of a better term) by putting the folder name into braces like so `(publiconly)`

Screenshot 2022-12-22 070222

But there seems to be some sort of escaping issue with the pathnames, since Windows uses `\` instead of `/` as directory separator.
So the `\` before the `(` is removed. I guess to escape the `(`.

Screenshot 2022-12-22 070356

### Neovim version

```text
NVIM v0.9.0-dev-530+gde90a8bfe
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Übersetzt von runneradmin@fv-az368-246

Features: -acl +iconv +tui
See ":help feature-compile"

          System-vimrc-Datei: "$VIM\sysinit.vim"
     Voreinstellung für $VIM: "C:/Program Files (x86)/nvim/share/nvim"

```

Windows 10
** Update ** I should clearify, that I'm NOT running WSL.

### nvim-tree version

'e14c289'

### Minimal config

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

In Windows
- create a folder called `(test)` 
- go into that folder 
- create a new text file called `test.txt`
- open that textfile in windows notepad
- write something
- save the file and close notepad
- now open NVIM
- call `:NvimTreeToggle`
- navigate into the `(test)` folder
- open `test.txt`
- press `CTRL + g` to print the full filename under the status bar.

The content of the opened file will be empty and before the `(` of `(test)` will be a `\` missing.

### Expected behavior

the path should look something like this `parentFolder\(test)\test.txt`

### Actual behavior

the backslash before `(test)` is removed (I susspect some sort of escaping issue), leading to the new name of the file being `parentFolder(test)\test.txt`
Since the folder `parentFolder(test)` does not exist, Nvim opens a new file instead of the choosen one.

DoodlingTurtle:
fixes #1848

Since Windows uses '\' instead of '/'
otherwise vim would throw out all '\' in front of folders that start with a "("

Tested in Nvim-qt, Windows CMD and PowerShell

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -234,9 +234,11 @@ local function open_in_new_window(filename, mode, win_ids)
     end
   end
 
-  local fname = vim.fn.fnameescape(filename)
-  if vim.fn.has("win32") then
-    fname = fname:gsub("%(", "\\(")
+  local fname = vim.fn.fnameescape(utils.canonical_path(filename))
+  fname = vim.loop.fs_realpath(fname);
+
+  if utils.is_windows then
+    fname =  fname:gsub("%(", "\\(")
   end
 
   local cmd
@@ -256,8 +258,7 @@ local function open_in_new_window(filename, mode, win_ids)
   else
     set_current_win_no_autocmd(target_winid, { "BufEnter" })
   end
-  
-  
+
   pcall(vim.cmd, cmd)
   lib.set_target_win()
 end
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -235,6 +235,9 @@ local function open_in_new_window(filename, mode, win_ids)
   end
 
   local fname = vim.fn.fnameescape(filename)
+  if vim.fn.has("win32") then
+    fname = fname:gsub("%(", "\\(")
+  end
 
   local cmd
   if create_new_window then
@@ -253,7 +256,8 @@ local function open_in_new_window(filename, mode, win_ids)
   else
     set_current_win_no_autocmd(target_winid, { "BufEnter" })
   end
-
+  
+  
   pcall(vim.cmd, cmd)
   lib.set_target_win()
 end


gegoune:
We probably should generalise windows support. It seems like different part of code are affected by windows' quirks. Would be good to factor out those path manipulations to separate functions. @alex-courtis What do you think?

alex-courtis:
> We probably should generalise windows support. It seems like different part of code are affected by windows' quirks. Would be good to factor out those path manipulations to separate functions. @alex-courtis What do you think?

Absolutely. OS feature flags have just been extracted to `utils`: https://github.com/nvim-tree/nvim-tree.lua/pull/1868/files#diff-7c68f0fc800988f242ac61f7db0c2167124815cb6b65f1d01a99df9a4bb609f1

`utils.is_windows` is appropriate here.

Let's move this (PowerShell specific) sub to `utils.canonical_path`

alex-courtis (changes requested):
Thanks for raising this. Please:

- [ ] stylua fixes
- [x] attempt to use the generic solution `vim.loop.fs_realpath` as per https://github.com/nvim-tree/nvim-tree.lua/issues/1848#issuecomment-1364495936
- [ ] move to `utils.canonical_path` https://github.com/nvim-tree/nvim-tree.lua/pull/1854#issuecomment-1368148471

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -234,7 +234,12 @@ local function open_in_new_window(filename, mode, win_ids)
     end
   end
 
-  local fname = vim.fn.fnameescape(filename)
+  local fname = vim.fn.fnameescape(utils.canonical_path(filename))
+  fname = vim.loop.fs_realpath(fname);
+
+  if utils.is_windows then
+    fname =  fname:gsub("%(", "\\(")
```
alex-courtis, line 241: Will we need this any more? The idea is for `fs_realpath` to escape these.

Also, we should move `fs_realpath` into `canonical_path`

alex-courtis (changes requested):

DoodlingTurtle:
- `vim.loop.fs_realpath` did not solve the issue. If you print the path, it looks fine, but Vim still opens the wrong thing. 
It is not an issue with the module per se.  
The issue happens after/within the `pcall` function call  on line 262 
An escape of  `\(` seems nessary. 

- I checked for `[` too, but those work fine without escape.

lua/nvim-tree/actions/node/open-file.lua
DoodlingTurtle: - `vim.loop.fs_realpath` did not solve the issue. If you print the path, it looks fine, but Vim still opens the wrong thing. 
It is not an issue with the module per se.  
The issue happens after/within the `pcall` function call  on line 262 
An escape of  `\(` seems nessary. 

- I checked for `[` too, but those work fine without escape.

alex-courtis: That is reasonable. I was naive in hoping that fs_realpath would do the job...

Let's put that gsub inside `canonical_path` behind the windows feature flag. If we discover any other such problem characters we can add them in that location.

gegoune: > The issue happens after/within the pcall function call on line 262
An escape of \( seems nessary.

If you are willing to do some more tests, would you mind replacing that string manipulation there and use

DoodlingTurtle: `nvim_cmd` requires some more rewrites but it delivers the same result as using `pcall`
In the End it seems to be just another way to call native VIM functions.

Using `nvim_parse_cmd` seems to make things worse by opening a 'unknown' file.

```lua

  --if utils.is_windows then
  --  -- fname =  fname:gsub("%(", "\\(")
  --end

  local cmd
  if create_new_window then
    cmd = new_window_side .. ' vsplit'
    --cmd = string.format("%s vsplit %s", new_window_side, fname)
  elseif mode:match "split$" then
    cmd = mode
    --cmd = string.format("%s %s", mode, fname)
  else
    cmd = 'edit'
    --cmd = string.format("edit %s", fname)
  end

  if mode == "preview" and view.View.float.enable then
    -- ignore "WinLeave" autocmd on preview
    -- because the registered "WinLeave"
    -- will kill the floating window immediately
    set_current_win_no_autocmd(target_winid, { "WinLeave", "BufEnter" })
  else
    set_current_win_no_autocmd(target_winid, { "BufEnter" })
  end

  vim.api.nvim_cmd({cmd = cmd, args = {fname}}, {})
  --pcall(vim.cmd, cmd)
  lib.set_target_win()
```

Having a   `(` in the name is nothing unusual/illegal. That is why none of the escape functions react to it.
I don't understand why Vim is handling `\(` that way. What kind of function do the escaped parentheses serve?
Google does not say anything about it. 

could it be a bug in https://github.com/neovim/neovim ?

alex-courtis:
Any progress on this one @DoodlingTurtle ? If it's an upstream issue we will close this and revisit later if necessary.

DoodlingTurtle:
> Any progress on this one @DoodlingTurtle ? If it's an upstream issue we will close this and revisit later if necessary.

I don't see what the problem in line 241 is, that Stylua is complaining about, so I did not touch this any further. As I said before, Lua, especially NVIMs Version of it, is not my strength. Sorry. 😔

alex-courtis:
> > Any progress on this one @DoodlingTurtle ? If it's an upstream issue we will close this and revisit later if necessary.
> 
> I don't see what the problem in line 241 is, that Stylua is complaining about, so I did not touch this any further. As I said before, Lua, especially NVIMs Version of it, is not my strength. Sorry. pensive

There's an unnecessary semicolon at the end of the line, and some extra space around the =

Run `stylua .` locally to see all the details.

alex-courtis:
PR was closed.

PR #1859, opened 2022-12-25
feat(#1814): added cursorline config to DEFAULT_OPTS

youkwhd:
Extends #1814

Currently, the config cursorline is set to `true` by default.

This behaviour can only be changed by a hacky way of listening to an event, as @alex-courtis mentioned that: "The user can change this default if they want via event".

This PR generalizes the configuration to be easier to config via the function `setup()`.


alex-courtis (approved):
Works beautifully, thank you for your contribution.

Apologies for the delay in review.

PR #1821, opened 2022-12-11
feat(event): add will rename node event.

antosha417:
Hey!
Some language servers support [workspace/willRenameFiles](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willRenameFiles) request. It is used in order to automatically fix imports. 
Currently neovim does not support it. So plugins must be used if one wants to use this feature. [Here](https://github.com/neovim/neovim/issues/20784?ysclid=lbgzrup04p974657325) is an open issue in neovim repo.

Since I use `nvim-tree.lua` for all file manipulations I thought it would be cool to add `WillRenameNode` event and do [workspace/willRenameFiles](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willRenameFiles) request in the handler.

PR #1870, opened 2022-12-31
feat: async rename and create file

yioneko:
Split of #1863.

New experimental options: 
```lua
  experimental = {
    async = {
      create_file = false,
      rename_file = false,
    },
  },
```

I select these two operations because their async version is nearly 1to1 copy of the original implementation, and the risk of breaking is bit lower. Also, I think we could turn on `rename_file` by default, and let users choose to enable `create_file` either if they feel well about the new async behavior.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -408,6 +408,12 @@ Subsequent calls to setup will replace the previous configuration.
           watcher = false,
         },
       },
+      experimental = {
+        async = {
+          create_file = false,
+          rename_file = false,
+        }
+      }
     } -- END_DEFAULT_OPTS
 <
 
@@ -1204,6 +1210,20 @@ Configuration for diagnostic logging.
 
         *nvim-tree.log.types.watcher*
         |nvim-tree.filesystem_watchers| processing, verbose.
+          Type: `boolean`, Default: `false`
+
+*nvim-tree.experimental*
+Configuration for experimental features.
+
+    *nvim-tree.experimental.async*
+    Control experimental async behavior.
+
+        *nvim-tree.experimental.async.create_file*
+        Toggle async behavior of create file operation.
+          Type: `boolean`, Default: `false`
+
+        *nvim-tree.experimental.async.rename_file*
+        Toggle async behavior of rename file operation.
           Type: `boolean`, Default: `false`
 
 ==============================================================================
--- a/lua/nvim-tree/actions/fs/create-file.lua
+++ b/lua/nvim-tree/actions/fs/create-file.lua
@@ -97,6 +97,11 @@ local function create_file(new_file_path)
   if not is_error then
     notify.info(new_file_path .. " was properly created")
   end
+  if M.enable_async then
+    async.schedule()
+  end
+  -- synchronously refreshes as we can't wait for the watchers
+  find_file(utils.path_remove_trailing(new_file_path))
 end
 
 local function get_containing_folder(node)
@@ -135,15 +140,12 @@ function M.fn(node, cb)
 
     if M.enable_async then
       async.exec(create_file, new_file_path, function(err)
-        find_file(utils.path_remove_trailing(new_file_path))
         if cb then
           cb(err)
         end
       end)
     else
       create_file(new_file_path)
-      -- synchronously refreshes as we can't wait for the watchers
-      find_file(utils.path_remove_trailing(new_file_path))
     end
   end)
 end
--- a/lua/nvim-tree/actions/fs/create-file.lua
+++ b/lua/nvim-tree/actions/fs/create-file.lua
@@ -5,32 +5,28 @@ local core = require "nvim-tree.core"
 local notify = require "nvim-tree.notify"
 
 local find_file = require("nvim-tree.actions.finders.find-file").fn
+local async = require "nvim-tree.async"
 
 local M = {}
 
+---@async
 local function create_and_notify(file)
-  local ok, fd = pcall(vim.loop.fs_open, file, "w", 420)
-  if not ok then
+  local fd, err
+  if M.enable_async then
+    err, fd = async.call(vim.loop.fs_open, file, "w", 420)
+  else
+    fd, err = vim.loop.fs_open(file, "w", 420)
+  end
+  if err then
     notify.error("Couldn't create file " .. file)
     return
   end
-  vim.loop.fs_close(fd)
-  events._dispatch_file_created(file)
-end
-
-local function create_file(file)
-  if utils.file_exists(file) then
-    local prompt_select = "Overwrite " .. file .. " ?"
-    local prompt_input = prompt_select .. " y/n: "
-    lib.prompt(prompt_input, prompt_select, { "y", "n" }, { "Yes", "No" }, function(item_short)
-      utils.clear_prompt()
-      if item_short == "y" then
-        create_and_notify(file)
-      end
-    end)
+  if M.enable_async then
+    async.call(vim.loop.fs_close, fd)
   else
-    create_and_notify(file)
+    vim.loop.fs_close(fd)
   end
+  events._dispatch_file_created(file)
 end
 
 local function get_num_nodes(iter)
@@ -41,6 +37,68 @@ local function get_num_nodes(iter)
   return i
 end
 
+local function create_file(new_file_path)
+  -- create a folder for each path element if the folder does not exist
+  -- if the answer ends with a /, create a file for the last path element
+  local is_last_path_file = not new_file_path:match(utils.path_separator .. "$")
+  local path_to_create = ""
+  local idx = 0
+
+  local num_nodes = get_num_nodes(utils.path_split(utils.path_remove_trailing(new_file_path)))
+  local is_error = false
+  for path in utils.path_split(new_file_path) do
+    idx = idx + 1
+    local p = utils.path_remove_trailing(path)
+    if #path_to_create == 0 and utils.is_windows then
+      path_to_create = utils.path_join { p, path_to_create }
+    else
+      path_to_create = utils.path_join { path_to_create, p }
+    end
+    if is_last_path_file and idx == num_nodes then
+      if M.enable_async then
+        async.schedule()
+      end
+      if utils.file_exists(new_file_path) then
+        local prompt_select = "Overwrite " .. new_file_path .. " ?"
+        local prompt_input = prompt_select .. " y/n: "
+        if M.enable_async then
+          local item_short = async.call(lib.prompt, prompt_input, prompt_select, { "y", "n" }, { "Yes", "No" })
+          utils.clear_prompt()
+          if item_short == "y" then
+            create_and_notify(new_file_path)
+          end
+        else
+          lib.prompt(prompt_input, prompt_select, { "y", "n" }, { "Yes", "No" }, function(item_short)
+            utils.clear_prompt()
+            if item_short == "y" then
+              create_and_notify(new_file_path)
+            end
+          end)
+        end
+      else
+        create_and_notify(new_file_path)
+      end
+    elseif not utils.file_exists(path_to_create) then
+      local err
+      if M.enable_async then
+        err = async.call(vim.loop.fs_mkdir, path_to_create, 493)
+      else
+        local _
+        _, err = vim.loop.fs_mkdir(path_to_create, 493)
+      end
+      if err then
+        notify.error("Could not create folder " .. path_to_create .. ": " .. err)
+        is_error = true
+        break
+      end
+      events._dispatch_folder_created(new_file_path)
+    end
+  end
+  if not is_error then
+    notify.info(new_file_path .. " was properly created")
+  end
+end
+
 local function get_containing_folder(node)
   if node.nodes ~= nil then
     return utils.path_add_trailing(node.absolute_path)
@@ -49,7 +107,8 @@ local function get_containing_folder(node)
   return node.absolute_path:sub(0, -node_name_size - 1)
 end
 
-function M.fn(node)
+--TODO: once async feature is finalized, use `async.wrap` instead of cb param
+function M.fn(node, cb)
   node = node and lib.get_last_group_node(node)
   if not node or node.name == ".." then
     node = {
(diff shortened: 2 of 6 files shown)


lua/nvim-tree/actions/fs/create-file.lua
```diff
@@ -49,7 +61,62 @@ local function get_containing_folder(node)
...
+  local path_to_create = ""
+  local idx = 0
+
+  local num_nodes = get_num_nodes(utils.path_split(utils.path_remove_trailing(new_file_path)))
+  local is_error = false
+  for path in utils.path_split(new_file_path) do
+    idx = idx + 1
+    local p = utils.path_remove_trailing(path)
+    async.schedule()
+    if #path_to_create == 0 and vim.fn.has "win32" == 1 then
```
gegoune, line 93: Please use newly added (moved, really) platform checks `is_*` from `utils`.

lua/nvim-tree/events.lua
```diff
@@ -18,22 +20,49 @@ M.Event = {
...
+      local success, error = pcall(handler, payload)
+      if not success then
+        notify.error("Handler for event " .. event_name .. " errored. " .. vim.inspect(error))
+      end
     end
   end
+
+  local evt_handlers = get_handlers(event_name)
+
+  safe_call_hanlders(evt_handlers.normal)
```
gegoune, line 56: Typo here and in following lines?

gegoune (changes requested):
Good work!

lua/nvim-tree/events.lua
yioneko: Thanks.

alex-courtis:
Thanks for the updates @yioneko 

This PR deserves significant review time which I can allocate this weeken.

yioneko:
Just post the discussion about async lib in core: https://github.com/neovim/neovim/issues/19624. 

My current implementation is quite minimal so it might be fairly convenient to migrate to it once standardized. Also, it's reasonable to suspend this pr unless the async lib is implemented in core, and we could use it directly. Open to any suggestions.

alex-courtis:
> Just post the discussion about async lib in core: [neovim/neovim#19624](https://github.com/neovim/neovim/issues/19624).
> 
> My current implementation is quite minimal so it might be fairly convenient to migrate to it once standardized. Also, it's reasonable to suspend this pr unless the async lib is implemented in core, and we could use it directly. Open to any suggestions.

It sounds a core implementation is quite far away.

I'm quite happy to have the implementation in nvim-tree and migrate later, which should be trivial.

lua/nvim-tree/actions/fs/create-file.lua
```diff
@@ -49,7 +60,61 @@ local function get_containing_folder(node)
...
+    return
+  end
+
+  -- create a folder for each path element if the folder does not exist
+  -- if the answer ends with a /, create a file for the last path element
+  local is_last_path_file = not new_file_path:match(utils.path_separator .. "$")
+  local path_to_create = ""
+  local idx = 0
+
+  local num_nodes = get_num_nodes(utils.path_split(utils.path_remove_trailing(new_file_path)))
```
alex-courtis, line 86: Please do not duplicate this code.

alex-courtis (changes requested):
I'm happy to run with this; experimental features are... experimental.

Please:

- [x] Revert :help formatting changes
- [x] Remove API
- [ ] Revert unnecessary events changes
- [x] Revert bulk move
- [x] Revert open file changes
- [ ] Normalise create-file - DRY
- [ ] Force and assume `filesystem_watchers.enable`
- [ ] Use "standard" windows feature flag detection: https://github.com/nvim-tree/nvim-tree.lua/pull/1870#discussion_r1059569916
- [ ] More useful error messages: https://github.com/nvim-tree/nvim-tree.lua/pull/1870#discussion_r1059580587

alex-courtis:
@yioneko I just discovered that review comments do not get posted until the review is complete.

Please accept my apologies; I assumed you were able to see them.

yioneko:
@alex-courtis Uhh, that's fine, I'm also busy during this period. Thanks for the detailed comments, I'll work on this next week.

alex-courtis:
> @alex-courtis Uhh, that's fine, I'm also busy during this period. Thanks for the detailed comments, I'll work on this next week.

No worries. This is valuable functionality, but there is no rush.

lua/nvim-tree/actions/fs/rename-file.lua
```diff
@@ -15,18 +16,27 @@ local function err_fmt(from, to, reason)
...
   if utils.file_exists(to) then
     notify.warn(err_fmt(node.absolute_path, to, "file already exists"))
     return
   end
 
   events._dispatch_will_rename_node(node.absolute_path, to)
-  local success, err = vim.loop.fs_rename(node.absolute_path, to)
+  local success, err
+  if M.enable_async then
+    err, success = async.call(vim.loop.fs_rename, node.absolute_path, to)
```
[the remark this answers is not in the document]
yioneko: Could you explain more about the relationship between fs watcher and these async file operations? As far as I know, reloading will be enabled on watcher disabled, I didn't see any potential explore sync issue here. The async version still ensures that reloading is triggered on the end of execution.

lua/nvim-tree/async.lua
```diff
@@ -0,0 +1,140 @@
...
+  local cb = table.remove(args) or function() end
+  local thread = co.create(func)
+  async_threads[thread] = true
+
+  local function step(...)
+    local res = { co.resume(thread, ...) }
+    local ok = table.remove(res, 1)
+    local err_or_next = res[1]
+    if co.status(thread) ~= "dead" then
+      local _, err = xpcall(err_or_next, debug.traceback, step)
```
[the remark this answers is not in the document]
yioneko: I prefer not adding extra runtime overhead as the case is unusual and only happens when user calls `coroutine.yield(non_function_value)` in the async context. This is actually an antipattern (we expect user to use `async.call`).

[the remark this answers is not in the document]
yioneko: > /nit step is unused

It's actually false positive of lua-language-server. `xpcall` in luajit does accpet additional arguments. 
 https://luajit.org/extensions.html

doc/nvim-tree-lua.txt
```diff
@@ -1277,6 +1302,10 @@ exists.
     - navigate.prev
     - navigate.select
 
+- api.async:                                            *nvim-tree.api.async*
```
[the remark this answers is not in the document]
yioneko: This is correlated with `events.subscribe_async` to make user able to interfere with async execution flow of file operations. My initial intention here is to supplement https://github.com/nvim-tree/nvim-tree.lua/pull/1821 to be able to use async version of lsp request instead of blocking [`request_sync`](https://github.com/antosha417/nvim-lsp-file-operations/blob/2d2d3466a7a92be86a596fc9751809825e60be64/lua/lsp-file-operations/will-rename.lua#L18). It's fine to leave it for next PR.

lua/nvim-tree/async.lua
```diff
@@ -0,0 +1,140 @@
...
+      if err then
+        cb(err)
+      end
+    else
+      async_threads[thread] = nil
+      if ok then
+        cb(nil, unpack(res))
+      else
+        cb(debug.traceback(thread, err_or_next))
```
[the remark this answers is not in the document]
yioneko: Have not figured out this yet, I failed to reproduce it locally and could see the full error message. It's likely caused by other stuff. `debug.traceback` should prepend the error message to the stack trace.

lua/nvim-tree/async.lua
```diff
@@ -0,0 +1,180 @@
...
+  local cur = vim.loop.hrtime()
+  if cur - self.last >= self.interval * 1000000 then
+    self:yield()
+    self.last = cur
+  end
+end
+
+M.Interrupter = Interrupter
+
+---This is useful for cancelling execution async function
```
yioneko, line 138: One more pending question to make this ready for merge: should we implement cancellation of these operations? This is impossible in sync version but achievable now.

lua/nvim-tree/async.lua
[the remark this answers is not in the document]
alex-courtis: Let's aim for parity: a message similar to existing:

`[NvimTree] "Cannot rename /home/alex/src/nvim-tree.lua/yioneko/README.md -> /home/alex/src/nvim-tree.lua/yioneko/inexistent/README.md: ENOENT: no such file or directory: /home/alex/src/nvim-tree.lua/yioneko/README.md -> /home/alex/src/nvim-tree.lua/yioneko/inexi
stent/README.md"`

I know that error message is not perfect as it repeats itself, however we can fix that in time.

lua/nvim-tree/actions/fs/create-file.lua
```diff
@@ -41,6 +37,73 @@ local function get_num_nodes(iter)
...
+    end
+    if is_last_path_file and idx == num_nodes then
+      if M.enable_async then
+        async.schedule()
+      end
+      if utils.file_exists(new_file_path) then
+        local prompt_select = "Overwrite " .. new_file_path .. " ?"
+        local prompt_input = prompt_select .. " y/n: "
+        if M.enable_async then
+          local item_short = async.call(lib.prompt, prompt_input, prompt_select, { "y", "n" }, { "Yes", "No" })
```
alex-courtis, line 65: Do we need to make the prompt async? Only `create_and_notify` etc. needs to be async.

lua/nvim-tree/async.lua
```diff
@@ -0,0 +1,180 @@
+---Idea taken from: https://github.com/ms-jpq/lua-async-await
```
alex-courtis, line 1: There's a lot of unused codepaths in here. Let's trim it down. If we do need functionality we can refer to the source.

It looks like:
`in_async`
`wrap`
`all`
interrupt/abort

lua/nvim-tree/async.lua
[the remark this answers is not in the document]
alex-courtis: Actually... on further real and synthetic testing I am getting sane messages.

`debug.traceback` is only used in exceptional conditions, so I'm happy to leave that in.

Nothing to do here.

alex-courtis (changes requested):
This is looking good; we will get there.

Please:

- [ ] remove unused codepaths
- [ ] resolve my naive questions around the need to schedule
- [ ] don't use async prompts

lua/nvim-tree/actions/fs/create-file.lua
yioneko: Yes, this is mandatory in async version. In fact, `lib.prompt` itself is already async because of usage of callback, the only difference is that the async context of `lib.prompt` is not controlled. If we use callback instead of wrapping with `async.call`, `create_and_notify` may not be called in the same async context as where `async.exec` begins and the immediate following `async.call` in `create_and_notify` will throw error.

You can test it will override `vim.ui.input` with https://github.com/stevearc/dressing.nvim or even a simple delayer:
```lua
local original = vim.ui.input
vim.ui.input = function(opts, on_confirm)
  original(opts, function(input)
     vim.defer_fn(function() on_confirm(input) end, 0)
  end)
end
```

lua/nvim-tree/async.lua
```diff
@@ -0,0 +1,180 @@
...
+        results[i] = { ... }
+        if finished == #tasks then
+          cb(results)
+        end
+      end)
+    end
+  end)
+end
+
+---Asynchronous `vim.schedule`
```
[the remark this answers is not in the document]
yioneko: As documented in `:h lua-loop-callbacks`, we cannot call some `vim.api.*` and all `vim.fn.*` in luv callbacks. So every time before these calls, we should manually call `vim.schedule` (`async.schedule` here to avoid callback hells). It's my fault to assume that we already get this; I'll add necessary comments about it.

lua/nvim-tree/actions/fs/create-file.lua
yioneko: In principle, all callback-style functions should be wrapped to make the async context consistent.

alex-courtis: Unfortunately I still don't understand. Yes, the vim.ui.select calls themselves are asynchronous, however the callbacks appear to be executed on the main thread / context.

```lua
function M.prompt(prompt_input, prompt_select, items_short, items_long, callback)
  ...
  if M.select_prompts then
    vim.ui.select(items_short, { prompt = prompt_select, format_item = format_item }, function(item_short)
      log.line("dev", "prompt select in_fast_event=%s is_thread=%s", vim.in_fast_event(), vim.is_thread())
      callback(item_short)
    end)
  else
    vim.ui.input({ prompt = prompt_input }, function(item_short)
      log.line("dev", "prompt input in_fast_event=%s is_thread=%s", vim.in_fast_event(), vim.is_thread())
      callback(item_short)
    end)
  end
end
```

```
[2023-01-23 10:29:44] [dev] prompt input in_fast_event=false is_thread=false
```

yioneko: Yes, because it might be called in main thread (not in async thread), we need to use `async.call` to transfer it to async thread, otherwise `create_and_notify` will throw error because `async.call` uses `coroutine.yield` under the hood.

lua/nvim-tree/async.lua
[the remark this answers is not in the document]
alex-courtis: I see. So this schedule is essentially a thread join.

I do not understand the timing; surely it should be called immediately after the `.call` or during the callback. It's called once at the start of `create_file` and once at the end.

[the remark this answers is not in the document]
yioneko: Certainly we could do that for every callback, but usually this is only needed when the callback contains vim apis (or we are uncertain about this, such as user provided callback). In some cases we may not want to use `vim.schedule` to delay the execution of callback like a sequence of consecutive luv calls and always delaying may have potential negative impact on performance.

lua/nvim-tree/actions/fs/create-file.lua
alex-courtis: All our vim.ui.input and vim.ui.select calls are from the main thread. If we made these calls from an async thread it would be an issue, however we have no need to do that.

Please revert the changes to the inputs. This change is about making the create and rename file system operations async, not the prompts. We can make a future change to allow async input calls if it is ever needed.

lua/nvim-tree/async.lua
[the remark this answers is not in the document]
alex-courtis: I see. In that case, please move the calls to schedule immediately after the calls that require this join. It should be clear exactly when and where it is required.

lua/nvim-tree/actions/fs/create-file.lua
yioneko: I'm sorry that it is required to call the prompt callback in async thread, otherwise we have to use `async.exec` before every calls to following async code in the callback and it doesn't deserve IMO. And additionally, in the current implementation `vim.ui.*` is actually called in the main thread, only the callback is called in coroutine. I think technically there would be no issue here.

alex-courtis: Let's just focus on `rename` for now as it is less complex than `create`.

Outcome: `M.rename` `vim.loop.fs_rename` call is asynchronous
Condition: `async.exec` must be called from the main thread
Problem: `M.rename` is executed from a `vim.ui.input` callback which may not be on the main thread

Proposal: call `M.rename` after the input, outside of the `vim.ui.input` callback

yioneko: `async.exec` is not required to be called from main thread. It's possible to create new coroutine in another coroutine. The coroutine (or thread, here) is just stackful statemachine and could be created anywhere.

Would you mind I make a simple guidance for understanding of current basic async implementation? I've seen that there are plenty of misunderstandings between us and possibly with future contributors.

alex-courtis: Yes please. A detailed explaination would be gratefully appreciated.

The outcome of this PR:
* file system operations for create / rename are asynchronous
* code changes are minimal and clear
* asynchronous operations are localised i.e. not spread over multiple callbacks / functions

Unless there is a very good reason, inputs should not be asynchronous. This adds unnecessary complexity.

yioneko: Sorry for that I'm gonna to be off-topic, I think the referred async input codepath would not be ever reached, as we already checked the existence of file before `async.exec`: https://github.com/nvim-tree/nvim-tree.lua/blob/c71f47a8fa955d985018ae607b10cc741b79b965/lua/nvim-tree/actions/fs/create-file.lua#L136-L142. 

Do you think it's safe to remove this? I just cp from the original implementation. There is no other place with async call to  user input.

alex-courtis: I see... that's a good catch!

It does look like this `file_exits` will fail and exit early. The `local function create_file(new_file_path)` call to `file_exists` will always pass and the `"Overwrite " .. new_file_path .. " ?"` prompt cannot occur.

This is existing behaviour: we can safely remove the `file_exists` check in `create_file`.

Please raise a separate PR to remove this dead code path. We will merge that first, keeping our changes small and atomic and easily revertable.

yioneko: I found that I referred the wrong variable, `new_file_path` -> `path_to_create`, sorry about that. However, the truth that this is a dead code remains; and I think we should early return instead of prompting for overwrite for already existing file.

alex-courtis:
Any updates on this one @yioneko ?

We are getting very close...

alex-courtis:
Closing due to inactivity.

The concept and implementation of async is sound; changes to the usages are required.

PR #1872, opened 2023-01-01
fix(#1785): retain focused node following filter toggle

### LINKED ISSUE #1785: Focus Incorrect After Filter Toggle
### Description

After toggling a filter the current line is unchanged, which can change selected file.

### Neovim version

```text
NVIM v0.8.1
Build type: Release
LuaJIT 2.1.0-beta3
```

Linux 6.0.10-arch2-1

### nvim-tree version

829e9f6

### Minimal config

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

open nvim-tree.lua
open doc
move cursor to `CONTRIBUTING.md`
`I`

### Expected behavior

cursor at `CONTRIBUTING.md`

### Actual behavior

cursor at `.stylua.toml`

alex-courtis:
fixes #1785

I was only briefly tempted to do this for all reloads, as there may be assumptions around the cursor location.

### DIFF
--- a/lua/nvim-tree/actions/tree-modifiers/toggles.lua
+++ b/lua/nvim-tree/actions/tree-modifiers/toggles.lua
@@ -1,3 +1,4 @@
+local lib = require "nvim-tree.lib"
 local view = require "nvim-tree.view"
 local filters = require "nvim-tree.explorer.filters"
 local renderer = require "nvim-tree.renderer"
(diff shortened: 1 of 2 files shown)


gegoune:
Seems to be working well when highlighted file is opened in nvim.
If file is not opened cursor will jump to first line of tree and remain there when switching filter on and off. Seem like it just doesn't have an entry to jump to.

You can test it by opening unchanged (git) file and filtering with `C`. It will work OK when you open and "find" changed file.

alex-courtis:
> If file is not opened cursor will jump to first line of tree and remain there when switching filter on and off. Seem like it just doesn't have an entry to jump to.

That is a problem. I'll have a play and try and come up with something "natural".

alex-courtis:
The problem I'd like to solve is the filter not retaining focus on the node, when it is visible before and after filter toggle.

Options for "node no longer visible" case:
1. Focus root
2. Focus lowest visible parent folder if present
3. Do nothing

1 is jarring, 3 is current behaviour.

I'll try 2 and see how it looks.

alex-courtis:
Abandoning. Things get complicated when it comes to folders and nothing felt quite right.

A solution will present itself at some point.

PR #1877, opened 2023-01-02
feat(event): add TreeAttachedPost

### LINKED ISSUE #1869: Event.TreeAttachedPost
Event fired following `on_attach` call.

Context:
#1858 
https://github.com/mrjones2014/legendary.nvim/issues/289#issuecomment-1364618688

[Legendary](https://github.com/mrjones2014/legendary.nvim) has a requirement to access / mutate user's mappings set via their `on_attach`.

alex-courtis:
closes #1869

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1639,6 +1639,11 @@ e.g. handler for node renamed: >
                 handler parameters: ~
                   {folder_name}  `{string}` Absolute path to the removed folder.
 
+- Event.OnAttachPost
+                Invoked after |nvim-tree.on_attach| has been called.
+                handler parameters: ~
+                  {buf}          `{number} `API buffer handle (buffer number)
+
 ==============================================================================
  9. BOOKMARKS                                            *nvim-tree-bookmarks*
 
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1640,7 +1640,8 @@ e.g. handler for node renamed: >
                   {folder_name}  `{string}` Absolute path to the removed folder.
 
 - Event.TreeAttachedPost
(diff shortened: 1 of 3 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -1639,6 +1639,11 @@ e.g. handler for node renamed: >
                 handler parameters: ~
                   {folder_name}  `{string}` Absolute path to the removed folder.
 
+- Event.OnAttachPost
```
gegoune, line 1642: Do we need `On` prefix here? Seems like other events don't have those. `AttachPost` sounds bit off though. How about `TreeAttached`? Might want to make it `TreeAttachedPost` if there is a chance of `TreeAttachedPre` being added as well.

alex-courtis: That makes sense. Telescope uses attach and gitsigns only adds `on_` for the callback itself.

alex-courtis: If we ever handle multiple trees/tabs, that could be added.

gegoune: Thanks, one more option I thought of: `TreeCreated` - maybe it's bit more descriptive than `Attached`. Up to you.

alex-courtis: on_attach happens in the middle of things.

We might want a TreeCreated event before or a TreeReady event after if a need arises :shrug:

gegoune: That makes sense then. Thanks once again.

alex-courtis: Thank you for the diligence.

I'm just super-paranoid about not walling us into any corners with API or Events.

Whilst I have you... could I please ask for a sanity check / test for #1872

Off until next weekend :wave:

gegoune (approved):

PR #1879, opened 2023-01-03
Fix nvim tree frozen with no name buffer

### LINKED ISSUE #1878: nvim 100% CPU And Frozen When `modified.enabled = true`
### Description

nvim will freeze and process will use 100% CPU when first editing a no name new buffer.

### Neovim version

```text
NVIM v0.8.2
Build type: Release
LuaJIT 2.1.0-beta3
```

Linux 6.1.1-arch1-1

### nvim-tree version

since dcc344c

### Minimal config

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {
	  modified = {
		  enable = true,
	  },
	  log = {
		  enable = true,
		  truncate = true,
		  types = {
			  all = true,
		  },
	  },
  }
end
```

### Steps to reproduce

nvim -nu /tmp/nvt-min.lua
```vim
:NvimTreeOpen
<C-w>w
isomething
```

### Expected behavior

_No response_

### Actual behavior

_No response_

chomosuke:
closes #1878.

Thanks for raising this one @alex-courtis.

The issue turned out to be path never becoming `cwd` or `"/"` no matter how many `fnamemodify(path, ":h")`. This bug would've also appeared if a file out side of the cwd is modified on windows as path would've stopped at `"C:\\"` instead of `"/"`.

I've made it so that:
- ignore buffer for modified if `buf.name` is `""`
- Don't keep looking for parents if the path is already recorded as modified, or if `fnamemodify` didn't change the path.

### DIFF
--- a/lua/nvim-tree/modified.lua
+++ b/lua/nvim-tree/modified.lua
@@ -9,10 +9,16 @@ function M.reload()
   local bufs = vim.fn.getbufinfo { bufmodified = true, buflisted = true }
   for _, buf in pairs(bufs) do
     local path = buf.name
-    M._record[path] = true
-    while path ~= vim.fn.getcwd() and path ~= "/" do
-      path = vim.fn.fnamemodify(path, ":h")
-      M._record[path] = true
+    if path ~= "" then -- not a [No Name] buffer
+      -- mark all the parent as modified as well
+      while
+        M._record[path] ~= true
+        -- no need to keep going if already recorded
+        -- This also prevents an infinite loop
+      do
+        M._record[path] = true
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Works nicely, thanks for the quick fix.

PR #1887, opened 2023-01-05
Implement turning off y/n prompt for file deletion

amamdemous:
I have a suggestion to give users the option to disable the prompt when deleting a file. For example, when working in small folders, there is no particular need to double-check the file you are deleting

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -374,7 +374,6 @@ Subsequent calls to setup will replace the previous configuration.
       },
       trash = {
         cmd = "gio trash",
-        require_confirm = true,
       },
       live_filter = {
         prefix = "[FILTER]: ",
@@ -389,6 +388,12 @@ Subsequent calls to setup will replace the previous configuration.
       },
       notify = {
         threshold = vim.log.levels.INFO,
+      },
+      ui = {
+        confirm = {
+          remove = true,
+          trash = true,
+        },
       },
       log = {
         enable = false,
@@ -993,10 +998,6 @@ Configuration options for trashing.
     The default is shipped with glib2 which is a common linux package.
       Type: `string`, Default: `"gio trash"`
 
-    *nvim-tree.trash.require_confirm*
-    Show a prompt before trashing takes place.
-      Type: `boolean`, Default: `true`
-
 *nvim-tree.actions*
 Configuration for various actions.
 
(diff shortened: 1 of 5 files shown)


gegoune (commented):
We should try to avoid adding new options if at all possible. If the is highly desirable I would like that option name to be changed to something like:

```lua
ui = {
  confirm = {
    node_deletion = true,
  }
}
```

(`node` can actually be called `file`, because everything is a file :)

Would you be willing to add other confirmations as well?

@alex-courtis

alex-courtis:
> Would you be willing to add other confirmations as well?

Yes. Some of these are annoying, but it would be good to be selective.

How about we add an enable to turn them all off?
```lua
ui = {
  confirm = {
    enable = true,
    node_deletion = true,
    paste = true,
    trash = true,
    create = true,
```

lua/nvim-tree.lua
```diff
@@ -495,6 +495,7 @@ local function setup_autocommands(opts)
 end
 
 local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
+  file_deletion_confirmation = true,
```
alex-courtis, line 498: Please refactor this to `ui.confirm.remove`.

alex-courtis (changes requested):
Many thanks for taking the initiative on this UX improvement.

Looks like a straightforward refactor.

If you are motivated to make the same change for other operations, and implement `ui.confirm.enable` I would be most grateful.

amamdemous:
Made refactors as suggested

doc/nvim-tree-lua.txt
```diff
@@ -390,6 +389,12 @@ Subsequent calls to setup will replace the previous configuration.
       notify = {
         threshold = vim.log.levels.INFO,
       },
+      ui = {
+        confirm = {
+          remove = true,
+          trash = true,
+        },
+      },
```
alex-courtis, line 397: Next time please remember to update help as per [CONTRIBUTING](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md)

Added doc for this section and renamed to match the action.

alex-courtis (changes requested):
Remove works beautifully, trash fails at cf36d35 with:

```
E5108: Error executing lua: ...cker/start/amamdemous/lua/nvim-tree/actions/fs/trash.lua:32: attempt to index field 'trash' (a nil value)
stack traceback:
        ...cker/start/amamdemous/lua/nvim-tree/actions/fs/trash.lua:32: in function 'handle_tree_actions'
        ...cker/start/amamdemous/lua/nvim-tree/actions/dispatch.lua:123: in function 'dispatch'
        ...k/packer/start/amamdemous/lua/nvim-tree/actions/init.lua:273: in function <...k/packer/start/amamdemous/lua/nvim-tree/actions/init.lua:272>
```

alex-courtis (approved):
Works nicely, thank you for your contribution.

Issue #1545, opened 2022-08-23
Resize event not dispatched when resizing using mouse


Tcholly:
In my init.vim script I subscribe to the resize event, if I use `:NvimTreeResize` the subscribed function gets called, but if I resize with the mouse it doesn't.

https://user-images.githubusercontent.com/35864488/186197406-3a3d54c5-176c-4295-90af-9bd987de25b6.mp4

From the above video you can see that if I use the command, the tabs above follow as well, but if I resize using the mouse they don't.
You can also see how I listen to the event.

alex-courtis:
In the future please raise a bug report, not an empty issue.

We are firing resize events for all nvim-tree operations, including adaptive_size and friends.

Mouse and `<C-W><` do not fire events.

This fix should not cause any unwanted behavioural changes or break anyone.

wlh320:
Hi! I noticed that recently neovim added a `WinResized` autocommand event in [neovim/neovim/issues/11532](https://github.com/neovim/neovim/issues/11532). This may be helpful for nvim-tree to firing Resize events for mouse and `<C-W>`.

alex-courtis:
Fantastic! Thank you @wlh320

PR #1915, opened 2023-01-12
feat: max_width for adaptive_size

### LINKED ISSUE #1797: adaptive_size Maximum Width
it is too wide when the file name is long, so i'd like to limit the width when in adaptive_size mode

ramezgerges:
resolves #1797

P.S> I'm not sure if this is an issue, but the width of NvimTree doesn't seem to be pixel perfect. The width is off by 2 every time. This is not caused by my changes (I tested this using the provided clean room config).


alex-courtis:
> [](https://github.com/nvim-tree/nvim-tree.lua/issues/1797)
> 
> P.S> I'm not sure if this is an issue, but the width of NvimTree doesn't seem to be pixel perfect. The width is off by 2 every time. This is not caused by my changes (I tested this using the provided clean room config).

That would be the sign column. We need to remember that the user may disable it.

alex-courtis:
This is a preexisting issue: 2 is added to width when setting apaptive size less than max:

20230114_145756 498x518

20230114_145827 501x514

This is not strictly part of this PR, however I would be most grateful if you resolved this.

lua/nvim-tree/view.lua
```diff
@@ -246,14 +246,20 @@ end
 local function grow()
   local starts_at = M.is_root_folder_visible(require("nvim-tree.core").get_cwd()) and 1 or 0
   local lines = vim.api.nvim_buf_get_lines(M.get_bufnr(), starts_at, -1, false)
-  local max_length = M.View.initial_width
+  local padding = 3
```
alex-courtis, line 249: No magic numbers please. Let's consider `view.signcolumn`.

alex-courtis: Out of interest: why 3 instead of 2?

alex-courtis:
Strange behaviour with small sizes:

```lua
    view = {
      adaptive_size = true,
      width = 30,
      max_width = 10,
```

[asciicast](https://asciinema.org/a/551509)

alex-courtis (changes requested):
Thank you for your contribution, this will greatly improve the user experience.

Tested mostly OK:
* max_width 30
* max_width function
* max_width function returning garbage, OK as same as width
* max_width percent
* max_width -1

Please:

- [x] Account for `view.signcolumn` and remove magic number https://github.com/nvim-tree/nvim-tree.lua/pull/1915#issuecomment-1382654075
- [ ] Resolve small window size incorrectness https://github.com/nvim-tree/nvim-tree.lua/pull/1915#issuecomment-1382658803

lua/nvim-tree/view.lua
ramezgerges: https://github.com/nvim-tree/nvim-tree.lua/blob/3ce0a8e770f70a135ef969e0a640bd8d05baf42c/lua/nvim-tree/view.lua#L251

I just used the preexisting 3 :).

Just to clarify, the intended behavior is to always have a right-padding of 1 character regardless of `view.signcolumn` and/or adaptive_size, is that right?

I have exams in the next couple of days, so I might not be very active, but I will try to fix these ASAP.

ramezgerges:
> Strange behaviour with small sizes:
> 
> ```lua
>     view = {
>       adaptive_size = true,
>       width = 30,
>       max_width = 10,
> ```
> 
> [asciicast](https://asciinema.org/a/551509)



I cannot reproduce this. On my device, `width` is practically completely ignored when `adaptive_size` is true. So, in this example, the width is at most 10 (+/- 2). Here's the example config for reassurance:

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "ramezgerges/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {
    view = {
      adaptive_size = true,
      width = 30,
      max_width = 10,
    }
  }
end
```

(pardon the weird characters, I have nerd fonts)
[asciicast](https://asciinema.org/a/gYBgnRNQd0HflnuRKs3tpWzRq)

gegoune:
How about `width_max`, just more pleasing visually. Or make `width` a table?

ramezgerges:
> Strange behaviour with small sizes:
> 
> ```lua
>     view = {
>       adaptive_size = true,
>       width = 30,
>       max_width = 10,
> ```
> 
> [asciicast](https://asciinema.org/a/551509)

After an unnecessarily long amount of digging, it turns out that the problem is
https://github.com/nvim-tree/nvim-tree.lua/blob/3ce0a8e770f70a135ef969e0a640bd8d05baf42c/lua/nvim-tree/view.lua#L300-L302

Moving it to the top solves the problem (still no idea why it was resized to that exact value though), but I don't really understand what `view.preserve_window_proportions` is supposed to be doing. I understand that it equalizes the other windows, but it does so on any `resize` invocation (so basically on any redraw) and not specifically when opening a file (as is written in the docs), so is there something wrong, or is that intended?

ramezgerges:
> How about `width_max`, just more pleasing visually. Or make `width` a table?

I'm not sure how to respond to this :). I prefer `max_window`, but don't mind it either way. Not sure who has the final call on naming conventions.

alex-courtis:
> > How about `width_max`, just more pleasing visually. Or make `width` a table?
> 
> I'm not sure how to respond to this :). I prefer `max_window`, but don't mind it either way. Not sure who has the final call on naming conventions.

Maybe `adaptive_max`...

A table would be nice:
```lua
    width = {
      adaptive = true,
      min = 30,
      max = 50,
    },
```

As usual we would need to do a silent refactor.

alex-courtis:
> Moving it to the top solves the problem (still no idea why it was resized to that exact value though), but I don't really understand what view.preserve_window_proportions is supposed to be doing. I understand that it equalizes the other windows, but it does so on any resize invocation (so basically on any redraw) and not specifically when opening a file (as is written in the docs), so is there something wrong, or is that intended?

Here be dragons. I now vaguely recall some vim dramas around `winwidth` and `winfixwidth` for small numbers. The live filter overlay is hardcoded to 20, probably to work around such issues.

Let's not deal with that here.

A pragmatic solution may simply be to always specify / percentage calculate a width >=20

alex-courtis:
> [](https://github.com/nvim-tree/nvim-tree.lua/issues/1797)
> 
> P.S> I'm not sure if this is an issue, but the width of NvimTree doesn't seem to be pixel perfect. The width is off by 2 every time. This is not caused by my changes (I tested this using the provided clean room config).

That's clearly a bug. The padding should be 2 or 0 based on sign column visibility.

Pushed a fix.

gegoune:
> We could even get rid of the adaptive variable. If the user specifies max > min it implies adaptive. Default both to 30.

Still need a way to specify static size. I propose:

```lua
width = number -- for static size
width = {min = number, max = number} -- for adaptive
```

alex-courtis:
yy

> > We could even get rid of the adaptive variable. If the user specifies max > min it implies adaptive. Default both to 30.
> 
> Still need a way to specify static size. I propose:
> 
> ```lua
> width = number -- for static size
> width = {min = number, max = number} -- for adaptive
> ```

Nice. No big refactoring required.

Users with `adaptive_size = true` can be silently migrated to `width = {min = 20, max = -1}`. <20 is broken anyway so they get a free fix.

alex-courtis:
Pushed some changes. Please feel free to revert!
* fix the sign column silliness
* apply a minimum width of 20
* use a table
* harden the function / string widths from the user

It was hard to resist further refactorings...

Needs review and extensive testing from @ramezgerges

alex-courtis (changes requested):
- [ ] @ramezgerges review

lua/nvim-tree/view.lua
```diff
@@ -246,7 +246,7 @@ end
 local function grow()
   local starts_at = M.is_root_folder_visible(require("nvim-tree.core").get_cwd()) and 1 or 0
   local lines = vim.api.nvim_buf_get_lines(M.get_bufnr(), starts_at, -1, false)
-  local padding = 3
+  local padding =  M.View.winopts.signcolumn and 2 or 0
```
ramezgerges: Do we want to add an extra character (I.e. 3 instead of 3) as it was previously? I personally find the extra character useful as it indicates that I haven't reached the maximum width yet (i.e. the filename is fully shown and not clipped)

gegoune:
I feel like switching between adaptive and not should be something interested users would do themselves with API. If current API doesn't allow it then it could be expanded.

lua/nvim-tree/view.lua
alex-courtis: Ah I think you got it... I didn't realise that the extra space was useful.

Yes, let's put that back and ensure that it works for max = -1 and max = 30

alex-courtis:
> I feel like switching between adaptive and not should be something interested users would do themselves with API. If current API doesn't allow it then it could be expanded.

Having the `adaptive_width` boolean doesn't quite feel right. It's not a configuration time thing, it's a runtime thing.

API works: allow the user to set width. They can set  `min=25, max=50` when they want adaptive, `min=25,max=25` when they want fixed.

They could even just set it to an integer when they want fixed and a table when they want adaptive.

I'm not quite sure how that API would look; it's a bit different to anything existing.

Proposal: release this now, with the padding put back, then we think about API.

ramezgerges:
Alright, I finally found time for testing.

Some observations:
* Firstly, the `MIN_WIDTH` hack is working fine for now. It does technically break behavior for old configs with width < 20. However, it was already buggy, so I guess that's reasonable. I'm just pointing it out as a reminder.

* Not sure if it's the same issue as- #1545, but on resizing by mouse, the window snaps back to its previous width on the next redraw. This is not caused by this PR, but it exists.

* Side-note: I'm able to reproduce the width < 10 snapping issue even before the PR commits, so this wasn't caused by this PR, which makes sense (I was wondering what I could've possibly changed that caused this).

* There seems to be an issue with window splitting. Splitting into a 2x2 setup doesn't seem to align perfectly. Obviously a minor issue, but I will leave it up to you to decide if an issue should be opened to tackle this.
This was reproduced with `width = 10` and `width = 40`. Not caused by this PR.

(Notice how I move bottom left window to the right manually to show the difference.)
<br/>
<br/>
<br/>
This are the ones that I could spot for now. I'm not the most perceptive person in the world, I may have missed a problem or two :).
<br/>
<br/>
One more thing about `MIN_WIDTH`. I can look into `vim.cmd ":wincmd ="` to see if we can find a fix for this. Is this something worth spending time on, or should I just forget about it? I assume that would be a separate PR, but I'm just asking if it's worth looking into. I found the `MIN_WIDTH` fix so uncomfortable :).

alex-courtis:
>     * Firstly, the `MIN_WIDTH` hack is working fine for now. It does technically break behavior for old configs with width < 20. However, it was already buggy, so I guess that's reasonable. I'm just pointing it out as a reminder.
>
> One more thing about `MIN_WIDTH`. I can look into `vim.cmd ":wincmd ="` to see if we can find a fix for this. Is this something worth spending time on, or should I just forget about it? I assume that would be a separate PR, but I'm just asking if it's worth looking into. I found the `MIN_WIDTH` fix so uncomfortable :).

I think you are right here. This is disruptive to the user, and there will be some who are using width < 20 and dealing with any issues.

I'll back that change out; width 20 is just a bad hack. I would be most grateful if you could look into the root issue it in a future PR.

alex-courtis (approved):
Thank you for your contribution and all the detailed analysis/testing @ramezgerges

alex-courtis:
Reverted my very broken "fix" for sign column:

`local padding = (M.View.winopts.signcolumn and 2 or 0) + 1`

PR #1930, opened 2023-01-21
fix: variable width accounts for sign/number columns

alex-courtis:
When calculating the variable width pad for the sign/number columns plus an additional 1 as indicator.


alex-courtis:
I'd be most grateful if you could quickly review/test this one @ramezgerges

HendrikPetertje:
Given that this one here is doing all my work automatically (no need to set padding manually) this one might be a lot better than my suggestion in https://github.com/nvim-tree/nvim-tree.lua/pull/1941.

alex-courtis:
superseded by #1941

PR #1936, opened 2023-01-23
fix: remove redundant file existence check in create file operation

yioneko:
See https://github.com/nvim-tree/nvim-tree.lua/pull/1870#discussion_r1083594425.

For backward compatibility, when the file already exists, use early return instead of prompting to overwrite.


alex-courtis (approved):
Tested OK:
* create normally
* create already exists
* create failure

alex-courtis:
Thanks for raising this one. It will make things simpler in the future.

alex-courtis:
I know that we should test the fd returned by the pcall, however that is a change for Another Day.

PR #1941, opened 2023-01-24
Add dynamic sizing padding options

HendrikPetertje:
with https://github.com/neovim/neovim/pull/20621 merged in it is now possible to fully customize the status-column in nvim (the column on the left containing line-numbers, fold info, signs and borders).

A fair few cool implementations have popped up like:
- https://github.com/CKolkey/config/blob/master/nvim/after/plugin/statuscolumn.lua
- https://github.com/luukvbaal/statuscol.nvim
- and my own personal one (based on CKolkey's fantastic work) https://git.hendrikpeter.net/hendrikpeter/pico-vim/-/blob/main/lua/peva/status_column.lua

The problem with nvim-tree however is that dynamic sizing doesn't take the custom size of a status column into account and the end of file names get clipped off. This little patch should fix that (and give some examples to help other status_column modders get started).

Thanks for looking at this and thanks for making this amazing plugin, I've been using it for a while and I really like it!

## Some screenshots:

Before:
Screenshot 2023-01-24 at 10 13 45

After:
Screenshot 2023-01-24 at 10 13 18

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ require("nvim-tree").setup()
 require("nvim-tree").setup({
   sort_by = "case_sensitive",
   view = {
-    width = 30
+    width = 30,
     mappings = {
       list = {
         { key = "u", action = "dir_up" },
--- a/README.md
+++ b/README.md
@@ -74,14 +74,7 @@ require("nvim-tree").setup()
 require("nvim-tree").setup({
   sort_by = "case_sensitive",
   view = {
-    -- adaptive width (use width = number instead for static width)
-    width = {
-      min = 30
-      -- -1 means no limit
-      max = -1,
-      -- set total padding space when resizing (handy for custom status columns)
-      padding = 3
-    }
+    width = 30
     mappings = {
       list = {
         { key = "u", action = "dir_up" },
--- a/README.md
+++ b/README.md
@@ -74,7 +74,14 @@ require("nvim-tree").setup()
 require("nvim-tree").setup({
   sort_by = "case_sensitive",
   view = {
-    adaptive_size = true,
+    -- adaptive width (use width = number instead for static width)
+    width = {
+      min = 30
+      -- -1 means no limit
+      max = -1,
+      -- set total padding space when resizing (handy for custom status columns)
+      padding = 3
+    }
     mappings = {
       list = {
         { key = "u", action = "dir_up" },
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -734,16 +734,16 @@ Window / buffer setup.
       Type: `string | number | function | table`, Default: `30`
 
         *nvim-tree.view.width.min*
-          Minimum dynamic width.
-            Type: `string | number | function`, Default: `30`
+        Minimum dynamic width.
+          Type: `string | number | function`, Default: `30`
 
         *nvim-tree.view.width.max*
-          Maximum dynamic width, -1 for unbounded.
-            Type: `string | number | function`, Default: `-1`
+        Maximum dynamic width, -1 for unbounded.
+          Type: `string | number | function`, Default: `-1`
 
         *nvim-tree.view.width.padding*
-          extra padding to the right.
-            Type: `string | number | function`, Default: `1`
+        Extra padding to the right.
+          Type: `string | number | function`, Default: `1`
 
     *nvim-tree.view.side*
     Side of the tree, can be `"left"`, `"right"`.
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -734,11 +734,16 @@ Window / buffer setup.
 
         *nvim-tree.view.width.min*
           Minimum dynamic width.
-          Type: `string | number | function`, Default: `30`
+            Type: `string | number | function`, Default: `30`
 
         *nvim-tree.view.width.max*
           Maximum dynamic width, -1 for unbounded.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -74,7 +74,14 @@ require("nvim-tree").setup()
 require("nvim-tree").setup({
   sort_by = "case_sensitive",
   view = {
-    adaptive_size = true,
```
HendrikPetertje, line 77: I noticed that this is now deprecated, so I updated the example code in the README with the new way of working with dynamic sizes and then tacked my padding options on

doc/nvim-tree-lua.txt
```diff
@@ -734,11 +734,16 @@ Window / buffer setup.
...
 
         *nvim-tree.view.width.max*
           Maximum dynamic width, -1 for unbounded.
-          Type: `string | number | function`, Default: `-1`
+            Type: `string | number | function`, Default: `-1`
+
+        *nvim-tree.view.width.padding*
+          dynamic padding when resizing. Useful for custom status-column
+          styles.
+            Type: `number`, Default: `3`
```
HendrikPetertje, line 746: should be the right docs, padding only really supports a number

lua/nvim-tree/view.lua
```diff
@@ -250,7 +251,7 @@ local function grow()
   local starts_at = M.is_root_folder_visible(require("nvim-tree.core").get_cwd()) and 1 or 0
   local lines = vim.api.nvim_buf_get_lines(M.get_bufnr(), starts_at, -1, false)
   -- 1 column of right-padding to indicate end of path
-  local padding = 3
+  local padding = M.View.padding
```
HendrikPetertje, line 254: The big change happens here.

HendrikPetertje (commented):
Notes from self

HendrikPetertje:
https://github.com/nvim-tree/nvim-tree.lua/pull/1930 might be better than my change here, I'll leave the decision to you. (having both the ability to do a bit of padding and read it dynamically could also be an option, i personally like 2 or 3 columns of space to the right.

alex-courtis:
> https://github.com/nvim-tree/nvim-tree.lua/pull/1930 might be better than my change here, I'll leave the decision to you. (having both the ability to do a bit of padding and read it dynamically could also be an option, i personally like 2 or 3 columns of space to the right)

We can do both. Padding at the end is quite nice and there's no reason not to make it customisable.

doc/nvim-tree-lua.txt
alex-courtis: It can be a function, even if just for the sake of consistency.

You might wish to increase padding as you increase max, to "balance things out".

README.md
alex-courtis: Thank you. It seems I forgot to update the README... please update to match help https://github.com/nvim-tree/nvim-tree.lua/blob/9e87ee2d6e86f37ff09cb74ec7dcf2ac984a01e9/doc/nvim-tree-lua.txt#L82

it's only there to provide an example of a simple option to pass to setup.

alex-courtis (changes requested):
Please:

- [ ] allow a padding function
- [ ] update readme
- [ ] merge `fix-variable-width-sign-number-columns`

HendrikPetertje:
Thanks, I'll give this a look and make the corrections after my workday is done!

HendrikPetertje:
That should be it @alex-courtis & @ramezgerges 

I think i went through all the todo-items and cleaned things up a bit. I tested the code myself just now and did some experiments with functions and the likes.

alex-courtis (approved):
Tested OK

width:
20
min/max 20/25
min/max 20/20

sign:
"no"
"yes:4"

number:
true
false

padding:
0
1
2
function

alex-courtis:
Many thanks @HendrikPetertje

zhourrr:
Hi all, I want to thank you for this great plugin and merged feature. I notice that when I toggle the file tree, it prints padding as well. Possibly related to this piece of code 
image
It's kind of interrupting the workflow.

dasupradyumna:
+1
Please remove the extra padding.

doc/nvim-tree-lua.txt
```diff
@@ -82,6 +82,7 @@ Setup should be run in a lua file or in a |lua-heredoc| if using in a vim file.
       sort_by = "case_sensitive",
       view = {
         mappings = {
+          width = 30,
```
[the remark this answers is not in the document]
HendrikPetertje: my bad

lua/nvim-tree/view.lua
```diff
@@ -249,16 +254,24 @@ end
...
-  local padding = 3
+  -- number of columns of right-padding to indicate end of path
+  local padding = get_size(M.View.padding)
+
+  -- account for sign/number columns etc.
+  local wininfo = vim.fn.getwininfo(M.get_winnr())
+  if type(wininfo) == "table" and type(wininfo[1]) == "table" then
+    padding = padding + wininfo[1].textoff
+  end
+  print(padding)
```
dasupradyumna, line 265: Please remove this stray debug `print` statement.
This gives me an extra "Press ENTER or type command to continue" prompt with the padding printed, whenever I toggle Nvim tree window.

gegoune:
That has been fixed on master. Please update.

lua/nvim-tree/view.lua
HendrikPetertje: Yeah sorry that was my mistake, the problem was fixed a few days ago in master ^_^

Issue #1942, opened 2023-01-24
`OptionSet` event is triggered when opening the tree


tummetott:
### Description

Hello,

I registered an autocmd on the `OptionSet` event with `pattern = "cursorline"` and noticed that nvim-tree triggers that event  when opening the tree. The expected behaviour is that no events are triggered when setting an option.

Nvim has the ex command `:noautocmd <cmd>` which would supress that events are fired. However I'm not aware of an LUA API call with the same effect.

`nvim-tree` sets the cursorline here:
https://github.com/nvim-tree/nvim-tree.lua/blob/7944e479c18b7e79ff88a74cd829cf716d92aea5/lua/nvim-tree/view.lua#L130

A possible fix would be:
```lua
-- Create a temporary wrapper function
function _G.tmp()
    vim.opt_local[k] = v 
end

-- Fire the wrapper function from vimscript with `noautocmd`
vim.cmd('noautocmd call v:lua.tmp()')

-- Delete the wrapper function
_G.tmp = nil
```
This workaround is not very elegant but I can't think of a better way since :noautocmd only existst as ex-command

However I'm not sure if that has any sideeffects in general. I greped the repo for the `OptionSet` event and could not find anything. So I guess `nvim-tree` does not rely on events triggerd by `setlocal`, so I guess this change should be fine

### Neovim version

```text
NVIM v0.8.1
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

macOS 10.15.7

### nvim-tree version

16f2806

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
[...]

PR #1945, opened 2023-01-26
Changing window local options does not trigger OptionSet event

tummetott:
As mentioned here https://github.com/nvim-tree/nvim-tree.lua/issues/1942#issuecomment-1404397677, the `eventignore` solves the bug, that the `OptionSet` event is triggered when `nvim-tree` updates local window options.

This PR solves https://github.com/nvim-tree/nvim-tree.lua/issues/1942#issue-1554840995


alex-courtis (approved):
Works nicely, thank you.

"all" is a bit scary, however we are only setting options, so it's OK.

PR #1948, opened 2023-01-28
fix: remove printing padding message on startup

nolleh:
when starting up, there is verbosing echo message for padding.
remove that message.

image

(the printing integer '3')


gegoune:
Thanks a lot, already fixed in https://github.com/nvim-tree/nvim-tree.lua/pull/1949/files.

alex-courtis:
Thank you. Please accept my apologies; I should have picked this up during review/test.

PR #1951, opened 2023-01-29
deprecate open_on_setup

### LINKED ISSUE #1669: New Startup Configuration Mechanism Incoming
The existing startup mechanisms such an `open_on_setup` are problematic, with recent changes resulting in many regressions #1668.

The options are nonatomic and difficult (sometimes impossible) to configure for each use case.

There are often race conditions around startup and events with vim itself and other plugins.

A new mechanisms to specify startup behaviour with more granularity is under way.

The existing mechanism will be retained if an original configuration options such as `open_on_setup` is specified, however the user will be notified of the new mechanism on startup and pointed to a wiki page to explain migration to the new.

**Update 20230109:** the new mechanism will be a set of recipes for the user to control behaviour programatically, along with the events to allow this.

**Update 20230129:** existing configuration will be deprecated in favour of a self serve mechanism: [Open At Startup](https://github.com/nvim-tree/nvim-tree.lua/wiki/Open-At-Startup)

alex-courtis:
fixes #1669

Wiki with all the details: https://github.com/nvim-tree/nvim-tree.lua/wiki/Open-At-Startup

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -13,6 +13,7 @@ local collapse_all = require "nvim-tree.actions.tree-modifiers.collapse-all"
 local git = require "nvim-tree.git"
 local filters = require "nvim-tree.explorer.filters"
 local modified = require "nvim-tree.modified"
+local notify = require "nvim-tree.notify"
 
 local _config = {}
 
@@ -216,8 +217,10 @@ end
 M.resize = view.resize
 
 function M.open_on_directory()
+  require "amc.log".line("open_on_directory")
   local should_proceed = M.initialized and (_config.hijack_directories.auto_open or view.is_visible())
   if not should_proceed then
+    require "amc.log".line("open_on_directory exit early")
     return
   end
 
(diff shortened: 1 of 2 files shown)


lua/nvim-tree.lua
```diff
@@ -900,7 +901,17 @@ function M.setup(conf)
...
 
   vim.schedule(function()
-    M.on_enter(netrw_disabled)
+    if
+      #opts.ignore_ft_on_setup > 0
+      or opts.open_on_setup == true
+      or opts.open_on_setup_file
+      or opts.ignore_buffer_on_setup
+    then
+      notify.info "open_on_setup behaviour has been deprecated, please see https://github.com/nvim-tree/nvim-tree.lua/wiki/Open-At-Startup"
```
alex-courtis, line 910: I suspect people might get upset, however I can't see any other way to do this.

gegoune (approved):

alex-courtis:
"Open in window from previous session" longer occurs on startup. I'll prepare a PR in case anyone misses it.

polyzen:
Shouldn't netrw be allowed to work now? Now attempting to open directories is treated as opening a file, as described in https://github.com/nvim-tree/nvim-tree.lua/issues/1956#issue-1563563124.

alex-courtis:
> Shouldn't netrw be allowed to work now? Now attempting to open directories is treated as opening a file, as described in https://github.com/nvim-tree/nvim-tree.lua/issues/1956#issue-1563563124.

That's a really good point! netrw does have useful functionality including/excluding its "tree" handling.

With full control of the startup process, `:h nvim-tree.hijack_netrw` and `hijack_directories.enable` a user specific configuration that will work with netrw may be possible. We could come up with some [Recipes](https://github.com/nvim-tree/nvim-tree.lua/wiki/Recipes) and enhance the documentation around what is and is not possible.

I'd be grateful if you could explore this; some extra functionality may be required however we can address that as it arises.

PR #1977, opened 2023-02-07
fix(#914): remove existence check before create-file, rename-file, harden clobbered buffer renaming

alex-courtis:
fixes #914

PR #1988, opened 2023-02-12
fix(#914): windows/macos file_exists uses fs_open/fs_close rather than fs_stat due to case sensitivity dramas

### LINKED ISSUE #914: Case sensitive rename not possible (cannot rename)
**OS**
macOS Monterey 12.1

**Neovim version**
v0.6.1

**Nvim-tree version**
`master`

**Describe the bug**
Renaming a file like `Something.md` to `something.md` does not work. **[NvimTree] Cannot rename: file already exists** If I rename it first to something different, it works: `Something.md` → `something2.md`  → `something.md` .

**Expected behavior**
Renaming a file respecting lower and upper case is possible. `Something.md` → `something.md`.

Thanks for your work 🙏🏻

alex-courtis:
fixes #914 

supersedes #1977

### DIFF
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -177,11 +177,24 @@ function M.rename_loaded_buffers(old_path, new_path)
   end
 end
 
+--- Path exists and is readable. fs_stat is used to test.
+--- On windows and macos there are case sensitivity issues with fs_stat; fs_open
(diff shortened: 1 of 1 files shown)


alex-courtis:
This is a bit slower however correctness >> speed

alex-courtis:
Failed. Trying #1995

PR #2000, opened 2023-02-15
feat(renderer): add NvimTreeSymlinkFolderName

qtyosaf:
**Changes**:
Color named `FolderSymlink` added in `colors.lua`.
Color applied to folder symlinks in `_build_folder`  function in `builder.lua`
Added color to Highlight Groups in documentation.

I believe this change is necessary to add more contrast for folder symlinks.


lua/nvim-tree/colors.lua
```diff
@@ -33,6 +33,7 @@ local function get_hl_groups()
   return {
     IndentMarker = { fg = "#8094b4" },
     Symlink = { gui = "bold", fg = colors.cyan },
+    FolderSymlink = { gui = "bold", fg = colors.orange },
```
alex-courtis, line 36: `SymlinkFolder` would be more consistent.

lua/nvim-tree/renderer/builder.lua
```diff
@@ -157,6 +151,15 @@ function Builder:_build_folder(node)
     foldername_hl = "NvimTreeEmptyFolderName"
   end
 
+  local foldername = get_folder_name(node) .. self.trailing_slash
```
alex-courtis, line 154: Any specific reason to move this block of code?

I'm probably missing something obvious...

lua/nvim-tree/colors.lua
alex-courtis, line 36: Unfortunately this is a breaking change: all users will see this new highlight where there was none previously.

We cannot set this group by default, however we can default it to `FolderName` to preserve current behaviour.

doc/nvim-tree-lua.txt
```diff
@@ -1615,6 +1615,7 @@ as per |:highlight|
 Default linked group follows name.
 
 NvimTreeSymlink
```
alex-courtis, line 1617: We now have some inconsistency, as this group applies specifically to files.

We could add `NvimTreeFileSymlink` and deprecate `NvimTreeSymlink` however I don't think it is worth the effort of migration. It should be clear enough to the user as `NvimTreeFolderSymlink` exists.

alex-courtis (changes requested):
This works and will be useful, thanks for your contribution.

Please:
* remove defaults, linking to `FolderName`

qtyosaf:
Sorry for taking so long to update the pull request! I changed things according to the comments.

**Changes:**
Redid the `_build_folder` function so that checks for being a symlink only happen once.
No default color for symlink folders, links to `Directory`.
Color name changed to `NvimTreeSymlinkFolderName` for parity.

alex-courtis (approved):
Tested:
- default to Directory
- default to NvimTreeFolderName
- uses NvimTreeSymlinkFolderName

lua/nvim-tree/renderer/builder.lua
```diff
@@ -149,7 +143,15 @@ function Builder:_build_folder(node)
   end
 
   local foldername_hl = "NvimTreeFolderName"
-  if vim.tbl_contains(self.special_files, node.absolute_path) or vim.tbl_contains(self.special_files, node.name) then
+  if node.link_to and self.symlink_destination then
+    local arrow = icons.i.symlink_arrow
+    local link_to = utils.path_relative(node.link_to, core.get_cwd())
+    foldername = foldername .. arrow .. link_to
+    foldername_hl = "NvimTreeSymlinkFolderName"
+    icon_hl = "NvimTreeSymlinkFolderName"
```
alex-courtis, line 151:
```suggestion
```

Let's make this consistent with regular folders, which use FolderIcon instead of FolderName.

In the future `NvimTreeOpenedSymlinkFolderIcon` and `NvimTreeClosedSymlinkFolderIcon` may be added.

alex-courtis (changes requested):

alex-courtis (approved):

PR #2018, opened 2023-02-24
Fix `vim.opt.shellslash = true` bug.

### LINKED ISSUE #2017: [BUG] Wrong path separator on windows if `vim.opt.shellslash = true`.
### Description

I know there is a method in your code to specify the path separator:

https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/utils.lua#L32

But this can lead to a huge problem if you set `vim.opt.shellslash = true`, this option tells nvim to use `/` even on windows system (`:h shellslash`).

Unfortunately the bug occurred when I set this option, and I also set `sync_root_with_cwd` option, so now it always jump to the buffer location, this is a wired behavior beacuse I also have another plugin like `project.nvim` to pin my root directory. So the nvim-tree messed up the cwd. I traced to the code below:

https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree.lua#L30-L69

In fact the function always return on the last line, it bypass all if-else statement beacuse of the wrong slash `/`.

This is my solution, not very elegant, but can use.

```diff
diff --git a/lua/nvim-tree.lua b/lua/nvim-tree.lua
index 507855b..56bd48b 100644
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -27,7 +27,16 @@ function M.focus()
   view.focus()
 end
 
+local function mysub(mystr)
+  if vim.fn.has('win32') == 1 and vim.opt.shellslash._value == true then
+    return mystr:gsub('/', package.config:sub(1, 1))
+  end
+  return mystr
+end
+
 function M.change_root(filepath, bufnr)
+  filepath = mysub(filepath)
+
   -- skip if current file is in ignore_list
   local ft = vim.api.nvim_buf_get_option(bufnr, "filetype") or ""
   for _, value in pairs(_config.update_focused_file.ignore_list) do
@@ -37,7 +46,7 @@ function M.change_root(filepath, bufnr)
   end
 
   local cwd = core.get_cwd()
-  local vim_cwd = vim.fn.getcwd()
+  local vim_cwd = mysub(vim.fn.getcwd())
 
   -- test if in vim_cwd
   if utils.path_relative(filepath, vim_cwd) ~= filepath then
@@ -52,7 +61,7 @@ function M.change_root(filepath, bufnr)
   end
 
   -- otherwise test M.init_root
-  if _config.prefer_startup_root and utils.path_relative(filepath, M.init_root) ~= filepath then
+  if _config.prefer_startup_root and utils.path_relative(filepath, mysub(M.init_root)) ~= filepath then
     change_dir.fn(M.init_root)
     return
   end
```

ysl2:
Try to fix #2017

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -27,7 +27,16 @@ function M.focus()
   view.focus()
 end
 
+local function mysub(mystr)
+  if vim.opt.shellslash._value == true then
+    return mystr:gsub('/', package.config:sub(1, 1))
+  end
+  return mystr
+end
+
 function M.change_root(filepath, bufnr)
+  filepath = mysub(filepath)
+
   -- skip if current file is in ignore_list
   local ft = vim.api.nvim_buf_get_option(bufnr, "filetype") or ""
   for _, value in pairs(_config.update_focused_file.ignore_list) do
@@ -37,7 +46,7 @@ function M.change_root(filepath, bufnr)
   end
 
   local cwd = core.get_cwd()
-  local vim_cwd = vim.fn.getcwd()
+  local vim_cwd = mysub(vim.fn.getcwd())
 
   -- test if in vim_cwd
   if utils.path_relative(filepath, vim_cwd) ~= filepath then
@@ -52,7 +61,7 @@ function M.change_root(filepath, bufnr)
   end
 
   -- otherwise test M.init_root
-  if _config.prefer_startup_root and utils.path_relative(filepath, M.init_root) ~= filepath then
+  if _config.prefer_startup_root and utils.path_relative(filepath, mysub(M.init_root)) ~= filepath then
     change_dir.fn(M.init_root)
     return
   end
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -28,7 +28,7 @@ function M.focus()
 end
 
 local function mysub(mystr)
-  if vim.opt.shellslash and vim.opt.shellslash._value == true then
+  if vim.fn.has('win32') == 1 and vim.opt.shellslash._value == true then
     return mystr:gsub('/', package.config:sub(1, 1))
   end
   return mystr
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -28,7 +28,7 @@ function M.focus()
 end
 
 local function mysub(mystr)
-  if vim.opt.shellslash._value == true then
+  if vim.opt.shellslash and vim.opt.shellslash._value == true then
     return mystr:gsub('/', package.config:sub(1, 1))
   end
   return mystr


lua/nvim-tree.lua
```diff
@@ -27,7 +27,16 @@ function M.focus()
   view.focus()
 end
 
+local function mysub(mystr)
```
alex-courtis, line 30: This is not a useful name.

It could be `utils.canonical_cwd`

lua/nvim-tree.lua
```diff
@@ -27,7 +27,16 @@ function M.focus()
...
+  if vim.fn.has('win32') == 1 and vim.opt.shellslash._value == true then
+    return mystr:gsub('/', package.config:sub(1, 1))
```
alex-courtis, line 32: What is the result of `:lua print(package.config:sub(1, 1))` when you have shellslash set and not set?

I don't have a windows environment so cannot test.

I'm concerned that we may need to make changes anywhere `path_separator` is used instead of just cwd.

lua/nvim-tree.lua
```diff
@@ -27,7 +27,16 @@ function M.focus()
   view.focus()
 end
 
+local function mysub(mystr)
+  if vim.fn.has('win32') == 1 and vim.opt.shellslash._value == true then
```
alex-courtis, line 31: Please use `utils.is_windows`

alex-courtis (changes requested):
I can only assume this works as I cannot test this.

Changes are isolated behind a windows feature flag so this is OK.

Please:

- [ ] refactor mysub
- [ ] test with WSL
- [ ] answer question on package.config so that we may determine whether further changes are required

ysl2:
Close beacuse I'm too lazy and have no time to do this. Sorry bro.

gegoune:
@alex-courtis Do we want to take this one over? I have no way of testing Windows either (un!?)fortunately.

ysl2:
I can sure the bug exists on windows, and what I do is just wrap the function to replace the “\” to “/“.  And also we cannot derectly replace the whole “path_separator” variable beacuse other behaviors rely on the “\” instead of “/“. So I say “a huge problem” beacuse only this behavior rely on “/“ but others rely on “\”.

alex-courtis:
There are a lot of changes needed to support shellslash: we have a lot of functionality e.g. git that deals with slashes.

We would also need to test the inbuilt conversions we use such as fnamemodify to check if they handle shellslash correctly.

I can't confidently make this change.

ysl2:
Yes, I understand you bro, this is really a hard work.

alex-courtis:
> @alex-courtis Yes, I understand you bro, this is really a hard work.

What's the use case for shellslash? Is it common?

A note in nvim-tree-os-specific should be added.

ysl2:
> What's the use case for shellslash? Is it common?

@alex-courtis https://github.com/nvim-orgmode/orgmode#jumping-to-file-path-is-not-working-for-paths-with-forward-slash

alex-courtis:
Thank you. We can reopen if more users experience issues and a windows developer is available to create a PR.

PR #2032, opened 2023-03-04
Fix typo on clipboard notify

thatlittleboy:
Hi there, just noticed one of the notification messages reads wrong, so this is a tiny PR to fix it.


alex-courtis (approved):
Many thanks!

PR #2041, opened 2023-03-07
add guide for lazy.nvim

sbecerek:
Added on README.md quickstart section for Lazy.nvim package manager

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,21 @@ use {
 }
 ```
 
+or with [lazy](https://github.com/folke/lazy.nvim.git) importing file below or directory it is contained on lazy setup.
+```
+return {
+  "nvim-tree/nvim-tree.lua",
+  version = "*",
+  dependencies = {
+    "nvim-tree/nvim-web-devicons" 
+  },
+  config = function ()
+    require("nvim-tree").setup {
+  }
+  end
+}
+```
+
 ## Setup
 
 Setup should be run in a lua file or in a lua heredoc [:help lua-heredoc](https://neovim.io/doc/user/lua.html) if using in a vim file.


gegoune:
Thank you. I, personally am against adding installation details to readme. They are always towards the top and require scrolling over to get to more important information. Instruction on how to use plugin managers should (and do) fall on plugin managers themselves. Instead of adding another 15 lines I would rather see current installation method replaced with simple wording like: install with your preferred plugin manager or however else you manage your plugins. 

No offence, just my opinion. Will leave it to @alex-courtis to decide.

sbecerek:
Sure, understandable. We may move it to docs section as well. My motivation behind the PR is that I recently struggled adding this plugin to my nvim setup with lazy PM, thought it may help new comers to nvim.

Additionally, waiting response from @alex-courtis

alex-courtis:
Lazy loading, especially lazy.vim seems to be quite popular. See recent discussion #2052

I am, however, struggling to understand the use case. I've not received any good answers beyond "the cool factor". The use case is definitely there. @sbecerek I would be grateful for some insight here, so that nvim-tree can better work with it.

One motivation I can guess is startup time: we don't want plugin setup calls to be executed synchronously in `init.lua`. That isn't a problem for nvim-tree setup as it doesn't do anything beyond validating config. Nothing happens until you open the tree.

I've clarified the help around the nature of setup: https://github.com/nvim-tree/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt#L170

alex-courtis:
> They are always towards the top and require scrolling over to get to more important information. Instruction on how to use plugin managers should (and do) fall on plugin managers themselves

Yes! This is the case for many plugins listing dozens of package manager setups and makes it very difficult to read the readme.

We can't put it in the help... chicken vs egg.

Proposal: create a wiki page for all the different package manager flavours. Users can add to that.

> ## Install
> 
> Install `"nvim-tree/nvim-tree.lua"` and `"nvim-tree/nvim-web-devicons"` (optional) plugins.
>
> See [wiki: TODO plugin manager things](https://github.com/nvim-tree/nvim-tree.lua) for instructions for specific package managers.

alex-courtis:
Done:

https://github.com/nvim-tree/nvim-tree.lua#install

https://github.com/nvim-tree/nvim-tree.lua/wiki/Installation

alex-courtis:
Hopefully other plugins will follow our example ;)

You have inspired me to move build from source instructions to a separate page, for another of my projects.

Issue #2043, opened 2023-03-08
Unrecognized git state "TM"


drdo:
### Description

Ran into this while working on my nvim config repo.

I narrowed it down to having staged changes with a `typechange` changing a file
from a symlink to a regular file and unstaged modifications for the same file.

### Neovim version

```text
NVIM v0.8.3
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

macOS 13.2.1

### nvim-tree version

bbb6d48

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

1. `mkdir repo && cd repo`
2. `git init`
3. `ln -s /tmp/foo foo`
4. `git add foo && git commit -a -m foo`
5. `rm foo && echo "something" > foo`
6. `git add foo`
7. `echo somethingelse >> foo`
8. `nvim -nu /tmp/nvt-min.lua`
9. `:NvimTreeOpen`

### Expected behavior

No error message.

### Actual behavior

Shows an error message with
```
[...]

PR #2045, opened 2023-03-10
feat(git): add `TM` git status

gegoune:
Potential fix for #2043.


drdo:
Should this state show just the unstaged icon?
Since it occurs when there are both staged and unstaged changes should it not show both like some of the other states?

alex-courtis:
> Should this state show just the unstaged icon?
> Since it occurs when there are both staged and unstaged changes should it not show both like some of the other states?

The first T definitely indicates that it is staged, the M indicates modified from staged.

I reckon that's correct. Removing `somethingelse` removes the M. We should be capturing that.

alex-courtis (approved):
This works for TA.

DA is for another day.

PR #2056, opened 2023-03-13
feat(api): add api.config.mappings.get_keymap and get_keymap_default

### LINKED ISSUE #1858: Expose DEFAULT_KEYMAPS And Current Mappings
I requet to export [default keymapping](../lua/nvim-tree/keymap.lua) to ensure stable api .e.g.:

```lua
local   keymaps = require("nvim-tree").DEFAULT_KEYMAPS
...
```

* Well, it can, but the following is not guaranteed to be stable:</br>

```lua
require("nvim-tree.keymap").keymaps
```

> **Note** The callbacks should be exported explicitly!

alex-courtis:
fixes #1858

### DIFF
--- a/lua/nvim-tree/keymap.lua
+++ b/lua/nvim-tree/keymap.lua
@@ -1,5 +1,24 @@
 local M = {}
 
+--- Apply mappings to a scratch buffer and return buffer local mappings
+--- @param fn function(bufnr) on_attach or default_on_attach
+--- @return table as per vim.api.nvim_buf_get_keymap
+local function generate_keymap(fn)
+  -- create a scratch buffer
+  local scratch_bufnr = vim.api.nvim_create_buf(true, true)
+
+  -- apply mappings
+  fn(scratch_bufnr)
+
+  -- retrieve all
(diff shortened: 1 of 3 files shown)


lua/nvim-tree/keymap.lua
```diff
@@ -1,5 +1,24 @@
 local M = {}
 
+--- Apply mappings to a scratch buffer and return buffer local mappings
+--- @param fn function(bufnr) on_attach or default_on_attach
+--- @return table as per vim.api.nvim_buf_get_keymap
+local function generate_keymap(fn)
+  -- create a scratch buffer
+  local scratch_bufnr = vim.api.nvim_create_buf(true, true)
```
mrjones2014, line 8: The parameters here should be `false, true`, as the first parameters is `listed` which sets the `buflisted` option. If we're just using it in the background to get keymaps, it does not need to be in the user's buffer list, especially if we're deleting it right away anyway.

alex-courtis: Thank you so much! Fixed.

alex-courtis:
Tested successfully: https://github.com/nvim-tree/nvim-tree.lua/issues/1858#issuecomment-1467947032

PR #2065, opened 2023-03-20
Fix duplicated help tags

LongerHV:
The help tag `nvim-tree.api.config.mappings.get_keymap()` was specified twice in the latest commit.


gegoune:
Fixed by #2066

PR #2072, opened 2023-03-22
fix(diagnostics): `severity` option do not work with coc

CsYakamoz:
The `diagnostics.severity` option do not work properly with coc

So let it work :)

[Here](https://github.com/neoclide/coc.nvim/blob/d35b5387ba9bfff58863fd0f7919fbb3572d7b03/src/diagnostic/manager.ts#L36) is the response interface of `vim.fn.CocAction "diagnosticList"`

For the old version coc, it do not have `level` field, So I still use `severity` field

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -41,6 +41,10 @@ local function from_nvim_lsp()
   return buffer_severity
 end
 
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Works nicely, many thanks for your contribution.

PR #2084, opened 2023-03-25
fix(#1858): deprecation warning for config.mappings.active and default

alex-courtis:
@hinell I would be most grateful for your review / test.


gegoune (approved):

hinell:
@alex-courtis LGTM. :smile:

PR #2093, opened 2023-03-27
feat(#2092): add api.node.navigate.open.next, prev

### LINKED ISSUE #2092: node.navigate.open.next, prev
As per `node.navigate.git` and `node.navigate.diagnostics`.

This could be achieved by the user via API.

@gegoune is this API feature worthy?

alex-courtis:
fixes #2092

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -868,7 +868,7 @@ UI rendering setup
       Type: `boolean`, Default: `false`
 
     *nvim-tree.renderer.highlight_opened_files*
-    Highlight icons and/or names for opened files using the
+    Highlight icons and/or names for |bufloaded()| files using the
     `NvimTreeOpenedFile` highlight group.
     Value can be `"none"`, `"icon"`, `"name"` or `"all"`.
       Type: `string`, Default: `"none"`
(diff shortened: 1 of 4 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -1446,10 +1446,26 @@ api.tree.toggle_help()                       *nvim-tree.api.tree.toggle_help()*
...
+
+                                   *nvim-tree.api.navigate.diagnostics.next()*
+api.node.navigate.diagnostics.next() 
+    Navigate to the next item showing diagnostic status.
+
+                                   *nvim-tree.api.navigate.diagnostics.prev()*
+api.node.navigate.diagnostics.prev()
+    Navigate to the next item showing diagnostic status.
+
+api.node.navigate.open.next()             *nvim-tree.api.navigate.open.next()*
```
gegoune, line 1464: Not sure is `open` is a good name here. It sorta points toward action rather than a state, such as `diagnostics` or `git`. I don't have a better idea though. `opened`, `buffers` or more verbose like `opened_buffers`, `listed_buffers`. Not sure.

alex-courtis: > `listed_buffers`

That's factual and exactly what we are testing for. Updated.

alex-courtis: Still not right... changed to `opened` to match `highlight_opened_files`.

Clarified that both of these actually show `bufloaded()`.

Noted that `filters.no_buffer` uses `buflisted()`, however that could be considered a breaking change and may also have side effects RE buffer unloading.

PR #2096, opened 2023-03-28
ci: ensure PR subjects follow semantic commit spec

gegoune:


alex-courtis:
Beautiful! This will push me to use proper commit messages.

Issue #1697, opened 2022-10-28
Support for other `vim.notify` wrappers


gegoune:
**Is your feature request related to a problem? Please describe.**
Currently only `nvim-notify` is supported. There are other plugins providing similar functionality, namely [vigoux/notifier.nvim](https://github.com/vigoux/notifier.nvim), [folke/noice.nvim](https://github.com/folke/noice.nvim) or [j-hui/fidget.nvim](https://github.com/j-hui/fidget.nvim).

`nvim-tree` only checks for existence of `nvim-notify` and if it's not installed all messages send via `vim.notify` will have  `[NvimTree]` prefix added (even if user has other plugin installed).

**Describe the solution you'd like**
Check if other notify plugins are installed, or if that is not feasible skip adding prefix and rely on `{title='NvimTree'}.

**Describe alternatives you've considered**
Perhaps some better mechanism exists.

alex-courtis:
Perhaps a notify callback could be configured by the user. Something like:

```lua
function my_notify(msg, level, title)
```

alex-courtis:
We can also add a `notify_once` callback, as we do have a couple of `vim.notify_once` added via #1693

ghostbuster91:
I would like to add another perspective to that feature request. Some people might have `nvim-notify` on their runtime path, but it doesn't necessarily mean that messages should be routed there. For example noice can be used as a message router that routes messages to different notifiers, `nvim-notify` among them. 

In my case I have `nvim-notify` on my runtime path that is not easily remove-able (let's not go into the details :sweat_smile:), but I don't actually want to use it for any kind of notifications. 

Having said that, a callback like `my_notify` would solve the problem.

alex-courtis:
> If no one is working on that at the moment I can give it a try in the upcoming days.

That would be fantastic! More configurable notifications and inputs would greatly improve UX.

PR #2117, opened 2023-04-09
feat: Add ability to customize notification handler

ghostbuster91:
add `notify.handler` to setup opts

refers to #1697

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -440,6 +440,7 @@ applying configuration.
       },
       notify = {
         threshold = vim.log.levels.INFO,
+        handler = "default",
       },
       ui = {
         confirm = {
@@ -1223,6 +1224,19 @@ Configuration for notification.
     `WARNING`: non-fatal errors e.g. unable to system open a file.
     `INFO:`    information only e.g. file copy path confirmation.
     `DEBUG:`   not used.
+
+    *nvim-tree.notify.handler*
+    Change the default notification handler, can be a string `"default"` or a function.
+    The default handler will use |vim.notify()|
+      Type: `string` | `function`, Default: `"default"`
+
+    Parameters: ~
+      • {msg}   (string) See |vim.notify()|
+      • {level} (number) See |vim.notify()|
+      • {opts}  (table)  Optional parameters
+
+    Options: ~
+      • {title} (string) Usually `"NvimTree"`
 
 *nvim-tree.ui*
 General UI configuration.
--- a/lua/nvim-tree/notify.lua
+++ b/lua/nvim-tree/notify.lua
@@ -1,7 +1,12 @@
 local M = {}
 
+local fallback_handler = function(msg, level, opts)
+  vim.notify(string.format("[%s] %s", opts.title, vim.inspect(msg)), level)
+end
+
 local config = {
   threshold = vim.log.levels.INFO,
+  handler = fallback_handler,
 }
 
 local modes = {
@@ -13,19 +18,13 @@ local modes = {
 }
 
 do
-  local has_notify, notify_plugin = pcall(require, "notify")
-
   local dispatch = function(level, msg)
     if level < config.threshold then
       return
     end
 
     vim.schedule(function()
-      if has_notify and notify_plugin then
-        notify_plugin(msg, level, { title = "NvimTree" })
-      else
-        vim.notify(string.format("[NvimTree] %s", vim.inspect(msg)), level)
-      end
+      config.handler(msg, level, { title = "NvimTree" })
     end)
   end
 
@@ -36,9 +35,23 @@ do
   end
 end
 
+local create_default_handler = function()
+  local has_notify, notify_plugin = pcall(require, "notify")
+  if has_notify and notify_plugin then
+    return notify_plugin
+  else
+    return fallback_handler
+  end
+end
+
 function M.setup(opts)
   opts = opts or {}
   config.threshold = opts.notify.threshold or vim.log.levels.INFO
+  if type(opts.notify.handler) == "function" then
+    config.handler = opts.notify.handler
+  else
+    config.handler = create_default_handler()
+  end
 end
 
 return M
--- a/lua/nvim-tree/notify.lua
+++ b/lua/nvim-tree/notify.lua
@@ -1,12 +1,12 @@
 local M = {}
 
-local fallback_handler = function(msg, level, opts)
+local default_handler = function(msg, level, opts)
   vim.notify(string.format("[%s] %s", opts.title, vim.inspect(msg)), level)
 end
 
 local config = {
   threshold = vim.log.levels.INFO,
-  handler = fallback_handler,
+  handler = default_handler,
 }
 
 local modes = {
@@ -35,22 +35,11 @@ do
   end
 end
 
-local create_default_handler = function()
-  local has_notify, notify_plugin = pcall(require, "notify")
-  if has_notify and notify_plugin then
-    return notify_plugin
-  else
-    return fallback_handler
-  end
-end
-
 function M.setup(opts)
   opts = opts or {}
   config.threshold = opts.notify.threshold or vim.log.levels.INFO
   if type(opts.notify.handler) == "function" then
     config.handler = opts.notify.handler
-  else
-    config.handler = create_default_handler()
   end
 end
 
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -629,6 +629,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
   },
   notify = {
     threshold = vim.log.levels.INFO,
+    handler = "default",
   },
   ui = {
     confirm = {
@@ -674,6 +675,7 @@ local FIELD_OVERRIDE_TYPECHECK = {
   sort_by = { ["function"] = true, string = true },
   root_folder_label = { ["function"] = true, string = true },
   picker = { ["function"] = true, string = true },
+  handler = { ["function"] = true, string = true },
 }
 
 local function validate_options(conf)


doc/nvim-tree-lua.txt
```diff
@@ -440,6 +440,7 @@ applying configuration.
       },
       notify = {
         threshold = vim.log.levels.INFO,
+        handler = "default",
```
gegoune, line 443: Do we need that default? Can't it be done "behind scenes" and only used if set to function?

ghostbuster91: I tried to do it in a similar way to how other options were done (e.g. window picker). I am happy to change it if you want.

gegoune: I think it would make sense to have default option implied, with ability to override it.

ghostbuster91: Done, let me know if this is what you had in mind.

alex-courtis: That's kind of a break from existing - everything is specified in `DEFAULT_OPTS`. It's useful if only for the "default schema" that users can base their configuration on.

That's how my local dev harness works: put all the default options in the init lua setup call.

@gegoune are you OK to back that change out? Happy to revisit later; trimming the default options to a skeleton with no values would be less overwhelming.

lua/nvim-tree/notify.lua
```diff
@@ -1,7 +1,12 @@
 local M = {}
 
+local fallbackHandler = function(msg, level, opts)
```
alex-courtis, line 3:
```suggestion
local default_handler = function(msg, level, opts)
```

No camel case please.

lua/nvim-tree/notify.lua
```diff
@@ -1,7 +1,12 @@
...
+  vim.notify(string.format("[%s] %s", opts.title, vim.inspect(msg)), level)
+end
+
 local config = {
   threshold = vim.log.levels.INFO,
+  handler = fallbackHandler,
```
alex-courtis, line 9: Unnecessary, set during setup.

lua/nvim-tree/notify.lua
```diff
@@ -36,9 +35,23 @@ do
   end
 end
 
+local createDefaultHandler = function()
```
alex-courtis, line 38:
```suggestion
local create_handler = function()
```

No camel case please.

alex-courtis (changes requested):
Tested ok both commits:
* default handler
* `rcarriga/nvim-notify`
* `notify.handler` :
```lua
local function h(msg, level, opts)
	print(vim.inspect(msg))
	print(vim.inspect(level))
	print(vim.inspect(opts))
end
```

Please:

- [x] Update help
- [ ] Remove camel case and rename
- [ ] Revert DEFAULT_OPTS pending @gegoune

lua/nvim-tree/notify.lua
```diff
@@ -36,9 +35,23 @@ do
...
+end
+
 function M.setup(opts)
   opts = opts or {}
   config.threshold = opts.notify.threshold or vim.log.levels.INFO
+  if type(opts.notify.handler) == "function" then
+    config.handler = opts.notify.handler
+  else
+    config.handler = createDefaultHandler()
+  end
```
alex-courtis, line 54: This is so much cleaner - many thanks for tidying.

doc/nvim-tree-lua.txt
gegoune: Missed that point. Of course. @ghostbuster91 Please reinstate state before my comment and accept my apologies for causing you extra work.

doc/nvim-tree-lua.txt
```diff
@@ -1224,6 +1224,18 @@ Configuration for notification.
...
+    check if the nvim-notify plugin is present on the runtime path and
+    use it, otherwise it will fallback to use `vim.notify`. 
+      Type: `function` 
+
+    The function will be called with following parameters:
+
+    `msg`:    notification message
+    `level`:  notification level (see `vim.log.levels` for possible values)
+    `opts`:   table with additional parameters (e.g. `title`)
+
```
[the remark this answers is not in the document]
gegoune: Why are we checking for nvim-notify at all? Wasn't idea behind vim.notify such that it handles notifications send to it per users' handler, if it happens to be nvim-notify so be it. 

Not near my computer so can't check much easily.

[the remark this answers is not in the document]
alex-courtis: Oh I see! We set handler directly to vim.notify.

Much cleaner. Just have to ensure that the arguments line up.

doc/nvim-tree-lua.txt
ghostbuster91: No worries, luckily it was just a small change :)

doc/nvim-tree-lua.txt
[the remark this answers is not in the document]
ghostbuster91: > Why are we checking for nvim-notify at all? Wasn't idea behind vim.notify such that it handles notifications send to it per users' handler, if it happens to be nvim-notify so be it.

Yes, that's correct. I did it like that only to preserve the previous behavior, but I would happy to simplify it. Is there any issue/way of notifying users of such change? 

> Much cleaner. Just have to ensure that the arguments line up.

They do :)

lua/nvim-tree/notify.lua
ghostbuster91: Indeed, removed.

ghostbuster91:
- [x] Remove camel case and rename
- [x] Revert DEFAULT_OPTS 
- [x] Update help

Waiting for the confirmation regarding auto detection of nvim-notify

lua/nvim-tree/notify.lua
ghostbuster91: I've now recalled why I did it like that, it was because of that comment https://github.com/nvim-tree/nvim-tree.lua/pull/1693#discussion_r1009294846 

Let me know if I should restore it

alex-courtis: > I've now recalled why I did it like that, it was because of that comment [#1693 (comment)](https://github.com/nvim-tree/nvim-tree.lua/pull/1693#discussion_r1009294846)
> 
> So the notify module could be called before `setup` function gets called, and in such a case a `fallback_handler` would be used. Let me know if I should restore it.

Ah yes, thank you for the diligence.

doc/nvim-tree-lua.txt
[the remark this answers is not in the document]
alex-courtis: Working:
`handler = require("notify").notify,`
and
`handler = vim.notify,`

the latter case does not prepend the title, however that's expected.

doc/nvim-tree-lua.txt
```diff
@@ -1224,6 +1225,20 @@ Configuration for notification.
     `INFO:`    information only e.g. file copy path confirmation.
     `DEBUG:`   not used.
 
+*nvim-tree.notify.handler*
```
alex-courtis, line 1228:
```suggestion
    *nvim-tree.notify.handler*
```

doc/nvim-tree-lua.txt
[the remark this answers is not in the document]
alex-courtis: > Yes, that's correct. I did it like that only to preserve the previous behavior, but I would happy to simplify it. Is there any issue/way of notifying users of such change?

If we were building this today, the user would need to specify `handler = require("notify").notify` however the existing auto-detection behaviour exists and I don't like to break behaviour.

How about:
* `notify.handler` is nil or "default" AND
* `require("notify")` successful

"nvim-notify plugin detected: please set notify.handler = require("notify").notify. See :help nvim-tree.notify"

Options:
1. Preserve existing behaviour, defaulting to nvim-notify
2. Remove nvim-notify fallback, requiring the user to explicitly set `notify.handler`

Your call @gegoune :)

alex-courtis (changes requested):

gegoune:
I am all in for breaking changes if justified, they are not as disruptive as people might think if done properly.

Why are we even checking for nvim-notify? So we can call it with some other parameters/if it't not present to prefix message ourselves? If messages sent change then it barely is a breaking change at all. Am I missing on how that is a breaking change? 

I am all in to setting default handler to `vim.notify` with an option to change it to custom function.

(Is `vim.notify` sender aware? If so users that have a need to change handler per plugin could do it in their handler, we wouldn't need to accommodate it at all. :)

ghostbuster91:
> Why are we even checking for nvim-notify? So we can call it with some other parameters/if it't not present to prefix message ourselves? If messages sent change then it barely is a breaking change at all. Am I missing on how that is a breaking change?

The breaking change I had in mind was regarding the auto-detection of nvim-notify. 
I can imagine that some people relay on that feature but for some reason they don't have the `nvim-notify` attached to `vim.notify` directly. I know that it is probably a very rare case but still. I am fine with whatever decision you make, I just wanted to highlight that.

gegoune:
I agree with the removal. If someone is doing something like that we can very easily assume they know what they are doing and will be able to adapt.

alex-courtis:
How about:
* remove it
* put a breaking change note in the readme
* add a note in the :help about `handler = require("notify").notify`

gegoune:
> add a note in the :help about handler = require("notify").notify

I am not sure about that bit. We should just send stuff to `vim.notify`, which is default mechanism de facto now. Purpose of this PR was to enable users to come up with their own handlers. I will need to review it again, why is it required exactly, as once we just use `vim.notify` users can override that handler in Neovim to their liking.

ghostbuster91:
> I will need to review it again, why is it required exactly, as once we just use vim.notify users can override that handler in Neovim to their liking.

Very much indeed! Because of the small changes in the direction that we did during that PR it was not obvious to spot it, but I think that you are right. Now, once the auto-detection of nvim-notify was removed, we can just simply sent everything to `vim.notify` and it is totally up to the user to configure it to their liking.

gegoune:
Ok, good to know I wasn't imagining things. With that change we wouldn't need a custom handler any more. 

Would you like to amend that, or create new, PR with those changes?

ghostbuster91:
I will create a new PR to keep the things clear and refer to this one.

PR #2123, opened 2023-04-11
feat(#2079): sort_by may return predefined sort

### LINKED ISSUE #2079: Enhance `sort_by` To Return Presets
User can currently define `sort_by` as a preset string or a function.

Allow the function to return a preset string or an order table.

alex-courtis:
fixes #2079 

I might post a Recipe after this is done.


alex-courtis:
@Joemilne I would be most grateful if you tested this change. You could try:

```lua
local name = false
local sort_by = function()
  name = not name
  if name then
    return "name"
  else
    return "modification_time"
  end
end
```

That's not much use by itself; you would need to maintain your own state etc.

petobens:
Hi @alex-courtis I was wondering if this can now be used to "toggle" sorting with a mapping? i.e to address this discussion https://github.com/nvim-tree/nvim-tree.lua/discussions/1922#discussioncomment-4702127

alex-courtis:
> Hi @alex-courtis I was wondering if this can now be used to "toggle" sorting with a mapping? i.e to address this discussion [#1922 (reply in thread)](https://github.com/nvim-tree/nvim-tree.lua/discussions/1922#discussioncomment-4702127)

Please see [Recipe: Cycle Sort Methods](https://github.com/nvim-tree/nvim-tree.lua/wiki/Recipes#cycle-sort-methods)

petobens:
Awesome! Thank you

PR #2124, opened 2023-04-11
feat(#1808): deprecate hide_root_folder

### LINKED ISSUE #1808: Deprecate `nvim-tree.view.hide_root_folder`
We have `nvim-tree.renderer.root_folder_label` which controls what gets displayed as 'root_folder'. Setting it to false (and/or returning empty string) should hide it completely.

That would allow deprecation/removal of 

```
    *nvim-tree.view.hide_root_folder*
    Hide the path of the current working directory on top of the tree.
```

_Originally posted by @gegoune in https://github.com/nvim-tree/nvim-tree.lua/issues/1803#issuecomment-1336477437_

alex-courtis:
fixes #1808

### DIFF
--- a/lua/nvim-tree/legacy.lua
+++ b/lua/nvim-tree/legacy.lua
@@ -43,6 +43,12 @@ local function refactored(opts)
   end
 end
 
+local function deprecated(opts)
+  if opts.view.hide_root_folder then
+    notify.info "view.hide_root_folder is deprecated, please set renderer.root_folder_label = false"
+  end
+end
+
 local function removed(opts)
   if opts.auto_close then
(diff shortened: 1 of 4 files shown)


lua/nvim-tree/legacy.lua
```diff
@@ -43,6 +43,12 @@ local function refactored(opts)
   end
 end
 
+local function deprecated(opts)
+  if opts.view.hide_root_folder then
+    notify.info "view.hide_root_folder is deprecated, please set renderer.root_folder_label = false"
```
alex-courtis, line 48: I've gone with info here. It's still functional for now.

PR #2131, opened 2023-04-12
feat: checking for slashed suffix

bragamat:


lua/nvim-tree/actions/fs/rename-file.lua
```diff
@@ -15,12 +15,28 @@ local function err_fmt(from, to, reason)
   return string.format("Cannot rename %s -> %s: %s", from, to, reason)
 end
 
+local function create_parent_dir(path)
+  local parent_dir = utils.path_remove_trailing(path:match("^(.*)/[^/]*$") or ".")
```
gegoune, line 19: How will that work on Windows? Are you able to test it?

alex-courtis:
What exactly is the problem we are trying to solve?

Please provide a reproducer so that we may better understand.

alex-courtis:
Incomplete. Please reopen and raise an issue with a replicator.

PR #2140, opened 2023-04-15
fix: custom window pickers that create windows should work

### LINKED ISSUE #2126: If a custom window_picker creates a new window for this buffer, nvim-tree doesn't use that new window
### Description

I have a fork of nvim-window-picker (https://github.com/IndianBoy42/nvim-window-picker/tree/pick_or_create) that supports a pick_or_create function. This would create a window if the correct labels are used and returns that window id.  I tried to use that for the nvim-tree window picker:
```
  actions = {
    open_file = {
      window_picker = {
        picker = function() return require("window-picker").pick_or_create() end,
      },
    },
  },
```
The problem is that it assumes this window id is invalid because it didn't exist before picker was called, so it just creates a new window. I don't think this behaviour is a good idea, at most a warning could be given if a window is chosen that seems invalid. If the default picker is used then by construction the window id is invalid. If a custom picker is used then clearly that's what the user wanted, or its a bug in the picker that nvim-tree doesn't have to try and work around. The only condition that the window id should meet is that its possible to switch to it and run `:edit`, this condition can be checked by just trying to do those things within pcall

I can submit a PR for this I just wanted to know what approach you think is right? how much 'error checking' should be done on the target_winid returned by pickers

### Neovim version

```text
NVIM v0.9.0-dev-1350+gdd80ee0ca
Build type: Release
LuaJIT 2.1.0-beta3
```

6.2.6-76060206-generic

### nvim-tree version

d1410cb

### Minimal config

```lua
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			"nvim-tree/nvim-tree.lua",
			{ "indianboy42/nvim-window-picker", branch = "release/2.0" },
		},
		config = {
			package_root = package_root,
		},
	})
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
	require("window-picker").setup()
	require("nvim-tree").setup({
		actions = {
			open_file = {
				window_picker = {
					picker = function()
						return require("window-picker").pick_or_create()
					end,
				},
			},
		},
	})
end
```

### Steps to reproduce

1. `nvim -u /tmp/nvt-min.lua`
2. `:NvimTreeOpen`
3. navigate to any file and `<CR>`
4. Use the labels to create a new window, one of  `hjkl` if only one existing window or `hjkl+` a label at the center of a window

### Expected behavior

New window is created in the direction specified by the labels (this happens because window-picker is responsible for it), and the file chosen in nvim-tree is opened in that new window

### Actual behavior

A new window right next to the nvim-tree is created with the file chosen in nvim-tree

[Screencast from 04-11-2023 09:40:57 PM.webm](https://user-images.githubusercontent.com/5981889/231181850-75082b4f-7382-47dd-aec4-a1087846ccd4.webm)

the labels I used for the picker were `j` and `js` respectively, you can see the new window created but the file going to a completely new window

IndianBoy42:
call nvim_list_wins again after the picker is run

also show a warning if it doesn't exist. I don't see why this would happen unless its a bug in the picker, perhaps something else snuck in and deleted the window

Fixes #2126

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -164,13 +164,13 @@ local function on_preview(buf_loaded)
   view.focus()
 end
 
-local function get_target_winid(mode, win_ids)
+local function get_target_winid(mode)
   local target_winid
   if not M.window_picker.enable or mode == "edit_no_picker" then
     target_winid = lib.target_winid
 
     -- first available window
-    if not vim.tbl_contains(win_ids, target_winid) then
+    if not vim.tbl_contains(vim.api.nvim_tabpage_list_wins(0), target_winid) then
       target_winid = first_win_id()
     end
   else
@@ -201,20 +201,21 @@ local function set_current_win_no_autocmd(winid, autocmd)
   vim.opt.eventignore = eventignore
 end
 
-local function open_in_new_window(filename, mode, win_ids)
+local function open_in_new_window(filename, mode)
   if type(mode) ~= "string" then
     mode = ""
   end
 
-  local target_winid = get_target_winid(mode, win_ids)
+  local target_winid = get_target_winid(mode)
   if not target_winid then
     return
   end
 
-  local create_new_window = #vim.api.nvim_list_wins() == 1
+  local win_ids = vim.api.nvim_list_wins()
+  local create_new_window = #win_ids == 1 -- This implies that the nvim-tree window is the only one
   local new_window_side = (view.View.side == "right") and "aboveleft" or "belowright"
 
-  -- Target is invalid or window does not exist in current tabpage: create new window
+  -- Target is invalid: create new window
   if not vim.tbl_contains(win_ids, target_winid) then
     vim.cmd(new_window_side .. " vsplit")
     target_winid = vim.api.nvim_get_current_win()
@@ -287,8 +288,6 @@ function M.fn(mode, filename)
     return edit_in_current_buf(filename)
   end
 
-  local tabpage = vim.api.nvim_get_current_tabpage()
-  local win_ids = vim.api.nvim_tabpage_list_wins(tabpage)
   local buf_loaded = is_already_loaded(filename)
 
   local found_win = utils.get_win_buf_from_path(filename)
@@ -297,7 +296,7 @@ function M.fn(mode, filename)
   end
 
   if not found_win then
-    open_in_new_window(filename, mode, win_ids)
+    open_in_new_window(filename, mode)
   else
     vim.api.nvim_set_current_win(found_win)
     vim.bo.bufhidden = ""
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1121,6 +1121,7 @@ Configuration for various actions.
             Change the default window picker, can be a string `"default"` or a function.
             The function should return the window id that will open the node,
             or `nil` if an invalid window is picked or user cancelled the action.
+            The picker may create a new window.
               Type: `string` | `function`, Default: `"default"`
               e.g. s1n7ax/nvim-window-picker plugin: >
                 window_picker = {


IndianBoy42:
@alex-courtis do you see any problems with this approach? I don't think this could break any existing users.

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -211,11 +211,16 @@ local function open_in_new_window(filename, mode, win_ids)
     return
   end
 
-  local create_new_window = #vim.api.nvim_list_wins() == 1
+  local win_list = vim.api.nvim_list_wins()
+  local create_new_window = #win_list == 1 -- This implies that the nvim-tree window is the only one
   local new_window_side = (view.View.side == "right") and "aboveleft" or "belowright"
 
-  -- Target is invalid or window does not exist in current tabpage: create new window
-  if not vim.tbl_contains(win_ids, target_winid) then
```
alex-courtis, line 218: We can remove the `win_ids` parameter.

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -211,11 +211,16 @@ local function open_in_new_window(filename, mode, win_ids)
...
+  -- Target is invalid: create new window
+  if not vim.tbl_contains(win_list, target_winid) then
+    vim.notify(
+      "That window doesn't exist, either this is a bug in the picker or it was just deleted",
```
alex-courtis, line 221: I don't think we need this; the fallback is sufficient.

alex-courtis (changes requested):
Looks great, thank you.

Please:

- [ ] remove win_ids
- [ ] remove warning

lua/nvim-tree/actions/node/open-file.lua
IndianBoy42: The silent fallback was actually quite annoying when I was trying to figure this out, and as I said, I don't see how we reach this case without a bug in the picker or some very exceptional circumstance. I was even wondering if we should call the picker again (for the case where the window somehow gets deleted between picking and using)

alex-courtis: I understand, however users are very sensitive about notifications. There might be a use case that we are not aware of.

> I was even wondering if we should call the picker again (for the case where the window somehow gets deleted between picking and using)

Let's keep it simple. If it becomes a problem we can revisit.

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -164,13 +164,13 @@ local function on_preview(buf_loaded)
...
 
-local function get_target_winid(mode, win_ids)
+local function get_target_winid(mode)
   local target_winid
   if not M.window_picker.enable or mode == "edit_no_picker" then
     target_winid = lib.target_winid
 
     -- first available window
-    if not vim.tbl_contains(win_ids, target_winid) then
+    if not vim.tbl_contains(vim.api.nvim_tabpage_list_wins(0), target_winid) then
```
alex-courtis, line 173: :+1: 0 is current tabpage

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -288,7 +289,6 @@ function M.fn(mode, filename)
   end
 
   local tabpage = vim.api.nvim_get_current_tabpage()
```
alex-courtis, line 291: unused

alex-courtis (changes requested):

lua/nvim-tree/actions/node/open-file.lua
IndianBoy42: do you want me to explicitly get_current_tabpage? i see that's how its done in some other places in the code but I don't really know why when 0 works

alex-courtis: That's fine. It was probably a copy/paste from a location that was also using tabpage.

alex-courtis (approved):

IndianBoy42:
no problem! glad to contribute to the neovim ecosystem

PR #2164, opened 2023-04-23
feat(api): add `node.open.drop()`

gegoune:


gegoune:
Similar to `tab_drop`. Has some jarring/redrawing issue at the moment. @alex-courtis Any idea why that might be?

doc/nvim-tree-lua.txt
```diff
@@ -1545,6 +1545,15 @@ node.open.vertical()                      *nvim-tree-api.node.open.vertical()*
 node.open.horizontal()                  *nvim-tree-api.node.open.horizontal()*
     |nvim-tree-api.node.edit()|, file will be opened in a new horizontal split.
 
+node.open.drop()                               *nvim-tree-api.node.open.tab()*
```
alex-courtis, line 1548:
```suggestion
node.open.drop()                              *nvim-tree-api.node.open.drop()*
```

doc/nvim-tree-lua.txt
```diff
@@ -1545,6 +1545,15 @@ node.open.vertical()                      *nvim-tree-api.node.open.vertical()*
...
+    Switch to window with selected file if it exists.
+    Open file in current window otherwise.
```
alex-courtis, line 1550: This is strictly correct, however `view._prevent_buffer_override()` will always fire to eject the newly opened file.

How about "Open file otherwise"?

alex-courtis (approved):
Definitely works:
`open_file.quit_on_open`
`update_focused_file`

Same behaviour as `:e something` in the tree window.

`view._prevent_buffer_override()` is firing. That is the source of the jarring.

alex-courtis:
> Similar to `tab_drop`. Has some jarring/redrawing issue at the moment. @alex-courtis Any idea why that might be?

See https://github.com/nvim-tree/nvim-tree.lua/pull/2164#pullrequestreview-1397144103

Idea:
* search buffers for file
* determine whether a window shows that buffer
* shown: drop
* not shown: open as normal

I'm not sure that's really worthwhile - adds complexity, as we are second-guessing `:drop`.

gegoune:
Another idea here is to generalise open functionality. Accept callback/function as an argument which will be responsible for opening files, users then will be able to do whatever they want and we can provider wrappers for currently existing functionality.

This would work very well for niche case such as `tab_drop` which seems very much workflow oriented (missing vim's tabs ;).

alex-courtis:
> Something like
> 
> ```lua
> node.open(function(node) vim.cmd("drop " .. node end)
> ```

I like that. Fits with our "API instead of specific functionality" push.

The other `node.open` follow `|nvim-tree.actions.open_file|` however there is no need here.

It's more generic than that even. It's a general "do something with a node" function. Maybe `node.run.fn` ?

alex-courtis:
> niche case such as `tab_drop` which seems very much workflow oriented (misusing vim's tabs ;).

We can remove that from documented API if we leave it in there, deprecated.

gegoune:
That definitely makes sense! Can you see any issues with functionality such as 'close on open'?

doc/nvim-tree-lua.txt
gegoune: Updated, good suggestion, thanks.

alex-courtis:
> That definitely makes sense! Can you see any issues with functionality such as 'close on open'?

Anything _could_ go wrong, depending on what the user does in the function.

Providing it works for the basics like `":e " .. node.absolute_path` we should be fine: "eject the buffer, quit nvim-tree" would be known to work. We know that `:drop` is OK.

This opens up all sorts of options for the user with no additional plumbing required.

alex-courtis:
@gegoune I'm happy to merge this if you are; it works and is consistent with the other `node.open` actions.

PR #2180, opened 2023-05-02
fix: fix type hint to add nil for bufnr of utils.is_nvim_tree_buf()

kang8:


gegoune (approved):

alex-courtis (approved):
Many thanks for picking this up!

Issue #2168, opened 2023-04-26
Tree is not reloading after performing actions


mskelton:
### Description

When performing actions in the tree (e.g. rename or delete), it is not refreshing to show the new name or hide the deleted item.

This previously did work, but broke at some point recently. The issue occurs with a stock installation of nvim-tree or a custom setup.

### Neovim version

```text
NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

macOS 13.2.1

### nvim-tree version

bb375fb

### Minimal config

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

1. nvim -nu /tmp/nvt-min.lua
2. :NvimTreeOpen
3. Delete a file using `d`
4. Notice that the file still remains in the list until you reload with `R`

alex-courtis:
I can replicate this with my working nvim-tree repository, however not with a fresh clone.

```
[...]

PR #2012, opened 2023-02-19
fix(#1976): improve `$GIT_DIR` handling

dmehala:
- Retrieve $GIT_DIR using `git rev-parse --absolute-git-dir`
- Move $GIT_DIR ignore in the project exploration part

Resolves #1976

Issue #1976, opened 2023-02-06
File system watcher throws error when opening git repo with $GIT_DIR set


austin5627:
### Description

When opening nvim in a git repo with `$GIT_DIR` set to something other than `.git` nvim-tree throws the error `[NvimTree] "Could not start the fs_event watcher for path <path_to_repo>/.git : ENOENT"`

### Neovim version

```text
NVIM v0.8.3
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux 6.1.9-arch1-1

### nvim-tree version

02fdc26

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

1. Make a git repo with `GIT_DIR=test git init`
2. Open nvim with `GIT_DIR=test GIT_WORK_TREE=. nvim`
3. Open NvimTree

alex-courtis:
Confirmed. That has been hardcoded.

We should use `git -C /path/to/repo rev-parse --absolute-git-dir`

mangelozzi:
Experiencing the same thing

alex-courtis:
[...]

Issue #2104, opened 2023-04-03
Experimental Features - Git Async - Enabled


alex-courtis:
# Experimental Features

nvim-tree has experimental features, previews of new functionality that will become default or optional functionality.

Please try these out and reply with your experiences. If you encounter a problem, please raise a [bug report](https://github.com/nvim-tree/nvim-tree.lua/issues/new?assignees=&labels=bug&template=bug_report.yml) and indicate the experimental feature you are using.

## Background Git Refresh

**2023 04 22** enabled by default

**2023 05 27** option removed

Direct file writes and `.git/` writes are executed asynchronously: the git process runs in the background. The tree updates on completion.
Other git actions such as first tree draw and explicit refreshes are still done in the foreground.

```lua
  experimental = {
    git = {
      async = true,
    },
  },
```

PR #2194, opened 2023-05-07
fix(#2168): Tree not reloading

dmehala:
The PR contains:
- #1976: no differences with #2012
- Fix for #2168 

## The issue
`is_folder_ignored` used to process the input in two steps. First, check the path against kernel filesystems with `vim.startswith`, then against any entry in the configuration file using `vim.fn.match`.

With #2012, kernel filesystems and user ignored directories share the same list and are fed to `vim.fn.match`. At that point I think you understand the issue, I forgot to transform the kernel filesystem entries to regex pattern. That lead to any path containing "/dev", "/sys" and "/proc" falsely marked as ignored.

### DIFF
--- a/lua/nvim-tree/explorer/watch.lua
+++ b/lua/nvim-tree/explorer/watch.lua
@@ -4,11 +4,7 @@ local Watcher = require("nvim-tree.watcher").Watcher
 
 local M = {}
 
-local function is_git(path)
-  return vim.fn.fnamemodify(path, ":t") == ".git"
-end
-
-local IGNORED_PATHS = {
+M.ignore_dirs = {
   -- disable watchers on kernel filesystems
   -- which have a lot of unwanted events
   "/sys",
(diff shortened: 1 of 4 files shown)


alex-courtis:
Thanks @dmehala , soaking for a week.

alex-courtis:
I have encountered an issue with regular file system watchers not updating the tree following a git commit.

I'll enable logging and try to replicate.

alex-courtis:
I just do not know what to do here:
I'm not able to get to the bottom of the issue I've seen only twice.
I don't want to release this and experience another regression.

Options to reduce the blast radius:
1. Release this as an experiment: #2104
2. Put this behind an option

I'm leaning towards 2, as it removes the extra git invocation for users that don't use it.

```help
    *nvim-tree.git.respect_git_dir*
    Attempt to resolve `$GIT_DIR`, falling back to `.git`
```

alex-courtis (changes requested):
Let's put this one behind an option please.

alex-courtis:
I reckon this is fixed by #2263

We can reopen if needed.

PR #2200, opened 2023-05-09
docs: add disclaimer for disabling netrw

howarddo2208:
I was confused how my `gx` key bind didn't work, but couldn't find it stated in the docs. Instead, I need to read through the issue section. So, I added these for later users who will be in the same situation as I was.


gegoune:
There is way more functionality lost when disabling netrw. Not sure we should/can list everything here. Maybe some more generic link to netrw help/features instead?

alex-courtis:
I wish netrw would just Go Away. The decision to remove it is stuck in decision-by-committee.

> Maybe some more generic link to netrw help/features instead?

That would be good, especially from the perspective of a user who actually gains value from it.

howarddo2208:
I just edited the disclaimer to look more general and added the link to netrw docs. Is it ok to merge now?

alex-courtis:
I don't want to pollute the readme or quickstart; netrw is a very niche feature that few actually use.

I've given it a section at the end of the help: #2222

Please review @howarddo2208

alex-courtis:
Resolved via #2222

Issue #2221, opened 2023-05-21
Generated :help Pages


alex-courtis:
[tree-sitter-vimdoc](https://github.com/neovim/tree-sitter-vimdoc) can generate pages like https://neovim.io/doc/user/pi_netrw.html

This can be added to the readme and all `:help` links pointed to specific sections of that doc.

alex-courtis:
Needs very specifically customised [neovim source](https://github.com/neovim/neovim/blob/master/scripts/gen_help_html.lua); too difficult to add to the devloop.

PR #2222, opened 2023-05-21
docs: add nvim-tree-netrw

alex-courtis:
replaces #2200

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,10 +46,12 @@ Please install via your preferred package manager. See [Installation](https://gi
 
 ## Quick Start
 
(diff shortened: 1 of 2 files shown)


gegoune (approved):
Better than readme. But perhaps let's link to it from readme?

howarddo2208:
yeah, I think there need to be something in readme, beginners like me won't dig deep at first.

alex-courtis:
Readme link is a good idea. Needs #2221 but I can at least get _a_ link in there for now.

gegoune (approved):
Looks great, thanks!

PR #2232, opened 2023-05-25
feat: support vim.diagnostic.is_disabled()

sstallion:
This PR adds support for checking `vim.diagnostic.is_disabled()` on nvim 0.9+. This function was added to support globally enabling or disabling diagnostics. This is especially useful for setups where diagnostics are enabled conditionally or on a project basis. There is no change in behavior for versions prior to 0.9.

Thanks for a great plugin!

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -27,13 +27,20 @@ end
 local function from_nvim_lsp()
   local buffer_severity = {}
 
-  for _, diagnostic in ipairs(vim.diagnostic.get(nil, { severity = M.severity })) do
-    local buf = diagnostic.bufnr
-    if vim.api.nvim_buf_is_valid(buf) then
-      local bufname = vim.api.nvim_buf_get_name(buf)
-      local lowest_severity = buffer_severity[bufname]
-      if not lowest_severity or diagnostic.severity < lowest_severity then
-        buffer_severity[bufname] = diagnostic.severity
+  local is_disabled = false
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Works nicely. We don't get a `DiagnosticChanged` event until switching buffers or writing, but that's Good Enough: the tree is refreshed reasonably quickly.

alex-courtis (changes requested):
Please resolve styling issue.

sstallion:
Done. Thanks again!

sstallion:
I wonder if adding the following autocommands (from `lsp.txt`) would be enough:
```
autocmd User LspProgressUpdate redrawstatus
autocmd User LspRequest redrawstatus
```
I'll play around with this a bit today to see if this can close the gap.

alex-courtis:
> I'll play around with this a bit today to see if this can close the gap.

That would be great, thank you. Instant updates would look great.

sstallion:
> That would be great, thank you. Instant updates would look great.

I've been looking into this, this evening and it looks like updates may already be instant. After `from_nvim_lsp()` is called, signs will be cleared, which should achieve the desired result.

Would it be okay to merge this PR as-is? I'm not seeing much else that can be done to improve the responsiveness at this point.

alex-courtis:
> Would it be okay to merge this PR as-is? I'm not seeing much else that can be done to improve the responsiveness at this point.

That's fine. The UX is great.

alex-courtis (approved):

sstallion:
My pleasure. Thanks for a great plugin!

PR #2238, opened 2023-05-28
fix(#1545): dispatch Event.Resize on all window resizes, requires nvim 0.9+

### LINKED ISSUE #1545: Resize event not dispatched when resizing using mouse

alex-courtis:
fixes #1545

@wlh320 @Tcholly I'd be most grateful if you tested this fix:

```sh
cd /path/to/nvim-tree.lua
git pull
git checkout 1545-resize-events
```

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -15,6 +15,7 @@ local modified = require "nvim-tree.modified"
 local keymap_legacy = require "nvim-tree.keymap-legacy"
 local find_file = require "nvim-tree.actions.tree.find-file"
 local open = require "nvim-tree.actions.tree.open"
+local events = require "nvim-tree.events"
 
 local _config = {}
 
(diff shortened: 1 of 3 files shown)


lua/nvim-tree.lua
```diff
@@ -337,6 +338,21 @@ local function setup_autocommands(opts)
       end,
     })
   end
+
+  -- TODO #1545 remove similar check from view.resize
+  if vim.fn.has "nvim-0.9" == 1 then
+    create_nvim_tree_autocmd("WinResized", {
+      callback = function()
+        if vim.v.event and vim.v.event.windows then
```
gegoune, line 346: Not sure if it makes any difference but fallback receives an event afair.

alex-courtis: Do you mean the fallback in view.lua?

Sending multiple events would be undesirable...

gegoune: > Do you mean the fallback in view.lua?
> 
> 
> 
> Sending multiple events would be undesirable...

No, meant that you should be able to do 

```lua
callback = function(event)
```

alex-courtis: Of course, I'll give that a go.

alex-courtis: Unfortunately no. The `v:event` data is still not passed to the callback. I hope they add that at some point...

alex-courtis:
nudge @wlh320 @Tcholly

I'd be most grateful if you tested this fix:

```sh
cd /path/to/nvim-tree.lua
git pull
git checkout 1545-resize-events
```

PR #2247, opened 2023-06-01
refactor: move `reload` function into `utils` module

Akmadan23:
This is the refactor we discussed in #2202, it will be useful for future fixes.
I also noticed that I called `utils.focus_file` when it was unnecessary, so I replaced it with a simple `view.set_cursor`.

### DIFF
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -354,6 +354,10 @@ function M.focus_file(path)
   require("nvim-tree.view").set_cursor { i + 1, 1 }
 end
 
+---Focus node passed as parameter if visible, otherwise focus first visible parent.
+---If none of the parents is visible focus root.
+---If node is nil do nothing.
+---@param node table|nil node to focus
 function M.focus_node_or_parent(node)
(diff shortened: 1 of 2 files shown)


lua/nvim-tree/utils.lua
```diff
@@ -354,6 +354,27 @@ function M.focus_file(path)
   require("nvim-tree.view").set_cursor { i + 1, 1 }
 end
 
+function M.focus_node_or_parent(node)
```
alex-courtis, line 357: Please add doc annotations. See `is_nvim_tree_buf` for an example.

alex-courtis (changes requested):
Many thanks; structural improvements are always very welcome.

lua/nvim-tree/utils.lua
Akmadan23: Sure, I didn't do it because most of the other functions don't have them. Will all of them eventually have annotations?

alex-courtis: Many thanks. Ideally we would like documentation everywhere but that will be a slow process...

alex-courtis (approved):

PR #2254, opened 2023-06-03
fix(#2250): Prevent filename from appearing in message box after picking a window.

### LINKED ISSUE #2250: Selecting a window with the "Pick window" prompt causes a "Press enter or type command to continue"
### Description

My cmdheight=1; when I have multiple windows open and select a file in nvim-tree it asks me to select a window ("Pick window" prompt), but then when I pick one, it causes neovim to show the "Press ENTER or type command to continue", I believe because my one-line message window now has two lines in it, one from the printed "Pick window", and another from the name of the file being printed.  Any way that we can fix this?  Ideally I'd like for the file to open in the window that I choose immediately after selecting the window, and don't want to have to hit enter again.

### Neovim version

```text
NVIM v0.9.1
```

Pop!_OS

### nvim-tree version

latest

### Minimal config

```lua
I don't think this is relevant here, but if someone wants it I can provide it.
```

### Steps to reproduce

Open a few vertical splits, open nvim-tree, select a file, then select a window, and then you have to hit enter before the file will open.

### Expected behavior

_No response_

### Actual behavior

_No response_

no-more-secrets:
Fixes #2250.

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -282,7 +282,7 @@ local function open_in_new_window(filename, mode)
   -- Prevents the filename from being displayed in the message
   -- box and thus causing a 'Press enter to continue...' when the
   -- cmdheight=1.
-  cmd = 'silent! ' .. cmd
+  cmd = "silent! " .. cmd
 
   if mode == "preview" and view.View.float.enable then
     -- ignore "WinLeave" autocmd on preview
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -279,6 +279,10 @@ local function open_in_new_window(filename, mode)
   else
     cmd = string.format("edit %s", fname)
   end
+  -- Prevents the filename from being displayed in the message
+  -- box and thus causing a 'Press enter to continue...' when the
+  -- cmdheight=1.
+  cmd = 'silent! ' .. cmd
 
   if mode == "preview" and view.View.float.enable then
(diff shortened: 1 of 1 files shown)


alex-courtis (commented):
This definitely works.

Unfortunately I'm not able to replicate the problem in a [Clean Room](https://github.com/nvim-tree/nvim-tree.lua/wiki/Troubleshooting#clean-room-replication)

Please:
- [ ] provide a clean room replicator
- [ ] resolve style issue

no-more-secrets:
Ok so I tried the clean room reproducer, and I got the same results as you (not able to reproduce).  That lead me to find an issue in one of my other config files that, when fixed, makes the problem go away.  The issue in question was that I was doing:
```lua
vim.opt.shortmess = 'I'
```
when it should have been:
```lua
vim.opt.shortmess:append( 'I' )
```
It turns out that when you clear a couple of options from `shortmess` it causes more output to the message box when a file is opened.  So therefore you can reproduce this by doing:
```
set cmdheight=1
set shortmess=
```
However, I don't think it is a problem anymore, so we can probably close this without merging, but I will let you decide.

alex-courtis:
> It turns out that when you clear a couple of options from shortmess it causes more output to the message box when a file is opened.  So therefore you can reproduce this by doing:
> ```

Nice digging.

> However, I don't think it is a problem anymore, so we can probably close this without merging, but I will let you decide.

Let's close. It is still useful to show messages when, say, a file cannot be opened, is open elsewhere etc.

alex-courtis:
Many thanks for your PR - future PRs are most gratefully appreciated.

PR #2261, opened 2023-06-10
fix: focus visible parent on collapse all

Akmadan23:
Here's a simple fix analogous to the ones made to toggles and live filter.


alex-courtis (approved):
Works beautifully! Many thanks for all your changes...

PR #2265, opened 2023-06-11
fix(#2243): fs_watcher not started for .git folders (Win)

### LINKED ISSUE #2243: fs_event watcher not started for .git folder on Windows
#### The problem

While starting the plugin for a git repo, it tells me that no fs_event watcher could be started for the `.git` folder.

The error states the path as `C:\Users\anton\Documents\ActivityRadar\backend\.git`, where the weird character before the last backslash seems to be the problem.

#### Tracking it down

I tracked the creation of the path back and found out, that the [git command to retrieve the project root](https://github.com/nvim-tree/nvim-tree.lua/blob/e2a4c9d09d205ebe5f071264f43f73a0077c43a3/lua/nvim-tree/git/utils.lua#LL9C22-L9C22) returns a newline.

This is of course handled at the end of that function by removing the trainling newline. But my problem lies in having msys2 installed and so the plugin thinks I am a cygwin user (which I am not). In [this call](https://github.com/nvim-tree/nvim-tree.lua/blob/e2a4c9d09d205ebe5f071264f43f73a0077c43a3/lua/nvim-tree/git/utils.lua#L25) the `cygpath` program does something weird with the trailing whitespace which is not yet removed (namely, producing the weird trailing character).

#### My thoughts on solutions

One easy thing that would fix the issue (at least as far as I see it), would be to remove the trailing newline directly after the `vim.fn.system` call. Like so:
```diff
-local toplevel = vim.fn.system(cmd)
+local toplevel = vim.fn.system(cmd):gsub("\n", "")
```
That way it doesnt get screwed up by `cygpath`.

Another approach would be to try fixing it on the cygpath side. As that use case is very niche, I'd prefer to opt for the first approach. As the one requesting the msys2-support (https://github.com/nvim-tree/nvim-tree.lua/issues/1290), @bstaint, could you tell us if you have the same issue currently? :)

### Neovim version

```text
NVIM v0.9.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

Win 11

### nvim-tree version

9ef6c3c

### Minimal config

```lua
-- I'll provide this if necessary
```

### Steps to reproduce

...

### Expected behavior

_No response_

### Actual behavior

_No response_

akriese:
Resolves #2243
The fix happens inside a `if vim.fn.has "win32"` block and thus, does not affect WSL.
If necessary, I could install cygwin to test it on that too. Also, I wasn't sure, what test case (in CONTRIBUTING.md) meant, as I didnt find a test setup for the repo. The issue describes the exact problematic case and why this fix works for it. If more description is needed, please tell me :)

### DIFF
--- a/lua/nvim-tree/git/utils.lua
+++ b/lua/nvim-tree/git/utils.lua
@@ -25,7 +25,7 @@ function M.get_toplevel(cwd)
   if vim.fn.has "win32" == 1 then
     -- msys2 git support
     if has_cygpath then
-      toplevel = vim.fn.system("cygpath -w " .. vim.fn.shellescape(toplevel))
+      toplevel = vim.fn.system("cygpath -w " .. vim.fn.shellescape(toplevel:sub(0, -2)))
       if vim.v.shell_error ~= 0 then
         return nil
       end


alex-courtis (approved):
Thank you for your contribution.

This is behind the windows feature flags and will only affect a small number of users.

I cannot review or test this one; I assume it is working for you.

Please be prepared to resolve any issues that may arise from this change.

PR #2280, opened 2023-06-18
feat(#2270): option to use item names with fs action notifications

### LINKED ISSUE #2270: Option to use file/directory names instead of absolute paths for notifications
### Issue
When using a notification plugin like [vim-notify](https://github.com/rcarriga/nvim-notify) that has floating window notifications, using absolute paths in the messages of the notifications take up too much of the available horizontal window space, which could be quite distracting. 

### Feature Request
Having shorter notification messages that will take up _less_ of the window space when using notification manager plugin with floating windows. 

### Potential Solution
Adding a boolean option to `opts.notify` that will be set to *false* by **default**. When enabled it will have the file system actions use file/directory names in the notification messages, and use absolute paths by default. 

### Drawbacks
When the option for using file/directory names is enabled, people who are using `nvim-tree.api`, more specifically the file system actions API, without visual context of the current file system, the notification messages might be quite ambiguous if there are files or directories with the same name within their file system. However, I don't believe that this is a significant limitation by any means, since the proposed solution will **not** be a breaking change.   

### References
#2266

asror1:
fixes #2270 

I've tested this new option with _create_, _rename_, _copy-paste_, and _remove_ file system actions, and they are working great for me. However, please do check if everything works as indented. 

Also... I just noticed I've changed ➜ arrow within the rename notification to  because using neovim on powershell with a **ttf** font doesn't have the best support for wider glyphs. But this was not intended to be in this feature implementation, if this is something that will be a breaking change for some, I'm more than happy to revert to "➜" and submit a feature request to add an option to let users change that specific arrow?


lua/nvim-tree/utils.lua
```diff
@@ -38,6 +38,14 @@ function M.path_split(path)
   return path:gmatch("[^" .. path_separator .. "]+" .. path_separator .. "?")
 end
 
+--- Retrieve last element of a path.
+--- @param path string Path to retrieve last element from.
+--- @return string Last element of the path.
+function M.get_last_path_elem(path)
```
alex-courtis, line 44: Could we please use `fnamemodify( , ":t")`

lua/nvim-tree.lua
```diff
@@ -360,7 +360,8 @@ local function setup_autocommands(opts)
   end
 end
 
-local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
+local DEFAULT_OPTS = {
+  -- BEGIN_DEFAULT_OPTS
```
alex-courtis, line 364: Please don't move these around; the help update is pretty primitive and expects exact formatting ;)

lua/nvim-tree/actions/fs/create-file.lua
```diff
@@ -6,12 +6,15 @@ local notify = require "nvim-tree.notify"
...
-local M = {}
+local M = {
+  config = {},
+}
 
 local function create_and_notify(file)
   local ok, fd = pcall(vim.loop.fs_open, file, "w", 420)
   if not ok then
-    notify.error("Couldn't create file " .. file)
+    local notify_file = M.config.notify.absolute_path and file or utils.get_last_path_elem(file)
```
alex-courtis, line 16: We are doing this a lot; could we extract a function to do this? `utils.lua` would be a good place.

alex-courtis (changes requested):
Looking great, thank you!

I'll give it a full test after changes have been addressed.

lua/nvim-tree.lua
asror1: Sorry about that! I think there might be an issue with my `stylua`, potentially because I'm on Windows... but its been telling me that **all** the files in the repository aren't formatted correctly even besides the files I've changed. And moving  "-- BEGIN_DEFAULT_OPTS" was stylua as well. I will try to see if it's possible for me to fix stylua on my end somehow.

lua/nvim-tree/utils.lua
asror1: Will do!

lua/nvim-tree/actions/fs/create-file.lua
asror1: Yea absolutely. If I understand correctly though, the function should take in a _boolean_ and an absolute path, and return the last element or the path itself based on the boolean value?

lua/nvim-tree.lua
asror1: Whoops, this is silly but it was actually my LSP formatter that was messing with the comment. I thought it was stylua because I also formatted `nvim-tree.lua` through stylua, but nope not stylua's fault there. And I found out why stylua was telling me that all the files weren't formatted correctly, which was because my git was configured to use CRLF line endings (it was probably set to auto), I've changed to use LF line endings instead.

alex-courtis: No worries. We can't even use `-- stylua: ignore start` `end` as that would also break the generator.

lua/nvim-tree/actions/fs/create-file.lua
alex-courtis: > the function should take in a _boolean_ and an absolute path

Ah yes, you are right. That's not very workable.

How about we put it in notify itself? That knows about `opts.notify`. The function could then just take the path and maybe strip.

asror1: Hmm, I see, maybe be can do something like this then? 
```lua
for _, x in ipairs(modes) do
  M[x.name] = function(msg, src, dest, ...)
    local args = { ... }
    local parse_path = function(path)
      return config.absolute_path and path or vim.fn.fnamemodify(path, ":t")
    end
    if src ~= nil and dest ~= nil then
      msg = string.format(msg, parse_path(src), parse_path(dest), unpack(args))
    elseif src ~= nil then
      msg = string.format(msg, parse_path(src), unpack(args))
    elseif dest ~= nil then
      msg = string.format(msg, parse_path(dest), unpack(args))
    else
      msg = string.format(msg, unpack(args))
    end
    return dispatch(x.level, msg)
  end
end

``` 
**New Params**
`src`  -> some element path 
`dest` -> another element path (primarily to be used with _rename_ action) 
`{...}` -> other arguments to be used to format the `msg` 

I believe this could potentially be used for various different use cases.
* Notifications with only source element but other format args e.g. `notify.error("Could not create %s, because %s", node.absolute_path, nil, some_err)`
* Only destination and other format args e.g.  `notify.error("Destination folder %s doesn't exist, try %s", nil, some_path, some_advice)` (Although probably not the most _useful_ use case) 
*  With both source and destination and other args e.g. `notify.error("Could not move %s into %s, because %s", node.absolute_path, to, some_err)`
* Just source element e.g. `notify.error("%s was properly created", node.absolute_path)`
* Just other args e.g. `notify.error("Something went wrong... %s", nil, nil, some_err)`
* And other use cases. 

But most importantly I think this won't be a breaking change, because you would still be able to pass in only a `msg` and have the same functionality as before?

alex-courtis: I might be missing something, but that's probably a bit complex.

I was thinking something like this, that open-file and friends could call and pass to `notify.info`.

```lua
function M.render_path(path)
  if M.config.notify.absolute_path then
    return fnamemodify ...
  else
    return path
  end
end
```

asror1: Ah! okay, yea I was definitely over complicating things here. If I understand correctly though, have `render_path` be a method of _notify_ module, and have the fs actions call upon `notify.render_path(some_path)` to retrieve the appropriate path, and then be able to use it in `notify.some_mode`?

alex-courtis: Yes please. Simplest/most readable solution.

asror1: I'm a fan of this solution and it works great! However, I've made an error when squashing my commits (I had way too many unnecessary commits), so I will close this PR, and raise a cleaner one if that's okay? Sorry about that.

PR #2286, opened 2023-06-20
feat(#2270): add notify.absolute_path - show file or absolute path (default) names with notifications

### LINKED ISSUE #2270: Option to use file/directory names instead of absolute paths for notifications

asror1:
fixes #2270

Sorry for duplicate PRs! I totally messed up with #2280. Hopefully no harm was done.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -520,6 +520,7 @@ applying configuration.
       },
       notify = {
         threshold = vim.log.levels.INFO,
+        absolute_path = true,
       },
       ui = {
         confirm = {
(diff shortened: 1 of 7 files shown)


alex-courtis (approved):
Works beautifully, many thanks! I'll use this now...

Issue #135, opened 2020-11-17
Floating window and preview


fragov:
Is there any chance to have nvim-tree in floating window with files preview in next versions?

kyazdani42:
i don't think this is a goal i wish to acomplish, although it should be quite easy to do with the api i'll provide in the next version i'm working on.

wkcosmology:
+1  Unlike the splitting window, the floating window won't mess the window layout.

dennisfrostlander:
Hi @kyazdani42, did you by any chance already add the 

> i don't think this is a goal i wish to acomplish, although it should be quite easy to do with the api i'll provide in the next version i'm working on.

Hi @kyazdani42, did you by chance already add the api you've mentioned?

kyazdani42:
hi, haven't really had the time to refactor this part yet :)

fanliver:
Looking forward to this feature. I really miss this feature since switching from coc-explorer.

stevenocchipinti:
Me too! `coc-explorer` is oddly the main thing keeping me from switching to neovim's native LSP 😅

kyazdani42:
i guess i've rewritten the view handling enough so this can be tackled.
something like extending the `view.lua` file to handle a `popup` view basically :)

tsakirist:
I would love that too, this is something that amazed me when I used coc-explorer and it's really helpful.
It would be great if you could add this :)

fragov:
I tried to mess up with view.lua, but my LUA knowledge is almost zero. So I had to switch to lir. :(

fragov:
Tried 3 another file-tree plugins and Telescope explorer and I'm back again. Too many bugs and issues. Last one had an interesting choice of hotkeys: d for creating a directory, and shift-d to delete a folder (and it even didn't ask if you really want to delete that folder). I deleted my work for full day today.

What a good way to learn how I can extend nvim-tree? Any suggestions?

kyazdani42:
hi @fragov sorry for the late answer.
Well i believe the view should be refactored first before implementing this, i'm not sure it would be too easy to do in the current state of the plugin :)
This will land at some point

alex-courtis:
@lu5je0 any chance you could put something together for this?

alex-courtis:
Floating nvim-tree window has been implemented. `:help nvim-tree.view.float`.

It doesn't seem that floating preview has gained much interest. Please re-open / create a new PR if that's something you have a good use case for.

[...]

PR #2290, opened 2023-06-22
feat(#135 float-preview): float preview

JMarkin:
Hello, I implemented this feature.

I think this feature would be used like `Recipe` like this:
```lua 
    local api = require("nvim-tree.api")
    local FloatPreview = require("nvim-tree.float-preview")

    local prev, float_close_wrap = FloatPreview:new()
    prev:attach(bufnr)
    --- There are keymaps must to wrap for correct work
    -- ...
    vim.keymap.set("n", "<C-t>", float_close_wrap(api.node.open.tab), opts("Open: New Tab"))
    vim.keymap.set("n", "<C-v>", float_close_wrap(api.node.open.vertical), opts("Open: Vertical Split"))
    vim.keymap.set("n", "<C-s>", float_close_wrap(api.node.open.horizontal), opts("Open: Horizontal Split"))
    vim.keymap.set("n", "<CR>", float_close_wrap(api.node.open.edit), opts("Open"))
    vim.keymap.set("n", "<Tab>", float_close_wrap(api.node.open.preview), opts("Open"))
    vim.keymap.set("n", "o", float_close_wrap(api.node.open.edit), opts("Open"))
    vim.keymap.set("n", "O", float_close_wrap(api.node.open.no_window_picker), opts("Open: No Window Picker"))
    vim.keymap.set("n", "q", float_close_wrap(api.tree.close), opts("Close"))
    vim.keymap.set("n", "a", float_close_wrap(api.fs.create), opts("Create"))
    vim.keymap.set("n", "d", float_close_wrap(api.fs.remove), opts("Delete"))
    vim.keymap.set("n", "r", float_close_wrap(api.fs.rename), opts("Rename"))
 end
```

Currently it has 2 surprise for me
1. If lsp enabled with auto_attach=true, after open any buffer and lsp attached, preview buffer didn't unload(or load twice, I don't understand)
2. If style == 'minimal', after open file it would be open in minimal mode(would be disable many options)

### DIFF
--- a/lua/nvim-tree/float-preview.lua
+++ b/lua/nvim-tree/float-preview.lua
@@ -1,0 +1,160 @@
+local get_node = require("nvim-tree.lib").get_node_at_cursor
+local FloatPreview = {}
+FloatPreview.__index = FloatPreview
+
+local preview_au = "float_preview_au"
+vim.api.nvim_create_augroup(preview_au, { clear = true })
+
+function FloatPreview:new(cfg)
+  local prev = {}
+  setmetatable(prev, FloatPreview)
+
+  cfg = cfg or {
+    scroll_lines = 20,
+    mapping = {
+      down = { "<C-d>" },
+      up = { "<C-e>", "<C-u>" }
+    }
+  }
+
+  prev.buf = nil
+  prev.win = nil
+  prev.path = nil
+  prev.current_line = 1
+  prev.max_line = 999999
+  prev.disabled = false
+  prev.cfg = cfg
+
+  local function action_wrap(f)
+    return function(...)
+      prev:close()
+      local r = f(...)
+      prev:preview_under_cursor()
+      return r
+    end
+  end
+
+  return prev, action_wrap
+end
+
+function FloatPreview:close()
+  if self.path ~= nil then
+    pcall(vim.api.nvim_win_close, self.win, { force = true })
+    pcall(vim.api.nvim_buf_delete, self.buf, { force = true })
+    self.win = nil
+    self.buf = nil
+    self.path = nil
+    self.current_line = 1
+    self.max_line = 999999
+  end
+end
+
+function FloatPreview:preview(path)
+  self.path = path
+  self.buf = vim.api.nvim_create_buf(false, false)
+
+  vim.api.nvim_buf_set_option(self.buf, "bufhidden", "delete")
+  vim.api.nvim_buf_set_option(self.buf, "readonly", true)
+
+  local width = vim.api.nvim_get_option("columns")
+  local height = vim.api.nvim_get_option("lines")
+  local prev_height = math.ceil(height / 2)
+  local opts = {
+    -- style = "minimal",
+    relative = "win",
+    width = math.ceil(width / 2),
+    height = prev_height,
+    row = vim.fn.line("."),
+    col = vim.fn.winwidth(0) + 1,
+    border = "rounded",
+    focusable = false,
+    noautocmd = true,
+  }
+
+  self.win = vim.api.nvim_open_win(self.buf, true, opts)
+  local cmd = string.format("edit %s", vim.fn.fnameescape(self.path))
+  vim.api.nvim_command(cmd)
+  self.max_line = vim.fn.line("$")
+
+  local ok, out = pcall(vim.filetype.match, { buf = self.buf, filename = self.path })
+  if ok and out then
+    cmd = string.format("set filetype=%s", out)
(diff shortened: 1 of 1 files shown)


alex-courtis:
This looks incredible! The preview is working with this PR as well as with `"JMarkin/nvim-tree.lua-float-preview"`.

I'm not sure if I'm doing it right but I'm seeing the contents appear in both the floating window as well as the "regular" window when I tab preview:

alex-courtis:
> May be to need only plugin and Recipe ? What do you think?

Rather than a recipe, we should make this an Official [wiki: Extension Plugin](https://github.com/nvim-tree/nvim-tree.lua/wiki/Extension-Plugins). An entry in [wiki: Installation](https://github.com/nvim-tree/nvim-tree.lua/wiki/Installation) should also be added.

Ideas for improvements before we publish:
* Add a `setup` as per nvim-tree's setup, with an `open_win_config` to match nvim-tree's `view.float.open_win_config`. Users love to modify that ;)
* Please use `api.tree.get_node_under_cursor` rather than lib

alex-courtis:
>    If lsp enabled with auto_attach=true, after open any buffer and lsp attached, preview buffer didn't unload(or load twice, I don't understand)

Could you add an `LspAttach` event handler to "fix" this behaviour?

>    If style == 'minimal', after open file it would be open in minimal mode(would be disable many options)

If you had a setup user configuration, you could disable / remove that with a warning that minimal is unavailable.

lua/nvim-tree/float-preview.lua
```diff
@@ -0,0 +1,160 @@
+local get_node = require("nvim-tree.lib").get_node_at_cursor
```
alex-courtis, line 1: Please use api.tree.get_node_under_cursor rather than lib

alex-courtis (changes requested):

JMarkin:
> > May be to need only plugin and Recipe ? What do you think?
> 
> Rather than a recipe, we should make this an Official [wiki: Extension Plugin](https://github.com/nvim-tree/nvim-tree.lua/wiki/Extension-Plugins). An entry in [wiki: Installation](https://github.com/nvim-tree/nvim-tree.lua/wiki/Installation) should also be added.
> 
> Ideas for improvements before we publish:
> 
>     * Add a `setup` as per nvim-tree's setup, with an `open_win_config` to match nvim-tree's `view.float.open_win_config`. Users love to modify that ;)
> 
>     * Please use `api.tree.get_node_under_cursor` rather than lib
> 
>     * Copy some nvim-tree configuration files into your repo:
>       
>       * `.editorconfig`
>       * `.luacheckrc`
>       * `.luarc.json`
>       * `.stylua.toml`

@alex-courtis I fixed all your comments in [repo](https://github.com/JMarkin/nvim-tree.lua-float-preview). Should this PR to closed? Or you think integrate this to nvim-tree.lua more correct?

alex-courtis:
> I fixed all your comments in [repo](https://github.com/JMarkin/nvim-tree.lua-float-preview).

This looks good. The user has multiple ways to configure i.e.
```
require("nvim-tree").setup({
  on_attach = my_on_attach,
})
require("float-preview").setup({
  scroll_lines = 123,
})

-- or

local prev, float_close_wrap = FloatPreview:new({
  scroll_lines = 123,
})
```

> Or you think integrate this to nvim-tree.lua more correct?

The level of integration is great: no nvim-tree configuration is required to start and use `"JMarkin/nvim-tree.lua-float-preview"`, other than the actual `on_attach` mappings.

alex-courtis:
Does the user have to `new` float preview inside their on_attach? Could they just create it once e.g.

```lua
local float_preview, float_preview_wrap = require("float-preview"):new({ ... })

function M.on_attach(bufnr)
---
  float_preview:attach(bufnr)
  vim.keymap.set("n", "<C-t>", float_preview_wrap(api.node.open.tab), opts("Open: New Tab"))
---
end
```

alex-courtis:
Unfortunately I still don't understand how the plugin works. I am doing it wrong.

* How do I close the preview window? It only closed when I `q` nvim-tree
* The preview window always show the currently opened buffer. How do I keep it on the previewed buffer? I'd like to have a different preview window than the buffer I am editing.

See [asciicast](https://asciinema.org/a/593422)

alex-courtis:
I've added a draft help file https://github.com/JMarkin/nvim-tree.lua-float-preview/pull/2

It needs some work.

alex-courtis:
The external plugin is looking good and can be added to the wiki when it's ready.

PR #2302, opened 2023-07-01
feat: add sort_by "filetype"

moniquelive:
Fixes nvim-tree/nvim-tree.lua#2288 by proposing a new kind of sorting: by vim's filetype.

PR #2306, opened 2023-07-02
fix: sort_by "extension" falls back to name

moniquelive:
As suggested by @alex-courtis on #2302


alex-courtis:
Thanks for the contribution. I'll review next weekend.

alex-courtis (approved):
Many thanks for the contribution!

PR #2335, opened 2023-07-23
feat(trash): add synchronized trash support for windows

### LINKED ISSUE #2331: Trash files on Windows as well as Unix system
**Is this a question?**
Move to Windows recycle bin instead of deleting files completely as I need to delete files. I know there is a function 'Trash' with shortcut `D`, but it told me:

> Trash is currently a UNIX only feature!

Sorry I'm not good at it and I don't know about it... But I know there's a PowerShell script can do that:

```powershell
$sh = new-object -comobject "Shell.Application"
$ns = $sh.Namespace(0).ParseName("PATH\TO\FILE\TO\DELETE")
$ns.InvokeVerb("delete")
```

Add Windows support of command 'Trash'.

Add Windows support of command 'Trash'.

#279 Shows the way I used it on WSL.

**Additional context**
Nothing.

linrongbin16:
fixes #2331

Support trash-cli on Windows, relate to discussion https://github.com/nvim-tree/nvim-tree.lua/discussions/2333 and #2331.

1. remove `Unix only` from trash action.
2. set `detach=false` and `jobwait { job }` on Windows to make trash operation sync on Windows. This is because: in my testing/practicing, when set `detach=true` in jobstart, nvim (running in Windows Terminal app) quickly switch to desktop, then switch back to nvim, that makes the screen blink. I guess the **process detech** behaviour is different under Windows, to avoid this blink I set sync wait under Windows.
3. manual tested on Windows 10 x86_64, nvim v0.9.1, latest nvim-tree master.

### DIFF
--- a/lua/nvim-tree/actions/fs/trash.lua
+++ b/lua/nvim-tree/actions/fs/trash.lua
@@ -52,12 +52,12 @@ function M.fn(node)
   local function trash_path(on_exit)
     local need_sync_wait = utils.is_windows
     local job = vim.fn.jobstart(M.config.trash.cmd .. ' "' .. node.absolute_path .. '"', {
-      detach = true,
+      detach = not need_sync_wait,
       on_exit = on_exit,
       on_stderr = on_stderr,
     })
     if need_sync_wait then
-      vim.fn.jobwait({job})
+      vim.fn.jobwait { job }
     end
   end
 
--- a/lua/nvim-tree/actions/fs/trash.lua
+++ b/lua/nvim-tree/actions/fs/trash.lua
@@ -29,8 +29,8 @@ function M.fn(node)
     return
   end
 
+  -- configs
   if utils.is_unix or utils.is_windows then
-    -- configs
     if M.config.trash.cmd == nil then
       M.config.trash.cmd = "trash"
     end
--- a/lua/nvim-tree/actions/fs/trash.lua
+++ b/lua/nvim-tree/actions/fs/trash.lua
@@ -30,16 +30,11 @@ function M.fn(node)
   end
 
   -- configs
-  if utils.is_unix then
-    if M.config.trash.cmd == nil then
-      M.config.trash.cmd = "trash"
-    end
-    if M.config.ui.confirm.trash == nil then
-      M.config.ui.confirm.trash = true
-    end
-  else
-    notify.warn "Trash is currently a UNIX only feature!"
-    return
+  if M.config.trash.cmd == nil then
+    M.config.trash.cmd = "trash"
+  end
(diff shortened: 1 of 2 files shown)


linrongbin16:
I mark this PR as draft, and try to support async.

let me take time to test the jobstart api.

alex-courtis (changes requested):
Looking good.

I'm happy with the sync: cautious but reliable behaviour is preferred. We don't need any more options.

What is the actual `trash.cmd`? We can set this for windows, as I have just done for macos: https://github.com/nvim-tree/nvim-tree.lua/pull/2336
It might be easiest to merge that PR before this one.

Please:
- [x] update help section 10
- [x] update help section 4 nvim-tree.trash.cmd
- [x] test under WSL and powershell (if you haven't already)

linrongbin16:
hi @alex-courtis ,

 1. the actual trash.cmd is `trash` (see https://github.com/sindresorhus/trash-cli), the two implementations (python and nodejs) use the same cmd name `trash`. so no other things need to change.
 2. as far as I know, these two trash cli implementations are most popular, and they all use `trash`, so it should be good to set default trash cmd to `trash`.
 3. todo: add help doc and other testing

lua/nvim-tree/actions/fs/trash.lua
```diff
@@ -30,16 +30,11 @@ function M.fn(node)
   end
 
   -- configs
-  if utils.is_unix then
```
alex-courtis, line 33: Can we please be conservative and use is_unix or is_windows ?

We'll sort out macos later: #2336

alex-courtis:
>     4. [ ]  test under WSL. -- Is this [WSL](https://learn.microsoft.com/en-us/windows/wsl/install#upgrade-version-from-wsl-1-to-wsl-2) or WSL2? I don't have WSL on my Windows 10 or Windows 11, because it's quite legacy, I only have WSL2, which is actually a ubuntu VM running on Windows 10 host.

RE WSL or WSL2 I'll leave that to you... I'm not familiar with windows.

A quick test on whatever is the current WSL would  be great.

alex-courtis (changes requested):
Looking good, many thanks for your contribution.

linrongbin16:
> A quick test on whatever is the current WSL would be great.

Test on WSL2 is done.

lua/nvim-tree/actions/fs/trash.lua
linrongbin16: update: use `is_unix or is_windows` check

linrongbin16: please check again, should be OK now.

alex-courtis (approved):

PR #2336, opened 2023-07-24
fix: trash.cmd defaults to 'trash' on macos

### LINKED ISSUE #1366: Warn users when trash command is not available
### Description

If `gio trash` is not available on the user's system, attempting to trash a file instead permanently deletes the file with no warning to the user. For anyone who was using `trash` before, after #1323 they would now be deleting files without knowing.

### Neovim version

```text
NVIM v0.8.0-dev+1894-gc57f6b28d
Build type: Release
LuaJIT 2.1.0-beta3
```

macOS 13.0 Beta

### nvim-tree version

79258f1

### Minimal config

```lua
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

1. touch delete-me.txt
2. nvim -nu nvt-min.lua
3. `:NvimTreeToggle`
4. hover "delete-me.txt" and press `D` (confirming the deletion)

### Expected behavior

Pressing `D` to trash a file should either trash the file, in this case `gio trash` must be available, or fail the operation.

### Actual behavior

Pressing `D` to trash a file will permanently a file if `gio trash` is not available.

alex-courtis:
fixes #1366

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -673,6 +673,8 @@ end
 local function localise_default_opts()
   if utils.is_macos then
     DEFAULT_OPTS.trash.cmd = "trash"
+  elseif utils.is_windows then
+    DEFAULT_OPTS.trash.cmd = "TODO @linrongbin16"
   end
 end
 
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -671,10 +671,8 @@ end
 
 --- Apply OS specific localisations to DEFAULT_OPTS
 local function localise_default_opts()
-  if utils.is_macos then
+  if utils.is_macos or utils.is_windows then
     DEFAULT_OPTS.trash.cmd = "trash"
-  elseif utils.is_windows then
-    DEFAULT_OPTS.trash.cmd = "TODO @linrongbin16"
   end
 end
 
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -663,6 +663,13 @@ local function validate_options(conf)
   end
 end
 
+--- Apply OS specific localisations to DEFAULT_OPTS
+local function localise_default_opts()
+  if utils.is_macos then
+    DEFAULT_OPTS.trash.cmd = "trash"
+  end
+end
+
 function M.purge_all_state()
   require("nvim-tree.watcher").purge_watchers()
   view.close_all_tabs()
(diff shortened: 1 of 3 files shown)


lua/nvim-tree.lua
```diff
@@ -663,6 +663,13 @@ local function validate_options(conf)
   end
 end
 
+--- Apply OS specific localisations to DEFAULT_OPTS
+local function localise_default_opts()
+  if utils.is_macos then
+    DEFAULT_OPTS.trash.cmd = "trash"
```
gegoune, line 669: Shouldn't that be `gio trash`?

alex-courtis: This is where I get confused. I was under the impression that gio would put the files in `~/.local/share/Trash/files` which is not recognised by macos.

Perhaps we should just update the doc, removing "Trash is unavailable" and noting the available homebrew options for macos users.

alex-courtis: I think it might be time for me to use https://github.com/sickcodes/Docker-OSX

kennyishihara: [Here](https://github.com/nvim-tree/nvim-tree.lua/issues/1366#issuecomment-1646788804), we found that neither `trash` nor `gio` comes preinstalled on macOS so it's best to do what you said, which is to warn the users that `D` is not available unless you install `trash` from homebrew. We shouldn't use `gio trash` on macOS as it moves the file to a different location than `~/.Trash` which is where the Bin icon is pointing to.

alex-courtis:
Updated to include the windows trash changes #2335

- [ ] @linrongbin16 what is the value of `trash.cmd` for windows?

linrongbin16:
Hi, @alex-courtis, there's no such *default* trash command on Windows.
People have to install third party command to make it work. For now, the most porpular one is: https://github.com/sindresorhus/trash-cli. The command is just: `trash`.

alex-courtis:
> 

Thanks mate. Adding to the doc.

linrongbin16:
sure thing

linrongbin16 (approved):
lgtm

PR #2337, opened 2023-07-24
feat: remove, copy, trash, bulk delete prompts default to "yes"

xundaoxd:


alex-courtis (changes requested):
This is a fantastic idea! Many thanks for your contribution.

- [x] Please apply the same to each of these other: copy, trash, bulk delete
- [ ] ~~When `select_prompts = true,` the default is not applied~~

xundaoxd:
@alex-courtis I update the code. use first option as default in vim.ui.input

alex-courtis:
> @alex-courtis I rewrite the code. Use the first option as default in lib.prompt when select_prompts = false

That's really nice!

I'm happy to leave `select_prompts` alone; users of that likely have their own select plugin.

alex-courtis (approved):

Issue #1628, opened 2022-10-07
File opens on the same nvim-tree window when quit_on_open = true


Almo7aya:
### Description

When opening a file using Telescope or :e [filename] while focusing on the nvim-tree window, the file opens the same window as nvim-tree

https://user-images.githubusercontent.com/17254073/194481452-6e013b10-c9b9-461d-872f-25c06bc40928.mp4

### Neovim version

```text
NVIM v0.9.0-dev-1214-ge54541f7f-dirty
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

MacOS 12.6

### nvim-tree version

main - 7282f7d

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "kyazdani42/nvim-tree.lua",
      "kyazdani42/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {
    actions = {
      open_file = {
      quit_on_open = true
    }
  }
}
end
```

### Steps to reproduce

1- nvim -nu /tmp/nvt-min.lua
2- e ./nvt-min.lua
3- :NvimTreeOpen
4- e ./nvt-min.lua

Almo7aya:
[...]

PR #2341, opened 2023-07-28
feat: add actions.open_file.eject

sharkov63:
Hello!

In issue #1628 and PR #1637 the following expected behavior was introduced. When opening a file using `:edit` (or Telescope) while being focused on nvim-tree window, the new file opens in another window (and, depending on `quit_on_open` option, the nvim-tree window either stays or closes). This is useful when using nvim-tree in "expected" way, i.e. as a narrow column window on the side, which can be toggled on and off.

However, I'd like to use nvim-tree as a replacement of netrw: I would mostly want to open files from the tree in-place, overriding the tree window. While there already exists an API function `api.node.open.replace_tree_buffer` which performs an in-place open, issuing `:e` or using Telescope does not make the desired in-place open.

This PR adds an option `nvim-tree.actions.open_file.eject`, which is `true` by default (current behavior). When it's `false`, it disables nvim-tree hijacking of such in-place opens: the new file will be opened in the same window, replacing nvim-tree.

Please provide feedback, whether this option would be useful (or whether there is already another way to achieve the desired effect). Thank you

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -494,6 +494,7 @@ applying configuration.
         },
         open_file = {
           quit_on_open = false,
+          prevent_buffer_override = true,
           resize_window = true,
           window_picker = {
             enable = true,
@@ -1203,8 +1204,11 @@ Configuration for various actions.
 
         *nvim-tree.actions.open_file.quit_on_open*
         Closes the explorer when opening a file.
-        It will also disable preventing a buffer overriding the tree.
           Type: `boolean`, Default: `false`
+
+        *nvim-tree.actions.open_file.prevent_buffer_override*
+        Prevent new opened file from opening in the same window as the tree.
+          Type: `boolean`, Default: `true`
 
         *nvim-tree.actions.open_file.resize_window*  (previously `view.auto_resize`)
         Resizes the tree when opening a file.
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -494,7 +494,7 @@ applying configuration.
         },
         open_file = {
           quit_on_open = false,
-          prevent_buffer_override = true,
+          eject = true,
           resize_window = true,
           window_picker = {
             enable = true,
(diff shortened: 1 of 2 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -494,6 +494,7 @@ applying configuration.
         },
         open_file = {
           quit_on_open = false,
+          prevent_buffer_override = true,
```
alex-courtis, line 497: buffer_override is used internally, however it's not a very clear description.

Perhaps we could call this `eject_buffer`?

alex-courtis (changes requested):
Tested ok, works as advertised.

`quit_on_open` behaviour is now correctly documented.

lua/nvim-tree.lua
```diff
@@ -192,15 +192,17 @@ local function setup_autocommands(opts)
...
-  -- prevent new opened file from opening in the same window as nvim-tree
-  create_nvim_tree_autocmd("BufWipeout", {
-    pattern = "NvimTree_*",
-    callback = function()
-      if utils.is_nvim_tree_buf(0) then
-        view._prevent_buffer_override()
-      end
-    end,
-  })
+  if opts.actions.open_file.prevent_buffer_override then
```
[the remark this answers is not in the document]
sharkov63: Done

doc/nvim-tree-lua.txt
sharkov63: Yes, that's a better name. Maybe also `eject_tree` to highlight the nvim-tree buffer is ejected?

alex-courtis: false indicates "do nothing" i.e. don't interfere with normal vim behaviour. We never really do anything with the tree itself.

Since this is under `open_file` we could just call it `open_file.eject`. Your description is quite accurate, describing what happens to the new buffer.

I do get quite critical when it comes to option names / places. It's easy to get right when adding but very hard to refactor later...

sharkov63: I agree naming things that you cannot change later should be discussed vigorously :) I will rename to `open_file.eject`

sharkov63:
Rebased with nvim-tree master

alex-courtis (approved):
Many thanks for your changes!

There have been many discussions in the past around such behaviour and this should have a wide impact.

PR #2344, opened 2023-07-30
fix(#2343): tree is now correctly abandoned upon an in-place open with eject=false

### LINKED ISSUE #2343: nvim-tree does not open after an in-place open with eject=false
### Description

In #2341 I  introduced a bug which happens only with `eject = false` option. Turns out, when making an in-place open with `eject = false`, the window, in which nvim-tree buffer is overridden with a new file, is still considered a nvim-tree window. Thus, when we try to open nvim-tree, we have `view.is_visible == true` an the tree is not opened.

I figured out the fix and will provide PR very soon. The issue is to put a clean room example in history.

### Neovim version

```text
NVIM v0.10.0-dev
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

Linux 5.16.15-051615-generic

### nvim-tree version

master

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set packpath=./nvt-min/site]]
local package_root = "./nvt-min/site/pack"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {
    actions = {
      open_file = {
        eject = false,
      }
    }
  }
end
```

### Steps to reproduce

1. nvim -nu /tmp/nvt-min.lua
2. :NvimTreeOpen
3. :e nvt-min.lua
4. :NvimTreeOpen

### Expected behavior

The nvim-tree window opens.

### Actual behavior

Nothing happens: nvim-tree window does not open.

sharkov63:
Closes #2343

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -196,8 +196,13 @@ local function setup_autocommands(opts)
   create_nvim_tree_autocmd("BufWipeout", {
     pattern = "NvimTree_*",
     callback = function()
-      if utils.is_nvim_tree_buf(0) and opts.actions.open_file.eject then
+      if not utils.is_nvim_tree_buf(0) then
+        return
+      end
+      if opts.actions.open_file.eject then
         view._prevent_buffer_override()
+      else
+        view.abandon_current_window()
       end
     end,
   })


alex-courtis (approved):
Works nicely, thanks for the quick report/fix.

PR #2357, opened 2023-08-06
fix: prompt uses first character of response - allow "yy"

alex-courtis:
follows #2337

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -148,7 +148,7 @@ function M.prompt(prompt_input, prompt_select, items_short, items_long, callback
     end)
   else
     vim.ui.input({ prompt = prompt_input, default = items_short[1] or "" }, function(item_short)
-      callback(item_short)
+      callback(item_short and item_short:sub(1, 1) or nil)
     end)
   end
 end


Danie-1:
Maybe using the last character would be more sensible? If I click n and then enter I would expect it to cancel.

alex-courtis:
> Maybe using the last character would be more sensible? If I click n and then enter I would expect it to cancel.

Agreed. See #2377

mohamedarish:
Shouldn't yy and all other things starting with y be accepted?
Having the last character be checked would result in ny["y"]/ns["yes"] being accepted
Not being able to accidentally delete is better than being easily able to accidentally delete

alex-courtis:
> Shouldn't yy and all other things starting with y be accepted?
> Having the last character be checked would result in ny["y"]/ns["yes"] being accepted
> Not being able to accidentally delete is better than being easily able to accidentally delete

Of course. I misread the comment from @Danie-1 , thinking it was first.

Your PR addresses all these concerns.

Danie-1:
I made that comment because I think at one point there was a feature that the letter "y" would be automatically present in the response, and so if we check the first letter I would have to be careful to make sure I delete that letter (because even if I type n then enter, the response would be y). It seems like the prompts have changed though so I don't think the comment is relevant anymore (and maybe it wasn't relevant in the first place and I had just misunderstood how the prompts work).

alex-courtis:
> I made that comment because I think at one point there was a feature that the letter "y" would be automatically present in the response, and so if we check the first letter I would have to be careful to make sure I delete that letter (because even if I type n then enter, the response would be y). It seems like the prompts have changed though so I don't think the comment is relevant anymore (and maybe it wasn't relevant in the first place and I had just misunderstood how the prompts work).

You're right... it wasn't good UX and was the driver of #2377

Issue #2364, opened 2023-08-10
add support to show the files first or after folder


sajjathossain:
In vscode there's an option to show the files first. which makes it easy to navigate between files if a folder is opened at the same level. It would be nice to have a feature of that sort.

Akmadan23:
Closed via #2366

PR #2366, opened 2023-08-11
feat(#2364): add option to show files first

Akmadan23:
This is a simple and straight forward implementation of the feature requested in #2364. For the configuration I think that since `sort.folders_first` is enabled by default, `sort.files_first = true` should override it. I made a minor refactoring in `explorer/sorter.lua`, moving the folders-first logic in a separate function so that it's easier to manage and enhance.


alex-courtis:
FYI: you're a repo member: you can push a branch to this repo instead of using a fork.

lua/nvim-tree/explorer/sorters.lua
```diff
@@ -124,12 +149,9 @@ local function node_comparator_name_ignorecase_or_not(a, b, ignorecase)
...
   end
 
-  if M.config.sort.folders_first then
-    if a.nodes and not b.nodes then
-      return true
-    elseif not a.nodes and b.nodes then
-      return false
-    end
+  local early_return = folders_or_files_first(a, b)
+  if early_return ~= nil then
```
alex-courtis, line 153: Could we just use `if folders_or_files_first(a, b) then return`?

lua/nvim-tree/explorer/sorters.lua
```diff
@@ -23,6 +23,31 @@ local function tbl_slice(t, first, last)
   return slice
 end
 
+---Evaluate `sort.folders_first` and `sort.files_first`
+---@param a table
+---@param b table
+---@param third_option_callback function|nil function should return `boolean`. Evaluated only if necessary
+---@return boolean|nil
+local function folders_or_files_first(a, b, third_option_callback)
```
alex-courtis, line 31: We could just use `comparator` for the third argument.

lua/nvim-tree/explorer/sorters.lua
```diff
@@ -23,6 +23,31 @@ local function tbl_slice(t, first, last)
...
+  if a.nodes and not b.nodes then
+    x = true
+  elseif not a.nodes and b.nodes then
+    x = false
+  elseif a.nodes and b.nodes and third_option_callback ~= nil then
+    x = third_option_callback()
+  end
+
+  if x ~= nil then
+    if M.config.sort.files_first then
```
alex-courtis, line 43: Could we flip this around? Test this config at the start of the function?

That would remove the need for the mystery `x`

alex-courtis (changes requested):
This is nice, simplifies things a little as well.

Tested:
* all permutations of `folders_first` and `files_first`
* sort function

lua/nvim-tree/explorer/sorters.lua
```diff
@@ -23,6 +23,31 @@ local function tbl_slice(t, first, last)
   return slice
 end
 
+---Evaluate `sort.folders_first` and `sort.files_first`
+---@param a table
```
alex-courtis, line 27: /nit

`---@param a table node`

lua/nvim-tree/explorer/sorters.lua
Akmadan23: Yeah, I stored in `x` what should be returned if `sort.folders_first` is enabled and reverse it with `sort.files_first`, but now that I think about it maybe something like this would be more readable and intuitive.
```lua
local a_and_not_b, not_a_and_b

if M.config.sort.files_first then
  a_and_not_b = false
  not_a_and_b = true
elseif M.config.sort.folders_first then
  a_and_not_b = true
  not_a_and_b = false
else
  return
end

if a.nodes and not b.nodes then
  return a_and_not_b
elseif not a.nodes and b.nodes then
  return not_a_and_b
elseif a.nodes and b.nodes and third_option_callback ~= nil then
  return third_option_callback()
end
```

lua/nvim-tree/explorer/sorters.lua
Akmadan23: Do you mean for the name of the variable instead of `third_option_callback`? If so I don't know if it's appropriate in this case since we aren't comparing anything, the function just returns the fallback value...

Akmadan23:
> FYI: you're a repo member: you can push a branch to this repo instead of using a fork.

I didn't think about that, next time I'll push a new branch here.

lua/nvim-tree/explorer/sorters.lua
alex-courtis: `fallback` it is :)

lua/nvim-tree/explorer/sorters.lua
alex-courtis: Much more readable. Could we go simpler?

```lua
if a.nodes and not b.nodes then
  return M.config.sort.folders_first
```

doc/nvim-tree-lua.txt
```diff
@@ -317,6 +317,7 @@ applying configuration.
       sort = {
         sorter = "name",
         folders_first = true,
+        files_first = false,
```
gegoune, line 320: Sorry for not checking myself, but what will happen when user sets both `folders_first` and `files_first` to `true`?

Maybe some extra sorters, perhaps configurable, should be provided instead of adding specialised options?

lua/nvim-tree/explorer/sorters.lua
alex-courtis: Ah of course. How about an early exit?

```lua
if not M.config.sort.files_first and not M.config.sort.folders_first then
  return
end

File: doc/nvim-tree-lua.txt (hunk 5 above)
Comment (alex-courtis) [reply doc/nvim-tree-lua.txt:311]: Tested OK

File: lua/nvim-tree/explorer/sorters.lua (hunk 3 above)
Comment (alex-courtis) [reply lua/nvim-tree/explorer/sorters.lua]: This should do it:

```lua
---Evaluate `sort.folders_first` and `sort.files_first`, files takes precedence.
---@param a table node
---@param b table node
---@param fallback function|nil function should return `boolean`. Evaluated when both are folders.
---@return boolean|nil
local function folders_or_files_first(a, b, fallback)
  if not M.config.sort.folders_first and not M.config.sort.files_first then
    return
  end

  if not a.nodes and b.nodes then
    -- file <> folder
    return M.config.sort.files_first
  elseif a.nodes and not b.nodes then
    -- folder <> file
    return not M.config.sort.files_first
  elseif fallback and a.nodes and b.nodes then
    -- folder <> folder
    return fallback()
  end
end
```

Yes, we could be tricky with ternaries but readability is more important.

Akmadan23: Oh, I didn't consider to completely ignore `sort.folders_first` but it works so that's fine.

Akmadan23:
With the last commit I removed the `fallback` because it's used only once, I think this way is even more simple and clear.

alex-courtis:
> With the last commit I removed the `fallback` because it's used only once, I think this way is even more simple and clear.

That's great... simple is best.

I'll test this one on the weekend.

alex-courtis (approved):
Tested:
* all permutations of `folders_first` and `files_first`
* sort function

PR #2371, opened 2023-08-14
chore: remove legacy view.mappings.list

alex-courtis:
I think it's time.


alex-courtis:
https://github.com/nvim-tree/nvim-tree.lua/wiki/Migrating-To-on_attach is no longer useful and can be removed following merge.

gegoune (approved):
Nice, let's pull the bandaid off. :)

PR #2374, opened 2023-08-14
fix(#2352): windows: escape special filename characters on edit

### LINKED ISSUE #2352: Cannot open file whose filename or folder name has parenthesis "(" on Windows
### Description

As the title says, I cannot open files that have "(" or ")" in their file name or folder structure.
It looks like the path xxx\(abc)\(edf).jsx will be interpreted incorrectly due to the "escaping" effect.

All file paths like xxx\(index)\(index).jsx will be interpreted to xxx(index)(index).jsx due to the "char escaping" mechanism...

See description and replies here:
https://github.com/neovim/neovim/issues/24542

But even when I `set shellslash`, and can open those files with `:e <filename>`, I still cannot open the files in `nvim-tree`.
It seems `nvim-tree` does not follow the setting of `shellslash`

### Neovim version

```text
NVIM v0.9.1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

Windows 11

### nvim-tree version

0a89dcb

### Clean room replication

```lua
raw nvim with nvim-tree.
```

### Steps to reproduce

- create a folder/file with path like: xxx\(index)\(index).jsx
- Open the file in `nvim-tree`.
- The file cannot be opened/showing empty file.

### Expected behavior

the file should be opened correctly.

### Actual behavior

nvim shows an empty file -- this should be caused by the "escape" issue.

davisthedev:
fixes #2352
- add check if environment is windows
- if windows, escape special characters

This works on windows, will need additional linux/mac testing.

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -270,7 +270,7 @@ local function open_in_new_window(filename, mode)
   end
 
   local fname = vim.fn.fnameescape(filename)
-  if vim.fn.has "win32" == 1 or vim.fn.has "win32unix" == 1 then
+  if utils.is_windows then
     fname = fname:gsub("%(", "\\("):gsub("%)", "\\)")
   end
 
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -270,9 +270,7 @@ local function open_in_new_window(filename, mode)
   end
 
   local fname = vim.fn.fnameescape(filename)
-  if utils.is_windows then
-    fname = fname:gsub("%(", "\\("):gsub("%)", "\\)")
-  end
+  fname = utils.escape_windows_special_chars(fname)
 
   local cmd
   if create_new_window then
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -270,6 +270,9 @@ local function open_in_new_window(filename, mode)
   end
 
   local fname = vim.fn.fnameescape(filename)
+  if vim.fn.has "win32" == 1 or vim.fn.has "win32unix" == 1 then
+    fname = fname:gsub("%(", "\\("):gsub("%)", "\\)")
+  end
 
   local cmd
   if create_new_window then
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -270,7 +270,7 @@ local function open_in_new_window(filename, mode)
   end
 
   local fname = vim.fn.fnameescape(filename)
-  fname = utils.escape_windows_special_chars(fname)
+  fname = utils.escape_special_chars(fname)
 
   local cmd
   if create_new_window then
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -211,10 +211,13 @@ function M.canonical_path(path)
 end
 
 -- Escapes special characters in string if windows else returns unmodified string.
--- @param string string
--- @return string
-function M.escape_windows_special_chars(string)
-  return M.is_windows and string:gsub("%(", "\\("):gsub("%)", "\\)") or string
+-- @param path string
+-- @return path
+function M.escape_special_chars(path)
+  if path == nil then
+    return path
+  end
+  return M.is_windows and path:gsub("%(", "\\("):gsub("%)", "\\)") or path
 end
 
 -- Create empty sub-tables if not present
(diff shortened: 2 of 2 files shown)


lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -270,6 +270,9 @@ local function open_in_new_window(filename, mode)
   end
 
   local fname = vim.fn.fnameescape(filename)
+  if vim.fn.has "win32" == 1 or vim.fn.has "win32unix" == 1 then
```
gegoune, line 273: Use https://github.com/nvim-tree/nvim-tree.lua/blob/18c7a3119839adc4599d838726deae662859c8b2/lua/nvim-tree/utils.lua#L12 instead.

gegoune (changes requested):

gegoune:
We definitely should have dedicated function to do that

davisthedev:
I am happy to add this in other areas or let someone with more knowledge handle it and I can test.

I am not sure if there would be any adverse effects to other pieces of code with the escaping. 
That is why I only made the change at the last point.

alex-courtis:
> I am happy to add this in other areas or let someone with more knowledge handle it and I can test.

I am confident it your abilities.

> I am not sure if there would be any adverse effects to other pieces of code with the escaping. That is why I only made the change at the last point.

We can't cover every possible case, however we can:
1. Test and maybe fix the different actions that edit e.g. open in tab, open in place etc.
2. Test all codepaths that lead to this common function.

This is behind the feature flag so we can be confident.

davisthedev:
I did not have any issues with ```node.open.replace_tree_buffer``` or ```node.open.tab```.
I did move the code to its own utils function so it can be used if the problem is found elswhere.

lua/nvim-tree/utils.lua
```diff
@@ -210,6 +210,16 @@ function M.canonical_path(path)
   return path
 end
 
+-- Escapes special characters in string if windows else returns unmodified string.
+-- @param string string
+-- @return string
+function M.escape_windows_special_chars(string)
+  if M.is_windows then
```
gegoune, line 217: How about ternary here instead?

davisthedev: Done.

lua/nvim-tree/utils.lua
```diff
@@ -210,6 +210,13 @@ function M.canonical_path(path)
   return path
 end
 
+-- Escapes special characters in string if windows else returns unmodified string.
+-- @param string string
+-- @return string
+function M.escape_windows_special_chars(string)
```
alex-courtis, line 216: Let's be a little more generic:

`escape_windows_special_chars(path)`

as string is the name of the string library.

lua/nvim-tree/utils.lua
```diff
@@ -210,6 +210,13 @@ function M.canonical_path(path)
...
+  return M.is_windows and string:gsub("%(", "\\("):gsub("%)", "\\)") or string
```
alex-courtis, line 217: Please guard against nil path: `string.gsub` will fail hard on nil.

alex-courtis (changes requested):
Many thanks for your contribution. I'll get back to this one on the weekend.

I was tempted to put this everywhere we use a `fnameescape` however it's safest just to fix known issues.

lua/nvim-tree/utils.lua
davisthedev: Updated.

lua/nvim-tree/utils.lua
davisthedev: Added nil check.

davisthedev:
I don't know if the change even needs to go everywhere fnameescape is used. The issue only seems to happen whenever the path is handed over to neovim to create a buffer. It seems like everything is fine inside nvim-tree.

I am working on a similar merge for telescope, and it has the same thing. It only has this problem when it tries to open the file in a buffer or window. So I think the root of might actually be fixable inside neovim, but I am not ready to work on that codebase yet.

alex-courtis:
> It only has this problem when it tries to open the file in a buffer or window. So I think the root of might actually be fixable inside neovim, but I am not ready to work on that codebase yet.

That sounds quite reasonable. We are escaping paths using vim's api, it's just these special cases that are not correct.

We might encounter an issue if this is fixed in nvim, however we cannot do anything until that time.

alex-courtis (approved):

xarthurx:
> > It only has this problem when it tries to open the file in a buffer or window. So I think the root of might actually be fixable inside neovim, but I am not ready to work on that codebase yet.
> 
> That sounds quite reasonable. We are escaping paths using vim's api, it's just these special cases that are not correct.
> 
> We might encounter an issue if this is fixed in nvim, however we cannot do anything until that time.

https://github.com/neovim/neovim/issues/24542 

I don't expect they will change it in the near future.

Issue #2369, opened 2023-08-13
Folder name specific colors


antoniosarosi:
**Is this a question?**
Original Q&A: https://github.com/nvim-tree/nvim-tree.lua/discussions/1498

**Can this functionality be implemented utilising API?**
I don't think so, I haven't found any function to change the color of an icon.

**Is your feature request related to a problem? Please describe.**
It's not related to a problem or bug, the feature would simply allow the user to change the color of a directory icon based on its name, for example, `"node_modules" = "green"`, `"lua" = "blue"`, etc.

**Describe the solution you'd like**
Either a setup option similar to `nvim-web-devicons`:

```lua
require'nvim-web-devicons'.setup {
 override_by_filename = {
  [".gitignore"] = {
    icon = "",
    color = "#f1502f",
    name = "Gitignore"
  }
 };
}
```

Maybe something like this:

```lua
require("nvim-tree").setup({
    renderer = {
        indent_markers = {
            enable = true,
        },
        override_by_directory_name = {
            ["node_modules"] = {
                icon = "󰎙",
                color = "#68A063",
                name = "NodeModules",         
            }
        }
    }
})
```

Or some sort of API function that gets called with the directory name and returns the color (not sure how that would work or how it would impact performance).

**Describe alternatives you've considered**
I tried configuring `nvim-web-devicons` directly but I think you can only to that with files, not directories.

**Additional context**
Basically trying to simulate icon themes like `Material Icon Theme` for VScode:

![](https://raw.githubusercontent.com/PKief/vscode-material-icon-theme/main/images/folderIcons.png)

alex-courtis:
This can be done. We can match the folder based on nvim-web-devicons name.

Something like `M.devicons.get_icon(node.name, node.extension)` in `icons.lua get_folder_icon`

The icon and colour would be provided from overrides in nvim-web-devicons

[Pull requests](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md) are always gratefully appreciated.

alex-courtis:
My initial notes around highlight groups were for folder colour, however as that is not needed, this much simpler solution is appropriate.

antoniosarosi:
You mean [here](https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/renderer/components/icons.lua#L12)? I'll try to implement it myself and submit a PR. I don't know much about Lua but seems to be a simple language.

[...]

Issue #2361, opened 2023-08-08
User should be able to change default icon even if nvim-web-devicons is available


2ax3ax5:
I'd like to set the nvim-tree.renderer.icons.glyphs.default to a desired value even if nvim-web-devicons are available.

From the docs
```
*nvim-tree.renderer.icons.glyphs.default*
Glyph for files. Will be overridden by `nvim-web-devicons` if available.
Type: `string`, Default: `""`
```

Note that this is the only icons.glyphs "option" that behaves that way and the user cannot custom it (if nvim-web-devicons are available).

alex-courtis:
> I'd like to set the nvim-tree.renderer.icons.glyphs.default to a desired value even if nvim-web-devicons are available.

Which functionality do you still require from nvim-web-devicons, if you are overriding the files?

alex-courtis:
The above change should enable your desired behaviour. Please comment if you experience issues.

PR #2375, opened 2023-08-14
feat(#2369): Add `webdev_colors_folder` option

antoniosarosi:
Implements feature https://github.com/nvim-tree/nvim-tree.lua/issues/2369

I started writing a little draft to provide an option that allows the user to change the icon and color of a directory based on its name. I added a new option `webdev_colors_folder` and I changed the `get_folder_icon` function as mentioned [here](https://github.com/nvim-tree/nvim-tree.lua/issues/2369#issuecomment-1676506929). Now you can do this:

```lua
require("nvim-web-devicons").setup {
    strict = true,
    override_by_filename = {
        [".git"] = {
            icon = "",
            color = "#ff0000",
            name = "Gitignore"
        },
        [".github"] = {
            icon = "",
            color = "#ff0000",
            name = "Github"
        },
    }
}

require("nvim-tree").setup {
  renderer = {
    icons = {
      webdev_colors_folder = true,
    }
  }
}
```

And get this result:

<details>
<summary>Click to show screenshot</summary>

2023-08-15-000051_819x651_scrot

</details>

But I have some questions before continuing. `nvim-web-devicons` doesn't distinguish between directories and files, so if you create a file that matches one of the `override_by_filename` keys it will get the same icon and color, which maybe you intended only for directories, not files. I am not sure how to override only directory names.

On the other hand, what should happen with overriden icons when the directory is open VS closed? Should it use the same overriden icon for both cases, use the overriden icon only for closed directories while using `glyphs.folder.open` for open directories, or should it allow the user to customize both closed and open icons/colors? I think the last option is not possible with `override_by_filename` as `nvim-web-devicons` doesn't know whether a directory is closed or not, that is handled by `nvim-tree`.

Finally, I'm not sure if `get_folder_icon` should return both the icon and highlight or if there should be an equivalent to `Builder:_get_highlight_override`, I implemented it as simply as I could.


alex-courtis:
> On the other hand, what should happen with overriden icons when the directory is open VS closed?

I'm happy to use the same one for open vs closed. The user can see the state via the arrows or indent markers.

> Finally, I'm not sure if get_folder_icon should return both the icon and highlight or if there should be an equivalent to Builder:_get_highlight_override, I implemented it as simply as I could.

We can follow the file pattern i.e. create a `get_folder_icon_webdev` which should simplify things.

alex-courtis:
We do have an issue:  folder icons will appear for users when they didn't previously. We need to preserve existing behaviour.

There are really two questions for files and folders:
1. show webdev icon
2. colourise webdev icon

For 1 we have no solution, which I think #2361 is about. The only way to prevent icons is to disable the plugin.

Proposal: add options: 
[0001-granular-web_devicons-config.patch.gz](https://github.com/nvim-tree/nvim-tree.lua/files/12340393/0001-granular-web_devicons-config.patch.gz)

We will need to add a silent refactor of `renderer.icons.webdev_colors` to `renderer.icons.web_devicons.file.color` in `legacy.lua` to not break existing users.

alex-courtis (changes requested):
Many thanks for the contribution; this will resolve a larger problem.

antoniosarosi:
> We do have an issue: folder icons will appear for users when they didn't previously. We need to preserve existing behaviour.

Isn't this controlled by the new `webdev_colors_folder` option? Since it's set to false by default, folder icons should behave the same as before unless the user sets the option to true , or am I misunderstanding?

alex-courtis:
Many thanks for the updates. I'll get to this on the weekend.

alex-courtis:
> Isn't this controlled by the new `webdev_colors_folder` option? Since it's set to false by default, folder icons should behave the same as before unless the user sets the option to true , or am I misunderstanding?

I'm scope creeping here. We never had any mechanism for controlling the display of icons, just their colour.

We require both controls for folders to prevent breaking existing folder icon behaviour. We should add both controls for file icons now, as the configuration and behaviour would be very inconsistent between files and folders.

antoniosarosi:
> I'm scope creeping here. We never had any mechanism for controlling the display of icons, just their colour.

Still not sure what you mean, if you use `override_by_filename` option when setting up `nvim-web-devicons` you can change the icon of any file. Then there's also `webdev_colors` in `nvim-tree` which is set to `true` by default. So you can control both icons and colors of files using a combination of `override_by_filename` and `webdev_colors`.

`webdev_colors_folder` does almost the same except you can't control icons and colors separately, is that the issue?

Maybe splitting it into two options like `webdev_colors_folder` and `webdev_icons_folder` would be better? That would allow the user to only change the icons or both icons and colors of directories.

alex-courtis:
> Maybe splitting it into two options like `webdev_colors_folder` and `webdev_icons_folder` would be better? That would allow the user to only change the icons or both icons and colors of directories.

Yes please.

Please `git apply 0001-granular-web_devicons-config.patch` the patch linked above. I've drafted all those options in help.

antoniosarosi:
I applied the patch and implemented `web_devicons.file` and `web_devicons.folder` options. I tested all the combinations and it seems to work pretty well. Let me know if you find any bug.

alex-courtis:
That's fantastic! I'll get to this on the weekend.

alex-courtis (approved):
Works beautifully:

```lua
file = { enable = false, color = false, }, folder = { enable = false, color = false, },
file = { enable = false, color = false, }, folder = { enable = false, color = true, },
file = { enable = false, color = false, }, folder = { enable = true, color = false, },
file = { enable = false, color = false, }, folder = { enable = true, color = true, },

file = { enable = false, color = true, }, folder = { enable = true, color = true, },
file = { enable = true, color = true, }, folder = { enable = true, color = true, },
file = { enable = true, color = false, }, folder = { enable = true, color = true, },
```

`renderer.icons.webdev_colors -> renderer.icons.web_devicons.file.color`

Many thanks for your contribution, you have gone well beyond the call of duty.

PR #2376, opened 2023-08-15
fix(#2365): make bufname filterable by shada

JosefLitos:
https://github.com/nvim-tree/nvim-tree.lua/discussions/2365

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -94,7 +94,7 @@ local function create_buffer(bufnr)
 
   local tab = vim.api.nvim_get_current_tabpage()
   BUFNR_PER_TAB[tab] = bufnr or vim.api.nvim_create_buf(false, false)
-  vim.api.nvim_buf_set_name(M.get_bufnr(), "NvimTree_" .. tab)
+  vim.api.nvim_buf_set_name(M.get_bufnr(), "tree://NvimTree_" .. tab)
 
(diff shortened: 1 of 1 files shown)


alex-courtis:
Many thanks for the contribution. I'll get to this on the weekend.

alex-courtis:
Thank you for your contribution, however please accept my apologies @JosefLitos : we won't implement this.

On reflection:
* `:help nvim-tree-api.tree.is_tree_buf()` needs to be fixed
* `utils.is_nvim_tree_buf` needs to be fixed and all its usages audited
* Users do rely on the naming without using the API e.g. https://github.com/nvim-tree/nvim-tree.lua/wiki/Auto-Close

I don't think this change is worth the risk as the blast radius is large.

My [unscientific measurements](https://github.com/nvim-tree/nvim-tree.lua/discussions/2365#discussioncomment-6711350) indicated nvim-tree occupies ~0.25 of the shada history sections and is not overwhelming them.

JosefLitos:
I understand. I haven't considered what losing the path in bufname would cause (although that part could be easily fixed). Anyway, thank you for considering the change.

alex-courtis:
> I understand. I haven't considered what losing the path in bufname would cause (although that part could be easily fixed, but I suppose the prefix is the main issue). Anyway, thank you for considering the change.

Thank you for your understanding and your efforts.

PR #2100, opened 2023-03-30
fix(api): `find_file` doesn't work when `group_empty` option is enabled

liljaylj:

PR #2380, opened 2023-08-16
fix: expand and collapse whole folder groups

Danie-1:
Currently when opening folder groups, the folder icon and arrow indicators don't change correctly.
Current behaviour:
image
With this fix:
image

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -74,19 +74,19 @@ end
 
 -- If node is grouped, return the last node in the group. Otherwise, return the given node.
 function M.get_last_group_node(node)
-  local next = node
-  while next.group_next do
-    next = next.group_next
+  local next_node = node
+  while next_node.group_next do
+    next_node = next_node.group_next
   end
-  return next
+  return next_node
 end
 
 function M.get_all_nodes_in_group(node)
-  local next = utils.get_parent_of_group(node)
+  local next_node = utils.get_parent_of_group(node)
   local nodes = {}
-  while next do
-    table.insert(nodes, next)
-    next = next.group_next
+  while next_node do
+    table.insert(nodes, next_node)
+    next_node = next_node.group_next
   end
   return nodes
 end
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -81,6 +81,16 @@ function M.get_last_group_node(node)
   return next
 end
 
+function M.get_all_nodes_in_group(node)
+  local next = utils.get_parent_of_group(node)
+  local nodes = {}
+  while next do
+    table.insert(nodes, next)
+    next = next.group_next
+  end
+  return nodes
+end
+
 function M.expand_or_collapse(node)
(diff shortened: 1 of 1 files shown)


alex-courtis:
Regression since https://github.com/nvim-tree/nvim-tree.lua/pull/2100

lua/nvim-tree/lib.lua
```diff
@@ -81,6 +81,16 @@ function M.get_last_group_node(node)
   return next
 end
 
+function M.get_all_nodes_in_group(node)
```
alex-courtis, line 84: We could simplify a bit `open_all_nodes_in_group` howeve this function will likely see use in the future.

lua/nvim-tree/lib.lua
```diff
@@ -81,6 +81,16 @@ function M.get_last_group_node(node)
...
+  local next = utils.get_parent_of_group(node)
```
alex-courtis, line 85: Please don't overload the builtin function `next`.

I know it's been used elsewhene...

alex-courtis (changes requested):
Many thanks for addressing this one!

- [x] replace `next`
- [x] bonus points: replace other usages of `next`

alex-courtis:
When removing the last file in a grouped or ungrouped open folder it still shows as opened.

No need to address that; it's an edge case that will likely be difficult to resolve and will have a high blast radius.

lua/nvim-tree/lib.lua
Danie-1: Would a function to "apply function to all nodes in group" be more useful instead maybe?

Danie-1:
I had a brief look at earning the bonus points but it looked like it would take me a little bit of time to figure what would be a suitable variable name to replace it (in some of the places). I can open a pull request for it at some point if noone else does it first.

Danie-1:
Also I was expecting there might still be some edge cases this didn't handle but I was thinking maybe some of those edge cases might be caused by a similar issue to this one, where the state of each node in a group should be changed rather than a single node (but I'm not at all certain about that because I haven't found buggy edge cases or looked at the code for stuff).

lua/nvim-tree/lib.lua
alex-courtis: > Would a function to "apply function to all nodes in group" be more useful instead maybe?

See `NodeIterator`

You could use that here.

alex-courtis:
> I had a brief look at earning the bonus points but it looked like it would take me a little bit of time to figure what would be a suitable variable name to replace it (in some of the places). I can open a pull request for it at some point if noone else does it first.

No worries, well out of scope.

alex-courtis:
> Also I was expecting there might still be some edge cases this didn't handle but I was thinking maybe some of those edge cases might be caused by a similar issue to this one, where the state of each node in a group should be changed rather than a single node (but I'm not at all certain about that because I haven't found buggy edge cases or looked at the code for stuff).

That's fine. Folder/grouping is complex and we can only fix what we know about.

alex-courtis (approved):

PR #2381, opened 2023-08-18
docs: fix typo in API Node section

devansh08:
There is a duplicate `node.` in the the API NODE section for `navigate.git.next`.


gegoune:
Thanks!

PR #2420, opened 2023-09-16
fix: add legacy options safety

Akmadan23:
While testing #2414 I noticed that setting `view` to anything but a table would crash due to this problem in `legacy` module, so I fixed it.


alex-courtis (approved):
Great catch! Works nicely.

PR #2432, opened 2023-09-25
fix: schedule notifications to avoid UI freeze on startup

Akmadan23:
Also added `{ title = "NvimTree" }` for notification plugins that support it.


alex-courtis (approved):
Many thanks - always improving the UX!

Issue #2440, opened 2023-10-02
validate view.width.padding as number|function only


IZ-ZI:
### Description

Getting a message on startup as follows, have been using the config for a long time without any issues:

[NvimTree]
Unknown option: view.width.padding

And when I looked at :help nvim-tree-opts-view, view.width.padding is definitely a valid option. I just finished migrating from packer to Lazy today and accidentally updated a lot of my plugins, is this related?

### Neovim version

```text
NVIM v0.9.1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Windows 10

### nvim-tree version

113e095

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {
    view = {
      width = {
        padding = 3,
        min = 33,
      },
      cursorline = true,
      centralize_selection = true,
    },
    renderer = {
      add_trailing = false,
    },
    actions = {
      file_popup = {
        open_win_config = {
          col = 1,
          row = 1,
          relative = "cursor",
          border = "single",
[...]

PR #2442, opened 2023-10-02
fix(#2440): `view.width.padding` may be a number or function returning a number

reyniersbram:
Update accepted types for view.width option so padding can be set. See #2440.
Overrides documentation change of #2441

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -746,7 +746,7 @@ longest line (previously `view.adaptive_size`).
 
     *nvim-tree.view.width.padding*
     Extra padding to the right.
-      Type: `string`, Default: `" "`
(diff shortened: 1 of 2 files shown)


Akmadan23:
Good job! This actually fixes the issue raised by @IZ-ZI.
By the way @alex-courtis I took a brief look to `nvim-tree/view.lua` and having `padding` as string or function actually results in 0 padding, so probably it's better to only accept numbers.

alex-courtis:
> By the way @alex-courtis I took a brief look to `nvim-tree/view.lua` and having `padding` as string

Yes. It's good to see the type validator working and protecting against this ;)

alex-courtis (approved):

alex-courtis:
> specify in the documentation that the function should return a number.

Done

PR #2453, opened 2023-10-08
feat: notify [NvimTree] prefix is multiline only if the message is multiline

alex-courtis:
This was a little annoying:
```
[NvimTree]
notify.lua added to clipboard.
Press ENTER or type command to continue
```


Akmadan23 (approved):
Right, I didn't think about single line messages, this way is really clever.

PR #2455, opened 2023-10-08
feat(#2415): highlight overhaul (#2454)

### LINKED ISSUE #2415: Highlight Overhaul
#2410 demonstrated that highlighting can be additive e.g. add underline with a different colour.

There are some inconsistencies in how highlighting is applied e.g. `highlight_opened_files` vs `highlight_git`

- [x] Make all highlights additive as per #2410
  - [x] git
  - [x] diagnostics
  - [x] opened
  - [x] modified
  - [x] bookmarks
  - [x] clipboard
- [x] Make all highlights granular `"none", "icon", "name", "all"`
  - [x] highlight_git
  - [x] highlight_diagnostics
  - [x] highlight_opened_files
  - [x] highlight_modified
  - [x] highlight_bookmarks
  - [x] highlight_clipboard
- [x] Refactor highlight groups to be consistent and end with `HL` and `Icon`
  - [x] git
  - [x] diagnostics
  - [x] opened
  - [x] modified
  - [x] bookmarks
  - [x] clipboard
- [x] clean up and standardise `get_colours`
- [x] cterm compatibility
- [x] investigate term compatibility (e.g. linux) - colours and highlights are distinct and readable, just missing icons
- [x] legacy highlight group linking
- [x] simplify renderer builder for efficiency

Refactor may be silently done as per options.

alex-courtis:
fixes #2415

@gegoune @Akmadan23 apologies for this very large change; breaking it up was not feasible.

I'd be most grateful if you could test this for a week or so...


Akmadan23:
I noticed that `renderer.highlight_git` was missing from the accepted strings, so I simply added it.

By the way, I am working on adding type annotations to almost all functions, and I was wondering if it would be worth it to do it now and deal with the conflicts with this branch or wait for this to be merged... What do you think?

lua/nvim-tree.lua
```diff
@@ -660,9 +660,11 @@ local ACCEPTED_STRINGS = {
     signcolumn = { "yes", "no", "auto" },
   },
   renderer = {
+    highlight_git = { "none", "icon", "name", "all" },
```
tomcri, line 663: I was working on something similar because I needed to highlight only names but not git icons, by the time I was writing the code I discovered icons.show.git = false that did the trick for me, I'm glad that you implemented this feature.

lua/nvim-tree/renderer/decorator/git.lua
```diff
@@ -0,0 +1,205 @@
...
+    ["T "] = "NvimTreeGitFileStagedHL",
+    ["TT"] = "NvimTreeGitFileStagedHL",
+    [" M"] = "NvimTreeGitFileDirtyHL",
+    ["CM"] = "NvimTreeGitFileDirtyHL",
+    [" C"] = "NvimTreeGitFileDirtyHL",
+    [" T"] = "NvimTreeGitFileDirtyHL",
+    ["MM"] = "NvimTreeGitFileDirtyHL",
+    ["AM"] = "NvimTreeGitFileDirtyHL",
+    dirty = "NvimTreeGitFileDirtyHL",
+    ["A "] = "NvimTreeGitFileNewHL",
```
tomcri, line 78: I found a bug here for that I made a pr (2534) , I see you moved the logic from components to decorator, if my pr don't pass, please can you fix this one in your pr? ["A "] = { icons.staged }, vs ["A "] = "NvimTreeFileStaged", can be reproduced: create a new file in your git repository and after that stage it (git add filename), I think the correct state is staged icons are good but highlight not.

alex-courtis: Many thanks @tomcri !

I'll make sure that fix is not lost...

lua/nvim-tree.lua
alex-courtis: Many thanks @Akmadan23

alex-courtis:
> I noticed that `renderer.highlight_git` was missing from the accepted strings, so I simply added it.
> 
> By the way, I am working on adding type annotations to almost all functions, and I was wondering if would be worth it to do it now and deal with the conflicts with this branch or wait for this to be merged... What do you think?

That's fantastic! It's been bothering me but I never seem to get a chance.

The big one would be a Node class... lot's of surface area last time I experimented...

You go first, this branch seems pretty localised and should merge without dramas.

Akmadan23:
> The big one would be a Node class...

That's the only one I haven't added yet, I'll try my best.

> You go first, this branch seems pretty localised and should merge without dramas.

Yeah, maybe I can skip the `renderer` module since you are rewriting it. That was my only concern.

alex-courtis:
> That's the only one I haven't added yet, I'll try my best.

Oh wow! Sounds like you're making great progress...

Akmadan23:
After some testing I noticed that the "partial" line highlight isn't consistent with all icon types, namely the default file icon is not included in the line highlight whereas (probably?) all other icons are. I took a quick look at the code but I didn't find the issue, but you surely know better where to look.

alex-courtis:
> Edit 2: turns out this "incomplete" line highlight is due to my `on_attach` function which overrides the `Normal` hl group. It works correctly when I open nvim-tree normally, but it doesn't on `nvim .`. This is weird because on master it works fine (kind of), however there is no need to fix it now since it's very much an edge case. I can eventually dive into it to and see what's the issue.

Many thanks for all the detailed testing!

That should probably be fixed before release - I'm sure you're not the only one. Can you please link your on_attach so that I can attempt a fix?

I have little available time right now however I will over the new year so I'll get onto it then.

Akmadan23:
This is my minimal `on_attach`:
```lua
on_attach = function()
  vim.schedule(function()
    local winnr = require("nvim-tree.view").get_winnr()
    if winnr then
      local ns = vim.api.nvim_create_namespace("NvimTree")
      vim.api.nvim_win_set_hl_ns(winnr, ns)
      vim.api.nvim_set_hl(ns, "Normal", { fg = "#F8F8F2", bg = "#272822" })
    end
  end)
end
```

I use this namespace trick to have a different background color in nvim-tree windows.

> This is weird because on master it works fine (kind of)

What I meant here is that on master the current line highlight isn't interrupted, but the background color is not the one I specified in my `on_attach`, instead uses the default `Normal` background color without transparency, so that's still pretty weird...

alex-courtis:
> Edit: I tested the master branch with `NvimTreeIndentMarker` linked to `Normal` and the interrupted line highlight comes up, so nothing in this PR created the issue, it's probably something going on with conflicting highlights in different namespaces, and it has always been around. This is why I think this doesn't need to be addressed here.

I see, thank you. So long as I'm not causing a regression...

alex-courtis:
There is a problem with additive highlights, which "limit 2" does not address properly, as it removes highlighting that should not be, such as icon colour.

```vim
:hi NvimTreeBookmarkHL guibg=blue
:hi NvimTreeCopiedHL guibg=magenta
:hi NvimTreeCutHL guibg=red
```

```lua
  vim.api.nvim_buf_add_highlight(0, -1, "DevIconC", 1, 2, 5)
  vim.api.nvim_buf_add_highlight(0, -1, "NvimTreeBookmarkHL", 1, 2, 5)
  vim.api.nvim_buf_add_highlight(0, -1, "NvimTreeCopiedHL", 1, 2, 5)
  vim.api.nvim_buf_add_highlight(0, -1, "NvimTreeCutHL", 1, 2, 5)
```

Icon is drawn in C blue, with blue bookmark HL.

Move the DevIconC to the end and it's drawn in magenta copied HL.

It seems that only one attribute may be overridden.

Possible solution: build and apply dynamic highlight groups based on combining highlight groups.

alex-courtis:
Pulling the trigger. It's been stably dogfooded for 2 weeks.

I have 3 days to deal with any regressions or questions.

PR #2448, opened 2023-10-07
chore: stylua column width 120 -> 140

alex-courtis:
No `"if$"` remaining, adds readability.

#2449 increases to 160 however those long lines are genuinely too long. Manually resolved those, see with comments.

PR #2456, opened 2023-10-08
refactor: format tables line by line for better readability

Akmadan23:
Discussed in #2448. Not sure if this is more of a refactor or a chore. Either way now almost every table with more than 3 members is formatted line by line, greatly improving readability.


alex-courtis (approved):
This is a lot more readable, especially API.

Many thanks!

PR #2464, opened 2023-10-12
feat: add preview with no window picker action

umlx5h:
Hi.

I often use preview feature, But if there is more than one window, It always prompts picker selection menu.
But I don't want to disable picker feature entirely.

I think that the option to not use the picker should be provided in preview as well, since it is the same kind of functionality as 'Open'.

When I set 'Open Preview: No Window Picker' to match 'Open: No Window Picker', the help message became too long horizontally.
So I chose 'Open Preview: No Picker'.

Feel free to change the name if you like.

I ran this script and githooks.
```
$ scripts/update-help.sh
```

no indent diff
https://github.com/nvim-tree/nvim-tree.lua/pull/2464/files?w=1


alex-courtis:
> So I chose 'Open Preview: No Picker'.

I like that.

Updating the API to match.

lua/nvim-tree/api.lua
```diff
@@ -171,16 +171,6 @@ local function open_or_expand_or_dir_up(mode)
   end
 end
 
-local function open_preview(node)
```
alex-courtis, line 174: Nice cleanup.

alex-courtis (approved):
Tested OK:

- [x] node.name == ".."
- [ ] node.nodes
- [ ] else

Issue #2466, opened 2023-10-13
git integration not working in Windows.


peter-lyr:
### Description

5 git jobs have timed out after git.timeout 400ms, disabling git integration.

### Neovim version

```text
NVIM v0.9.2
Build type: RelWithDebInfo
LuaJIT 2.1.1694082368
```

### Operating system and version

Windows 10

### nvim-tree version

53b0bca

### Clean room replication

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {
	git = {
		enable = true,
		ignore = false,
	},
	filesystem_watchers = {
		enable = false,
		debounce_delay = 50,
	},
	log = {
		enable = true,
		types = {
			all = false,
			config = false,
			copy_paste = false,
			dev = false,
			diagnostics = false,
			git = true,
			profile = false,
			watcher = true,
        },
	},
	diagnostics = {
		enable = true,
		show_on_dirs = false,
	},
	update_focused_file = {
		enable = true
	},
	renderer = {
		group_empty = true,
	},
	filters = {
		dotfiles = false,
	},
  }
end
```

### Steps to reproduce

[...]

PR #2478, opened 2023-10-17
fix(#2467): remove newline in git paths

### LINKED ISSUE #2467: Any git directory will report [NvimTree] Could not start the fs_event watcher for path C:\test_project\.git  : ENOENT
### Description

After a few months without updating any plugins, today i did a :PackerUpdate, opened a random git directory, got this warning:
`[NvimTree]
Could not start the fs_event watcher for path C:\test_project\.git
 : ENOENT`
After some reading, and comparing the logs with a no-issue ubuntu with mostly the same config. i think there is a newline/return issue there.
I don't think i'm skillful enough to dig deeper, I hope the info i provided here is useful, Thanks.

### Neovim version

```text
NVIM v0.9.1
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

Windows 10

### nvim-tree version

4054fc4

### Clean room replication

```lua
unmodified
```

### Steps to reproduce

1. nvim -nu /tmp/nvt-min.lua C:/any_dir_has_a_.git
2. :NvimTreeOpen
3. console pop up, issue arise:
[NvimTree]
Could not start the fs_event watcher for path C:\test_project\.git
 : ENOENT

### Expected behavior

_No response_

### Actual behavior

[nvim-tree.zip](https://github.com/nvim-tree/nvim-tree.lua/files/12906681/nvim-tree.zip)

mj2068:
Fixes #2467 (potentially also for #2466), regarding failed git integration on Windows.
Background: In file git/utils.lua, the two paths variables toplevel and git_dir are returned by vim.fn.system(...), which will add a trailing newline character, causing them to be invalid paths.
Fix: Added two lines of gsub function to remove any newline characters in these paths.

After preliminary tests on Ubuntu, I suggest caution for the general behavior of vim.fn.system() on all systems, where its return value seems to always contain a trailing newline, therefore its direct return value can't be used as paths.

### DIFF
--- a/lua/nvim-tree/git/utils.lua
+++ b/lua/nvim-tree/git/utils.lua
@@ -41,10 +41,14 @@ function M.get_toplevel(cwd)
       if vim.v.shell_error ~= 0 then
         return nil, nil
       end
+      -- remove trailing newline(\n) character added by vim.fn.system
+      toplevel = toplevel:gsub("\n", "")
       git_dir = vim.fn.system("cygpath -w " .. vim.fn.shellescape(git_dir))
       if vim.v.shell_error ~= 0 then
         return nil, nil
       end
+      -- remove trailing newline(\n) character added by vim.fn.system
+      git_dir = git_dir:gsub("\n", "")
     end
     toplevel = toplevel:gsub("/", "\\")
     git_dir = git_dir:gsub("/", "\\")
--- a/lua/nvim-tree/git/utils.lua
+++ b/lua/nvim-tree/git/utils.lua
@@ -38,13 +38,13 @@ function M.get_toplevel(cwd)
     -- msys2 git support
     -- cygpath calls must in array format to avoid shell compatibility issues
     if M.use_cygpath then
-      toplevel = vim.fn.system({ "cygpath", "-w", toplevel })
+      toplevel = vim.fn.system { "cygpath", "-w", toplevel }
       if vim.v.shell_error ~= 0 then
         return nil, nil
       end
       -- remove trailing newline(\n) character added by vim.fn.system
       toplevel = toplevel:gsub("\n", "")
-      git_dir = vim.fn.system({ "cygpath", "-w", git_dir })
+      git_dir = vim.fn.system { "cygpath", "-w", git_dir }
       if vim.v.shell_error ~= 0 then
         return nil, nil
       end
--- a/lua/nvim-tree/git/utils.lua
+++ b/lua/nvim-tree/git/utils.lua
@@ -36,14 +36,15 @@ function M.get_toplevel(cwd)
   -- git always returns path with forward slashes
   if vim.fn.has "win32" == 1 then
     -- msys2 git support
+    -- cygpath calls must in array format to avoid shell compatibility issues
     if M.use_cygpath then
-      toplevel = vim.fn.system("cygpath -w " .. vim.fn.shellescape(toplevel))
+      toplevel = vim.fn.system({ "cygpath", "-w", toplevel })
       if vim.v.shell_error ~= 0 then
         return nil, nil
       end
       -- remove trailing newline(\n) character added by vim.fn.system
       toplevel = toplevel:gsub("\n", "")
-      git_dir = vim.fn.system("cygpath -w " .. vim.fn.shellescape(git_dir))
+      git_dir = vim.fn.system({ "cygpath", "-w", git_dir })
       if vim.v.shell_error ~= 0 then
         return nil, nil
       end
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -29,9 +29,10 @@ CONTENTS                                                           *nvim-tree*
   5.14 Opts: Trash                              |nvim-tree-opts-trash|
   5.15 Opts: Tab                                |nvim-tree-opts-tab|
   5.16 Opts: Notify                             |nvim-tree-opts-notify|
-  5.17 Opts: UI                                 |nvim-tree-opts-ui|
-  5.18 Opts: Experimental                       |nvim-tree-opts-experimental|
-  5.19 Opts: Log                                |nvim-tree-opts-log|
+  5.17 Opts: Help                               |nvim-tree-opts-help|
+  5.18 Opts: UI                                 |nvim-tree-opts-ui|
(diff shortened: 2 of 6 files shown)


lua/nvim-tree/git/utils.lua
```diff
@@ -40,10 +40,14 @@ function M.get_toplevel(cwd)
       if vim.v.shell_error ~= 0 then
         return nil, nil
       end
+      -- remove trailing newline(\n) character added by vim.fn.system
+      toplevel = toplevel:gsub("\n", "")
```
alex-courtis, line 44: This is inside the cygpath block.

I was under the impression that you are using PowerShell. Does it see cygpath?

alex-courtis (approved):
Looks good to me, I'm just not certain how this works for PowerShell

lua/nvim-tree/git/utils.lua
mj2068: Apologize for didn't fully describe my environment before.
Yes, I use PowerShell, but I also have msys2 installed on my Windows 10.
Yes, PowerShell sees "cygpath", because the POSIX compatible environment provided msys2 is based on Cygwin. So Cygwin tools like "cygpath"(exe files you can run directly) is in the Windows path.

alex-courtis: > PowerShell sees "cygpath" is because it's inside X:/msys2/usr/bin, and this bin path is in the Windows path environment variable.

That explains a lot - a valid use case that we never expected or handled.

I'm prepared for disappointment after your testing, however we will fix this one way or another.

Big favour, if possible: please run `has.lua` for neovim inside msys2. It would be great to fill out the entirety of [OS Feature Flags](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development#os-feature-flags)

mj2068: Had to do some reading about how msys2 works. Then I did the following inside the msys2 official shell, with nvim 0.9.4 installed from [MSYS2 Packages](https://packages.msys2.org/package/mingw-w64-x86_64-neovim?repo=mingw64) using pacman, following their doc. 

Flags: [has_list_msys2_mingw.log](https://github.com/nvim-tree/nvim-tree.lua/files/13063808/has_list_msys2_mingw.log)

I then did quick git integration tests regarding the new `cygwin_support` manual option, false and true, using the latest [commit](https://github.com/nvim-tree/nvim-tree.lua/commit/83b699533b279af6d207d6c3860d4301556d65b4).

With the "Clean Room" lua file, test folder "project1" git initialized, which contains one untracked file "README.md".
After "Ready!" message, :NvimTreeOpen, select "project1", hit Enter, the results are:

cygwin_support = false(default):
shown error message: `[NvimTree] Could not start the fs_event watcher for path \tmp\project1\.git : ENOENT`
log: [nvim-tree_cygwin_false.log](https://github.com/nvim-tree/nvim-tree.lua/files/13063965/nvim-tree_cygwin_false.log)

cygwin_support = true:
no error, but no git tracking.
log: [nvim-tree_cygwin_true.log](https://github.com/nvim-tree/nvim-tree.lua/files/13063984/nvim-tree_cygwin_true.log)

I think we need to normalize the paths.

Thx, alex.

Attachs the two screenshots, false and true respectively:
Screenshot 2023-10-23 004555

Screenshot 2023-10-23 004741

alex-courtis: That looks great, many thanks for the depth.

Confirming:
* we do need the path fixing from this PR
* `cygwin_support = true` is needed for msys2 and now functions correctly

alex-courtis: Thank you for all your diligent testing with this unfamiliar environment. This is quite the difficult issue...

It looks like the `cygwin_support = true` case is not matching any toplevel and returning nils. It's definitely hitting the `use_cygpath` branch as the `win32` feature flag is set and we are seeing different results.

To confirm:
* does this PR fix the above?
* do we need to do more right now or can we merge this?

alex-courtis:
Can I ask a big favour:

#2491 is a guess to attempt some path normalisation however may be incorrect or destructive.

Could you please attempt to incorporate the above into this PR?
It would be best to have all the pieces tested together in known environment(s). If #2491 is not useful please disregard it.

lua/nvim-tree/git/utils.lua
mj2068: > * does this PR fix the above?

It did, because the new "cypwin_support" option affected this.

> * do we need to do more right now or can we merge this?

I don't think we can merge this. But the issue is getting clearer to me. I'm working on a fix, can I have few days?

mj2068:
It's okay, alex, glad to help.

As to #2491, I don't think "\r\n" is the source of the issue.

vim.fn.system will always convert "\r\n" to "\n", according to:
:help system()

> 		Result is a String, filtered to avoid platform-specific quirks:
>		- <CR><NL> is replaced with <NL>
>		- NUL characters are replaced with SOH (0x01)

I confirmed this vim.fn.system behavior in my previous tests.

Anyway, will be back. Peace.

mj2068:
And @alex-courtis, I'm not very familiar with git PR workflow, I did some git things in my fork, like, fetch upstream, merge, rebase, etc., according to some tutorials. Now, this PR page shows me, it has changed a lot files. I hope I don't break anything, did I do something wrong?

alex-courtis:
This sounds fantastic, don't worry about CI etc. We can get this right then we tidy.

I'll have a chance to look / review on the weekend.

mj2068:
Finally.
The compatibility issues are really complicated.

But I think this fixed it.
I tested git integration on Windows and Ubuntu.
Windows:
  PowerShell cygwin_support = false, OK.
  PowerShell cygwin_support = true, OK.
  MSYS2 MinGW shell cygwin_support = true, OK.
  MSYS2 MinGW shell cygwin_support = false, NOT. (paths incompatible)
Ubuntu, OK(not affected).

Take a look.

alex-courtis:
This PR is seeing up phantom changes; this branch is definitely up to date: https://github.com/nvim-tree/nvim-tree.lua/compare/master...mj2068:nvim-tree.lua:bugfix

alex-courtis (approved):
This is just fantastic, thank you for the great consideration and testing!

Closed #2491

PR #2482, opened 2023-10-20
feat: add option to sort entries in help window

Akmadan23:
Just an option that is nice to have

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -29,9 +29,10 @@ CONTENTS                                                           *nvim-tree*
   5.14 Opts: Trash                              |nvim-tree-opts-trash|
   5.15 Opts: Tab                                |nvim-tree-opts-tab|
   5.16 Opts: Notify                             |nvim-tree-opts-notify|
-  5.17 Opts: UI                                 |nvim-tree-opts-ui|
-  5.18 Opts: Experimental                       |nvim-tree-opts-experimental|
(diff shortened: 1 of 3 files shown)


alex-courtis (approved):
Works beautifully. I think I'll enable this one....

Could we go one step further and add a mapping to the window to cycle sort methods?

Akmadan23:
The only minor issue is that the keymap is hardcoded (just like `q` to quit), not sure if it's worth to give the user a way to customize them.

alex-courtis:
Works beautifully, no stray buffers.

> The only minor issue is that the keymap is hardcoded (just like `q` to quit), not sure if it's worth to give the user a way to customize them.

I think that's fine; it's very clear what the mappings are and they are mnemonics.

alex-courtis:
Now that we have the clear mapping do we need the `help.sort_by` option?

We have too many options already...

Akmadan23:
> Now that we have the clear mapping do we need the `help.sort_by` option?

Well, the option can still be useful for those (like me) that don't mind changing the sorting method each time but rather have the preferred one active by default.

alex-courtis:
> Well, the option can still be useful for those (like me) that don't mind changing the sorting method each time but rather have the preferred one active by default.

That's fair enough; it is something that is used frequently. We can remove or refactor if needed.

I've also got a half formed idea of executing commands that may need a help config.

alex-courtis (approved):

alex-courtis (approved):
Fantastic work as always!

PR #2483, opened 2023-10-20
feat: use IEC binary size prefixes

Andrew15-5:
I'm very happy that this project uses 1024 as a divisor instead of 1000. But to explicitly inform user about it the prefixes should consist of 3 letters:
- multiple (K, M, G, T)
- 'i' (bInary)
- 'B' (Byte)

Exception is any number of bytes that are less than 1024, then the prefix is just the 'B' letter.

As an addition I've completed the list with the missing prefixes: P, E, Z, Y.

Quick reference: https://en.wikipedia.org/wiki/Binary_prefix.

P.S. Ri and Qi are not yet confirmed (only SI counterparts exist starting from 2022: R, Q).

### DIFF
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -292,7 +292,7 @@ function M.format_bytes(bytes)
 
   pow = pow + 1
 
-  return (units[pow] == nil) and (bytes .. "B") or (value .. units[pow])
+  return (units[pow] == nil) and (bytes .. units[1]) or (value .. units[pow] .. "i" .. units[1])
 end
 
 function M.key_by(tbl, key)
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -281,7 +281,7 @@ function M.move_missing_val(src, src_path, src_pos, dst, dst_path, dst_pos, remo
 end
 
 function M.format_bytes(bytes)
-  local units = { "B", "K", "M", "G", "T" }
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):

We will indeed be using decimal prefixes over my dead body.

alex-courtis:
I'll need you to rebase as I don't have push privileges to your fork.

Andrew15-5:
Wait-wait-wait, hold on. So, you're approving of my changes, but saying that you will use decimal prefixes? And the title now has "SI binary size"?

If you look at the table in the link, there is a separate decimal column and binary column or SI and IEC. Decimal is for multiples of 1000 and binary is for multiples of 1024.

So tell me, which pill do you take? (red or blue) /jk

alex-courtis:
> Wait-wait-wait, hold on. So, you're approving of my changes, but saying that you will use decimal prefixes? And the title now has "SI binary size"?
> 
> If you look at the table in the link, there is a separate decimal column and binary column or SI and IEC. Decimal is for multiples of 1000 and binary is for multiples of 1024.
> 
> So tell me, which pill do you take? (red or blue) /jk

Bad joke; binary it is.

I'm happy, just need a rebase before I can merge.

Andrew15-5:
I have found a bug (because I didn't know how the logic works exactly) and 2 more improvements. Expect a new PR in the next 24 hours.

alex-courtis:
> I have found a bug (because I didn't know how the logic works exactly) and 2 more improvements. Expect a new PR in the next 24 hours.

Many thanks for your diligence.

I don't think we need to revert as we'll fix it quickly.

Issue #2460, opened 2023-10-09
Navigation via `diagnostics.next` and `diagnostics.prev` doesn't work


hinell:
### Description

I can't find similar issue.

Hitting both `]e` and `[e` doesn't move cursor to the next/previous diagnostic.
Neither it works via programmatic calls (you have to open `nvim-tree` window):
```lua
lua =require("nvim-tree.api").node.navigate.diagnostics.prev()

```

### Neovim version

```text
NVIM v0.10.0-dev-ac1c23442f
Build type: MinSizeRel
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux/Kubuntu

### nvim-tree version

53b0bcaadaffb505acff230578b56a86ec1ab38a

### Clean room replication

```lua
Experiencing same behavior
```

### Steps to reproduce

1. cd into git repo with changes
2. `:NvimTreeToggle`
3. Hit `]e` or `[e`

### Expected behavior

I've used that feature much and not sure how exactly it should work really.

alex-courtis:
I'm not able to replicate, although that doesn't really prove anything.

`]e` and `[e` will navigate to `commands.lua` and `lib.lua`

![20231015_115105](https://github.com/nvim-tree/nvim-tree.lua/assets/1505378/22b42fbb-75ae-4136-abff-c51cc36ceb1f)

Let's dig deeper:
* Please test with stable nvim, `0.9.4` at time of writing.
* Please provide a clean room reproducer

hinell:
> Please provide a clean room reproducer

It fails in clean room as well.

alex-courtis:
> It fails in clean room as well.

In that case it could be data related - files/language or the LSP/COC in use.

Please provide a replicator that I can attempt:
* project to create / check out
* nvt-min.lua with the LSP/COC configured
* instructions to replicate e.g. run `nvim -nu /tmp/nvt-min.lua` in X directory, modify Y file, open tree etc.

alex-courtis:
Here's an LSP nvt-min.lua template you could start from, which uses `lua-language-server`

Insert your own language server as required.

[nvt-min.lua.gz](https://github.com/nvim-tree/nvim-tree.lua/files/12910945/nvt-min.lua.gz)

hinell:
Do the `[e` bindings open a file or just focus a ronw in `nvim-tree` window? I didn't use `[e` before

gegoune:
[...]

PR #2484, opened 2023-10-21
docs: add (disabled) diagnostics config to bug report template

alex-courtis:
Drafted during #2460

### DIFF
--- a/.github/ISSUE_TEMPLATE/nvt-min.lua
+++ b/.github/ISSUE_TEMPLATE/nvt-min.lua
@@ -35,3 +35,13 @@ _G.setup = function()
   require("nvim-tree").setup {}
 end
 
+-- UNCOMMENT this block for diagnostics issues, substituting pattern and cmd as appropriate.
+-- Requires diagnostics.enable = true in setup.
+--[[
+vim.api.nvim_create_autocmd("FileType", {
+  pattern = "lua",
+  callback = function()
+    vim.lsp.start { cmd = { "lua-language-server" } }
+  end,
+})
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/nvt-min.lua
```diff
@@ -11,6 +11,10 @@ local function load_plugins()
       "wbthomason/packer.nvim",
       "nvim-tree/nvim-tree.lua",
       "nvim-tree/nvim-web-devicons",
+
+      -- UNCOMMENT for diagnostics issues
+      -- "neovim/nvim-lspconfig",
```
gegoune, line 16: Shouldn't we request native lsp start function call over additional plugin?

alex-courtis: Yes please!

Do you know how to get that going?

Akmadan23 (approved):
This makes total sense. Seeing this I even thought to add a separate bug report template for diagnostics related issues but maybe that's too much. This seems an optimal middle ground.

.github/ISSUE_TEMPLATE/nvt-min.lua
gegoune: I guess we can just direct to `:h vim.lsp.start` which has an example right there.

alex-courtis: Well... I was overthinking that; I didn't realise it was that simple.

Please test drive the new commented out template.

PR #2492, opened 2023-10-22
fix: Fixed "BiB" bug and improved binary prefixes

Andrew15-5:
This is a new patch after #2483.

What's new:
- Now there is a whitespace between value and unit (this is the correct way, same as in git, crates.io; see https://github.com/typst/typst/discussions/2100#discussioncomment-6961087).
- Now values >= 1024 YiB are shown in YiB instead of B.
- To reuse same code a new local function was added: round().

I've also added comments to clarify some things. The `round()` function is local, but the plugin works correctly (I thought it wouldn't). I don't know how exactly `M` from `utils.lua` is used, so I didn't do `M.round()`.

### DIFF
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -280,19 +280,38 @@ function M.move_missing_val(src, src_path, src_pos, dst, dst_path, dst_pos, remo
   end
 end
 
+local function round(value)
+  -- Amount of digits to round to after floating point.
+  local digits = 2
+  local round_number = 10 ^ digits
+  return math.floor((value * round_number) + 0.5) / round_number
+end
+
 function M.format_bytes(bytes)
   local units = { "B", "K", "M", "G", "T", "P", "E", "Z", "Y" }
+  local i = "i" -- bInary
 
   bytes = math.max(bytes, 0)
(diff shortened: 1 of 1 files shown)


Andrew15-5:
I also created a separate `i` variable that holds the `"i"` string. I just wanted the function to be as useful as possible. The way it is written right now, you can, for example, change language very easely, if someone wants to. And since `units[1]` is used instead of `B`, I figured why not use another variable for (repetitive) `"i"` as well.

alex-courtis:
> I've also added comments to clarify some things. The `round()` function is local, but the plugin works correctly (I thought it wouldn't). I don't know how exactly `M` from `utils.lua` is used, so I didn't do `M.round()`.

That's fine, round should be private i.e. not a member of utils

alex-courtis (approved):

Works as intended, space is more readable.

Andrew15-5:
Oh, right, I forgot to include this (now fixed):

> * I also increased number of digits to round to from 1 to 2 (same as in git, crates.io).

alex-courtis:
> Oh, right, I forgot to mention this (now fixed):
> 
> > * I also increased number of digits to round to from 1 to 2 (same as in git, crates.io).

Many thanks. I'm quite happy with the results.

PR #2500, opened 2023-10-26
feat(#2498): add option for "Y/n" prompts

### LINKED ISSUE #2498: change default value to ui.confirm.remove/trash
Would it be possible to add a default for the `ui.confirm` so that when prompted on `delete` or `trash` it would be `Y` instead of `N`?
`Remove file.txt ? y/N:` -> `Remove file.txt ? Y/n:`

I tried setting `ui.confirm.remove` to `false`, but then there is no prompt at all. I'd prefer to just hit enter on prompt so I don't have to type the `y`

Akmadan23:
Closes #2498.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -573,6 +573,7 @@ Following is the default configuration. See |nvim-tree-opts| for details.
         confirm = {
           remove = true,
           trash = true,
+          default_yes = false,
         },
       },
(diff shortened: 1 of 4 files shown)


alex-courtis (approved):
Looking good as always Akmadan

Tested combinations of:
* default_yes
  * enter
  * n
  * y
* confirm.trash
* confirm.remove

PR #2502, opened 2023-10-30
feat(full_name): do not exclude root node

przepompownia:


przepompownia:
What do you think about including the parent node to `full_name`?

image
image
image

lua/nvim-tree/renderer/components/full-name.lua
```diff
@@ -60,8 +57,9 @@ local function show()
   end
 
   M.popup_win = vim.api.nvim_open_win(vim.api.nvim_create_buf(false, false), false, {
-    relative = "win",
-    bufpos = { vim.fn.line "." - 2, 0 },
+    relative = "cursor",
+    row = 0,
+    col = 1 - vim.fn.getcursorcharpos()[3],
```
alex-courtis, line 62: This is a more sensible way to achieve positioning.

alex-courtis (approved):
Great catch; this has been broken a long time, possibly forever.

We don't see the window when the tree buffer is scrolled horizontally, however that's existing behaviour and not really a sensible use case.

lua/nvim-tree/renderer/components/full-name.lua
przepompownia:
```lua
vim.wo.number = true
vim.api.nvim_buf_set_lines(0, 0, 1, true, {'', '', ''})
vim.bo[0].buftype = 'nofile'

local buf = vim.api.nvim_create_buf(false, true)
vim.api.nvim_buf_set_lines(buf, 0, 0, true, {'>---------------->'})
vim.bo[buf].buftype = 'nofile'

local winId = nil

local function test()
  if winId then
    vim.api.nvim_win_close(winId, true)
  end
  local bufline = vim.api.nvim_win_get_cursor(0)[1]
  print(bufline)
  local winConfig = {
    relative = 'win',
    width = 50,
    height = 1,
    style = 'minimal',
    bufpos = { 1, 0 },
  }
  winId = vim.api.nvim_open_win(buf, false, winConfig)
end

test()
vim.api.nvim_create_autocmd('CursorMoved', {callback = test})
```
Initially I tried `relative = 'win'` but never didn't able to cover the the first line. I then switched to try `cursor`.

przepompownia:
Thanks for merging. I hope it works also on cases not used usually by me.

Issue #2480, opened 2023-10-18
Floating: Cannot close last window


geril2207:
### Description

Error when deleting opened file with floating tree window.

https://github.com/nvim-tree/nvim-tree.lua/assets/62308020/fb9ccb80-19ef-47c9-a1f1-84b13624fb33

### Neovim version

```text
NVIM v0.10.0-dev-9f32deb
Build type: Release
LuaJIT 2.1.1696795921
```

### Operating system and version

Fedora 6.5.5-200.fc38.x86_64

### nvim-tree version

40b9b88

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvt-min/site]])
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			"nvim-tree/nvim-tree.lua",
			"nvim-tree/nvim-web-devicons",
			-- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
		},
		config = {
			package_root = package_root,
			compile_path = install_path .. "/plugin/packer_compiled.lua",
			display = { non_interactive = true },
		},
	})
end
if vim.fn.isdirectory(install_path) == 0 then
	print("Installing nvim-tree and dependencies.")
	vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
vim.cmd([[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]])
vim.opt.termguicolors = true
vim.opt.cursorline = true

local width_ratio = 0.5
local height_ratio = 0.8

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
	require("nvim-tree").setup({
		view = {
			float = {
				enable = true,
				quit_on_focus_loss = false,
				open_win_config = function()
					local screen_w = vim.opt.columns:get()
					local screen_h = vim.opt.lines:get() - vim.opt.cmdheight:get()

					local window_w = screen_w * width_ratio
					local window_h = screen_h * height_ratio
[...]

PR #2503, opened 2023-10-30
fix: error when deleting opened file from floating window

geril2207:
https://github.com/nvim-tree/nvim-tree.lua/issues/2480

### DIFF
--- a/lua/nvim-tree/actions/fs/remove-file.lua
+++ b/lua/nvim-tree/actions/fs/remove-file.lua
@@ -9,7 +9,10 @@ local M = {
 }
 
 local function close_windows(windows)
-  if view.View.float.enable and #vim.api.nvim_list_wins() == 1 then
+  -- Prevent from closing when the win count equals 1 or 2,
+  -- where the win to remove could be the last opened.
+  -- For details see #2503.
+  if view.View.float.enable and #vim.api.nvim_list_wins() < 3 then
     return
   end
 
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/actions/fs/remove-file.lua
```diff
@@ -22,17 +22,17 @@ end
...
       if buf.hidden == 0 and (#bufs > 1 or view.View.float.enable) then
-        local winnr = vim.api.nvim_get_current_win()
         vim.api.nvim_set_current_win(buf.windows[1])
         vim.cmd ":bn"
-        if not view.View.float.enable then
-          vim.api.nvim_set_current_win(winnr)
-        end
       end
       vim.api.nvim_buf_delete(buf.bufnr, { force = true })
+      if not view.View.float.quit_on_focus_loss then
```
geril2207, line 33: found that `vim.api.nvim_buf_delete` does not delete the buffer if there is no focus on it, so firstly delete buffer then return focus.

lua/nvim-tree/actions/fs/remove-file.lua
```diff
@@ -9,7 +9,7 @@ local M = {
 }
 
 local function close_windows(windows)
-  if view.View.float.enable and #vim.api.nvim_list_wins() == 1 then
+  if view.View.float.enable and #vim.api.nvim_list_wins() < 3 then
```
alex-courtis, line 12: Please leave a comment explaining the magic number 3

alex-courtis (changes requested):
Fantastic work, tested OK:

not float.enable
- [x] remove only tree open
- [x] remove open
- [x] remove not open

float.enable and float.quit_on_focus_loss
- [x] remove only tree open
- [x] remove open
- [x] remove not open

float.enable and not float.quit_on_focus_loss
- [x] remove only tree open
- [x] remove open
- [x] remove not open

Please:
- [ ] add a comment

lua/nvim-tree/actions/fs/remove-file.lua
geril2207: Initial problem was in case where list of wins is 2 (float tree, and opened buffer), so when trying to delete opened buffer window, vim throws an error cannot close last window. Is there a case where list can be 1? If it possible we save previous condition and add `if not 2` which prevents from closing the last window.

geril2207: Additionally we could check if the window count is 2, and there is no floating tree, in which case we could close the window. But even with `quit_on_focus_loss`, `list_wins` length would be 2 including floating tree.

alex-courtis: Safety first, thank you.

alex-courtis (approved):

PR #2506, opened 2023-11-01
feat: allow cycling on git/diagnostic/opened files navigation

Akmadan23:
It was really annoying to me not to have this capability

### DIFF
--- a/lua/nvim-tree/actions/moves/item.lua
+++ b/lua/nvim-tree/actions/moves/item.lua
@@ -37,7 +37,7 @@ function M.fn(where, what)
 
       if node == node_cur then
         cur = line
-      elseif valid and cur and nex == nil then
+      elseif valid and cur then
         nex = line
         break
       end
--- a/lua/nvim-tree/actions/moves/item.lua
+++ b/lua/nvim-tree/actions/moves/item.lua
@@ -37,7 +37,7 @@ function M.fn(where, what)
 
       if node == node_cur then
         cur = line
-      elseif valid and cur and not nex then
+      elseif valid and cur and nex == nil then
         nex = line
         break
       end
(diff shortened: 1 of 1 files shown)


Akmadan23:
I don't understand why it complains about an access to uninitialized variable when I'm doing a nil check...

alex-courtis (changes requested):
Works beautifully. Tested OK:
* 3 navigation cases
* `root_folder_label = false`

Can I please be a pain in the arse and ask for this to be optional?
1. Existing users will complain about a change in behaviour
2. `:nowrapscan` people like me can't cope with cycling...

Maybe something like:
```help
node.navigate.git.next(opts)          *nvim-tree-api.node.navigate.git.next()*
    Navigate to the next item showing git status.

    Parameters: ~

    Options: ~
      • {wrap} (boolean) wrap around the bottom of the tree, default false

node.navigate.git.prev(opts)          *nvim-tree-api.node.navigate.git.prev()*
    Navigate to the previous item showing git status.

    Parameters: ~

    Options: ~
      • {wrap} (boolean) wrap around the top of the tree, default false
```

alex-courtis:
We probably should have done something like `node.navigate.git(opts)` and specify direction and wrap however probably not worth it...

@Akmadan23 many, many thanks for all the fantastic reviews recently - quality is increasing!

Akmadan23:
> Can I please be a pain in the arse and ask for this to be optional?
> 
>   1. Existing users will complain about a change in behaviour
>   2. `:nowrapscan` people like me can't cope with cycling...

Sure, I initially considered to add an option for it. Also, I was not aware of the existence of `nowrapscan`, maybe instead of adding another option to nvim-tree we can use exactly that, wouldn't it be better?

alex-courtis (approved):
> we can use exactly that, wouldn't it be better?

Brilliant. Intuitive and doesn't need any explanation. I would never have thought of that...

As per my mail I'll leave it for you to merge :)

Issue #2507, opened 2023-11-01
Icon in message after rename file


Dongyomi:
Hi.
Thanks for your plugin.

This is not important but it is for no nvim-web-devicons.

I suggest changing rename-file.lua.

before 

![스크린샷 2023-11-02 08-25-04](https://github.com/nvim-tree/nvim-tree.lua/assets/101002432/c06988fb-eb62-4c8e-8689-584a7364d4a1)

after

![스크린샷 2023-11-02 08-25-15](https://github.com/nvim-tree/nvim-tree.lua/assets/101002432/de81feff-2b0c-4796-8c42-6f9c43ff0d25)

Akmadan23:
Seems reasonable, feel free to submit a PR

PR #2508, opened 2023-11-03
fix(#2507): icon in message after rename-file

Dongyomi:


lua/nvim-tree/actions/fs/rename-file.lua
```diff
@@ -34,7 +34,7 @@ function M.rename(node, to)
   if not success then
     return notify.warn(err_fmt(notify_from, notify_to, err))
   end
-  notify.info(notify_from .. "  " .. notify_to)
```
gegoune, line 37: Please use `string.format()` instead of string concatenation.

lua/nvim-tree/actions/fs/rename-file.lua
```diff
@@ -53,7 +53,9 @@ function M.fn(default_modifier)
 
     -- support for only specific modifiers have been implemented
     if not ALLOWED_MODIFIERS[modifier] then
-      return notify.warn("Modifier " .. vim.inspect(modifier) .. " is not in allowed list : " .. table.concat(ALLOWED_MODIFIERS, ","))
```
gegoune, line 56: Same here.

gegoune (changes requested):

Issue #2512, opened 2023-11-06
Cannot create file in empty directory without root label


geril2207:
### Description

Cannot create file using `a` map in empty folder with `renderer.root_folder_label = false`

### Neovim version

```text
NVIM v0.10.0-dev-324fad1
Build type: Release
LuaJIT 2.1.1699180677
```

### Operating system and version

Fedora 38

### nvim-tree version

20a0707

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvt-min/site]])
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			"nvim-tree/nvim-tree.lua",
			"nvim-tree/nvim-web-devicons",
			-- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
		},
		config = {
			package_root = package_root,
			compile_path = install_path .. "/plugin/packer_compiled.lua",
			display = { non_interactive = true },
		},
	})
end
if vim.fn.isdirectory(install_path) == 0 then
	print("Installing nvim-tree and dependencies.")
	vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
vim.cmd([[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]])
vim.opt.termguicolors = true
vim.opt.cursorline = true

vim.g.mapleader = " "

vim.keymap.set("n", "<leader>e", ":NvimTreeFocus<CR>", { noremap = true, silent = true })

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
	require("nvim-tree").setup({
		renderer = {
			root_folder_label = false,
		},
	})
end

-- UNCOMMENT this block for diagnostics issues, substituting pattern and cmd as appropriate.
-- Requires diagnostics.enable = true in setup.
--[[
vim.api.nvim_create_autocmd("FileType", {
  pattern = "lua",
  callback = function()
    vim.lsp.start { cmd = { "lua-language-server" } }
  end,
})
[...]

PR #2514, opened 2023-11-06
fix: file creation in empty folder without root_folder_label

geril2207:
Fixes: #2512

### DIFF
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -40,14 +40,23 @@ local function wrap(f)
   end
 end
 
---- Inject the node as the first argument if absent.
---- f function to invoke
-local function wrap_node(f)
+---Inject the node as the first argument if absent.
+---@param fn function function to invoke
+local function wrap_node(fn)
   return function(node, ...)
     node = node or require("nvim-tree.lib").get_node_at_cursor()
     if node then
-      f(node, ...)
+      fn(node, ...)
     end
+  end
+end
+
+---Inject the node or nil as the first argument if absent.
+---@param fn function function to invoke
+local function wrap_node_or_nil(fn)
+  return function(node, ...)
+    node = node or require("nvim-tree.lib").get_node_at_cursor()
+    fn(node, ...)
   end
 end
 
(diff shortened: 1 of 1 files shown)


alex-courtis (changes requested):
Thank you for your contribution:

Tested OK except there is an issue:

`root_folder_label =  ":~:s?$?/..?"`
- [ ] add multiple files 
- [ ] add multiple folders
- [ ] add mix

`root_folder_label = false`
- [ ] add multiple files 
- [ ] add multiple folders
- [ ] add mix

issue:
<CR> on empty root node navigates up a directory

lua/nvim-tree/lib.lua
alex-courtis, line 1: I'm not sure we can make this fix here: `get_node_at_cursor` is used in a great many places and there are assumptions that it will return nil if the tree is empty and root folder hidden.

Could we make this change specifically in the add function?

geril2207:
@alex-courtis Could you please describe `on empty root node navigates up a directory`. I guess i have no trouble with it.

lua/nvim-tree/lib.lua
geril2207: The main problem in this function which used for all api functions:
```
local function wrap_node(f)
  return function(node, ...)
    node = node or require("nvim-tree.lib").get_node_at_cursor()
    if node then
      f(node, ...)
    end
  end
end
```
`get_node_at_cursor` returns `nil` and stops execution. in `add` function handle already exists
```
function M.fn(node)
  node = node and lib.get_last_group_node(node)
  if not node or node.name == ".." then
    node = {
      absolute_path = core.get_cwd(),
      nodes = core.get_explorer().nodes,
      open = true,
    }
  end
``` 
There are few options we can do:
1. Make another wrapper without `if node then `;
2. Add second param to `wrap_node` which would be true by default, kind of `skip_if_nil`, we can add to ther api calls if need to skip or not and it would be look like this:
```
local function wrap_node(f, skip_if_nil_node)
  return function(node, ...)
    node = node or require("nvim-tree.lib").get_node_at_cursor()
    if skip_if_nil_node or node then
      f(node, ...)
    end
  end
end
```

geril2207:
@alex-courtis So what do you think about current solution?

lua/nvim-tree/api.lua
```diff
@@ -41,12 +41,13 @@ local function wrap(f)
 end
 
 --- Inject the node as the first argument if absent.
---- f function to invoke
-local function wrap_node(f)
+---@param fn function function to invoke
```
alex-courtis, line 44: Thank you.

alex-courtis (changes requested):
Pragmatic and safe.

API is a bit of a mess and adding boolean arguments isn't the best for readability.

Could you please add a specific method, maybe `wrap_node_or_nil`

alex-courtis (approved):
Fantastic work, thank you!

PR #2521, opened 2023-11-07
feat(#2515): add option to change grouped folders name with custom function

### LINKED ISSUE #2515: group_empty with custom label
`group_empty` option allows for joining simple paths into one, resulting in a long path. The same can happen with `root_folder_label`, but there we can use a custom function that suits our needs and provides usually a shorter label. Such option would be very useful for grouped paths (for me at least).

the simplest would be to allow `group_empty` option to be a function. Or maybe add a separate option in the likes of `grouped_folder_label` just for the function.

I tried looking into the code if there was an easy way to replace something, but I didn't get beyond [some form of recursion](https://github.com/nvim-tree/nvim-tree.lua/blob/c763861afb32f839555f9e797f4392f54ef4ad23/lua/nvim-tree/explorer/explore.lua#L73C22-L73C22). I didn't find a place where the names of the directories would get joined and rendered as a single point to inject my function.

**Additional context**
I shorten all paths to just their first letters of each word to fit into my little NvimTree window:
```lua
function(path)
	path = path:gsub(os.getenv 'HOME', '~', 1)
	return path:gsub('([a-zA-Z])[a-z]+', '%1') .. path:gsub('.*[^a-zA-Z].?', '', 1)
end
```

Akmadan23:
Closes #2515


doc/nvim-tree-lua.txt
```diff
@@ -795,7 +795,10 @@ Appends a trailing slash to folder names.
 
 *nvim-tree.renderer.group_empty*
 Compact folders that only contain a single folder into one node.
-  Type: `boolean`, Default: `false`
+Can be a custom function that takes one argument (the relative path
```
gegoune, line 798:
```suggestion
Boolean or a function that takes single argument (the relative path
```

doc/nvim-tree-lua.txt
```diff
@@ -795,7 +795,10 @@ Appends a trailing slash to folder names.
...
+of the grouped folders) and returns a string (the newly elaborated name
```
gegoune, line 799:
```suggestion
of grouped folders) and returns a string to be displayed.
```

Not 100% sure about grammar here.

doc/nvim-tree-lua.txt
```diff
@@ -795,7 +795,10 @@ Appends a trailing slash to folder names.
...
+that will be shown in the tree)
```
gegoune, line 800:
```suggestion

```

gegoune (changes requested):
Nit pick on documentation. Let's keep it impersonal and as short as possible while conveying necessary meaning.

Thanks for that PR! Interested to see what people will come up with.

alex-courtis (approved):
- [x] false
- [x] "invalid"
- [x] function returning number
- [x] function returning nil
- [x] function returning hardcoded string
- [x] function returning decorated

doc/nvim-tree-lua.txt
Akmadan23: I agree with all the corrections, I would just leave "one" instead of "single" since it's more consistent with two other occurrences of something like "the function takes _one_ argument" throughout the docs

gegoune: Absolutely, consistency is the key. Thanks!

PR #2522, opened 2023-11-09
fix: live filter delete buffer when removing overlay

geril2207:
Problem: When leaving live filter overlay we delete window, but not buffer, you can debug this and found that these buffers living all time untill you close neovim.
Solution: Add buffer deletion.
Also small improvement, `vim.api.nvim_win_close` as second argument takes `force` as `boolean` not `table`(but maybe this also working) https://neovim.io/doc/user/api.html#nvim_win_close()

### DIFF
--- a/lua/nvim-tree/live-filter.lua
+++ b/lua/nvim-tree/live-filter.lua
@@ -37,7 +37,8 @@ local function remove_overlay()
     })
   end
 
-  vim.api.nvim_win_close(overlay_winnr, { force = true })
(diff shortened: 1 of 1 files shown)


alex-courtis:
> second argument takes `force` as `boolean` not `table`(but maybe this also working)

Look accidental to me... table -> boolean?

alex-courtis (approved):
Fantastic work - all those scratch buffers are remove.

Thank you for finding and fixing.

PR #2524, opened 2023-11-09
fix: live filter overlay width calculation

### LINKED ISSUE #2523: Live filter window exceeds window limit with `relativenumber`
### Description

With `view.relativenumber` live filter window exceeds window width limit.

image
image

### Neovim version

```text
NVIM v0.10.0-dev-43b0e27
Build type: Release
LuaJIT 2.1.1699392533
```

Fedora

### Windows variant

_No response_

### nvim-tree version

latest master

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

vim.g.mapleader = " "

vim.keymap.set("n", "<leader>e", ":NvimTreeFocus<CR>", {noremap = true, silent = true})

_G.setup = function()
  require("nvim-tree").setup {
    view = {
      relativenumber = true
    }
  }
end

-- UNCOMMENT this block for diagnostics issues, substituting pattern and cmd as appropriate.
-- Requires diagnostics.enable = true in setup.
--[[
vim.api.nvim_create_autocmd("FileType", {
  pattern = "lua",
  callback = function()
    vim.lsp.start { cmd = { "lua-language-server" } }
  end,
})
]]
```

### Steps to reproduce

1. open nvim with clean room
2. open nvim-tree(:NvimTreeFocus or `<space>e`)
3. press `f`
4. Check filter window width

### Expected behavior

_No response_

### Actual behavior

_No response_

geril2207:
Closes #2523

### DIFF
--- a/lua/nvim-tree/live-filter.lua
+++ b/lua/nvim-tree/live-filter.lua
@@ -104,8 +104,23 @@ local function configure_buffer_overlay()
   vim.api.nvim_buf_set_keymap(overlay_bufnr, "i", "<CR>", "<cmd>stopinsert<CR>", {})
 end
 
-local function create_overlay()
+local function calculate_overlay_win_width()
+  local wininfo = vim.fn.getwininfo(view.get_winnr())[1]
+
+  if wininfo then
+    return wininfo.width - wininfo.textoff - #M.prefix
+  end
+
   local min_width = 20
+
+  if view.View.float.enable then
+    return min_width - 2
+  end
+
+  return min_width
+end
+
+local function create_overlay()
   if view.View.float.enable then
     -- don't close nvim-tree float when focus is changed to filter window
     vim.api.nvim_clear_autocmds {
@@ -113,8 +128,6 @@ local function create_overlay()
       pattern = "NvimTree_*",
       group = vim.api.nvim_create_augroup("NvimTree", { clear = false }),
     }
-
-    min_width = min_width - 2
   end
 
   configure_buffer_overlay()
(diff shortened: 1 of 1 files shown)


lua/nvim-tree/live-filter.lua
```diff
@@ -104,25 +104,38 @@ local function configure_buffer_overlay()
   vim.api.nvim_buf_set_keymap(overlay_bufnr, "i", "<CR>", "<cmd>stopinsert<CR>", {})
 end
 
-local function create_overlay()
+local function calculate_overlay_win_width()
+  local wininfo = vim.fn.getwininfo(view.get_winnr())[1]
```
alex-courtis, line 108: `getwininfo` is hardened against nil winid.

I'm not sure how we'd get a nil however thank you for checking.

Simly returning 20 would be enough; subtracting 2 doesn't really do much.

alex-courtis (changes requested):
Many thanks for all the fantastic contributions you've made recently - quality is greatly improving!

OK:
- [ ] defaults
- [ ] view.relativenumber = true
- [ ] view.signcolumn = "no"
- [ ] view.float.enable = true

lua/nvim-tree/live-filter.lua
geril2207: By doc it could be `nil`, even if  it happens we can secure by simple return 20. I don't think it reduces the readability of the code.

alex-courtis: > By doc it could be nil, maybe 1 time in 1 million and if it happens we can secure by simply return 20. I don't think it reduces the readability of the code.

Yup, that's the safest and most resilient solution.

alex-courtis (approved):
Many thanks for your contrubution.

Issue #1389, opened 2022-07-02
Next / Prev Git File Item


alex-courtis:
**Is your feature request related to a problem? Please describe.**
`next_git_item` will take you to the directory containing the git item. It would be useful to navigate directly to the file.

In this case, you need to `next_git_item` 5 times to get to the file.

![20220703_094542 409x466](https://user-images.githubusercontent.com/1505378/177019228-2410885b-335c-4e8f-afb4-cfb6218b3547.png)

**Describe the solution you'd like**
`next_git_file_item` to navigate directly to the file.

**Describe alternatives you've considered**
Directories could also be recursively opened until the file item is found.

**Additional context**
I find myself using `next_git_item` `edit` `next_git_item` `edit` `next_git_item` ... very frequently.

alex-courtis:
An alternative might be simply a `show_on_dirs` option for git, as per diagnostics.

Both would be nice.

kyazdani42:
Maybe just skipping if the file is a directory in the next/prev action would work too.

kyazdani42:
@alex-courtis is this working now ?

alex-courtis:
Closing. `show_on_dirs` is a good enough solution.

I'd still like some means of finding the git items down the tree, however I don't have a user story for that.

PR #2525, opened 2023-11-09
feat(#1389): add recurse opt for git and diag moves

antznin:
Would like having comments on this feature that I personally find useful. This solves the initial problem stated in #1389, I think. Also works when `nvim-tree.git.show_on_dirs` or `nvim-tree.git.show_on_open_dirs` are set to `true`/`false`.


lua/nvim-tree/actions/moves/git.lua
```diff
@@ -0,0 +1,62 @@
+local explorer_node = require "nvim-tree.explorer.node"
+local lib = require "nvim-tree.lib"
+local moves_item = require "nvim-tree.actions.moves.item"
+
+local M = {}
+local MAX_DEPTH = 100
+
+function M.next_item_recursive()
```
alex-courtis, line 8: This could be generic - we could move this into `item.lua` to apply to all the 3 similar moves: git, diagnostics and opened.

lua/nvim-tree/actions/moves/git.lua
```diff
@@ -0,0 +1,62 @@
+local explorer_node = require "nvim-tree.explorer.node"
+local lib = require "nvim-tree.lib"
+local moves_item = require "nvim-tree.actions.moves.item"
+
+local M = {}
+local MAX_DEPTH = 100
```
alex-courtis, line 6: Sensible, thank you.

doc/nvim-tree-lua.txt
```diff
@@ -2118,6 +2119,7 @@ You are encouraged to copy these to your own |nvim-tree.on_attach| function.
   vim.keymap.set('n', 'C',       api.tree.toggle_git_clean_filter,    opts('Toggle Filter: Git Clean'))
   vim.keymap.set('n', '[c',      api.node.navigate.git.prev,          opts('Prev Git'))
   vim.keymap.set('n', ']c',      api.node.navigate.git.next,          opts('Next Git'))
+  vim.keymap.set('n', ']g',      api.node.navigate.git.next_recurse,  opts('Next Git Recursive'))
```
alex-courtis, line 2122: We can reuse next/prev with an opt. Something like:

```help
node.navigate.git.next({opts})        *nvim-tree-api.node.navigate.git.next()*
    Navigate to the next item showing git status.

    Parameters: ~

    Options: ~
      • {recurse}     (boolean)  recurse directories containing items showing status, default false
```

alex-courtis (changes requested):
This is great! I've been wanting this for a long time...

Works beautifully:
- [ ] one deep
- [x] many deep
- [x] deltas at multiple levels
- [x] `group_empty = true`
- [x] `show_on_dirs = true` `show_on_open_dirs = false`

Except for:
- [ ] `show_on_dirs = false` - does not recurse

lua/nvim-tree/actions/moves/git.lua
antznin: I agree. I'll look into making this more generic.

lua/nvim-tree/actions/moves/git.lua
antznin: This is a limit I chose arbitrarily. If you see another value that would make more sense, lmk.

doc/nvim-tree-lua.txt
antznin: This`recurse` opt would also need to be added for diagnostics and opened then - unless a global `recurse` opt would make more sense? And apply to all three categories.

antznin:
> Except for:
> 
>     * [ ]  `show_on_dirs = false` - does not recurse

This cannot work with `show_on_dirs` set to `false` because `move_to_next_git_item()` does not see the inner nodes statuses. Is this behavior OK? I don't see another approach that would make this work with `show_on_dirs = false`.

lua/nvim-tree/actions/moves/git.lua
alex-courtis: That's fine. We can add an option like `actions.expand_all.max_folder_discovery` only if we need it.

doc/nvim-tree-lua.txt
alex-courtis: There are too many options;  we're preferring API to options wherever possible.

alex-courtis:
> > Except for:
> > ```
> > * [ ]  `show_on_dirs = false` - does not recurse
> > ```
> 
> @alex-courtis This cannot work with `show_on_dirs` set to `false` because `move_to_next_git_item()` does not see the inner nodes statuses. Is this behavior OK? I don't see another approach that would make this work with `show_on_dirs = false`.

I see. It looks like we have a short-circuit to not retrieve git status on folders: https://github.com/nvim-tree/nvim-tree.lua/blob/c1466f991a3c4cbf77f44282e3baddd4b7432f8c/lua/nvim-tree/explorer/node.lua#L64

That's OK; we can remove that one: git status are now very cheap and will likely be fetched at some point.

doc/nvim-tree-lua.txt
alex-courtis: I'd be most grateful if you could add a type for all navigations:

```lua
---@class ApiNodeNavigateOpts
---@field recurse boolean|nil default false
```

lua/nvim-tree/keymap.lua
```diff
@@ -57,6 +57,7 @@ function M.default_on_attach(bufnr)
   vim.keymap.set('n', 'C',       api.tree.toggle_git_clean_filter,    opts('Toggle Filter: Git Clean'))
   vim.keymap.set('n', '[c',      api.node.navigate.git.prev,          opts('Prev Git'))
   vim.keymap.set('n', ']c',      api.node.navigate.git.next,          opts('Next Git'))
+  vim.keymap.set('n', ']g',      api.node.navigate.git.next_recurse,  opts('Next Git Recursive'))
```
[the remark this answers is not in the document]
gegoune: There literally is no risk, those are buffer local mappings.

antznin:
My previous commit defines a new `recurse` api opt, and moves the logic to `actions/moves/item.lua`. However there were some limitations, and perhaps bugs I faced. Let me summarize here below:

- The `opened` case is different because there is no `show_on_dirs`/`show_on_open_dirs` for this type of move. The way it is currently implemented, it does not propagate that information to the parent. The only way I would see `opened` being recursive at the moment would be for the parent nodes (directories) to hold that opened information on their children.

- The `show_on_dirs`/`show_on_open_dirs` options are acting strange for diagnostics. If `show_on_dirs = true` and `show_on_open_dirs = false`, the diagnostic mark is not propagated to the parent directories when they are **closed**. When both options are set to `true` the parent directory *does* show the diagnostic mark and the `recurse` works the same as for git. Git does not have that issue: the status is correctly displayed on the closed directory when `show_on_dirs = true` and `show_on_open_dirs = false`.

- Taking my previous comment in consideration, diagnostics and git recurse moves do work well together though when both are enabled :thumbsup:.

I was able to map these next moves with the recurse opt like so:

```lua
  vim.keymap.set('n', ']e', function() api.node.navigate.diagnostics.next({recurse=true}) end, opts('Next Diagnostic'))
  vim.keymap.set('n', ']c', function() api.node.navigate.git.next({recurse=true}) end, opts('Next Git'))
```

It this the way you suggested it to be @alex-courtis?

lua/nvim-tree/actions/moves/item.lua
```diff
@@ -5,48 +5,135 @@ local lib = require "nvim-tree.lib"
...
+      recurse = opts.recurse
+    end
+
+    if recurse == false then
+      move(where, what)
+      return
+    end
+
+    -- recurse logic below.
+    -- local move_to_next_git_item = moves_item.fn("next", "git")
```
antznin, line 87: Oops junk there, will remove.

lua/nvim-tree/api.lua
```diff
@@ -207,9 +207,14 @@ Api.node.navigate.sibling.first = wrap_node(require("nvim-tree.actions.moves.sib
...
 Api.node.navigate.parent_close = wrap_node(require("nvim-tree.actions.moves.parent").fn(true))
+
+---@class ApiNodeNavigateOpts
+---@field recurse boolean|nil default false
+
 Api.node.navigate.git.next = wrap_node(require("nvim-tree.actions.moves.item").fn("next", "git"))
-Api.node.navigate.git.prev = wrap_node(require("nvim-tree.actions.moves.item").fn("prev", "git"))
 Api.node.navigate.diagnostics.next = wrap_node(require("nvim-tree.actions.moves.item").fn("next", "diag"))
+
+Api.node.navigate.git.prev = wrap_node(require("nvim-tree.actions.moves.item").fn("prev", "git"))
```
alex-courtis, line 217: Ouch... this doesn't quite work as we expect:  when `opts` is not passed the node will be injected.

We've hit the same roadblock as #2583

The solution above was to add specific new API: https://github.com/nvim-tree/nvim-tree.lua/pull/2583/files#diff-b5b2d1abcb86ead5a2319f7e065a7545b834a407cc9f420590d7acebb285fca9R1865

We'll just have to cope with the extra API. Apologies for steering you on the wrong path.

alex-courtis:
> The show_on_dirs/show_on_open_dirs options are acting strange for diagnostics. If show_on_dirs = true and show_on_open_dirs = false, the diagnostic mark is not propagated to the parent directories when they are closed. When both options are set to true the parent directory does show the diagnostic mark and the recurse works the same as for git. Git does not have that issue: the status is correctly displayed on the closed directory when show_on_dirs = true and show_on_open_dirs = false.

I'm completely fine with that: there are many factors at play with diagnostics, which are usually quite lazy. Git is always known. We could try and tweak if it becomes a problem.

alex-courtis:
> * The `opened` case is different because there is no `show_on_dirs`/`show_on_open_dirs` for this type of move. The way it is currently implemented, it does not propagate that information to the parent. The only way I would see `opened` being recursive at the moment would be for the parent nodes (directories) to hold that opened information on their children.

Many thanks for looking into that one; YAGNI

alex-courtis (changes requested):
Looking good; all the pieces are there and work, just needs API tidy.

Many thanks for all your work on this one and apologies for lacking the time to review.

lua/nvim-tree/api.lua
antznin: Alright, NP. I'll try to review my changes, rebase and update the PR in the coming days/weeks depending on the time I have.

alex-courtis: Thanks Mate

There is no hurry, this is a tricky one.

antznin:
@alex-courtis Should be good now.

lua/nvim-tree/api.lua
```diff
@@ -220,7 +220,11 @@ Api.node.navigate.git.prev = wrap_node(require("nvim-tree.actions.moves.item").f
 Api.node.navigate.git.next_skip_gitignored = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "next", what = "git", skip_gitignored = true })
 -- stylua: ignore
 Api.node.navigate.git.prev_skip_gitignored = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "prev", what = "git", skip_gitignored = true })
+-- stylua: ignore
+Api.node.navigate.git.next_recursive = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "next", what = "git", recurse = true })
 Api.node.navigate.diagnostics.next = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "next", what = "diag" })
+-- stylua: ignore
```
alex-courtis, line 226: This will Go Away with #2593

lua/nvim-tree/actions/moves/item.lua
```diff
@@ -14,46 +85,73 @@ local M = {}
...
+    -- recurse only available for "next" git and diag moves.
+    if opts.where == "next" and (opts.what == "git" or opts.what == "diag") and opts.recurse ~= nil then
+      recurse = opts.recurse
+    end
+
+    if opts.skip_gitignored ~= nil then
+      skip_gitignored = opts.skip_gitignored
+    end
+
+    if recurse == false then
```
alex-courtis, line 100: / nit `not recurse`

alex-courtis (changes requested):
This is looking great - change is well contained with a very small blast radius.

Same test results as https://github.com/nvim-tree/nvim-tree.lua/pull/2525#pullrequestreview-1726162597

Is it intended for this functionality to work when `show_on_dirs = false`? Did you look at always retrieving git status as per https://github.com/nvim-tree/nvim-tree.lua/pull/2525#issuecomment-1807478295 ?

If we don't intend it to work we could:
* document the limitation
* document the fallback behaviour

- [ ] `show_on_dirs = false`
- [ ] missing prevs
- [ ] minor nits

lua/nvim-tree/api.lua
```diff
@@ -220,7 +220,11 @@ Api.node.navigate.git.prev = wrap_node(require("nvim-tree.actions.moves.item").f
...
+Api.node.navigate.diagnostics.next_recursive = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "next", what = "diag", recurse = true })
```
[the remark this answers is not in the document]
antznin: I didn't implement a prev recursive move. A recursive move is always going down a directory, and thus moving forward which is next here.

I would sort of see a "prev_recursive" move as:

1. Go to the previous node with a non-recursive move.
2. Go to the next node with a recursive move.

But it's kind of strange IMO.

antznin:
> Is it intended for this functionality to work when `show_on_dirs = false`? Did you look at always retrieving git status as per [#2525 (comment)](https://github.com/nvim-tree/nvim-tree.lua/pull/2525#issuecomment-1807478295) ?

@alex-courtis No, it will not work when `show_on_dirs = false` atm because the recursive mechanism relies on the original move mechanism: if no status is shown anywhere in the tree, moving is not possible even if one of the directory contains a  file with a status.

Retrieving the git status for a directory when `show_on_dirs = false` is possible though, and would remove this limitation. I could work on this later but right now I will I will document the limitation in this PR instead.

alex-courtis:
> Retrieving the git status for a directory when `show_on_dirs = false` is possible though, and would remove this limitation. I could work on this later but right now I will I will document the limitation in this PR instead.

No worries. I will hold you to that ;)

lua/nvim-tree/api.lua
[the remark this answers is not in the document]
alex-courtis: How about something like this:



recursive or regular prev from `builder.lua` will move to `bulk-move.lua`

marks is closed



recursive prev from `builder.lua` will still move to `bulk-move.lua`

lua/nvim-tree/api.lua
antznin: Ok, will update once merged.

doc/nvim-tree-lua.txt
```diff
@@ -1889,6 +1894,12 @@ node.navigate.git.prev_skip_gitignored()
 node.navigate.diagnostics.next()
     Navigate to the next item showing diagnostic status.
 
+                              *nvim-tree-api.node.navigate.diagnostics.next_recursive()*
+node.navigate.diagnostics.next_recursive()
+    Same as |node.navigate.diagnostics.next()|, but navigates to next item
+    showing diagnostic status recursively. Needs `diagnostics.show_on_dirs`
+    and `diagnostics.show_on_open_dirs` set to `true`.
```
alex-courtis, line 1901: Thank you - that is very clear.

alex-courtis (changes requested):
This is great, it's working exactly as advertised.

However, we should still implement the prev use cases. They will inevitably be requested as soon as users discover this new functionality, which will be surprisingly quick.

I'm not seeing any reason why we can't do that with `move_next_recursive` - it's quite comprehensive.

Also... merge conflicts, sorry.

lua/nvim-tree/api.lua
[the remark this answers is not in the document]
antznin: We're on the same page, I'll work on that!

antznin:
Hi @alex-courtis, here are 4 commits implementing next and prev recursive moves as discussed.

doc/nvim-tree-lua.txt
antznin: I'm wondering now: my latest tests showed success for diagnostics when `diagnostics.show_on_dirs = true` and `diagnostics.show_on_open_dirs = false`… what about you?

alex-courtis: Yes... that's the most useful use case IMO

doc/nvim-tree-lua.txt
```diff
@@ -1875,13 +1875,23 @@ node.open.preview_no_picker()    *nvim-tree-api.node.open.preview_no_picker()*
 node.navigate.git.next()              *nvim-tree-api.node.navigate.git.next()*
     Navigate to the next item showing git status.
 
+                      *nvim-tree-api.node.navigate.git.next_recursive()*
+node.navigate.git.next_recursive()
+    Same as |node.navigate.git.next()|, but navigates to next item showing git
+    status recursively. Needs `git.show_on_dirs` set to `true`.
```
alex-courtis, line 1881: OK when `show_on_open_dirs = false`

lua/nvim-tree/actions/moves/item.lua
```diff
@@ -6,6 +6,223 @@ local explorer_node = require "nvim-tree.explorer.node"
...
+
+    i = i + 1
+  end
+end
+
+-- Find the line number of a node.
+-- Return -1 is node is nil or not found.
+---@param node Node|nil
+---@return integer
+local function find_node_line(node)
```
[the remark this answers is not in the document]
alex-courtis: It's nice and generic. In utils.lua please.

alex-courtis (changes requested):
Everything works and is documented, great job!

- [ ] move to utils.lua
- [ ] remove limitation from diagnostics help

alex-courtis (approved):
Working nicely, many thanks for all your work.

Changing my git bindings now  :)

PR #2527, opened 2023-11-11
fix: reload tree on `BufEnter` if cwd is different

### LINKED ISSUE #2488: Retain Tree State on DirChanged
### Description

So here is an interesting problem, i have. I would usually open a new tree in a separate tab to explore or mutate a given working directory, say in the first vim tab i open a file and a normal side view of nvim tree for the same working directory of the file just opened, on the second tab i would open nvim tree only in the home directory (for example :e ~/). Now in the second tab i will expand the tree view in some folder from the home. I navigate back to the first tab, my configuration would trigger a DirChanged event, and the tree in the first tab would show correctly the view it did initially. However going back to the tab displaying the home folder, DirChanged will trigger again, and will be for the home directory, however the tree would auto-colapse, losing its expanded state it had before i left that home-dir tab, which is quite annoying to be frank, the directory for that tree technically did not change it is still /home, but i do know that nvim tree somehow reuses the same buffer and triggers redraws. Looking at the code i could see that if i disable `opts.sync_root_with_cwd`, can help, but breaks other things. This function `M.change_dir(vim.loop.cwd())` in its implementation redraws the tree. So i assume it does not store the state, or knows that. 

### Neovim version

```text
NVIM 0.9.4
```

Fedora 38

### Windows variant

_No response_

### nvim-tree version

master

### Clean room replication

```lua
on_attach = default_on_attach,
        auto_reload_on_write = false,
        sync_root_with_cwd = true,
        respect_buf_cwd = true,
        hijack_cursor = true,
        disable_netrw = true,
        hijack_netrw = true,
        hijack_unnamed_buffer_when_opening = true,
        hijack_directories = {
            enable = true,
            auto_open = true,
        },
        renderer = {
            full_name = true,
            indent_width = 1,
            root_folder_label = function(path)
                return vim.fn.fnamemodify(path, ":t")
            end,
            highlight_opened_files = "name",
        },
        modified = {
            enable = false,
            show_on_dirs = false,
            show_on_open_dirs = false,
        },
        diagnostics = {
            enable = false,
            show_on_dirs = false,
            show_on_open_dirs = false,
        },
        git = {
            enable = false,
            ignore = true,
        },
        actions = {
            change_dir = {
                enable = false,
                global = false,
            },
            open_file = {
                quit_on_open = true,
                resize_window = true,
            },
        },
        update_focused_file = {
            enable = false,
            update_root = false,
        },
```

### Steps to reproduce

1. open file in a random working directory
2. open the tree for that directory using the api.tree.open
3. open new tab in another directory, open the tree there
4. start expanding, collapsing, navigating, mutating the tree view
5. navigate to the first tab, tcd path-to-work-dir
6. navigate back to the second, tcd path-to-second-dir

### Expected behavior

The tree state should be remembered, when going back and forth between directories.

### Actual behavior

The tree state is not remembered, rather going back to the same tree, the entire tree only shows the root folders of the directory for which it was opened.

Akmadan23:
Not sure if this entirely fixes #2488, but at least the tree gets reloaded when cwd has been changed by another instance of nvim-tree.

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -283,8 +283,8 @@ local function setup_autocommands(opts)
   create_nvim_tree_autocmd("BufEnter", {
     pattern = "NvimTree_*",
     callback = function()
-      if opts.reload_on_bufenter and not opts.filesystem_watchers.enable then
-        if utils.is_nvim_tree_buf(0) then
+      if utils.is_nvim_tree_buf(0) then
+        if vim.fn.getcwd() ~= core.get_cwd() or (opts.reload_on_bufenter and not opts.filesystem_watchers.enable) then
           reloaders.reload_explorer()
         end
       end


alex-courtis (approved):
Works nicely:
- [x] defaults - changes
- [x] sync_root_with_cwd - changes
- [x] respect_buf_cwd - no change

This may not fix the world but it's a great start.

PR #2534, opened 2023-11-16
fix: new staged files

tomcri:
sync icons with highlights:
["A "] = { icons.staged },
and 
["A "] = "NvimTreeFileStaged",


alex-courtis (approved):
Many thanks for finding and fixing.

Perhaps we might one day add more icons/colours to git staged statuses....

PR #2540, opened 2023-11-19
docs: update quick start example

dvdksn:
update sort configuration key


alex-courtis (approved):
Great catch. Thank you so much!

PR #2541, opened 2023-11-19
feat(log): add node inspection function

Akmadan23:
I often use this little function, exposing it in the api might be useful to someone.

### DIFF
--- a/lua/nvim-tree/log.lua
+++ b/lua/nvim-tree/log.lua
@@ -61,6 +61,26 @@ function M.line(typ, fmt, ...)
   end
 end
 
+local inspect_opts = {}
+
+--- @param opts table
+function M.set_inspect_opts(opts)
+  inspect_opts = opts
+end
+
+--- Write to log file the inspection of a node
+--- defaults to the node under cursor if none is provided
+--- @param typ string as per log.types config
+--- @param node table|nil node to be inspected
+--- @param fmt string for string.format
+--- @vararg any arguments for string.format
+function M.node(typ, node, fmt, ...)
+  if M.enabled(typ) then
+    node = node or require("nvim-tree.lib").get_node_at_cursor()
+    M.raw(typ, string.format("[%s] [%s] %s\n%s\n", os.date "%Y-%m-%d %H:%M:%S", typ, (fmt or "???"), vim.inspect(node, inspect_opts)), ...)
+  end
+end
+
 --- Logging is enabled for typ or all
 --- @param typ string as per log.types config
 --- @return boolean
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,28 +8,35 @@ on:
     branches:
       - master
 
+permissions:
+  contents: read
+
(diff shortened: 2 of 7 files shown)


alex-courtis (changes requested):
I like the idea - it will be most useful to have something that I always end up adding log lines for.

I am a bit hesitant:
1. Inspecting nodes is _massive_ as it goes up and down the tree.
2. I don't want to encourage users to look into nodes. They are private and definitely shouldn't be mutated.

How about:
* Change this to log raw instead of print
* Accept an `options` parameter to pass directly to vim.inspect

alex-courtis:
RE 2. I wish we'd never let the nodes escape into public API. I'd like to provide a sanitised copy of nodes like `sort.sorter` however that's going to be difficult...

Akmadan23:
I understand your concerns

> Change this to log raw instead of print

This can be a good idea, but I don't know if it should go in the API, I'd rather put it in the `log` module...

alex-courtis:
> This can be a good idea, but I don't know if it should go in the API, I'd rather put it in the `log` module...

Ah, that would work nicely. I imagine myself doing something like this in an event or mapping:
```lua
local log = require("nvim-tree.log")
log.node("dev", node, "Event.NodeRenamed %s", data.new_name)
```

I'm happy with your best judgement. This will be useful.

alex-courtis (approved):
pre-emptive approval

lua/nvim-tree/log.lua
```diff
@@ -61,6 +61,26 @@ function M.line(typ, fmt, ...)
   end
 end
 
+local inspect_opts = {}
+
+--- @param opts table
+function M.set_inspect_opts(opts)
```
Akmadan23, line 67: Since the inspect options are rarely changed, I chose to add this little function to set them instead of another parameter to the `node` function.

lua/nvim-tree/log.lua
```diff
@@ -61,6 +61,26 @@ function M.line(typ, fmt, ...)
...
+
+--- Write to log file the inspection of a node
+--- defaults to the node under cursor if none is provided
+--- @param typ string as per log.types config
+--- @param node table|nil node to be inspected
+--- @param fmt string for string.format
+--- @vararg any arguments for string.format
+function M.node(typ, node, fmt, ...)
+  if M.enabled(typ) then
+    node = node or require("nvim-tree.lib").get_node_at_cursor()
```
alex-courtis, line 79: That is very handy...

alex-courtis (approved):
Works beautifully, many thanks.

Do you want me to just merge after approval or do you want to do it?

Akmadan23:
> Do you want me to just merge after approval or do you want to do it?

Oh, it's really no different to me, just merge whenever you feel like the PR is ready

alex-courtis:
> > Do you want me to just merge after approval or do you want to do it?
> 
> Oh, it's really no different to me, just merge whenever you feel like the PR is ready

I am also happy either way.

PR #2543, opened 2023-11-19
ci: ubuntu packages -> luarocks action, pin versions

alex-courtis:
Brings in line with https://github.com/nvim-tree/nvim-web-devicons/pull/341

* versions pinned
* avoids ubuntu packages

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,28 +8,35 @@ on:
     branches:
       - master
 
-jobs:
-  luacheck:
-    name: luacheck
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
+permissions:
+  contents: read
 
-      - name: Prepare
-        run: |
-          sudo apt-get update
-          sudo add-apt-repository universe
-          sudo apt install luarocks -y
-          sudo luarocks install luacheck
-      - name: Run luacheck
-        run: luacheck .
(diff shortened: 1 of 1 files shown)


Akmadan23 (approved):
Awesome

Akmadan23:
Not sure why there are these standalone `luackeck` and `stylua` checks here...

alex-courtis:
> Not sure why there are these standalone `luackeck` and `stylua` checks here...

Do you mean two jobs instead of one? I'm not sure either; it seems to be the pattern people follow.

The only reason I can think of is to prevent cross-contamination between jobs.

Akmadan23:
> Do you mean two jobs instead of one? I'm not sure either; it seems to be the pattern people follow.

I mean that besides the three `CI / *` jobs, there are two named respectively `luacheck` and `stylua` with the "Expected - Waiting for status to be reported" label, which prevent me from merging the PRs (any PR after this has this problem)

gegoune:
@Akmadan23 It should be fixed now. :)

Akmadan23:
> It should be fixed now. :)

Thanks! What was the matter?

gegoune:
Required checks are set in branch protection (settings), it still had old jobs configured as required.

alex-courtis:
> Required checks are set in branch protection (settings), it still had old jobs configured as required.

Many thanks!

PR #2542, opened 2023-11-19
feat(actions): git-aware cut / paste / rename fs actions

hinell:
https://github.com/nvim-tree/nvim-tree.lua/assets/8136158/a46aa792-b6d6-41fb-9af6-679e918ef5c9

> git-aware file renames right in your nvim-tree; refactor your source tree easily
> NOTE: Demo uses rebound default `x,c,p,r,e,u,<c-r>` keymap bindings
> NOTE: Letter `R` signifies git renamed file (setup by `confi.renderer.icons.glyphs.git`

This PR allows:
* Moving/renaming git-tracked files by using `fs.*` like file actions for cut / paste / rename
*  Moving/renaming non-tracked files by `fs.{cut,paste}` as usual
* Wraps around `fs.*` APIs nicely, minimize redundancy
* Utilizes log / info utils
* Closes #2474
### Usage

No keymaps setup by default. Example:
```lua
-- $HOME/.config/nvim/lua/user/nvim-keymaps.lua
-- put this inside autocmd callback for NvimTree filetype, see `h autocmd` and `h nvim_create_autocmd()`
local opts = { buffer = autoCmdEvent.buf, nowait = true }
local api  = require "nvim-tree.api"
vim.keymap.set("n", "x"    , api.git.cut            , vim.tbl_extend("keep", { desc = "nvim-tree: git: cut file into clipboard"   }, opts)
vim.keymap.set("n", "p"    , api.git.paste          , vim.tbl_extend("keep", { desc = "nvim-tree: git: past file from clipboard"  }, opts)
vim.keymap.set("n", "r"    , api.git.rename         , vim.tbl_extend("keep", { desc = "nvim-tree: git: rename: filename"          }, opts)
vim.keymap.set("n", "e"    , api.git.rename_basename, vim.tbl_extend("keep", { desc = "nvim-tree: git: rename: filename base"     }, opts)
vim.keymap.set("n", "u"    , api.git.rename_absolute, vim.tbl_extend("keep", { desc = "nvim-tree: git: rename: filename absolute" }, opts)
vim.keymap.set("n", "<C-r>", api.git.rename_sub     , vim.tbl_extend("keep", { desc = "nvim-tree: git: rename: omit filename"	  }, opts)
```

### API
* Adds:    
`api.git.cut(node)` - store a file under cursor into a git actions clipboard (just like `api.fs.cut`)
`api.git.paste(node)` - pop files from clipboard into a folder under cursor
`api.git.rename(node)` - rename file under cursor
[...]

PR #2550, opened 2023-11-22
refactor(actions): common code for rename-file

hinell:
https://github.com/nvim-tree/nvim-tree.lua/assets/8136158/bbddff00-e27a-435d-8ca9-cdba6ae58280

### 
* required to proceed with https://github.com/nvim-tree/nvim-tree.lua/pull/2542 PR
* Refactor `lua/nvim-tree/actions/fs/rename-file.lua`:
  * feat: `Input_path_editor` - new class
  * Deprecate private `fn()` in case it's used elsewhere
  * Replace `fn()` by `rename_*()` based on `Input_path_editor` new class
  * feat: new `rename_relative`, unbound by default

## `Input_path_editor` new class  
* Encapsulates reusable code from `rename-file.lua` used for editing chunks of absolute path per https://github.com/nvim-tree/nvim-tree.lua/pull/2542#discussion_r1399472803
* Despite looking complex in fact has only two methods that are used to `prepare()` / `restore()` path intended to be modified by user in `vim.ui.input` (see example)
* Reusable, extensible, well documented

### Why class?
* it isolates temp data in instance-namespace, easy to use / extend
* it encapsulates state that can be passed around / referenced from across async callback functions without need to pollute local scope

### Why prototype-based class?
* Prototype is a namespace of a class where instance methods are gathered / localized.


alex-courtis:
See https://github.com/nvim-tree/nvim-tree.lua/pull/2551#issuecomment-1828908354

PR #2546, opened 2023-11-21
ci: lua language server and Makefile

alex-courtis:
Adds lua-language-server check

Add Makefile

Document contributing.

Rework CI.

PR #2555, opened 2023-11-24
chore: add type annotations and resolve LSP warnings

Akmadan23:
Initially discussed in #2455
WIP: There are still some decisions to be made here


lua/nvim-tree.lua
```diff
@@ -17,6 +17,31 @@ local open = require "nvim-tree.actions.tree.open"
...
+---@class DirNode: BaseNode
+---@field has_children boolean
+---@field group_next Node|nil
+---@field nodes Node[]
+---@field open boolean
+
+---@class FileNode: BaseNode
+---@field extension string
+
+---@alias Node ParentNode|DirNode|FileNode
```
Akmadan23, line 43: This is the `Node` type I managed to come up with, let me know if it's ok and most importantly where it should go. I put it here just to put it temporarily somewhere.

lua/nvim-tree/actions/fs/remove-file.lua
```diff
@@ -44,10 +46,13 @@ local function clear_buffer(absolute_path)
...
 end
 
+---@param cwd string
+---@return boolean|nil
 local function remove_dir(cwd)
   local handle = vim.loop.fs_scandir(cwd)
   if type(handle) == "string" then
-    return notify.error(handle)
+    notify.error(handle)
+    return
```
Akmadan23, line 55: In some cases I added an explicit nil return instead of returning the call of a function that returns nil.

lua/nvim-tree/actions/root/dir-up.lua
```diff
@@ -3,13 +3,14 @@ local core = require "nvim-tree.core"
...
+---@param node Node
 function M.fn(node)
   if not node or node.name == ".." then
-    return require("nvim-tree.actions.root.change-dir").fn ".."
+    require("nvim-tree.actions.root.change-dir").fn ".."
   else
     local newdir = vim.fn.fnamemodify(utils.path_remove_trailing(core.get_cwd()), ":h")
     require("nvim-tree.actions.root.change-dir").fn(newdir)
-    return require("nvim-tree.actions.finders.find-file").fn(node.absolute_path)
+    require("nvim-tree.actions.finders.find-file").fn(node.absolute_path)
```
Akmadan23, line 13: Removed pointless and misleading `return`s as well

lua/nvim-tree/git/runner.lua
```diff
@@ -62,19 +73,25 @@ function Runner:_getopts(stdout_handle, stderr_handle)
...
+---@param callback function|nil
 function Runner:_run_git_job(callback)
   local handle, pid
   local stdout = vim.loop.new_pipe(false)
   local stderr = vim.loop.new_pipe(false)
   local timer = vim.loop.new_timer()
 
+  if stdout == nil or stderr == nil or timer == nil then
+    return
+  end
```
Akmadan23, line 93: Added this little check to silence the "Need nil check" warnings.

Akmadan23 (commented):
Then there two general discussion points:

- Style inconsistency: the majority of docs comments is of type `---@` but there are still some `--- @`. I don't know if one should be enforced instead of the other (`---@` should be the official one), unfortunately stylua hasn't an option to check it.
- Setup functions: this is almost irrelevant, I did not add type annotations to the various `M.setup` functions since they are all the same and are intuitive, but I would still add them just for completeness

lua/nvim-tree.lua
```diff
@@ -17,6 +17,31 @@ local open = require "nvim-tree.actions.tree.open"
 local events = require "nvim-tree.events"
 local notify = require "nvim-tree.notify"
 
+---@class ParentNode
+---@field name string
+
+---@class BaseNode
```
alex-courtis, line 23: This is fantastic... finding all these bits every time always slows me down.

alex-courtis:
>     * Style inconsistency: the majority of docs comments is of type `---@` but there are still some `--- @`. I don't know if one should be enforced instead of the other (`---@` should be the official one), unfortunately stylua hasn't an option to check it.

`---@` does seem the convention.

It doesn't look like the LLS formatter has it either: https://github.com/CppCXY/EmmyLuaCodeStyle/blob/master/docs/format_config_EN.md

We could run `find lua -type f -exec sed -i 's/^--- @/---@/g' {} \;` every now and then if people aren't following it.

Alternatively, add `grep -r "^--- @" lua` to the Makefile.

lua/nvim-tree/git/runner.lua
alex-courtis: This is where we really gain the value.

alex-courtis:
> Setup functions: this is almost irrelevant, I did not add type annotations to the various `M.setup` functions since they are all the same and are intuitive, but I would still add them just for completeness

Agreed, minimal value. I'm happy enough to do nothing.

In the future we could do an Options class however that would be a lot of maintenance...

lua/nvim-tree.lua
alex-courtis: > where it should go.

Where is indeed difficult. We could just create `nvim-tree/node.lua` just to hold the definitions.

When we eventually get to creating proper node classes they can live in there.

alex-courtis: > This is the Node type I managed to come up with, let me know if it's ok 

It's great! It contains all that we know about.

LLS can tell us if we're using an unknown field and we can add as we discover them.

alex-courtis (commented):
This is incredible. This will pay off immediately and continually.

alex-courtis:
What do you think about a `SymlinkDirNode` and `SymlinkFileNode`? lua does support multiple inheritence...

Akmadan23:
> What do you think about a `SymlinkDirNode` and `SymlinkFileNode`?

Right, I completely forgot about symlinks.

> Alternatively, add `grep -r "^--- @" lua` to the Makefile.

Or better yet add it to the style check so that it's done automatically.

> We could just create `nvim-tree/node.lua` just to hold the definitions.

Yeah, I'm not sure if LLS cares about files that are not imported tho...
Otherwise we can put it in the `lib` module which is the most generic I guess.

alex-courtis:
> Yeah, I'm not sure if LLS cares about files that are not imported tho...
> Otherwise we can put it in the `lib` module which is the most generic I guess.

Ah yes. That could be messy. `nvim-tree.lua` works for me...

> Or better yet add it to the style check so that it's done automatically.

Yes please! If you can find a simple way that would be fantastic.

> Right, I completely forgot about symlinks.

Only if it's a small amount of effort. We can iterate on this.

alex-courtis:
I had a play with #2546 against this branch... only 16 warnings! Nice work.

[uvlib.patch.gz](https://github.com/nvim-tree/nvim-tree.lua/files/13465542/uvlib.patch.gz) adds libuv so that we can reference `uv.uv_fs_t` and friends.

Akmadan23:
> It looks like they've added the [@meta](https://luals.github.io/wiki/annotations/#meta) annotation to handle this case.

Nice, I didn't know that.

> If you can find a simple way that would be fantastic.

I'm working on it, seems to be pretty straight forward.

> [uvlib.patch.gz](https://github.com/nvim-tree/nvim-tree.lua/files/13465542/uvlib.patch.gz) adds libuv so that we can reference `uv.uv_fs_t` and friends, which takes us to 11.

Feel free to push to this branch if you need to.

hinell:
Nice. We actually need to document basic strcture: Node.

It seems like there is no class for that.

alex-courtis:
> Feel free to push to this branch if you need to.

Done.

alex-courtis:
I've taken the liberty of adding `scripts/luals-check` and fixing some nits.

I'd like to free you up to continue the actual work.

Some recipes:
`scripts/luals-check | grep "file:"`
`scripts/luals-check | jq --color-output`
`scripts/luals-check | yq --colors -p json`

alex-courtis:
> Nice. We actually need to document basic strcture: Node.
> 
> It seems like there is no class for that.

Please Seek First To Understand: https://github.com/nvim-tree/nvim-tree.lua/blob/type-annotations/lua/nvim-tree/node.lua

Akmadan23:
> I've taken the liberty of adding `scripts/luals-check` and fixing some nits.

Will it be added in this PR? I'd rather add it with a separate PR... By the way I haven't commented on #2546 yet because I tested it on this branch and it runs as expected, but on master it doesn't find any problem, where that is clearly not the case. I also experienced a similar behavior when I was adding type annotations in some files, and only then LLS displayed warnings about something completely unrelated to the annotations just added...

alex-courtis:
> > I've taken the liberty of adding `scripts/luals-check` and fixing some nits.
> 
> Will it be added in this PR? I'd rather add it with a separate PR... By the way I haven't commented on #2546 yet because I tested it on this branch and it runs as expected, but on master it doesn't find any problem, where that is clearly not the case. I also experienced a similar behavior when I was adding type annotations in some files, and only then LLS displayed warnings about something completely unrelated to the annotations just added...

Pulling it out now.

Are you cool with the little fixes?

alex-courtis:
Do you want me to revert 8393f7cc0c0056844f0403b4983324fc20f8571a as well?

Maybe just the `workspace` bit...

Akmadan23:
> Are you cool with the little fixes?

Yeah, thank you for those!

> Do you want me to revert 8393f7c as well?

No no, that's fine. I was curious about your output of the LLS check on master: does it display the warnings?

alex-courtis:
> No no, that's fine. I was curious about your output of the LLS check on master: does it display the warnings?

It does... but it needs some polishing. It can come much later.

Apologies again for rudely pushing to your branch and interrupting your work!

lua/nvim-tree/marks/init.lua
```diff
@@ -4,21 +4,24 @@ local NvimTreeMarks = {}
 
 local M = {}
 
----@param node Node
+---@class MinimalNode
```
alex-courtis, line 7: Wow... I like this one... I'm seeing many possibilities.

lua/nvim-tree/actions/fs/create-file.lua
```diff
@@ -8,17 +8,20 @@ local find_file = require("nvim-tree.actions.finders.find-file").fn
...
+---@param file string
 local function create_and_notify(file)
   events._dispatch_will_create_file(file)
   local ok, fd = pcall(vim.loop.fs_open, file, "w", 420)
   if not ok then
     notify.error("Couldn't create file " .. notify.render_path(file))
     return
   end
-  vim.loop.fs_close(fd)
+  vim.loop.fs_close(tonumber(fd) or error "This will never happen")
```
Akmadan23, line 19: Maybe it's not the best solution, but it works...

lua/nvim-tree/core.lua
```diff
@@ -31,10 +34,12 @@ function M.reset_explorer()
   TreeExplorer = nil
 end
 
+---@return string
```
Akmadan23, line 37: This should be `string|nil`, but many internal functions rely on always receiving a string, so that's something I need to fix.

lua/nvim-tree/node.lua
```diff
@@ -0,0 +1,33 @@
...
+---@field extension string
+
+---@class SymlinkDirNode: DirNode
+---@field link_to string
+
+---@class SymlinkFileNode: FileNode
+---@field link_to string
+
+---@alias SymlinkNode SymlinkDirNode|SymlinkFileNode
+---@alias Node ParentNode|DirNode|FileNode|SymlinkNode|Explorer
```
Akmadan23, line 33: Adding `Explorer` here seems to be the best solution, otherwise we would have many `Node`/`Explorer` inconsistencies.

lua/nvim-tree/lib.lua
```diff
@@ -81,6 +85,8 @@ function M.get_last_group_node(node)
   return next_node
```
Akmadan23, line 85: I really don't get why it complains about a missing nil check here... And consequently a possible nil return

lua/nvim-tree/explorer/init.lua
```diff
@@ -7,11 +7,20 @@ local M = {}
...
+---@field open boolean
+
 local Explorer = {}
 Explorer.__index = Explorer
 
+---@param cwd string|nil
+---@return Explorer
 function Explorer.new(cwd)
-  cwd = vim.loop.fs_realpath(cwd or vim.loop.cwd())
+  cwd = vim.loop.fs_realpath(cwd or vim.fn.getcwd())
```
Akmadan23, line 21: This always returns a string instead of `string|nil`

Akmadan23 (commented):
According to my tests we should be down to just 4 warnings, 3 of which start from a non-issue in my opinion. More details in the comments

Akmadan23:
> Apologies again for rudely pushing to your branch and interrupting your work!

Oh, don't worry, I was just a bit confused by the unexpected output on my side.

lua/nvim-tree/actions/fs/create-file.lua
alex-courtis: This is a problem. We're struggling with similar issues at my workplace right now...

alex-courtis:
The issues with `vim` and `uv` are getting in the way of the real goal here. I apologise for steering us in that direction.

Could be be practical and remove them from the libraries?
We can add temporary definitions for now and deal with vim, uv etc. later.

node.lua
```lua
-- TODO add "${3rd}/luv/library" to "workspace.library"
---@class uv.uv_req_t: table
---@class uv.uv_fs_t: uv.uv_req_t
```

runner.lua
```lua
-- TODO add "${3rd}/luv/library" to "workspace.library"
---@class uv.uv_handle_t : table
---@class uv.uv_stream_t : uv.uv_handle_t
---@class uv.uv_pipe_t : uv.uv_stream_t
```

Akmadan23:
Now we should have 0 warnings related to our internal type annotations.

> We could even back out the changes to get around the spurious warnings from vim.cmd etc.

Yes, makes sense to do this since those changes aren't directly related to the annotations added here. I can stash them and open a new PR with other corrections of that kind...

lua/nvim-tree/lib.lua
```diff
@@ -68,19 +69,25 @@ local function clone_node(node)
...
 function M.get_last_group_node(node)
-  local next_node = node
-  while next_node.group_next do
-    next_node = next_node.group_next
+  while node and node.group_next do
+    node = node.group_next
   end
-  return next_node
+
+  ---@diagnostic disable-next-line: return-type-mismatch -- it can't be nil
```
Akmadan23, line 85: I can't see the possibility of a nil return given that the `node` parameter is not nil, or am I wrong?

alex-courtis: Looks like luals is getting confused.

We could refactor but this is fine.

We can revisit suppressions once we have types annotations present and correct.

alex-courtis:
All the warnings are resolved!

What's left to do?

lua/nvim-tree/actions/fs/create-file.lua
alex-courtis: Can we revert this one? It's correct and it passes checks.

Dealing with the `vim.` luals issues is a job for another day, if we even want to do it.

lua/nvim-tree/view.lua
```diff
@@ -133,7 +133,10 @@ local function setup_tabpage(tabpage)
 end
 
 local function set_window_options_and_buffer()
-  pcall(vim.cmd, "buffer " .. M.get_bufnr())
+  pcall(function()
+    vim.cmd("buffer " .. M.get_bufnr())
+  end)
```
alex-courtis, line 138: Could we revert this one as well? I'm the idiot who put it in but it's no longer necessary.

This function needs a rewrite anyway - we need to listen for failures and handle them.

Akmadan23:
> What's left to do?

- [x] Doc comments style
- [X] Revert unrelated changes

Everything should be good to go now.

lua/nvim-tree/node.lua
```diff
@@ -0,0 +1,37 @@
...
+---@class FileNode: BaseNode
+---@field extension string
+
+---@class SymlinkDirNode: DirNode
+---@field link_to string
+
+---@class SymlinkFileNode: FileNode
+---@field link_to string
+
+---@alias SymlinkNode SymlinkDirNode|SymlinkFileNode
```
alex-courtis, line 36: I do like this inheritence approach; it removes the need for annoying/unnecessary casts.

alex-courtis (approved):
Lets go!

alex-courtis:
Just incredible work @Akmadan23
The devspeed improvements will be immediately gained and the safety is a massive win.

PR #2558, opened 2023-11-25
refactor: api and command focus call tree.open(), soft deprecate tree.focus()

alex-courtis:
see #2554


gegoune (commented):
Great job! Quick tests worked just fine. Thanks!

Issue #1645, opened 2022-10-14
The cmdheight setting is not kept when closing when open_replacing_current_buffer is used.


vheon:
### Description

The `cmdheight` settings doesn't seems to be kept when `open_replacing_current_buffer` is used and then the `close` action is issued.

### Neovim version

```text
NVIM v0.9.0-dev-91-gd339b4aad
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux 5.15.0-48-generic

### nvim-tree version

b01e7be

### Minimal config

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
vim.opt.cmdheight = 2
_G.setup = function()
  require("nvim-tree").setup {
    view = {
      mappings = {
        list = {
          { key = "<CR>", action = "edit_in_place" },
          { key = "<C-e>", action = "edit" }
        }
      }
    },
    actions = {
      open_file = {
        quit_on_open = true
      }
    }
  }
end
```

### Steps to reproduce

1. vim -nu /tmp/nvt-min.lua
2. :e /tmp/nvt-min.lua
3. :lua require"nvim-tree".open_replacing_current_buffer()
4. q

[...]

PR #2570, opened 2023-12-03
refactor(#1645): remove unused open_replacing_current_buffer

alex-courtis:
discovered via #1645

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -77,30 +77,6 @@ function M.change_root(path, bufnr)
   change_dir.fn(vim.fn.fnamemodify(path, ":p:h"))
 end
 
----@param cwd string|nil
-function M.open_replacing_current_buffer(cwd)
-  if view.is_visible() then
-    return
-  end
-
-  local buf = vim.api.nvim_get_current_buf()
-  local bufname = vim.api.nvim_buf_get_name(buf)
-  if bufname == "" or vim.loop.fs_stat(bufname) == nil then
-    return
-  end
-
-  if cwd == "" or cwd == nil then
-    cwd = vim.fn.fnamemodify(bufname, ":p:h")
-  end
-
-  if not core.get_explorer() or cwd ~= core.get_cwd() then
-    core.init(cwd)
-  end
-  view.open_in_win { hijack_current_buf = false, resize = false }
-  require("nvim-tree.renderer").draw()
-  require("nvim-tree.actions.finders.find-file").fn(bufname)
-end
-
 function M.tab_enter()
   if view.is_visible { any_tabpage = true } then
     local bufname = vim.api.nvim_buf_get_name(0)


sharkov63:
lmao, this broke my config since I mapped `open_replacing_current_buffer` on my key. Is there any quick workaround to get the same functionality?

alex-courtis:
See `:help nvim-tree-api.node.open.replace_tree_buffer`

steliyan:
My workflow is broken as well, but currently, I don't believe it's supported.

1. Open a file
2. Use `open_replacing_current_buffer` (now removed)
3. Nvim-tree opens in the current buffer
4. Select a new file
5. Press ENTER
6. The new file is opened

Step 3 is not supported. What is described in the docs is step 4 onwards, but this re-arranges the windows in neovim which is not great.

If `NvimTreeFindFile` can use the current window, this will cover my (and probably @sharkov63 ) use cases.

javiertury:
I'm on the same boat as @sharkov63 and @steliyan, but I have to agree that the maintenance burden may not be worth it.

The function `open_replacing_current_buffer` was introduced into nvim-tree as an utility (particularly useful for those who want to make nvim-tree behave more like netrw), so the workaround is simply to re-define the function ourselves:

```lua
local core = require "nvim-tree.core"

local function openReplacingCurrentBuffer(cwd)
  if view.is_visible() then
    return
  end

  local buf = vim.api.nvim_get_current_buf()
  local bufname = vim.api.nvim_buf_get_name(buf)

  if bufname == "" or vim.loop.fs_stat(bufname) == nil then
    return
  end

  if cwd == "" or cwd == nil then
    cwd = vim.fn.fnamemodify(bufname, ":p:h")
  end

  if not core.get_explorer() or cwd ~= core.get_cwd() then
    core.init(cwd)
  end
  view.open_in_win { hijack_current_buf = false, resize = false }
  require("nvim-tree.renderer").draw()
  require("nvim-tree.actions.finders.find-file").fn(bufname)
end

-- Then use it for whatever purpose you see fit, for instance reimplementing netrw functionality

local function openInPlace(cwd)
  if cwd == "" or cwd == nil then
    local buf = vim.api.nvim_get_current_buf()
    local bufname = vim.api.nvim_buf_get_name(buf)

    if bufname ~= "" and vim.loop.fs_stat(bufname) ~= nil then
      cwd = vim.fn.fnamemodify(bufname, ":p:h")
    else
      cwd = vim.loop.cwd()
    end
  end

  openReplacingCurrentBuffer(cwd)
end

--
-- Commands
--

vim.api.nvim_create_user_command("Explore", function(res)
  openInPlace(res.args)
end, {
  desc = 'Open file explorer in current buffer',
  nargs = "?", complete = "dir"
})
```

gegoune:
And why wouldn't you use provided api (https://github.com/nvim-tree/nvim-tree.lua/pull/2570#issuecomment-1862094961)?

javiertury:
@gegoune, that api function doesn't seem to work for my usecase. It requires `nvim-tree` to already be present in the buffer, and when you call `replace_tree_buffer`, it opens the selected file in the same buffer.

My goal is to replicate netrw functionality. I want to open a nvim-tree window even if it didn't exist and use the provided argument as the working directory for `nvim-tree`. But I don't want to open any file.

gegoune:
You can use `:h nvim-tree-api.tree.open()` for that.

javiertury:
Thank you @gegoune, that is exactly what I needed!

sharkov63:
> You can use `:h nvim-tree-api.tree.open()` for that.

This is what I was looking for, thank you very much!

PR #2575, opened 2023-12-04
ci: add style check for doc comments

Akmadan23:
Discussed in https://github.com/nvim-tree/nvim-tree.lua/pull/2555#issuecomment-1826209931

### DIFF
--- a/scripts/doc-comments.sh
+++ b/scripts/doc-comments.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+out=$(grep -nr "^--- @" lua)
+
+if [ "$out" ]; then
+	last_file=""
+	while read -r line; do
+		file="$(echo "$line" | cut -d: -f1)"
+		if [[ "$file" != "$last_file" ]]; then
+			echo "$file:" >&2
+			last_file="$file"
+		fi
+		echo "$line" | awk -F: '{ printf("  line %s: %s\n", $2, $3) }' >&2
+	done <<< "$out"
+	exit 1
+fi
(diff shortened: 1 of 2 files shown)


scripts/doc-comments.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
alex-courtis, line 1: I always try to use `/bin/sh` after I caused a shell compatibility production incident 20 years ago.

However, we need bash features here and we know that CI provides bash so we are OK.

alex-courtis (changes requested):
Fails nicely.

shellcheck :+1: 

We're going to need to fix all these before we can merge: `find lua -type f -exec sed -i 's/^--- @/---@/g' {} \;`

scripts/doc-comments.sh
gegoune: If you insist on `sh` (it's fine either way):

```sh
#!/bin/sh

out=$(grep -nr "^--- @" lua)

if [ "$out" ]; then
	last_file=""
	printf %s "$out" | while read -r line; do
		file="$(echo "$line" | cut -d: -f1)"
		if [ "$file" != "$last_file" ]; then
			echo "$file:" >&2
			last_file="$file"
		fi
		echo "$line" | awk -F: '{ printf("  line %s: %s\n", $2, $3) }' >&2
	done
	exit 1
fi
```

alex-courtis: Works nicely, shellcheck is happy.

I will sleep better at night.

alex-courtis (approved):
CI all passing now

PR #2583, opened 2023-12-09
feat: add option to skip gitignored files on git navigation

Akmadan23:
This would be handy since usually we don't want to focus gitignored files. We have to decide if this option should be local to the function (in this case we would need to add a couple of extra API wrappers, something like `node.navigate.git.next_skip_gitignored`) or if we want it to be global...


lua/nvim-tree/actions/moves/item.lua
```diff
@@ -8,8 +8,9 @@ local M = {}
 
 ---@param where string
 ---@param what string
+---@param opts table|nil
 ---@return fun()
-function M.fn(where, what)
+function M.fn(where, what, opts)
```
Akmadan23, line 13: Moreover, if we all agree, I'd do a little refactor here, having just an `opts` table with the `where` and `what` fields, so that we can easily add new ones if necessary.

alex-courtis: Yes please. The API could be altered to pass `opts`.

We would need to retain backwards compatibility as per `ApiTreeToggleOpts` so as not to break existing users.

alex-courtis, line 13: nil check please. Existing mappings pass nil and blow up :)

gegoune: Unless we release 0.x.x and introduce whatever breaking changes we want for 1.0.

alex-courtis: > Unless we release 0.x.x and introduce whatever breaking changes we want for 1.0.

Interesting... the highlight overhaul could also go before 1.0.0. Perhaps we might do an audit and do any "final breaks" before the big release.

alex-courtis (changes requested):
Looking good, skips nicely:
* `nowrapscan`
* changes before and after
* changes before
* changes after
* `skip_gitignored` false

just need to get default mappings / API sorted.

lua/nvim-tree/actions/moves/item.lua
Akmadan23: I know, it was in fact temporary, with #2584 everything will work fine :)

alex-courtis:
What are you thinking about now for API? Maybe:

`node.navigate.git.next({skip_gitignored = true})`

or even full user control:

`node.navigate({ where = "prev", what = "diag" , skip_gitignored = true})`

I'm not the biggest fan of more options...

Akmadan23:
> `node.navigate.git.next({skip_gitignored = true})`

Between the two I would go with this one, but I would even prefer to add a separate `node.navigate.git.next_skip_gitignored` to easily use it in keymaps without a `function() ... end` wrapper...

alex-courtis:
> Between the two I would go with this one, but I would even prefer to add a separate `node.navigate.git.next_skip_gitignored` to easily use it in keymaps without a `function() ... end` wrapper...

Fair enough; that's more consistent with the current no-arg navigation methods, indeed all of our default mappings. The user wouldn't really gain anything from the extra complication of args.

Skip ignored args are also very git specific and any other possible navigation skips would be equally specific.

Let's keep it simple.

alex-courtis (approved):
Approved following the actual addition to API etc.

Akmadan23:
We have a couple of choices to address this style issue:
- Split lines on `=` like stylua suggests
- Format tables line by line
- Save `require("nvim-tree.actions.moves.item")` in a local variable

I'd personally go with the latter, but that would introduce an inconsistency with the rest of API entries. However, if we agree on this, I could work on a refactor to make it the standard. It would also greatly improve readability in my opinion...

alex-courtis:
> We have a couple of choices to address this style issue:
> 
>     * Split lines on `=` like stylua suggests
> 
>     * Format tables line by line
> 
>     * Save `require("nvim-tree.actions.moves.item")` in a local variable
> 
> 
> I'd personally go with the latter, but that would introduce an inconsistency with the rest of API entries. However, if we agree on this, I could work on a refactor to make it the standard. It would also greatly improve readability in my opinion...

It would be much more readable however we are in circular dependency land. API is a big problem.

>     * Format tables line by line

Do you mean something like
```lua
Api.node = {
  navigate = {
    git = {
      ["next"] = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "next", what = "git" }),
      prev = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "prev", what = "git" }),
      next_skip_gitignored = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "next", what = "git", skip_gitignored = true }),
      prev_skip_gitignored = wrap_node(require("nvim-tree.actions.moves.item").fn { where = "prev", what = "git", skip_gitignored = true }),
    },
  },
}
```

That would enhance readability.

There are a hundred ways we can refactor / split / tidy / lazy bind API however I'm happy enough with a `-- stylua: ignore
` above the bindings. It's a special case.

alex-courtis (approved):
Looking good; looking forward to using this one...

Akmadan23:
> There are a hundred ways we can refactor / split / tidy / lazy bind API however I'm happy enough with a `-- stylua: ignore` above the bindings. It's a special case.

Alright, let's merge this way for now.

> It would be much more readable however we are in circular dependency land. API is a big problem.

Well, the `api` module is only required in the `commands` module, so there aren't many circular dependencies. I already worked on a refactor proposal, pushing PR soon.

PR #2584, opened 2023-12-10
refactor: take single `opts` param on node navigation

Akmadan23:
Discussed in #2583


alex-courtis (approved):
That worked out nicely! No API changes needed.

PR #2593, opened 2023-12-20
refactor: tidy `actions` submodules and improve API readability

Akmadan23:
Discussed in #2583


lua/nvim-tree/api.lua
```diff
@@ -205,54 +189,59 @@ Api.node.open.tab = wrap_node(open_or_expand_or_dir_up "tabnew")
...
+local live_filter = require "nvim-tree.live-filter"
+
+Api.live_filter.start = wrap(live_filter.start_filtering)
+Api.live_filter.clear = wrap(live_filter.clear_filter)
+
+local marks = require "nvim-tree.marks"
+local marks_navigation = require "nvim-tree.marks.navigation"
+local marks_bulk_delete = require "nvim-tree.marks.bulk-delete"
+local marks_bulk_trash = require "nvim-tree.marks.bulk-trash"
+local marks_bulk_move = require "nvim-tree.marks.bulk-move"
```
Akmadan23, line 227: I tried to add the various `init.lua`s in the `marks` module but dependency loops start to come up, so this is best solution I could find.

Akmadan23 (commented):
The move of `actions/tree-modifiers` to `actions/tree/modifiers` is just a proposal, I think it makes sense to put it there but I can revert this change if it's better not to do that.
I would also move `actions/reloaders/reloaders.lua` to `actions/reloaders/init.lua` or even `actions/reloaders.lua`, as long as it's not a problem...

lua/nvim-tree/api.lua
```diff
@@ -205,54 +189,59 @@ Api.node.open.tab = wrap_node(open_or_expand_or_dir_up "tabnew")
...
+Api.node.navigate.diagnostics.next = wrap_node(actions.moves.item.fn { where = "next", what = "diag" })
+Api.node.navigate.diagnostics.prev = wrap_node(actions.moves.item.fn { where = "prev", what = "diag" })
+Api.node.navigate.opened.next = wrap_node(actions.moves.item.fn { where = "next", what = "opened" })
+Api.node.navigate.opened.prev = wrap_node(actions.moves.item.fn { where = "prev", what = "opened" })
+
+Api.git.reload = wrap(actions.reloaders.reload_git)
+
+local events = require "nvim-tree.events"
+
+Api.events.subscribe = wrap(events.subscribe)
```
alex-courtis, line 215: We can't wrap this one - we alllow users to subscribe before setup is called.

lua/nvim-tree/api.lua
```diff
@@ -205,54 +189,59 @@ Api.node.open.tab = wrap_node(open_or_expand_or_dir_up "tabnew")
...
+Api.marks.bulk.move = wrap(marks_bulk_move.bulk_move)
+Api.marks.navigate.next = wrap(marks_navigation.next)
+Api.marks.navigate.prev = wrap(marks_navigation.prev)
+Api.marks.navigate.select = wrap(marks_navigation.select)
+
+local keymap = require "nvim-tree.keymap"
+
+Api.config.mappings.get_keymap = wrap(keymap.get_keymap)
+Api.config.mappings.get_keymap_default = wrap(keymap.get_keymap_default)
+Api.config.mappings.default_on_attach = wrap(keymap.default_on_attach)
```
alex-courtis, line 244: No wrap for this one either; we let the users call this for before setup, even though it's an odd use case.

lua/nvim-tree/api.lua
```diff
@@ -205,54 +189,59 @@ Api.node.open.tab = wrap_node(open_or_expand_or_dir_up "tabnew")
...
+Api.marks.toggle = wrap_node(marks.toggle_mark)
+Api.marks.clear = wrap(marks.clear_marks)
+Api.marks.bulk.delete = wrap(marks_bulk_delete.bulk_delete)
+Api.marks.bulk.trash = wrap(marks_bulk_trash.bulk_trash)
+Api.marks.bulk.move = wrap(marks_bulk_move.bulk_move)
+Api.marks.navigate.next = wrap(marks_navigation.next)
+Api.marks.navigate.prev = wrap(marks_navigation.prev)
+Api.marks.navigate.select = wrap(marks_navigation.select)
+
+local keymap = require "nvim-tree.keymap"
```
alex-courtis, line 240: Could we move these requires to the top, or is order important here?

alex-courtis:
> The move of `actions/tree-modifiers` to `actions/tree/modifiers` is just a proposal, I think it makes sense to put it there but I can revert this change if it's better not to do that. I would also move `actions/reloaders/reloaders.lua` to `actions/reloaders/init.lua` or even `actions/reloaders.lua`, as long as it's not a problem...

That's great - it is a hierarchy which makes sense.

lua/nvim-tree/api.lua
alex-courtis: Marks are not quite like the other actions. They need a refactor / tidy but not today.

alex-courtis (changes requested):
This is incredible! Api is finally readable. There's very little more we can do for now...

Now you have me thinking of ways we could get rid of wrap...

lua/nvim-tree/api.lua
Akmadan23: I put the definitions just before the references to easily trace back which module each variable refers to, of course we can move them to the top.

lua/nvim-tree/actions/fs/copy-paste.lua
```diff
@@ -5,7 +5,7 @@ local core = require "nvim-tree.core"
 local events = require "nvim-tree.events"
 local notify = require "nvim-tree.notify"
 local renderer = require "nvim-tree.renderer"
-local reloaders = require "nvim-tree.actions.reloaders.reloaders"
+local reloaders = require "nvim-tree.actions.reloaders"
```
alex-courtis, line 8: Nice

lua/nvim-tree/actions/init.lua
```diff
@@ -1,18 +1,18 @@
...
-  require("nvim-tree.actions.fs.remove-file").setup(opts)
-  require("nvim-tree.actions.fs.copy-paste").setup(opts)
-  require("nvim-tree.actions.tree-modifiers.expand-all").setup(opts)
-  require("nvim-tree.actions.tree.find-file").setup(opts)
-  require("nvim-tree.actions.tree.open").setup(opts)
-  require("nvim-tree.actions.tree.toggle").setup(opts)
+  M.fs.setup(opts)
+  M.node.setup(opts)
+  M.root.setup(opts)
+  M.tree.setup(opts)
```
alex-courtis, line 15: This `@PostConstruct` style is an unfortunate necessity for lua's "dependency injection".

I'll be mindful to add more of this as I touch files with odd instantiation mechanisms.

alex-courtis (approved):
Love your work. Let's get this in!

Issue #2519, opened 2023-11-06
Diagnostics Not Updated When Tree Not Visible


alex-courtis:
### Discussed in https://github.com/nvim-tree/nvim-tree.lua/discussions/2517

<div type='discussions-op-text'>

<sup>Originally posted by **juhaku** November  7, 2023</sup>
Hey,

I'd like a support for functionality where the diagnostics status and what not would be updated / reloaded upon NvimTree  reveal? Is that currently possible? I did not see any configuration option that would enable such thing.

EDIT: There is `nvim-tree.auto_reload_on_write` that allows tree to update while being opened and a file is being written to. Also there is `tree.reload()` API that is tied to `R` command command and `reload_on_bufenter` configuration flag but non of these seem to give remedy to the situation. Like hitting `R` button does not seem to affect much to the state of the tree at least the diagnostics does not get updated. Same happens no matter I travel between different buffers. The only way for diagnostics to actually get updated seems to be by way of writing on buffer while keeping the NvimTree open. 

-Juha</div>

alex-courtis:
Thanks for finding this one @juhaku

Diagnostic events are being received whilst the tree is closed however the nodes are not being updated:

```
[2023-11-07 10:32:58] [diagnostics] DiagnosticChanged
[2023-11-07 10:32:58] [diagnostics] DiagnosticChanged
```

Should see
```
[2023-11-07 10:33:40] [diagnostics] DiagnosticChanged
[2023-11-07 10:33:40] [profile] START diagnostics update
[2023-11-07 10:33:40] [diagnostics] update
...
```

juhaku:
Anytime :slightly_smiling_face:

PR #2597, opened 2023-12-22
fix(#2519): Diagnostics Not Updated When Tree Not Visible

iusmac:
The diagnostic changed event has been untied from the current tree state to ensure the data availability at any time. Now, the diagnostics will be picked up "on the fly" whenever the tree is about to be (re-)drawn, which makes absolute sense, especially, when incrementally exploring the file system structure or when (re-)opening the tree and the diagnostic changed event has already occurred.

All in all, this effectively solves any inconsistencies (see https://github.com/nvim-tree/nvim-tree.lua/discussions/2517), and ensures that the tree shows the freshest diagnostics, regardless of the tree's state and when diagnostics change.

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -1,6 +1,5 @@
 local utils = require "nvim-tree.utils"
 local view = require "nvim-tree.view"
-local core = require "nvim-tree.core"
 local log = require "nvim-tree.log"
 
 local M = {}
@@ -11,6 +10,16 @@ local severity_levels = {
   Information = 3,
   Hint = 4,
 }
+
+--- A dictionary tree containing buffer-severity mappings.
+---@type table
+local buffer_severity_dict = {}
+
+---@param path string
+---@return string
+local function uniformize_path(path)
+  return utils.canonical_path(path:gsub("\\", "/"))
+end
 
 ---@return table
 local function from_nvim_lsp()
@@ -25,7 +34,7 @@ local function from_nvim_lsp()
     for _, diagnostic in ipairs(vim.diagnostic.get(nil, { severity = M.severity })) do
       local buf = diagnostic.bufnr
       if vim.api.nvim_buf_is_valid(buf) then
-        local bufname = vim.api.nvim_buf_get_name(buf)
+        local bufname = uniformize_path(vim.api.nvim_buf_get_name(buf))
         local lowest_severity = buffer_severity[bufname]
         if not lowest_severity or diagnostic.severity < lowest_severity then
           buffer_severity[bufname] = diagnostic.severity
@@ -67,7 +76,7 @@ local function from_coc()
   local buffer_severity = {}
   for bufname, severity in pairs(diagnostics) do
     if is_severity_in_range(severity, M.severity) then
-      buffer_severity[bufname] = severity
+      buffer_severity[uniformize_path(bufname)] = severity
     end
   end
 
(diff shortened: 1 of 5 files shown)


iusmac:
I've just noticed that fetching diagnostics with coc is completely broken when setting `nvim-tree.diagnostics.severity.max` to something else than `ERROR` (default).

The current implementation will find the highest severity, then discard items that are out of range.
Reversing the order fixes and simplifies everything.

Now, the lsp and coc functions look pretty much the same, which makes sense, as there's no big difference between the two, except that with coc we have to bother about the severity limit of each item.

The commit that introduced this bug: 874b7be5d053f1b31f545099d6fcbe8ae81e9e03.

lua/nvim-tree/diagnostics.lua
```diff
@@ -12,6 +11,16 @@ local severity_levels = {
   Hint = 4,
 }
 
+--- A dictionary tree containing buffer-severity mappings.
+---@type table
+local buffer_severity_dict = {}
+
+---@param path string
+---@return string
+local function uniformize_path(path)
```
alex-courtis, line 20: Nice name!

lua/nvim-tree/renderer/builder.lua
```diff
@@ -413,6 +413,8 @@ end
 function Builder:_build_line(node, idx, num_children, unloaded_bufnr)
   local copy_paste = require "nvim-tree.actions.fs.copy-paste"
 
+  require("nvim-tree.diagnostics").update_node_severity_level(node)
```
alex-courtis, line 416: This one is very concerning from a performance perspective;  we go out of our way to avoid such calls on drawing each line.

Could we improve via one of:
* do this once for all nodes prior to rendering
* short-circuit this so that it's only called when there is no diagnostic info

lua/nvim-tree/diagnostics.lua
```diff
@@ -79,47 +81,52 @@ local function is_using_coc()
...
+      buffer_severity_dict = from_nvim_lsp()
     end
-
-    local nodes_by_line = utils.get_nodes_by_line(core.get_explorer().nodes, core.get_nodes_starting_line())
-    for _, node in pairs(nodes_by_line) do
-      node.diag_status = nil
+    log.node("diagnostics", buffer_severity_dict, "update")
+    log.profile_end(profile)
+    if view.is_buf_valid(view.get_bufnr()) then
+      require("nvim-tree.renderer").draw()
```
alex-courtis, line 97: / nit please require once at the start of the file

lua/nvim-tree/diagnostics.lua
```diff
@@ -79,47 +81,52 @@ local function is_using_coc()
...
     return
   end
   utils.debounce("diagnostics", M.debounce_delay, function()
     local profile = log.profile_start "diagnostics update"
-    log.line("diagnostics", "update")
-
-    local buffer_severity
     if is_using_coc() then
-      buffer_severity = from_coc()
+      buffer_severity_dict = from_coc()
```
alex-courtis, line 90: / nit no dict please - we're not writing python here ;)

alex-courtis (changes requested):
This is looking great - many thanks for dealing with the regressive behaviour.

iusmac:
While we're at it, implement error handling for coc.nvim.

If you're lucky and hit `CTRL-C` while NvimTree is fetching diagnostics from coc.nvim, it can cause a deadlock state in the `utils.debounce()` that will never release its `debouncer.executing` flag due the exception in the callback. All subsequent diagnostic update events will be discarded.

```log
Error executing vim.schedule lua callback: function CocAction[4]..coc#rpc#request[4]..<SNR>38_request, line 31: Error on request: Vim:Interrupt
stack traceback:
        [C]: in function 'CocAction'
        ....vim/plugged/nvim-tree.lua/lua/nvim-tree/diagnostics.lua:69: in function 'from_coc'
        ....vim/plugged/nvim-tree.lua/lua/nvim-tree/diagnostics.lua:123: in function 'callback'
        ...usmac/.vim/plugged/nvim-tree.lua/lua/nvim-tree/utils.lua:400: in function <...usmac/.vim/plugged/nvim-tree.lua/lua/nvim-tree/utils.lua:399>
```

lua/nvim-tree/diagnostics.lua
```diff
@@ -12,6 +11,24 @@ local severity_levels = {
   Hint = 4,
 }
 
+---@class DiagStatus
+---@field value integer|nil
+---@field cache_version integer
+
+--- The buffer-severity mappings derived during the last diagnostic list update.
+---@type table
+local BUFFER_SEVERITY = {}
```
alex-courtis, line 20: This is looking good; we have a cache similar to git's.

It'll take me a while to properly review and test this one.

alex-courtis (approved):
Looks great, works well.

I did not test COC. In the unlikely event of any issues I'll call on you ;)

Added some doc and some minor tidies.

Many thanks for all your work!

lua/nvim-tree/diagnostics.lua
```diff
@@ -1,18 +1,44 @@
 local utils = require "nvim-tree.utils"
 local view = require "nvim-tree.view"
-local core = require "nvim-tree.core"
 local log = require "nvim-tree.log"
 
 local M = {}
 
-local severity_levels = {
+---TODO add "$VIMRUNTIME" to "workspace.library" and use the @enum instead of this integer
```
alex-courtis, line 7: That one's a work in progress.

Issue #2599, opened 2023-12-25
`hijack_cursor` with `update_focused_file.enable` does not set cursor on focus/toggle


geril2207:
### Description

After opening file and focus nvim-tree, cursor does not set to node name for the first time.

### Neovim version

```text
NVIM v0.10.0-dev-1951+ga42df1178
Build type: Release
LuaJIT 2.1.1703358377
```

### Operating system and version

arch

### nvim-tree version

master

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvt-min/site]])
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
	require("packer").startup({
		{
			"wbthomason/packer.nvim",
			"nvim-tree/nvim-tree.lua",
			"nvim-tree/nvim-web-devicons",
			-- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
		},
		config = {
			package_root = package_root,
			compile_path = install_path .. "/plugin/packer_compiled.lua",
			display = { non_interactive = true },
		},
	})
end
if vim.fn.isdirectory(install_path) == 0 then
	print("Installing nvim-tree and dependencies.")
	vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
vim.cmd([[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]])
vim.opt.termguicolors = true
vim.opt.cursorline = true

vim.g.mapleader = " "

vim.keymap.set("n", "<leader>e", ":NvimTreeFocus<CR>", { silent = true, noremap = true })

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
	require("nvim-tree").setup({
		hijack_netrw = true,
		disable_netrw = true,
		hijack_cursor = true,
		update_focused_file = {
			enable = true,
		},
	})
end

-- UNCOMMENT this block for diagnostics issues, substituting pattern and cmd as appropriate.
-- Requires diagnostics.enable = true in setup.
--[[
vim.api.nvim_create_autocmd("FileType", {
[...]

PR #2600, opened 2023-12-25
fix: hijack_cursor with update_focused_file

geril2207:
Fix: #2599

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -112,13 +112,11 @@ function M.reset_highlight()
   renderer.render_hl(view.get_bufnr())
 end
 
-local prev_line
 function M.place_cursor_on_node()
-  local l = vim.api.nvim_win_get_cursor(0)[1]
-  if l == prev_line then
+  local search = vim.fn.searchcount()
+  if search and search.exact_match == 1 then
     return
   end
-  prev_line = l
 
   local node = lib.get_node_at_cursor()
   if not node or node.name == ".." then


alex-courtis:
Unfortunately this doesn't always work. It seems the issue might be with `place_cursor_on_node`

[2599.log.patch.gz](https://github.com/nvim-tree/nvim-tree.lua/files/13767423/2599.log.patch.gz)

```
[2023-12-26 09:51:52] [dev] hijacking {
  buf = 2,
  event = "BufEnter",
  file = "/home/alex/src/linux/NvimTree_1",
  group = 8,
  id = 14,
  match = "/home/alex/src/linux/NvimTree_1"
}
[2023-12-26 09:51:52] [dev] place_cursor_on_node
[2023-12-26 09:51:52] [dev] hijacking {
  buf = 2,
  event = "CursorMoved",
  file = "/home/alex/src/linux/NvimTree_1",
  group = 8,
  id = 14,
  match = "/home/alex/src/linux/NvimTree_1"
}
[2023-12-26 09:51:52] [dev] place_cursor_on_node
...
[2023-12-26 09:52:00] [dev] hijacking {
  buf = 2,
  event = "CursorMoved",
  file = "/home/alex/src/linux/NvimTree_1",
  group = 8,
  id = 14,
  match = "/home/alex/src/linux/NvimTree_1"
}
[2023-12-26 09:52:00] [dev] place_cursor_on_node
[2023-12-26 09:52:00] [dev] nvim_win_set_cursor(0, { 35, 6 })
```

alex-courtis:
It also looks like the `BufEnter` is redundant, as `CursorMoved` is documented to fire when moved to another window.

geril2207:
> Unfortunately this doesn't always work. It seems the issue might be with `place_cursor_on_node`

I don't find any issues, but yeah, there are better ways to resolve.
Why do we need the following? Perf reasons? For me removing this check fixes the issue.
https://github.com/nvim-tree/nvim-tree.lua/blob/50f30bcd8c62ac4a83d133d738f268279f2c2ce2/lua/nvim-tree.lua#L117C2-L121

alex-courtis:
> For me removing this check fixes the issue. Git blame says it was search fix, can you describe more about this? How can we change it. Actually it is, without this check you can't go next in search using `n` but can go prev with `N`. Seems that we need to find better way to handle if search with hijack_cursor.

That is very old code indeed, unfortunately I have no insight. It could be from a time when there was no root node or there was something else on the top line.

It's actually blocking hijack on the root node itself.

This is only called in one place and we can easily test it. Do you want to just rewrite that method? It's clear what it needs to do and can be very simple.

geril2207:
> Correct me if I'm wrong, but it looks like we don't have any search n or N mappings or functionality, just a one off search and find.

I am about native neovim `/` or `?` search where you have native n and N binds for jump next/prev. I don't think we should break it. And with removing line check, we will break n bind completely, you can't jump to the next occurence if current one not on the start.

alex-courtis:
> I am about native neovim `/` or `?` search where you have native n and N binds for jump next/prev. I don't think we should break it. And with removing line check, we will break n bind completely, you can't jump to the next occurence if current one not on the start.

Ah I see, I completely misunderstood. I was thinking about `S` search rather than reglar vim search. Yes, we should not break that.

This is quite tricky. The `prev_line` local is quite hacky and will break when switching nvim-tree windows / tabs etc.

Do you want to try writing a "clean" implementation? It seems clear what the requirements are now.

geril2207:
> This is quite tricky. The prev_line local is quite hacky and will break when switching nvim-tree windows / tabs etc.
Do you want to try writing a "clean" implementation? It seems clear what the requirements are now.

I don't think we need completely rewrite impl, just to figure out how to handle search better. Any ideas, how to detect if search, or maybe detect params of text under cursor? Currently I have no way to solve it.

alex-courtis:
> Any ideas, how to detect if search, or maybe detect params of text under cursor? Currently I have no way to solve it.

Ah yes, that would do it.

Looking at lualine it's using `hlsearch` and seems functional: https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/components/searchcount.lua

geril2207:
My assumption is that we can get search entries through extmarks and if cursor under search extmark we skip hijacking.

alex-courtis:
> My assumption is that we can get search entries through extmarks and if cursor under search extmark we skip hijacking.

Yes please... that will inform us whether we are searching inside the nvim-tree window AND we are on the match.

geril2207:
Seems that we can't do that with extmarks. I checked `vim.fn.searchcount` and seems that we can use `exact_match` from the result. From my tests, if cursor under search word `exact_match` eq to 1. In this case we can skip hijacking. I can do rewrite with `vim.fn.searchcount` but not now.

geril2207:
So what do you think about current solution?

alex-courtis:
> So what do you think about current solution?

Well, that's rather elegant. Cursor is always correctly placed on n/N and is hijacked on j/k
* `update_focused_file` T/F
* `hijack_cursor` T/F
* `&wrapscan` on/off
* `root_folder_label = false`

Regression: user can't move horizontally on a node with h/l.
Looking at the documentation "Keeps the cursor on the first letter of the filename when moving in the tree."
I'm happy to run with this new behaviour - it matches the documentation.

alex-courtis (approved):
Fantastic work as always @geril2207

PR #2619, opened 2024-01-06
ci: run on push release-please--branches--master--components--nvim-tree

alex-courtis:


.github/workflows/ci.yml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - master
+        release-please--branches--master--components--nvim-tree
```
gegoune, line 10: Shouldn't that be an element of the list?

alex-courtis: Right you are... I fail at YAML

#2621

PR #2627, opened 2024-01-13
fix(#2624): open file from docked floating window

### LINKED ISSUE #2624: Float view doen't autoclose on opening a file
### Description

When hitting enter with the cursor on a line with some file name, the file opens in a new buffer, but the nvim-tree doesn't autoclose. Upon trying to move cursor in the nvim-tree window, a number of errors pop up. For some reason the problem does not occur with ```hijack_cursor = false```. The last commit without this bug is 96a783fbd606a458bcce2ef8041240a8b94510ce.

### Neovim version

```text
NVIM v0.9.2
Build type: RelWithDebInfo
LuaJIT 2.1.1694082368
```

Windows 10

### Windows variant

_No response_

### nvim-tree version

b8c3a23

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set packpath=F:/nvt-min/site]]
local package_root = "F:/nvt-min/site/pack"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup {
    hijack_cursor = true,
    view = {
      float = {
        enable = true,
      },
    },
  }
end
```

### Steps to reproduce

1. ```nvim -nu F:/nvt-min/nvt-min.lua```
2. ```:e .```
3. Position cursor on some file
4. Hit Enter

### Expected behavior

_No response_

### Actual behavior

_No response_

alex-courtis:
fixes #2624

### DIFF
--- a/lua/nvim-tree/lib.lua
+++ b/lua/nvim-tree/lib.lua
@@ -24,7 +24,7 @@ function M.get_node_at_cursor()
     return
   end
 
-  local cursor = vim.api.nvim_win_get_cursor(view.get_winnr())
+  local cursor = vim.api.nvim_win_get_cursor(winnr)
   local line = cursor[1]
 
   if line == 1 and view.is_root_folder_visible(core.get_cwd()) then
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -473,7 +473,7 @@ end
 function M.get_winnr(tabpage)
   tabpage = tabpage or vim.api.nvim_get_current_tabpage()
   local tabinfo = M.View.tabpages[tabpage]
-  if tabinfo ~= nil then
+  if tabinfo ~= nil and vim.api.nvim_win_is_valid(tabinfo.winnr) then
     return tabinfo.winnr
   end
 end


lua/nvim-tree/lib.lua
```diff
@@ -24,7 +24,7 @@ function M.get_node_at_cursor()
     return
   end
 
-  local cursor = vim.api.nvim_win_get_cursor(view.get_winnr())
```
alex-courtis, line 27: Not strictly necessary.

lua/nvim-tree/view.lua
```diff
@@ -473,7 +473,7 @@ end
 function M.get_winnr(tabpage)
   tabpage = tabpage or vim.api.nvim_get_current_tabpage()
   local tabinfo = M.View.tabpages[tabpage]
-  if tabinfo ~= nil then
+  if tabinfo ~= nil and vim.api.nvim_win_is_valid(tabinfo.winnr) then
```
shiroyasha, line 476: @alex-courtis I'm hunting down a bug that seems to be related to this change. Can you help me out in figuring out what is the problem? 🙏

What I currently know is that when I call:

```
api.node.open.replace_tree_buffer
```

I get the following error:

```
E5108: Error executing lua: ...config/nvim/plugged/nvim-tree.lua/lua/nvim-tree/view.lua:476: Expected Lua number
stack traceback:
        [C]: in function 'nvim_win_is_valid'
        ...config/nvim/plugged/nvim-tree.lua/lua/nvim-tree/view.lua:476: in function 'get_winnr'
        ....config/nvim/plugged/nvim-tree.lua/lua/nvim-tree/lib.lua:22: in function 'get_node_at_cursor'
        ....config/nvim/plugged/nvim-tree.lua/lua/nvim-tree/api.lua:61: in function <....config/nvim/plugged/nvim-tree.lua/lua/nvim-tree/api.lua:60>
```

Do you maybe have an idea where should I look next?

alex-courtis: This change did indeed surface some issues, many thanks for digging into this!

Can you log the value of tabinfo.winnr? As of current master it's nil checked, however anything could be in there.

We could just do a `type(tabinfo.winnr) == "number"`...

#2628 will do away with a lot of these issues.

PR #2636, opened 2024-01-20
fix: use highlight default for DEFAULT_DEFS

kevmuko:
In the previous `colors.setup()` the highlight groups were setup using `hi def` but that was missing for those groups under `DEFAULT_DEFS`. 

I use catppuccin's `custom_highlights` to override `NvimTreeFolderIcon` and `NvimTreeWindowPicker`, which did not work in overriding the default highlights after the recently pushed commits.

### DIFF
--- a/lua/nvim-tree/appearance.lua
+++ b/lua/nvim-tree/appearance.lua
@@ -191,7 +191,7 @@ local LEGACY_LINKS = {
 function M.setup()
   -- non-linked
   for k, d in pairs(DEFAULT_DEFS) do
-    vim.api.nvim_command("hi " .. k .. " " .. d)
+    vim.api.nvim_command("hi def " .. k .. " " .. d)
   end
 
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Many thanks for the fast find and fix!

Tested before and after nvim-tree setup:
```lua
require("catppuccin").setup({
  flavour = "macchiato",
  background = {
    light = "latte",
    dark = "mocha",
  },
  highlight_overrides = {
    all = function(colors)
      return {
        NvimTreeNormal = { fg = colors.red, bg = colors.flamingo },
        NvimTreeFolderIcon = { fg = colors.green, bg = colors.yellow },
        NvimTreeWindowPicker = { fg = colors.blue, bg = colors.pink },
      }
    end,
  },
  integrations = {
    -- nvimtree = true,
    nvimtree = false,
  },
})
vim.cmd.colorscheme("catppuccin")
```

Issue #2643, opened 2024-01-23
after #2415 NonText and WinSeparator have no highlight


JosefLitos:
### Description

![obrazek](https://github.com/nvim-tree/nvim-tree.lua/assets/54900518/1899ef81-b5fb-4242-8d77-81927e831178)

you can also see the weird `^^^^` in my statusline that appear only whenever I enter the NvimTree window (I have `laststatus=3`)

### Neovim version

```text
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1702233742
```

### Operating system and version

Linux IP5Pro 6.7.0-arch3-1 #1 SMP PREEMPT_DYNAMIC Sat, 13 Jan 2024 14:37:14 +0000 x86_64 GNU/Linux

### nvim-tree version

e9c5abe073a973f54d3ca10bfe30f253569f4405

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd 'hi link WinSeparator LineNr'

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end

[...]

PR #2652, opened 2024-01-27
fix(#2651): window's LineNrAbove and LineNrBelow map to NvimTreeLineNr

### LINKED ISSUE #2651: LineNrAbove/Below Not Applying LineNr
### Discussed in https://github.com/nvim-tree/nvim-tree.lua/discussions/2640

<sup>Originally posted by **TroySigX** January 23, 2024</sup>
Before this commit e9c5abe073a973f54d3ca10bfe30f253569f4405, my column for relative number (left-most column) will look like this:
image

But after, it look like this:
image

How can I change the color of the this column to be like before?

P.S: I'm using tokyonight.nvim colorscheme
</div>

alex-courtis:
fixes #2651

### DIFF
--- a/lua/nvim-tree/appearance.lua
+++ b/lua/nvim-tree/appearance.lua
@@ -138,6 +138,9 @@ local NS_LINKS = {
   Normal = "NvimTreeNormal",
   NormalNC = "NvimTreeNormalNC",
   NormalFloat = "NvimTreeNormalFloat",
+  -- Not mapped to cyclic LineNr when relativenumber set.
+  LineNrAbove = "NvimTreeLineNr",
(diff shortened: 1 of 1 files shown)


petobens:
I had the same issue and this fixes it. Thanks!

alex-courtis:
That's great, thank you.

We could use default groups like
```
  NvimTreeLineNrAbove = "LineNrAbove",
  NvimTreeLineNrBelow = "LineNrBelow",
```
however they only apply if LineNrAbove/Below have been defined. The default to LineNr exhibits the same issue.

Let's deal with #2643 and merge this if no further issues arise.

alex-courtis:
Fixed by #2653

PR #2655, opened 2024-01-29
feat(#2654): filters.custom may be a function

### LINKED ISSUE #2654: Add a `binaries` field to `nvim-tree-opts-filters` or `nvim-tree-opts-sort`
No

Binary executables are currently shown in the tree, alongside normal files; however they are unlikely to be useful.
They cannot be hidden with a filename pattern, since they don't have a file extension or other common pattern.

I would like to hide binary files, or sort them below other files.
Ideally binary+executable files (such as ELFs) are distinct from other file types such as images.

Filtering with a `nvim-tree.filters` or `nvim-tree.filters.custom`

Related to #2648 and similar to #2616

dxrcy:
Fixes #2654

Related to #2654 and #2648.

1. Adds support for Lua functions in `filters.custom`.
Functions take the absolute file path (`string`), and return `bool`.

2. Adds `filters.binaries` option to ignore ELF files.
Files are read 4 bytes to check for `\x7fELF` header.
No other executable/binary formats are currently supported, but can be easily added.
Toggle hotkey is currently `h` as a placeholder.

Example usage:

```lua
local opts = {
    filters = {
        custom = {
            --- This:
            "*.lock",
            -- ...is equivelant to this:
            function(path)
                return path:sub(-5) == ".lock"
            end
        },
        -- New field
        binaries = true,
    },
}
```


alex-courtis:
Many thanks, I'll get to this on the weekend.

lua/nvim-tree/explorer/filters.lua
```diff
@@ -69,6 +69,30 @@ local function dotfile(path)
   return M.config.filter_dotfiles and utils.path_basename(path):sub(1, 1) == "."
 end
 
+---@param path string
+---@return boolean
+local function binary(path)
```
alex-courtis, line 74: I'm sorry, we can't include this OS specific functionality. It will also be a big performance overhead.

Fortunately we have `:help executable()` that is portable and fast. That is persisted on the Node.

lua/nvim-tree/keymap.lua
```diff
@@ -69,6 +69,7 @@ function M.default_on_attach(bufnr)
   vim.keymap.set('n', 'g?',      api.tree.toggle_help,                opts('Help'))
   vim.keymap.set('n', 'gy',      api.fs.copy.absolute_path,           opts('Copy Absolute Path'))
   vim.keymap.set('n', 'H',       api.tree.toggle_hidden_filter,       opts('Toggle Filter: Dotfiles'))
+  vim.keymap.set('n', 'h',       api.tree.toggle_binaries_filter,       opts('Toggle Filter: Binaries'))
```
alex-courtis, line 72: I'm sorry, this is too niche a case to warrant its own filter and mapping.

`filters.custom` will work nicely to cover the binary case.

lua/nvim-tree/explorer/filters.lua
```diff
@@ -87,8 +111,14 @@ local function custom(path)
   -- filter custom regexes
   local relpath = utils.path_relative(path, vim.loop.cwd())
   for pat, _ in pairs(M.ignore_list) do
-    if vim.fn.match(relpath, pat) ~= -1 or vim.fn.match(basename, pat) ~= -1 then
-      return true
+    if type(pat) == "function" then
+      if pat(path) then
```
alex-courtis, line 115: Let's give the user all the information here to allow creation of any sort of filter.

We can use `lib` `clone_node` to render the subtree to pass. That contains executable.

alex-courtis (changes requested):
Looking good!

Let's get this generic.

lua/nvim-tree/explorer/filters.lua
dxrcy: I did not know of this function, thanks! :)

lua/nvim-tree/keymap.lua
dxrcy: Reverted 👍

lua/nvim-tree/explorer/filters.lua
dxrcy: Is it possible to get the node from the path, at the stage the filter is ran?
I believe the explorer has not yet loaded, so `utils.get_node_from_path` returns `nil`.

I am not very familiar with this codebase, so apologies if I am misunderstanding something.

lua/nvim-tree/explorer/filters.lua
dxrcy: One downside to this option is that executable text files, such as shell scripts, will be hidden with this filter.

alex-courtis: I see. That is a good use case, however as you say it is a good case for a freely defined user function.

doc/nvim-tree-lua.txt
```diff
@@ -1241,7 +1241,7 @@ Enabling this is not useful as there is no means yet to persist bookmarks.
 Custom list of vim regex for file/directory names that will not be shown.
 Backslashes must be escaped e.g. "^\\.git". See |string-match|.
 Toggle via |nvim-tree-api.tree.toggle_custom_filter()|, default `U`
-  Type: {string}, Default: `{}`
+  Type: {string | function(path)}, Default: `{}`
```
alex-courtis, line 1244: Needs to be added to `ACCEPTED_STRINGS`

alex-courtis:
On adding to `ACCEPTED_TYPES` and setting
```lua
  filters = {
    custom = function()
      return true
    end,
```

```
Error detected while processing /tmp/nvt-dev/nvt-dev.lua:
E5113: Error while calling lua chunk: ...ck/packer/start/dxrcy/lua/nvim-tree/explorer/filters.lua:162: attempt to get length of local 'custom_filter' (a function value)
stack traceback:
        ...ck/packer/start/dxrcy/lua/nvim-tree/explorer/filters.lua:162: in function 'setup'
        .../pack/packer/start/dxrcy/lua/nvim-tree/explorer/init.lua:62: in function 'setup'
        /tmp/nvt-dev/site/pack/packer/start/dxrcy/lua/nvim-tree.lua:796: in function 'setup'
        /tmp/nvt-dev/nvt-dev.lua:85: in main chunk
```

lua/nvim-tree/explorer/filters.lua
alex-courtis: It will often be null; we just have to test for that and ignore it.

alex-courtis (changes requested):
Looking good.

Once we have the plumbing in to actually execute the function we should be OK.

dxrcy:
> On adding to `ACCEPTED_TYPES` and setting
> 
> ```lua
>   filters = {
>     custom = function()
>       return true
>     end,
> ```
> 
> ```
> Error detected while processing /tmp/nvt-dev/nvt-dev.lua:
> E5113: Error while calling lua chunk: ...ck/packer/start/dxrcy/lua/nvim-tree/explorer/filters.lua:162: attempt to get length of local 'custom_filter' (a function value)
> stack traceback:
>         ...ck/packer/start/dxrcy/lua/nvim-tree/explorer/filters.lua:162: in function 'setup'
>         .../pack/packer/start/dxrcy/lua/nvim-tree/explorer/init.lua:62: in function 'setup'
>         /tmp/nvt-dev/site/pack/packer/start/dxrcy/lua/nvim-tree.lua:796: in function 'setup'
>         /tmp/nvt-dev/nvt-dev.lua:85: in main chunk
> ```

To accept a function or a table of functions or strings, is the syntax `custom = { { "string", "function" }, "function" }` ?

alex-courtis:
> To accept a function or a table of functions or strings, is the syntax `custom = { { "string", "function" }, "function" }` ?

Unfortunately we don't have that level of control over types. This is the best we can do:

```lua
  filters = {
    custom = { "table", "function", },
  },
```

alex-courtis (changes requested):
Looking good:
```lua
    custom = function(path, node)
      if node then
        if node.nodes then
          log.raw("dev", "custom path=%s node.nodes=%d\n", path, #node.nodes)
        else
          log.raw("dev", "custom path=%s node.nodes=-1\n", path)
        end
      else
        log.raw("dev", "custom path=%s node=nil\n", path)
      end
    end,
```

Let's flatten out `for pat, _ in pairs(M.ignore_list) do` and we can test in anger.

Later: we'll also need to do some optimisation, passing `should_filter` a node rather than a path, as retrieving / generating nodes is expensive.

doc/nvim-tree-lua.txt
```diff
@@ -1241,7 +1241,7 @@ Enabling this is not useful as there is no means yet to persist bookmarks.
 Custom list of vim regex for file/directory names that will not be shown.
 Backslashes must be escaped e.g. "^\\.git". See |string-match|.
 Toggle via |nvim-tree-api.tree.toggle_custom_filter()|, default `U`
-  Type: {string | function(path)}, Default: `{}`
+  Type: {string | function(path, node)}, Default: `{}`
```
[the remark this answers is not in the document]
dxrcy: What about the case where `node` is null, though? :/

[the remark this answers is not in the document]
alex-courtis: Ah I see, sorry, I did not look properly at `populate_children`. It doesn't hydrate the node until after the filter.

Let's see how it looks like this. If necessary we can refactor `reload` and `populate_children` to remove the premature optimisation.

alex-courtis:
This has gotten complicated. I'm the one who insisted on node, which is not strictly needed for your use case.

I'm happy if you remove node and just pass path. It matches the existing filter pattern and does allow you to use `executable()`

dxrcy:
> This has gotten complicated. I'm the one who insisted on node, which is not strictly needed for your use case.
> 
> I'm happy if you remove node and just pass path. It matches the existing filter pattern and does allow you to use `executable()`

Thanks :-)

dxrcy:
> > To accept a function or a table of functions or strings, is the syntax `custom = { { "string", "function" }, "function" }` ?
> 
> Unfortunately we don't have that level of control over types. This is the best we can do:
> 
> ```lua
>   filters = {
>     custom = { "table", "function", },
>   },
> ```

I am getting an error when trying to set the accepted type to `{ "table", "function" }`, and settings `filters.custom` to a table with a string.
Trying to pass a `string` as the first argument of `vim.tbl_contains`.
The error goes away when changing it to `{ "function" }`, but I suppose this is not properly checking types.

alex-courtis:
> The error goes away when changing it to `{ "function" }`, but I suppose this is not properly checking types.

I see. It looks like we're not handling the table or something case.

I'll see what I can put together.

alex-courtis:
Raised #2665 

We can use 
```lua
  filters = {
    custom = { "function" },
  },
```
for now. The value will be validated as an arraylike table or function.

The only issue is the error message
```
[NvimTree]
Invalid option: filters.custom. Expected function, got number
```
That's too bad and should not hold up this change.

doc/nvim-tree-lua.txt
```diff
@@ -1241,7 +1241,7 @@ Enabling this is not useful as there is no means yet to persist bookmarks.
 Custom list of vim regex for file/directory names that will not be shown.
 Backslashes must be escaped e.g. "^\\.git". See |string-match|.
 Toggle via |nvim-tree-api.tree.toggle_custom_filter()|, default `U`
-  Type: {string}, Default: `{}`
+  Type: {string | function(path)} | function(path), Default: `{}`
```
alex-courtis, line 1244:
```suggestion
  Type: {string} | `function(absolute_path)`, Default: `{}`
```
We'll get to luals annotation but not today.

alex-courtis (changes requested):
Looks like it's working nicely for the initial case, however stops working after toggling.

Test case:

- directory containing A B C
- open tree
- toggle twice
- filter does not apply

```lua
---@param path string
---@return boolean
local function custom(path)
  log.line("dev", "filters.custom %s", path)
  if not M.config.filter_custom then
    log.line("dev", "filters.custom %s NOT PRESENT", path)
    return false
  end
```

```lua
  filters = {
    custom = function(absolute_path)
      log.line("dev", "user    custom %s", absolute_path)
      return absolute_path:match("B")
    end,
  },
```

Output:
```
[2024-02-05 16:39:33] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/A
[2024-02-05 16:39:33] [dev] user    custom /home/alex/src/nvim-tree/r/2655/A
[2024-02-05 16:39:33] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/B
[2024-02-05 16:39:33] [dev] user    custom /home/alex/src/nvim-tree/r/2655/B
[2024-02-05 16:39:33] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/C
[2024-02-05 16:39:33] [dev] user    custom /home/alex/src/nvim-tree/r/2655/C

[2024-02-05 16:39:41] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/A
[2024-02-05 16:39:41] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/A NOT PRESENT
[2024-02-05 16:39:41] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/B
[2024-02-05 16:39:41] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/B NOT PRESENT
[2024-02-05 16:39:41] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/C
[2024-02-05 16:39:41] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/C NOT PRESENT

[2024-02-05 16:39:46] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/A
[2024-02-05 16:39:46] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/B
[2024-02-05 16:39:46] [dev] filters.custom /home/alex/src/nvim-tree/r/2655/C
```

Looks like `M.config.filter_custom` is being mutated.

dxrcy:
> Looks like M.config.filter_custom is being mutated.

My bad. Should be fixed.

lua/nvim-tree/explorer/filters.lua
```diff
@@ -164,9 +164,8 @@ function M.setup(opts)
   M.exclude_list = opts.filters.exclude
 
   local custom_filter = opts.filters.custom
-
   if type(custom_filter) == "function" then
-    M.config.filter_custom = custom_filter
+    M.ignore_list[custom_filter] = true
```
alex-courtis, line 168: Should we be using ignore_list at all; can't we just use `M.config.custom_filter`  directly?

Unfortunately I'm out of time until the weekend. I'll give it a solid test then.

dxrcy: By `M.config.custom_filter`, I assume you mean `M.config.filter_custom`?

It looks like `M.config.filter_custom` is a boolean, for whether all custom filters are enabled.
Adding the single custom filter function to `M.ignore_list` would be consistent with adding each item of the custom filter table to it, wouldn't it? As `M.ignore_list` is a set, with each item checked in the loop in the `custom()`.

Perhaps I should set `M.ignore_list` to `opts.filters.custom` (a function), and check the type in `custom()`, if this is more intuitive or performant.

alex-courtis: > It looks like M.config.filter_custom is a boolean, for whether all custom filters are enabled.

Right you are, please accept my apologies.

> Perhaps I should set `M.ignore_list` to `opts.filters.custom` (a function), and check the type in `custom()`, if this is more intuitive or performant.

That is more intuitive, however it's not good practice to use a member with variable types.

Can we please go one further and explicitly separate, something like

```diff
diff --git a/lua/nvim-tree/explorer/filters.lua b/lua/nvim-tree/explorer/filters.lua
index f3beb61..24747c2 100644
--- a/lua/nvim-tree/explorer/filters.lua
+++ b/lua/nvim-tree/explorer/filters.lua
@@ -4,6 +4,7 @@ local marks = require "nvim-tree.marks"
 local M = {
   ignore_list = {},
   exclude_list = {},
+  custom_function = nil,
 }

 ---@param path string
@@ -84,18 +85,17 @@ local function custom(path)

   local basename = utils.path_basename(path)

+  -- filter by user's custom function
+  if M.custom_function and M.custom_function(path) then
+    return true
+  end
+
   -- filter custom regexes
   local relpath = utils.path_relative(path, vim.loop.cwd())
   for pat, _ in pairs(M.ignore_list) do
-    if type(pat) == "function" then
-      if pat(path) then
-        return true
-      end
-    else
       if vim.fn.match(relpath, pat) ~= -1 or vim.fn.match(basename, pat) ~= -1 then
         return true
       end
-    end
   end

   local idx = path:match ".+()%.[^.]+$"
@@ -160,7 +160,7 @@ function M.setup(opts)

   local custom_filter = opts.filters.custom
   if type(custom_filter) == "function" then
-    M.ignore_list[custom_filter] = true
+    M.custom_function = custom_filter
   else
     if custom_filter and #custom_filter > 0 then
       for _, filter_name in pairs(custom_filter) do
```

dxrcy: That looks like a better solution.
Though this won't let a user define a both a function and a pattern, but I don't think this is a problem, as a custom function can of course contain a pattern check.

```lua
custom = function(path)
    if vim.fn.match(path, ".*\\.lock") ~= -1 then -- Could be a specific function like substring
        return true
    end
    return is_binary(path)
end
```

Previously:

```lua
custom = {
    "*.lock",
    function(path) -- Or simply pass `is_binary`
        return is_binary(path)
    end,
}
```

dxrcy: Is it possible to go back to expecting a `table<function | string>`, instead of a `function | table<string>`?
A singular function would just be wrapped in a table: `custom = { is_binary }`.

If this is not preferable, I do not mind.

alex-courtis: > Though this won't let a user define a both a function and a pattern, but I don't think this is a problem, as a custom function can of course contain a pattern check.

> Is it possible to go back to expecting a table<function | string>, instead of a function | table<string>?
A singular function would just be wrapped in a table: custom = { is_binary }.

Sorry, I'd prefer we keep it in the style of the other similar code - one (fast) function only.

alex-courtis (approved):
Many thanks for all your great work and bearing with all the changes!

PR #2671, opened 2024-02-11
ci: triggers, nvim stable version & env vars

gegoune:
This PR updates CI workflow such that
- `push` is no longer triggering runs (we don't need to run it on master as it's guarded on PR level, release-please branch should also be triggered by `pull_request`)
- nvim version is no longer hardcoded but `stable` tag is used instead
- `nightly` is added to nvim versions matrix - this might prepare us for upcoming releases
- environment variables are set vie `env` key instead of inlining them

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,13 +2,6 @@ name: CI
 
 on:
   pull_request:
-    branches:
-      - '*'
-  push:
-    branches:
-      - master
-      - release-please--branches--master--components--nvim-tree
-
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


alex-courtis:
> * `push` is no longer triggering runs (we don't need to run it on master as it's guarded on PR level, release-please branch should also be triggered by `pull_request`)

Thank you... I made a real mess of that.

alex-courtis (approved):
This definitely works - many thanks!

Nightly should be interesting; as we get more and more of vim into luals we will get very early warnings of any API breakages. There was one between 0.8 and 0.9 with highlighting.

PR #2693, opened 2024-03-04
fix: searchcount exception on invalid search regex

xVermillionx:
Hi there, I encountered an error message generated by `vim.fn.searchcount` in `M.place_cursor_on_node` function when having `hijack_cursor` option enabled. 

I don't know if this fix is wanted or not, its just something that bothered me when using `vim.fn.searchcount` and I noticed it occurring when opening NvimTree.

Using `pcall` should fix the occurrence of the error message generated by `vim.fn.searchcount` as showing this error seems to be irrelevant to the `M.place_cursor_on_node` function:

```
E54: Unmatched (                                                                                                                                                    
Error detected while processing CursorMoved Autocommands for "NvimTree_*":                                                                                          
Error executing lua callback: Vim:E54: Unmatched (                                                                                                                  
stack traceback:                                                                                                                                                    
        [C]: in function 'searchcount'                                                                                                                              
        ...e/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:108: in function 'place_cursor_on_node'                                                         
        ...e/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:236: in function <...e/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:234>   
```

To reproduce I just search for `(` in my neovim, then open nvim-tree.
1. `/\v(`
2. `:NvimTreeOpen`

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -105,8 +105,8 @@ function M.open_on_directory()
 end
 
 function M.place_cursor_on_node()
-  local search = vim.fn.searchcount()
-  if search and search.exact_match == 1 then
+  local ok, search = pcall(vim.fn.searchcount())
+  if ok and search and search.exact_match == 1 then
     return
   end
 
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -105,7 +105,7 @@ function M.open_on_directory()
 end
 
 function M.place_cursor_on_node()
-  local ok, search = pcall(vim.fn.searchcount())
+  local ok, search = pcall(vim.fn.searchcount)
   if ok and search and search.exact_match == 1 then
     return
   end


geril2207:
I can reproduce, this fix seems reasonable

gegoune:
Doesn't wrapping in `pcall` just hide underlying issue? Perhaps more fundamental fix is needed instead?

geril2207:
Idea with `searchcount` was taken from lualine, and they also use `searchcount` with `pcall` wrap https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/components/searchcount.lua#L22

> Perhaps more fundamental fix is needed instead?

I think the problem here is error which interrupts the default behaviour, but the behaviour is correct(if we hide the error), isn't it? Are you about adding displaying info log or something like this or more complex changes?

alex-courtis:
> Idea with `searchcount` was taken from lualine, and they also use `searchcount` with `pcall` wrap https://github.com/nvim-lualine/lualine.nvim/blob/master/lua/lualine/components/searchcount.lua#L22

Looks like I made that change: https://github.com/nvim-lualine/lualine.nvim/pull/1088

> Doesn't wrapping in pcall just hide underlying issue? Perhaps more fundamental fix is needed instead?

I reckon this should be classed as a vim bug: the initial search throws the `E54: Unmatched (` exception which is correct. Having `vim.fn.searchcount` throw an error instead of returning the "zero" dictionary goes against the documentation and adds no value.

alex-courtis (approved):
Works nicely for `hijack_cursor`:
* valid, no match
* valid, match
* invalid

geril2207 (approved):

PR #2698, opened 2024-03-08
feat: add api.fs.copy.basename, default mapping ge

itsramiel:
Hi there 👋

I would like to propose a new keymap, which is copying basename of a file. I believe in most OOP and functional programming language it is very common to create a file for a class/function and name the file and the class/function the same, so the process goes like creating a `somelongNameThatIDontWantToTypeAgain.js` file and then wanting to create a function with the same name. Currently I use `y` and then manually delete the extension. I personally dont think I have ever used the `y` as it is and only to copy and then delete the extension. 

I hope you can see the value in having such a keymap. The mapping `l` was random. I just used what is still available and can change it to whatever you like.

Hope you find it useful too. I think it'll impact lots of users.


alex-courtis (changes requested):
That will be really handy. Works as expected.

We do need to change the mapping:
* users expect `l` to act as per default vim behaviour

alex-courtis:
Our copy mappings aren't great:

```
 Y              Copy Relative Path
 c              Copy
 gy             Copy Absolute Path
 y              Copy Name
```

I'm using
```
c              Copy                      
yn             Copy Name                
yr             Copy Relative Path      
yy             Copy Absolute Path
```

Unfortunately we can't change the defaults. `ge` might be the best we can do; it's somewhat consistent with `gy` and `e`.

itsramiel:
> Our copy mappings aren't great:
> 
> ```
>  Y              Copy Relative Path
>  c              Copy
>  gy             Copy Absolute Path
>  y              Copy Name
> ```
> 
> I'm using
> 
> ```
> c              Copy                      
> yn             Copy Name                
> yr             Copy Relative Path      
> yy             Copy Absolute Path
> ```
> 
> Unfortunately we can't change the defaults. `ge` might be the best we can do; it's somewhat consistent with `gy` and `e`.

Nice mappings. I think I'll use them.
Btw updated mapping

alex-courtis (changes requested):
Working, just needs style and help fixes.

First time contributors don't get automated CI ;)

itsramiel:
I believe it is ready

alex-courtis (approved):
Many thanks for all your work!

PR #2701, opened 2024-03-10
ci: set concurrency and trigger ci on master

gegoune:
This PR sets concurrency for all workflows and also runs ci workflow on master branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,13 @@ name: CI
 
 on:
   pull_request:
+  push:
+    branches: [master]
   workflow_dispatch:
+
+concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -2,8 +2,14 @@ name: CI
 
 on:
   pull_request:
+  push:
+    branches: [master]
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: true
```
alex-courtis, line 11: That's really handy...

alex-courtis (approved):
Concurrency is a good find; something that is desirable regardless.

Let's see how it behaves.

PR #2704, opened 2024-03-12
fix: bookmark filter should include parent directory

### LINKED ISSUE #2703: Bookmark filter should include parent directory
### Description

If we bookmark a file that is inside a directory, the bookmark filter should include the parent directory of that file

**Example:**
1. File `bookmark_me.md` are being marked
image

2. Toggle bookmark filter with `M`
image

### Neovim version

```text
NVIM v0.9.4
Build type: Release
LuaJIT 2.1.1692716794
```

Linux 5.15.146.1-microsoft-standard-WSL2

### Windows variant

WSL

### nvim-tree version

041dbd1

### Clean room replication

```lua
Don't need to change anything
```

### Steps to reproduce

Use latest commit, bookmark some file `inside a directory` then toggle bookmark filter with M

### Expected behavior

The filter should include the parent directory of files are being marked

### Actual behavior

_No response_

remvn:
this PR fixes #2703


lua/nvim-tree/explorer/filters.lua
```diff
@@ -73,7 +73,20 @@ end
...
 ---@param bookmarks table<string, boolean> absolute paths bookmarked
 local function bookmark(path, bookmarks)
-  return M.config.filter_no_bookmark and not bookmarks[path]
+  if not M.config.filter_no_bookmark then
+    return false
+  end
+
+  -- add trailing slash to make it match only mark's parent directory
+  -- not it's siblings
+  local dict = utils.path_add_trailing(path)
```
alex-courtis, line 82: `dict` does not mean anything.

You can mutate the argument e.g. `path = utils.path_add_trailing(path)`. That's an accepted lua idiom.

remvn:
```lua
  -- add trailing slash to make it match only mark's parent directory
  -- not it's siblings
  local parent = utils.path_add_trailing(path)
  for mark, _ in pairs(bookmarks) do
    if path == mark or vim.fn.stridx(mark, parent) == 0 then
      return false
    end
  end
```

I used `path` (unmodified) for matching with mark itself with 
```lua 
path == mark  -- stop here, no need to run match string
```

`dict` (now is `parent` for clear understand) is to avoid matching with mark's siblings 
for example:
```
/foo/bar/bookmark_me (marked)
/foo/bar/book  (this will show up)
```

We can simply avoid it by adding a trailing slash to `path`
```lua
local parent = utils.path_add_trailing(path)
```

Now mark's siblings can't match 
```
/foo/bar/bookmark_me (marked)
/foo/bar/book/
```

lua/nvim-tree/explorer/filters.lua
```diff
@@ -73,7 +73,20 @@ end
...
+  for mark, _ in pairs(bookmarks) do
+    if path == mark or utils.str_find(mark, dict) then
```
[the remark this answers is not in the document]
remvn: I replaced it with 
```lua
vim.fn.stridx(mark, parent) == 0
```

alex-courtis (approved):
Works nicely, many thanks for your contribution!

PR #2711, opened 2024-03-15
chore: make View.side compatible with float windows

John-Bush14:
So you don't have to do it yourself


alex-courtis:
Fantastic idea however I don't think we can do this:

1. This will break users who expect the existing behaviour e.g. users that dock to the right and float to the left
2. This can be achieved via API - in your own `view.float.open_win_config`

This would be a great addition to [Recipes](https://github.com/nvim-tree/nvim-tree.lua/wiki/Recipes)

Issue #2710, opened 2024-03-15
Mappings/functionality discoverability discussion


gegoune:
Since you touched on a topic of keybindings, currently it's hard to get a sense of what's possible. 

1. The help panel is hidden under "g?" keybinding by default, and only when nvim-tree updated its way to change keybindings, I discovered that there's a help panel (I was smashing ? and since it did nothing, I just assumed there's no help panel). 
2. As a user, you think everything useful (or even possible) is in this panel. 

So I think the ideal solution would be to:
1. Make it clear that g? is a default help panel keybinding. Show "g? - help" somewhere in nvim-tree panel or make it visible in readme.
2. Show every possible API command in this help panel, even if it has no keybinding + keybinding if there's one
3. Since there are lots of commands, it would be great to group them

This way, you can have minimal keybinding config and it will be extremely easy to copy-paste api function to your on_attach without reading :h nvim-tree

_Originally posted by @DeH4er in https://github.com/nvim-tree/nvim-tree.lua/issues/2706#issuecomment-1999552699_

gegoune:
Let's discuss it here, to keep linked PR clean.

I think issue here is of a documentation rather than anything else.
Being an Neovim plugin and having the same target audience I am pretty happy to assume certain level of IT-related knowledge, mainly ability to find and read documentation.

DeH4er:
It's true. My point is to make it easier to see all useful features in place to set/change a keybinding. Setting/changing a keybinding is probably one of the most common config changes people would make.

I doubt that someone will ever notice that they can use `toggle_enable_filters` or any other API feature which currently has no keybinding

And since there's already a help panel, why not make it better? 

Here's what comes to mind

```
nvim-tree mappings                   exit: q


Navigation
  not assigned  Next Sibling         api.node.navigate.sibling.next
  <             Previous Sibling     api.node.navigate.sibling.prev
...

Files
  d             Delete               api.fs.remove
...
```

gegoune:
I am not agains showing all available api calls in help window. If grouped together, as you showed above, it's actually quite nice and useful.

I would not really like any statement regarding existance of `g?` in nvim-tree's window, because next person will want an option to remove it (that person could as well be me) and we will have to yet another option. We can improve documentation/readme to make it more obvious though.

DeH4er:
[...]

PR #2712, opened 2024-03-15
feat: improve help

DeH4er:
Linking discussion https://github.com/nvim-tree/nvim-tree.lua/issues/2710

This is roughly what the help panel looks like. I didn't add all apis and I'd need some help here since I'm not aware of some functionality. 

Screenshot 2024-03-15 at 21 42 59

It shouldn't change drastically how users write `on_attach`. The only thing is that they don't need to pass eg. `nvim-tree: Open` in opts and it'll catch up if they use api.x.y function. If they pass a custom function as a keybinding, currently it'll not be displayed in help panel

### DIFF
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -41,6 +41,7 @@ local Api = {
   },
   commands = {},
   diagnostics = {},
+  meta = {},
 }
 
 --- Do nothing when setup not called.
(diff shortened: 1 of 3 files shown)


alex-courtis:
This is looking good - the layout is logical and discoverable.

I'm assuming that this will be another `s` mode, as we don't want to break users that expect the current layout. We can, however, add the actual api functions to the existing `s` modes.

alex-courtis:
The groupings are almost broken up as per the 10 API groups: `tree`, `fs`, `marks` etc.

In the interest of simplicity and maintainability we should use those groupings, being consistent with help etc.

That feeds into the next comment.

alex-courtis:
There is too much boilerplate code configuration here and it's not going to be very maintainable. We were in such a place before we completely moved to `on_attach` and it caused difficulties.

Rather than using boilerplate we can be lazy and do something like walking the `Api` tree. We can generate strings like `validate_options` does and map the functions themselves to the vim mappings. We could create a small table that maps `Api.xxx` to a human readable header.

Goals:
* no change to `api.lua`
* no change to `keymap.lua`
* default and user `on_attach` still look and function the same
* help and all its generation remains unchanged

I reckon this change would only need to hit `help.lua` and `utils.lua` for all the desired functionality and would maintain itself.

alex-courtis:
I'm still keen to get this one in, however it's been inactive for a while.

Please reopen when you'd like to continue @DeH4er

PR #2735, opened 2024-03-31
fix(#2733): escape trash path

### LINKED ISSUE #2733: Parent folder deleted when deleting a folder starting with $
### Description

If deleted folder starts with `$` then parent folder deleted instead!

In video trying delete `api/$crud` but whole `api` deleted instead!

https://github.com/nvim-tree/nvim-tree.lua/assets/74734701/7713f462-993f-4a23-be84-70c9d495787a

### Neovim version

```text
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1702233742
```

Linux 6.5.13-7-MANJARO x86_64

### Windows variant

_No response_

### nvim-tree version

ddd1d6e

### Clean room replication

```lua
my lazy config:

  {
    "nvim-tree/nvim-tree.lua",
    cmd = {
      "NvimTreeOpen",
      "NvimTreeClose",
      "NvimTreeToggle",
      "NvimTreeFindFile",
      "NvimTreeFindFileToggle",
    },
    keys = {
      { "<leader>e", "<cmd>lua require('nvim-tree.api').tree.toggle()<cr>", desc = "Tree" },
    },
    config = function()
      require("nvim-tree").setup()
    end
  }
```

### Steps to reproduce

```
mkdir api/\$crud
nvim
```

open tree and delete `$crud` inside `api` folder

### Expected behavior

should delete `$crud` folder

### Actual behavior

it is deleting whole `api` folder

alex-courtis:
fixes #2733


gegoune:
There were some changes to paths handling on Neovim master recently that might be relevant here, but I didn't follow them up very closely, just FYI.

alex-courtis:
> There were some changes to paths handling on Neovim master recently that might be relevant here, but I didn't follow them up very closely, just FYI.

Any hints as to what I could look for?

Rather than using fnameescape directly we could have vim.fn.jobstart escape it for us, something like this:
```lua
    local job = vim.fn.jobstart({ "gio", "trash", node.absolute_path }, {
```

That's how we escape paths for `vim.fn.system`, `vim.loop.spawn` etc.

Unfortunately we would need to find a way to split `config.trash.cmd` into a list.

gegoune:
> Any hints as to what I could look for?

I think those are the two that I meant:
https://github.com/neovim/neovim/pull/28105
https://github.com/neovim/neovim/pull/28064

alex-courtis:
> I think those are the two that I meant: [neovim/neovim#28105](https://github.com/neovim/neovim/pull/28105) [neovim/neovim#28064](https://github.com/neovim/neovim/pull/28064)

Today I learned about vim.fs. Guessing it's recently added.

We're not using it but should probably migrate some of our `fnamemodify` calls and utils like `path_basename`

alex-courtis:
On further review it seems we should use the fit for purpose `shellescape` rather than `fnameescape`.

PR #2738, opened 2024-04-02
fix(#2774): add winfixbuf view option for nvim 0.10

petobens:
Hi @alex-courtis I've only added the option for the View class. Dunno if this is useful/needed elsewhere. Feel free to modify the PR as you deem fit.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -572,6 +572,10 @@ function M.setup(opts)
   end
 
   M.View.initial_width = get_width()
+
+  if vim.fn.has("nvim-0.10") == 1 then
+    M.View.winopts.winfixbuf = true
+  end
 end
 
 return M


alex-courtis (changes requested):
Looking good; this should greatly simplify things.

We are going to have to be careful: 0.10 is not stable and there will be unforseen consequences, due to existing assumptions and workarounds.

Please:
* Put this behind an experiment opt in "feature flag": `:h nvim-tree-opts-experimental`
* Disable `_prevent_buffer_override` and all the other workarounds in place that currently achieve winfixbuf functionality

Update help etc. etc.

alex-courtis:
@petobens we're officially supporting release 0.10 now.

Can you please continue with this PR with the experimental guards in place?

petobens:
Hi @alex-courtis, unfortunately I won't be able to followup this in the short term.  Feel free to continue it if you deem fit (otherwise I'll revisit it in the next months).

alex-courtis:
> Hi @alex-courtis, unfortunately I won't be able to followup this in the short term. Feel free to continue it if you deem fit (otherwise I'll revisit it in the next months).

No worries at all. Your contributions are welcome whenever you have the time.

kutsan:
What will be the status of the feature now that the issue is closed?

alex-courtis:
> What will be the status of the feature now that the issue is closed?

Still open. Waiting on a kind contributor.

emmanueltouzery:
Otherwise from a user point of view, nvim-tree support is not really needed. I just have this in my config and it works great:
```lua
-- winfixbuf
if vim.fn.has("nvim-0.10") == 1 then
    callback = function()
      if vim.list_contains({"NvimTree", "OverseerList", "aerial", "toggleterm", "dbui", "dbout"}, vim.bo.filetype) then
        vim.wo.winfixbuf = true
      end
    end,
  })
end
```

But I guess it would be handier if nvim-tree had it built-in.

PR #2774, opened 2024-05-13
fix: nil access exception with git integration when changing branches

jacobkania:
This PR fixes an infrequent (once every couple days), but annoying, nil access exception.

Occasionally, when switching git branches, to where some open files no longer existed, nvim-tree would raise a nil access exception, which would disable git integration until nvim was restarted. The stack trace brought me to this line.

I made this change a few weeks ago and have stopped seeing the error entirely, with no other changes to my local config. Seemed like a small enough fix to just contribute without any further research, as it's helped me on my local setup! Though, if you would prefer to dig into a root cause of why `status` was being passed as nil here, I would not be offended 😄

### DIFF
--- a/lua/nvim-tree/explorer/node.lua
+++ b/lua/nvim-tree/explorer/node.lua
@@ -29,7 +29,7 @@ end
 ---@param absolute_path string
 ---@return GitStatus
 local function get_git_status(parent_ignored, status, absolute_path)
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
> Though, if you would prefer to dig into a root cause of why `status` was being passed as nil here, I would not be offended

That would be difficult; the code paths to get there defy static analysis.

Nil check works nicely, many thanks for your contribution!

PR #2779, opened 2024-05-21
feat: allow changing root to dir

sarmong:


sarmong:
Also, is there a reason why it doesn't allow changing directory if it's under cwd? I kinda want to remove it

https://github.com/nvim-tree/nvim-tree.lua/blob/385514aabe526c8a5408b494735004dbf4e1cf4f/lua/nvim-tree.lua#L56-L59

gegoune:
Could you please describe what problem does it solve?

sarmong:
Changing root to dir? Well, first of all, I think it simply makes sense and nowhere in the docs does it say that only providing filename is allowed. I am using this api method directly and want to set nvim-tree root to a directory.

Regarding setting path under cwd - my problem is following. I made my own command `:Cd` that sets project root based on custom heuristics. Example: my project root is in `/foo/bar`. Current directory is `/foo/bar/baz`. In nvim-tree I navigate to check something upwards and end up in `/foo`. Now I want to call my command to reset root to project root everywhere. Buy I can't for nvim-tree.

Yes, I know I can use `sync_root_with_cwd`, but it works on DirChanged. But it might happen so that my vim cwd is already `/foo/bar`, but I still want to call `:Cd` to reset root everywhere. However, not for nvim-tree.

alex-courtis:
I have a similar use case: `:CD` changes to the initial directory however I do use `sync_root_with_cwd`.

I like the change, however we can't make it at this point: it will break open, find_file etc. which pass a file path.

Could you use the `nvim-tree-api.tree.change_root()` API? We could enhance it if we need to.

sarmong:
I am not sure how it would break the existing API. It calls to `fnamemodify` and it will return dirname whether path is a dir or a file

https://github.com/sarmong/nvim-tree.lua/blob/e60118db9a072bb5cfb9353fab1d63f1a819848b/lua/nvim-tree.lua#L75-L75

alex-courtis:
> Could you use the nvim-tree-api.tree.change_root() API? We could enhance it if we need to.

Firstly, please accept my apologies; I was confused and thought this method was private, not API.

> I am not sure how it would break the existing API. It calls to `fnamemodify` and it will return dirname no matter if path is a dir or a file

Sorry mate, I'm not willing to take the risk. This area functionality is very fragile and has broken frequently in the past.

I'm still not quite sure what the issue is; I'm not experiencing any issues changing to a directory.

Please provide a replicator and instructions on how to reproduce problem behaviour as a [bug report](https://github.com/nvim-tree/nvim-tree.lua/issues/new?assignees=&labels=bug&projects=&template=bug_report.yml)

alex-courtis:
I'm still keen to get this one in, however it's been inactive for a while.

Please reopen when you'd like to continue @sarmong

Issue #2731, opened 2024-03-30
0.10 luadoc and lsp integration


alex-courtis:
0.10 adds full luadoc and function definitions for `vim.*`, lsp, treesitter etc.

Add these to luals:
```json
    "library": [
      "$VIMRUNTIME/lua",
      "${3rd}/luv/library"
    ]
```

Resolve luals detected issues:
* type mismatches - no changes, just doc
* missing nil checks - fix
* incorrect returns/params - fix

Update documentation and function signatures to be properly typed.

PR #2781, opened 2024-05-24
refactor: replace deprecated use of vim.diagnostic.is_disabled() 

ZacharyRizer:
According to [Neovims Release news](https://neovim.io/doc/user/deprecated.html) in the newly released v.0.10 `vim.diagnostic.is_disabled()` is deprecated in favor of `vim.diagnostic.is_enabled()`. 

I wasn't sure if there should be backwards compatibility left in for v0.9 or not. I'm not very solid with Lua, so I don't know if that `if/else` makes sense to 'solve' backwards compatibility.

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -39,12 +39,14 @@ end
 local function from_nvim_lsp()
   local buffer_severity = {}
 
-  local is_disabled = false
-  if vim.fn.has "nvim-0.9" == 1 then
-    is_disabled = vim.diagnostic.is_disabled()
+  local is_enabled = false
+  if vim.fn.has "nvim-0.10" == 1 then
+    is_enabled = vim.diagnostic.is_enabled()
+  else
+    is_enabled = not vim.diagnostic.is_disabled()
   end
 
-  if not is_disabled then
+  if is_enabled then
     for _, diagnostic in ipairs(vim.diagnostic.get(nil, { severity = M.severity })) do
       local buf = diagnostic.bufnr
       if vim.api.nvim_buf_is_valid(buf) then
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -39,7 +39,7 @@ end
 local function from_nvim_lsp()
   local buffer_severity = {}
 
-  local is_enabled = false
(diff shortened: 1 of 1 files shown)


alex-courtis:
> I wasn't sure if there should be backwards compatibility left in for v0.9 or not. I'm not very solid with Lua, so I don't know if that `if/else` makes sense to 'solve' backwards compatibility.

Thank you, we do need to retain compatibility.

Those tests work - has 0.10 will still be true for 0.11...

alex-courtis (approved):

Now that 0.10 is released we can start on #2731 - any assistance on that would be most gratefully appreciated. That will allow us to pick up on these sorts of issues.

AntonC9018:
This gives me an error that `is_enabled` is not defined.
```
Error executing vim.schedule lua callback: ...re/nvim/lazy/nvim-tree.lua/lua/nvim-tree/diagnostics.lua:44: attempt to call field 'is_enabled' (a nil value)
stack traceback:
	...re/nvim/lazy/nvim-tree.lua/lua/nvim-tree/diagnostics.lua:44: in function 'from_nvim_lsp'
	...re/nvim/lazy/nvim-tree.lua/lua/nvim-tree/diagnostics.lua:158: in function 'callback'
	...al/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree/utils.lua:422: in function <...al/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree/utils.lua:421>
```
```
❯ nvim --version
NVIM v0.10.0-dev-2499+g0d553c834
Build type: RelWithDebInfo
LuaJIT 2.1.1707061634
```

alex-courtis:
Let's revert this one. It looks like it's not present in older 0.10 builds, which people do actively use.

alex-courtis:
Reverted in #2784

gegoune:
We need some other implementation of that check then as it affects users on new enough versions.

alex-courtis:
> We need some other implementation of that check then as it affects users on new enough versions.

Yes. Now that stable is 0.10 we should add 0.9 back to the matrix.

The 0.10 dev versions are a problem; API broke many times.

alex-courtis:
Fortunately, API levels have been implemented and are available since 0.9

#2787

We'll need to increase the minimum supported version to 0.9.

Issue #2789, opened 2024-06-01
Allow nodes to expand until a certain condition is met.


ghostbuster91:
**Can this functionality be implemented utilising API?**
To some degree, see my [PR](https://github.com/nvim-tree/nvim-tree.lua/pull/2790).

**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I work with java/scala projects. In these languages directory structure should reflect the namespace hierarchy which usually results in multiple nested directories with only a single child directory.

E.g. a namespace `com.example.something` will have the following directory structure:
```
modules/src
└── main
    └── com
        └── example
            └── something
                ├── bar.txt
                └── foo.txt
```

**Describe alternatives you've considered**

I have two options if I want to explore this structure and get to the bottom of it:
1. expand nodes manually one by one until I reach the bottom
2. use `expand_all` function but this has following downsides:
   - when I don't know the directory structure it might result in a lot of nodes being opened unnecessarily, which I will have to close later
   - when I do know the directory structure and I know that there are some additional directories I might still use `expand_all` knowing that I will have to close some of them
3. use `group_empty` feature. I prefer to have a tree like structure. One reason for that is that it is more horizontally compact that the same structure rendered with `group_empty` enabled. 

**Describe the solution you'd like**

I would like to have an option that allows me to recursively expand all directories as long as there is only one children that is also a directory. If a directory that has more children is encountered the expansion should be stopped (after expanding that directory).

This can be implemented in more generic way - expand until a certain, user provided, condition is met.

PR #2790, opened 2024-06-01
feat: Allow to expand until certain condition is met

ghostbuster91:
This is my attempt to solve #2789 

With this change I can now define the following mapping:
```lua
local function stop_expansion(_, node)
    return false, stop_expansion
end
local function expand_until_non_single(count, node)
    local cwd = node.link_to or node.absolute_path
    local handle = vim.loop.fs_scandir(cwd)
    if not handle then
        return false, stop_expansion
    end
    local status = git.load_project_status(cwd)
    populate_children(handle, cwd, node, status)
    local child_folder_only = explorer_node.has_one_child_folder(node) and node.nodes[1]
    if count > 1 and not child_folder_only then
        return true, stop_expansion
    elseif child_folder_only then
        return true, expand_until_non_single
    else
        return false, stop_expansion
    end
end
map('n', 'E', function()
    api.tree.expand_all(nil, expand_until_non_single)
end, opts("Expand until not single"))
```

This works as expected however it has some downsides:
1. the api is very flexible but at the same time quite complex. I had to do it in this way because the `expand_all` function will not expand the last directory. It stops as soon as `should_expand` returns `false` but I wanted the last directory to be expanded.
2. ~I had to copy `populate_children` method which is not a part of public api~ replaced with `explorer.explore(node, status)`

Btw if this got accepted in some form I am going to replace the default expand behavior in my config with this, since I find it so useful.

### DIFF
--- a/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
+++ b/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
@@ -39,8 +39,7 @@ end
 local function expand_until_max_or_empty(expansion_count, node)
   local should_halt = expansion_count >= M.MAX_FOLDER_DISCOVERY
   local should_exclude = M.EXCLUDE[node.name]
-  local result = not should_halt and node.nodes and not node.open and not should_exclude
-  return result
+  return not should_halt and node.nodes and not node.open and not should_exclude
 end
 
 local function gen_iterator(should_expand_fn)
@@ -74,8 +73,9 @@ local function gen_iterator(should_expand_fn)
 end
 
 ---@param base_node table
-function M.fn(base_node, expand_until)
-  expand_until = expand_until or expand_until_max_or_empty
+---@param expand_opts ApiTreeExpandAllOpts|nil
+function M.fn(base_node, expand_opts)
+  local expand_until = (expand_opts and expand_opts.expand_until) or expand_until_max_or_empty
   local node = base_node.nodes and base_node or core.get_explorer()
   if gen_iterator(expand_until)(node) then
     notify.warn("expansion iteration was halted after " .. M.MAX_FOLDER_DISCOVERY .. " discovered folders")
--- a/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
+++ b/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
@@ -27,12 +27,12 @@ local function populate_node(node)
   local cwd = node.link_to or node.absolute_path
   local handle = vim.loop.fs_scandir(cwd)
   if not handle then
-      return
+    return
   end
   local status = git.load_project_status(cwd)
 
   if #node.nodes == 0 then
-      core.get_explorer():expand(node, status)
+    core.get_explorer():expand(node, status)
   end
 end
 
--- a/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
+++ b/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
@@ -40,10 +40,6 @@ local function gen_iterator(should_expand_fn)
   local expansion_count = 0
 
   return function(parent)
-    -- if parent.parent and parent.nodes and not parent.open then
-    --   expansion_count = expansion_count + 1
-    --   expand(parent)
-    -- end
     expand(parent)
 
     Iterator.builder({parent})
@@ -55,14 +51,8 @@ local function gen_iterator(should_expand_fn)
           expansion_count = expansion_count + 1
           expand(node)
         end
-        -- print("iterator " .. node.name  .. " " .. vim.inspect(should_expand))
       end)
       :recursor(function(node)
-        -- print(vim.inspect(expansion_count < M.MAX_FOLDER_DISCOVERY))
-        -- print(vim.inspect(node.group_next and { node.group_next }))
-        -- print(vim.inspect((node.open and node.nodes)))
-
-        -- print("recursor " .. node.name .. " " .. vim.inspect(expansion_count < M.MAX_FOLDER_DISCOVERY and (node.group_next and { node.group_next } or (node.open and node.nodes))))
         return expansion_count < M.MAX_FOLDER_DISCOVERY and (node.group_next and { node.group_next } or (node.open and node.nodes))
       end)
       :iterate()
--- a/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
+++ b/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
@@ -24,14 +24,13 @@ local function populate_node(node)
   if node.nodes == nil then
     return
   end
-  local cwd = node.link_to or node.absolute_path
-  local handle = vim.loop.fs_scandir(cwd)
-  if not handle then
-    return
-  end
-  local status = git.load_project_status(cwd)
-
   if #node.nodes == 0 then
+    local cwd = node.link_to or node.absolute_path
+    local handle = vim.loop.fs_scandir(cwd)
+    if not handle then
+      return
+    end
+    local status = git.load_project_status(cwd)
     core.get_explorer():expand(node, status)
   end
 end
@@ -51,6 +50,7 @@ local function gen_iterator(should_expand_fn)
   local expansion_count = 0
 
   return function(parent)
+    populate_node(parent)
     parent.open = true
 
     Iterator.builder({ parent })
@@ -60,6 +60,7 @@ local function gen_iterator(should_expand_fn)
         should_expand_fn = should_expand_next_fn
         if should_expand then
           expansion_count = expansion_count + 1
+          populate_node(node)
           node = lib.get_last_group_node(node)
           node.open = true
         end
--- a/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
+++ b/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
@@ -39,12 +39,14 @@ end
 ---@param node Node
 ---@param populate_node function
 ---@return boolean, function
+-- luacheck: push ignore populate_node
 local function expand_until_max_or_empty(expansion_count, node, populate_node)
   local should_halt = expansion_count >= M.MAX_FOLDER_DISCOVERY
   local should_exclude = M.EXCLUDE[node.name]
   local result = not should_halt and node.nodes and not node.open and not should_exclude
   return result, expand_until_max_or_empty
 end
+-- luacheck: pop
 
 local function gen_iterator(should_expand_fn)
   local expansion_count = 0
--- a/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
+++ b/lua/nvim-tree/actions/tree/modifiers/expand-all.lua
@@ -44,19 +44,24 @@ end
 
 local function gen_iterator(should_expand)
   local expansion_count = 0
+  local function expand(node)
+    populate_node(node)
+    node = lib.get_last_group_node(node)
+    node.open = true
+  end
 
   return function(parent)
-    populate_node(parent)
-    parent.open = true
+    if parent.parent and parent.nodes and not parent.open then
+      expansion_count = expansion_count + 1
+      expand(parent)
+    end
 
-    Iterator.builder({ parent })
+    Iterator.builder(parent.nodes)
       :hidden()
       :applier(function(node)
         if should_expand(expansion_count, node, populate_node) then
           expansion_count = expansion_count + 1
-          populate_node(node)
-          node = lib.get_last_group_node(node)
-          node.open = true
+          expand(node)
         end
       end)
       :recursor(function(node)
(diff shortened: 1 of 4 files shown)


alex-courtis:
I love the concept and it is something we should do. Tried your example after a lot of hacking and it looks to work nicely.

- [ ] Does nothing when called from the root node
- [ ] Needs documentation

Use of private functions in your example aside, this is far too complex and not approachable for users. Having to populate the children is the key problem.

Suggestion: come up with a simple use case and example - one that would go in the documentation. We can start simplifying from there.

Perhaps I'm misunderstanding: why does this need to be an iterator? I was imagining the user would supply a simple boolean returning function to test against a node.

alex-courtis:
Any updates on this one @ghostbuster91 ?

ghostbuster91:
Sorry, I was quite busy. I am going to take a look at it. 

> I love the concept and it is something we should do.

I am glad to hear that!

> Perhaps I'm misunderstanding: why does this need to be an iterator? I was imagining the user would supply a simple boolean returning function to test against a node.

Because the simple `fn: node => boolean` is insufficient in my case. Take a look at the following example:

```
a
└── b
    └── c
        ├── c2
        │   └── d2 (file)
        └── d (file)
```

When expanding `a` I would like to expand `b` and `c`. 
However `c` has multiple children and also directories inside so it wouldn't pass a simple `should_expand` function. 

In my case the condition is following:
Keep expanding until you reach a node that has more than one children but expand this node as well. In other words the expansion is inclusive. 

I suppose this could be expressed in simpler way for example by using a flag - `inclusive=true` on the user side. 

> come up with a simple use case and example - one that would go in the documentation. 

Sure, let's conclude first what to do with the optionally inclusive expansion.

ghostbuster91:
In the end I was able to simply replace the iterator with accessing node's siblings via `node.parent.nodes`. With this the user side function for my particular use case looks as follow:
```lua
local function expand_until_non_single(count, node, populate_node)
    populate_node(node)
    if node.nodes == nil or not node.parent.open then
        return false
    end
    local has_one_child_folder = explorer_node.has_one_child_folder(node)
    local has_no_siblings = explorer_node.has_one_child_folder(node.parent)
    if has_one_child_folder and count == 0 then
        return true
    elseif has_no_siblings then
        return true
    else
        return false
    end
end
```

The CI keeps failing with following error:
```
{
    "file:///home/runner/work/nvim-tree.lua/nvim-tree.lua/lua/nvim-tree/actions/tree/modifiers/expand-all.lua": [
        {
            "code": "redundant-parameter",
            "message": "This function expects a maximum of 2 argument(s) but instead it is receiving 3.",
            "range": {
                "end": {
                    "character": 43,
                    "line": 33
                },
                "start": {
                    "character": 37,
                    "line": 33
                }
            },
            "severity": 2,
            "source": "Lua Diagnostics."
        }
    ]
}
```
Any idea how to fix it?

lua/nvim-tree/explorer/node.lua
```diff
@@ -36,7 +36,7 @@ end
 ---@param node Node
 ---@return boolean
 function M.has_one_child_folder(node)
-  return #node.nodes == 1 and node.nodes[1].nodes and vim.loop.fs_access(node.nodes[1].absolute_path, "R") or false
+  return node.nodes ~= nil and #node.nodes == 1 and node.nodes[1].nodes and vim.loop.fs_access(node.nodes[1].absolute_path, "R") or false
```
ghostbuster91, line 39: **thought**: guarding against calling this function with a file node

alex-courtis:
This minimal setup worked nicely:

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

local api = require("nvim-tree.api")
local lib = require("nvim-tree.lib")
local explorer_node = require("nvim-tree.explorer.node")

local function expand_until_non_single(count, node, populate_node)
  populate_node(node)
  if node.nodes == nil or not node.parent.open then
    return false
  end
  local has_one_child_folder = explorer_node.has_one_child_folder(node)
  local has_no_siblings = explorer_node.has_one_child_folder(node.parent)
  if has_one_child_folder and count == 0 then
    return true
  elseif has_no_siblings then
    return true
  else
    return false
  end
end

local function expand_until()
  local node = api.tree.get_node_under_cursor()
  if node.open then
    lib.expand_or_collapse(node, nil)
  else
    if node.nodes then
      api.tree.expand_all(node, expand_until_non_single)
    else
      api.node.open.edit(node)
    end
  end
end

  local function opts(desc)
    return { desc = "nvim-tree: " .. desc, buffer = bufnr, noremap = true, silent = true, nowait = true }
  end

  api.config.mappings.default_on_attach(bufnr)

  vim.keymap.set("n", "O", expand_until, opts("Expand until"))
end

_G.setup = function()
  require("nvim-tree").setup({
    on_attach = on_attach,
  })
end
```

alex-courtis:
Suggestions:

```lua
lib.expand_or_collapse(node, nil)
```

Replace with `node.open.edit` after testing for children i.e. `node.nodes`

```lua
  local has_one_child_folder = explorer_node.has_one_child_folder(node)
  local has_no_siblings = explorer_node.has_one_child_folder(node.parent)
```

Write your own test in this case. It will be a lot simpler as you have tested most of the conditions already.

alex-courtis:
> Any idea how to fix it?



status is not a parameter for `function Explorer:expand(node)`, looks like it's unused.

alex-courtis:
That parameter is not used and thus can be removed.

lua/nvim-tree/actions/tree/modifiers/expand-all.lua
```diff
@@ -64,9 +79,10 @@ local function gen_iterator()
 end
 
 ---@param base_node table
-function M.fn(base_node)
+function M.fn(base_node, expand_until)
```
alex-courtis, line 82: Please use an options table similar to `api.tree.resize`:
* add a class to `api.lua` to describe this
* docuement this at `*nvim-tree-api.tree.expand_all()*`

lua/nvim-tree/actions/tree/modifiers/expand-all.lua
```diff
@@ -3,6 +3,7 @@ local renderer = require "nvim-tree.renderer"
 local Iterator = require "nvim-tree.iterators.node-iterator"
 local notify = require "nvim-tree.notify"
 local lib = require "nvim-tree.lib"
+local git = require "nvim-tree.git"
```
alex-courtis, line 6: Appears unused.

lua/nvim-tree/actions/tree/modifiers/expand-all.lua
```diff
@@ -18,42 +19,56 @@ local function to_lookup_table(list)
...
-    Iterator.builder(parent.nodes)
+    Iterator.builder({ parent })
       :hidden()
       :applier(function(node)
-        if should_expand(expansion_count, node) then
+        local should_expand = should_expand_fn(expansion_count, node, populate_node)
+        if should_expand then
           expansion_count = expansion_count + 1
-          expand(node)
+          populate_node(node)
```
alex-courtis, line 64: I'm not quite following the logic here, but it looks like we're not actually making many changes e.g.
* `expand` was renamed to `populate_node`
* unnecessary `should_expand` boolean was extracted
* `get_last_group_node` was moved out of `expand`

Please can you refactor this down to the minimum number of changes necessary. Integration with the existing structure will result in simpler code that is more easily reviewable and maintainable.

I appreciate that we've been through many revisions to get to this point, with complexity from earlier commits removed, however we need to revert some structural changes. Perhaps consider how you'd make these changes from scratch, knowing what you have learned over the course of this PR.

alex-courtis (changes requested):
This is looking good, we'll get there:

- [x] simplify `expand-all.lua` preserving existing structure
- [x] remove unused variables etc. as per luals warnings
- [x] remove any usages of internals like `lib`
- [x] add options to `api.tree.expand_all`, documenting in help and api.lua
- [x] add an example to help, the absolute bare minimum, try to keep to 10 lines. If not, we can add to the wiki.

lua/nvim-tree/actions/tree/modifiers/expand-all.lua
ghostbuster91: it is used in the line https://github.com/nvim-tree/nvim-tree.lua/pull/2790/files#diff-b9bfb789bf93265f2e073fa63bcb9783b86a39e42c47b42b1e7193f87a03b489R33

ghostbuster91:
> 
> 
> That parameter is not used and thus can be removed.

Right, I always forget that this is possible in dynamically-typed languages :+1:

lua/nvim-tree/actions/tree/modifiers/expand-all.lua
ghostbuster91: > Please can you refactor this down to the minimum number of changes necessary. Integration with the existing structure will result in simpler code that is more easily reviewable and maintainable.

I totally agree with such approach. I think I was able to cut down amount of changes to the bare minimum.

There is one thing that I would like to discuss - changes in the `recursor` function. Before, when the logic was only to expand everything recursively, the condition was to check if `node.open==true` (putting `expansion_count aside`). In my case that is invalid as some nodes have been opened but they shouldn't be traversed further. 

In both cases reusing `should_expand` function results in correct behavior. First of all, this doesn't have to be true in general. Maybe sometimes the logic for `should_recurse` will be different than the logic for `should_expand`. 

Second. So, for each node we call `should_expand` twice, once to check if we should expand and second time to check if we should recurse. Between these calls, if should `should_expand` returns `true`, we call `expand` that changes the state of a given node. This means that users *must not* relay on `node.open` in their `should_expand` implementation as this would result in incorrect behavior. 

I don't have enough experience in developing nvim plugins to judge if this is a problem, but wanted to bring this to your attention. Maybe this should at least be mentioned in the documentation?

ghostbuster91:
> > Any idea how to fix it?
> 
> 
> 
> status is not a parameter for `function Explorer:expand(node)`, looks like it's unused.

It must have changed once I rebased on latest master. Fixed :+1:

ghostbuster91:
With all these changes my custom should expand function looks as follow:
```lua
local function has_one_child_folder(node)
    return #node.nodes == 1 and node.nodes[1].nodes and
        vim.loop.fs_access(node.nodes[1].absolute_path, "R") or false
end

local function expand_until_non_single(_, node, populate_node)
    populate_node(node)
    if node.nodes == nil or not node.parent.open then
        return false
    end
    return has_one_child_folder(node.parent)
end
```

Thank you for helping me getting this PR across the finish line. I truly appreciate your time and attitude :bowing_man:

alex-courtis:
> > status is not a parameter for `function Explorer:expand(node)`, looks like it's unused.
> 
> It must have changed once I rebased on latest master. Fixed 👍

Sorry about that; we are doing some significant refactoring of explorer/reload etc.

lua/nvim-tree/actions/tree/modifiers/expand-all.lua
```diff
@@ -47,13 +59,14 @@ local function gen_iterator()
     Iterator.builder(parent.nodes)
       :hidden()
       :applier(function(node)
-        if should_expand(expansion_count, node) then
+        if should_expand(expansion_count, node, populate_node) then
```
alex-courtis, line 62: This appears unused. Can we remove it?

alex-courtis (changes requested):
This is looking incredible - the sample gives the user exactly what they need without overwhelming them.

I hope you don't mind: in the interest of speed I've pushed a couple of doc / example tweaks.

- [ ] Issue: Sibling directories are not expanded:

```sh
mkdir -p foo1/bar/baz/foo/bar/baz
mkdir -p foo2/bar/baz/foo/bar/baz
```

`E` at root only expands one level deep:


`Z` at root expands `foo1` correctly but does not expand `foo1`:

lua/nvim-tree/actions/tree/modifiers/expand-all.lua
alex-courtis: > In both cases reusing should_expand function results in correct behavior. First of all, this doesn't have to be true in general. Maybe sometimes the logic for should_recurse will be different than the logic for should_expand.

I like the consistency, it's more reason-able and will make maintenance easier.

> This means that users _must not_ relay on `node.open` in their `should_expand` implementation as this would result in incorrect behavior.

So long as we document that we are providing a consistent user experience.

Is `node.nodes` correct; can they test that?

ghostbuster91:
@alex-courtis just fyi: I didn't forget about this. Just trying to finish my new apartment, so I can finally move in. I will try to finish this soon.

alex-courtis:
Closing as this PR is stale. Please reopen if you wish to continue this work.

Issue #2803, opened 2024-06-13
The path to open files is always displayed as an absolute path


epheien:
### Description

The path to open files is always displayed as an absolute path.
NERDTree work fine.

### Neovim version

```text
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info
```

### Operating system and version

macOS 12.7.5

### nvim-tree version

2086e56

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end
```

### Steps to reproduce

1. nvim -u nvt-min.lua
2. cd ~/.vim
3. :NvimTreeOpen
4. select `init.vim` and press `<CR>`
5. `<C-g>`

### Expected behavior

`"init.vim" xxxx lines --0%--` message in cmdline

### Actual behavior

`"~/.vim/init.vim" xxxx lines --0%--` message in cmdline

CcccX2017:
I have the same problem with nvim 0.9.5 and lunarvim 1.4-release
[...]

PR #2805, opened 2024-06-16
feat(#2127): add experimental.actions.open_file.relative_path to open files with a relative path rather than absolute

### LINKED ISSUE #2127: Open files with relative paths
Opening files uses their absolute paths which makes them very difficult to read. Example: the file I am currently editing has a path that is nearly 160 characters long, which is a little much for the statusline of an 80-column terminal.

(An option to?) open files by their relative paths when possible, falling back to the absolute path of files which are not descendants of the current working directory (relative paths technically also work here, but IMO, `../../../../x/y/z` is probably not ideal...).

**Additional context**
This was originally added in #487 and reverted in #524 after it caused a number of bugs. I unfortunately do not have the time to dedicate to this right now, but I suspect these bugs are likely relatively trivial to fix.

epheien:
fixes #2127

Try to fix #2127 and #2803

add `experimental.open.relative_path` option

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -621,8 +621,10 @@ Following is the default configuration. See |nvim-tree-opts| for details.
         },
       },
       experimental = {
-        open = {
-          relative_path = false,
+        actions = {
+          open_file = {
+            relative_path = false,
+          },
         },
       },
       log = {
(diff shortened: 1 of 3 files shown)


alex-courtis (approved):
Apologies for the delay. This works beautifully; I'll be turning this on today and watching `:ls`

alex-courtis:
Added #2819 

I'd be most grateful if you watched that pinned issue @epheien

PR #2811, opened 2024-06-24
feat(#2598): add api.tree.resize

### LINKED ISSUE #2598: API: tree.resize
Background: #2595

Add API similar to `:NvimTreeResize`.

```help
tree.resize({opts})                              *nvim-tree-api.tree.resize()*
    Resize the tree, persisting the new size.
    Resets to |nvim-tree.view.width| when no {opts} provided.
    See |:NvimTreeResize|

    Parameters: ~

    Options: ~
      • {width}    (table)  new |nvim-tree.view.width| value
      • {absolute} (number) set the width
      • {relative} (number) increase or decrease the width

    Only one option is supported, in the priority order above.
    {absolute} and {relative} do nothing when {width} is a function.
```

It may be best to have 3 methods due to the exclusive nature of the options. Look at api.txt and lua.txt for guidance.

opa-oz:
Resolved #2598 

## Testing environment

I slightly modified `useState` analogy to follow proposed API specification:
```lua
local resize = require("nvim-tree.api").tree.resize
local function use_nvim_tree_width(options)
	if type(options) ~= "table" then
		return
	end

	---@type number
	local offset = options.offset

	if type(offset) ~= "number" then
		offset = 10
	end

	---@type number|table
	local width = 30

	local function get_width()
		return width
	end

	local function increment_width()
		resize({ relative = offset }) -- <==== here
	end

	local function decrement_width()
		resize({ relative = -offset }) -- <==== here
	end

	local function dynamic_width()
		width = {}
		resize({ width = width }) -- <==== here
	end

	return get_width, increment_width, decrement_width, dynamic_width
end
```

Small showcase:
ScreenRecording2024-06-24at13 13 30-ezgif com-video-to-gif-converter

GIF explanation:
1. Increase
2. Increase
3. Increase
4. Decrease
5. Dynamic width

### DIFF
--- a/lua/nvim-tree/actions/tree/resize.lua
+++ b/lua/nvim-tree/actions/tree/resize.lua
@@ -34,8 +34,12 @@ function M.fn(opts)
 
   local relative = options.relative
   if type(relative) == "number" then
-    local new_size = view.get_width() + relative
-    view.resize(new_size)
+    local relative_size = tostring(relative)
+    if relative > 0 then
+      relative_size = "+" .. relative_size
+    end
+
+    view.resize(relative_size)
     return
   end
 end
--- a/lua/nvim-tree/actions/tree/resize.lua
+++ b/lua/nvim-tree/actions/tree/resize.lua
@@ -1,0 +1,47 @@
+local view = require "nvim-tree.view"
+
+local M = {}
+
+---Resize the tree, persisting the new size.
+---@param opts ApiTreeResizeOpts|nil
+function M.fn(opts)
+  if opts == nil then
+    -- reset to config values
+    view.configure_width()
+    view.resize()
+    return
+  end
+
+  local options = opts or {}
+  local width_cfg = options.width
+
+  if width_cfg ~= nil then
+    view.configure_width(width_cfg)
+    view.resize()
+    return
+  end
+
+  if not view.is_width_determined() then
+    -- {absolute} and {relative} do nothing when {width} is a function.
+    return
+  end
+
+  local absolute = options.absolute
+  if type(absolute) == "number" then
+    view.resize(absolute)
+    return
+  end
+
+  local relative = options.relative
+  if type(relative) == "number" then
+    local new_size = view.get_width() + relative
+    view.resize(new_size)
+    return
+  end
+end
+
+function M.setup(opts)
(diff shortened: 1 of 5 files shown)


lua/nvim-tree/view.lua
```diff
@@ -548,6 +548,34 @@ function M.reset_winhl()
...
 
+---Check if width determined or calculated on-fly
+---@return boolean
+function M.is_width_determined()
+  return type(M.View.width) ~= "function"
+end
+
+---Configure width-related config
+---@param width string|function|number|table|nil
+function M.configure_width(width)
```
alex-courtis, line 559: Thank you for extracting; this was too magically stateful.

lua/nvim-tree/actions/tree/resize.lua
```diff
@@ -0,0 +1,51 @@
...
+  local options = opts or {}
+  local width_cfg = options.width
+
+  if width_cfg ~= nil then
+    view.configure_width(width_cfg)
+    view.resize()
+    return
+  end
+
+  if not view.is_width_determined() then
```
alex-courtis, line 24: Rather than using this function, could we simply test `type(M.config.view.width) == "function"` ?

More encapsulation...

alex-courtis (changes requested):
Apologies for the delay.

Works beautifully.

Tested base cases and changing width from and to function.

A couple of small questions and you'll need to update help.

lua/nvim-tree/actions/tree/resize.lua
opa-oz: @alex-courtis 
Hmmmmmmmm

In this case we wanna use 
```lua
M.config.view.configure_width(...)
M.config.view.resize()
```
as well (because we anyway using it)

In this case we can get rid of 
```lua
```
and just do 
```lua
local view = M.config.view
```
which actually won't work...

IMO let's stay consistent (current solution). But if you OK with having `local view = require(...)` and `M.config.view` in the same function - I don't mind changing. it :)

alex-courtis: Fair enough, thanks for digging.

That change would indeed "taint" resize; view should be fixed but not today.

alex-courtis (approved):

Issue #2813, opened 2024-06-28
`nvim-tree` does not allow renaming the file even with changed letter capitalization


samueldurantes:
### Description

I'm trying to rename a file by changing the first letter from lowercase to uppercase, but I'm getting an error that says the file cannot be renamed because it already exists.

### Neovim version

```text
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1710088188
```

### Operating system and version

macOS 14.3

### nvim-tree version

2086e56

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup {}
end

-- UNCOMMENT this block for diagnostics issues, substituting pattern and cmd as appropriate.
-- Requires diagnostics.enable = true in setup.
--[[
vim.api.nvim_create_autocmd("FileType", {
  pattern = "lua",
  callback = function()
    vim.lsp.start { cmd = { "lua-language-server" } }
  end,
})
]]
```

### Steps to reproduce

1. mkdir test
2. cd test
3. touch test.js
[...]

PR #2814, opened 2024-06-28
fix(#2813): enable file renaming in `nvim-tree` with changed capitalization

samueldurantes:
This PR fixes an issue in `nvim-tree` on macOS that prevented file renaming when only the capitalization of letters was changed. Now, `nvim-tree` allows renaming files regardless of changes in letter capitalization in the existing file name.

### DIFF
--- a/lua/nvim-tree/actions/fs/rename-file.lua
+++ b/lua/nvim-tree/actions/fs/rename-file.lua
@@ -30,13 +30,21 @@ local function err_fmt(from, to, reason)
   return string.format("Cannot rename %s -> %s: %s", from, to, reason)
 end
 
+local function rename_file_exists(node, to)
+  if string.lower(node) == string.lower(to) then
+    return false
+  end
+
+  return utils.file_exists(to)
+end
+
 ---@param node Node
 ---@param to string
 function M.rename(node, to)
   local notify_from = notify.render_path(node.absolute_path)
   local notify_to = notify.render_path(to)
 
-  if utils.file_exists(to) then
+  if rename_file_exists(notify_from, notify_to) then
     notify.warn(err_fmt(notify_from, notify_to, "file already exists"))
     return
   end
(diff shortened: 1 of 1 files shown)


gegoune:
It doesn't work for me. This is not a limitation of nvim-tree but your case insensitive filesystem.

samueldurantes:
> It doesn't work for me. This is not a limitation of nvim-tree but your case insensitive filesystem.

Exactly, this PR aims to turn nvim-tree more resilient to this limitation.

gegoune:
Getting `[NvimTree] Cannot rename [...]/a [...]/A : file already exists`.

samueldurantes:
> Getting `[NvimTree] Cannot rename [...]/a [...]/A : file already exists`.

What is the your OS?

alex-courtis:
We've been down this path before and had unresolvable issues with MacOS:
discussion: https://github.com/nvim-tree/nvim-tree.lua/issues/914
attempts:  https://github.com/nvim-tree/nvim-tree.lua/pull/1977 https://github.com/nvim-tree/nvim-tree.lua/pull/1988

alex-courtis:
If we do this, it will need to be absolutely bulletproof: existing codepaths for windows and macos will need to be unaffected.

You can use utils.lua  `is_macos` `is_wsl` `is_windows`

samueldurantes:
> If we do this, it will need to be absolutely bulletproof: existing codepaths for windows and macos will need to be unaffected.
> 
> You can use utils.lua `is_macos` `is_wsl` `is_windows`

Fair enough, I will add a new check for this.

samueldurantes:
@alex-courtis Done!

alex-courtis (approved):

itsramiel:
While this seems to work, I tried copying a file and pasting it, when promted to rename since file already exsists I change the name but it does nothing

https://github.com/user-attachments/assets/10825c10-2930-468c-b498-00b979924684

alex-courtis:
> While this seems to work, I tried copying a file and pasting it, when promted to rename since file already exsists I change the name but it does nothing

I am not confident that can work; we are restricted by filesystem limitations.

@samueldurantes is it somehow possible?

PR #2839, opened 2024-07-14
feat(#2349): add "right_align" option for renderer.icons.*_placement

evertonse:
Closes #2349

#### This PR augments options for icons placement allowing rendering of theses decorations on the far right using the ``right_align`` option from neovim ``ext_marks``

image

##  Setup for testing
```lua
require('nvim-tree').setup {
  renderer = {
    icons = {
      git_placement = 'right_align',
      modified_placement = 'right_align',
      diagnostics_placement = 'right_align',
      bookmarks_placement = 'right_align',
      padding = ' ',
      show = {
        file = true,
        folder = true,
        diagnostics = true,
        bookmarks = true,
        git = true,
        modified = true,
      },
    },
  },
}
```

If there's anything you'd like me to change about documentation or code, let me know, I'll do it.

Issue #2349, opened 2023-08-01
Icon Placement Right


alex-courtis:
**Can this functionality be implemented utilising API?**
No, it's internal behaviour.

**Describe the solution you'd like**
Show icons: git, diagnostic, modified at the far right side of the tree.

**Describe alternatives you've considered**
Right sign column, not possible.

**Additional context**
#2348

alex-courtis:
Requires #1510

alex-courtis:
Possible values: `"right"` `"end"` `"rightest"`

Silently move placement from `renderer.icons` to `git` `modified` `diagnostics.

vollowx:
I currently know nothing about how this plugin renders nor writing complex lua plugin, so I'm sorry but still not able to make a PR to implement this feature. The only thing I know is that this may be implemented by `vim.api.nvim_buf_set_extmark`. If nobody makes PR, will you or when will you implement it by yourself?

alex-courtis:
This issue is open and is a candidate for implementation, however will likely only be implemented when a kind contributor builds it.

mr2rm:
@alex-courtis If you still have not picked that up, I think I can work on it.

alex-courtis:
That would be fantastic @mr2rm !

I have been wanting this one for a while.

mr2rm:
I've just started working on this issue. I came up with the idea to add a new option to `renderer.icons` which only applies to all the placements with an `after` value. If implemented like this, we would have more flexibility to place some of the icons in the before position or on the sign column. I would suggest `renderer.icons.after_alignment` with values `left` and `right`.

@alex-courtis What do you think about this idea? Does it work for you?

mr2rm:
Also, we should decide when there is not enough room for the icons on the right. First, should we place icons on the right relative to the current window width or the max width? In both cases, what should be the behavior when there is no space to show all the icons on the right?

alex-courtis:
> I came up with the idea to add a new option to `renderer.icons` which only applies to all the placements with an `after` value. If implemented like this, we would have more flexibility to place some of the icons in the before position or on the sign column. I would suggest `renderer.icons.after_alignment` with values `left` and `right`.

That might not be enough; renderer.icons doesn't have visibility or control over placement.

We'd need to create a new value for the `ICON_PLACEMENT` enum, say, `right`. We'd add that to `Decorator` and `Builder:format_line` could use it.

alex-courtis:
[...]

PR #2846, opened 2024-07-22
feat(#2349): add "right_align" option for renderer.icons.*_placement

evertonse:
### Context:
The feature from [PR #2839](https://github.com/nvim-tree/nvim-tree.lua/pull/2839) was merged but quickly reverted due to a bug caused by a namespace collision with ``full_name`` rendering when ``renderer.full_name`` is enabled. This PR addresses the issue by adding a separate namespace and also syncs with the current nvim-tree master branch.

- Closes [Issue #2349](https://github.com/nvim-tree/nvim-tree.lua/issues/2349)
- Relevant coment: [Comment](https://github.com/nvim-tree/nvim-tree.lua/pull/2839#issuecomment-2242092528)


lua/nvim-tree/renderer/init.lua
```diff
@@ -13,12 +13,13 @@ local M = {}
 local SIGN_GROUP = "NvimTreeRendererSigns"
 
 local namespace_id = vim.api.nvim_create_namespace "NvimTreeHighlights"
```
alex-courtis, line 15:
```suggestion
local namespace_highlights_id = vim.api.nvim_create_namespace "NvimTreeHighlights"
```

lua/nvim-tree/renderer/init.lua
```diff
@@ -13,12 +13,13 @@ local M = {}
...
+local namespace_extmarks_id = vim.api.nvim_create_namespace "NvimTreeExtmarks"
```
alex-courtis, line 16: Nice find.

Could we play this a bit safer and have a separate namespace for [full-name](https://github.com/nvim-tree/nvim-tree.lua/blob/26632f496e7e3c0450d8ecff88f49068cecc8bda/lua/nvim-tree/renderer/components/full-name.lua#L69) as well?

It's using `"NvimTreeHighlights"` which is a little concerning now that we've learned of the namespace sharing difficulties.

alex-courtis (changes requested):
Many thanks for the quick turnaround!

lua/nvim-tree/renderer/init.lua
gegoune: Would be nice to rename it too, indeed, doesn't have to happen here though.

lua/nvim-tree/renderer/init.lua
evertonse: Done.

lua/nvim-tree/renderer/init.lua
evertonse: I looked at the the namespace usage in ``full_name.lua``, and it seems to rely on the ``NvimTreeHighlights`` namespace being set up beforehand. I thought this because of line 69:
```lua
local extmarks = vim.api.nvim_buf_get_extmarks(0, ns_id, { line_nr - 1, 0 }, { line_nr - 1, -1 }, { details = true })
```
Because of that I don't think I can introduce a separate namespace for full_name rn, also
I did not wrote any code in that file I'm not so sure what I could safely change.

alex-courtis (approved):
Looking good; spacing preserved for long files and full name working:

lua/nvim-tree/renderer/init.lua
alex-courtis: Fair enough; there may all sorts of assumptions there.

Issue #2819, opened 2024-07-07
Experimental Features: actions.open_file.relative_path


alex-courtis:
# Experimental Features

nvim-tree has experimental features, previews of new functionality that will become default or optional functionality.

Please try these out and reply with your experiences. If you encounter a problem, please raise a [bug report](https://github.com/nvim-tree/nvim-tree.lua/issues/new?assignees=&labels=bug&template=bug_report.yml) and indicate the experimental feature you are using.

## #2127 Open Buffers With Relative Path (#2805)

Buffers opened by nvim-tree will use with relative paths instead of absolute.
Execute `:ls` to see the paths of all open buffers.
`:help nvim-tree.experimental.actions.open_file.relative_path`

Enable:
```lua
      experimental = {
        actions = {
          open_file = {
            relative_path = true,
          },
        },
      },
```

**2024 07 07** introduced, false by default

epheien:
I have been using this feature here for a while without any issues, but my usage scenario is relatively simple.

As long as the absolute path of the file is converted to a relative path before specifying the: edit (etc.) command, it should theoretically be sufficiently secure. Based on this, I rechecked the code and found that there is an area that needs to be fixed.

In `local function open_in_new_window(filename, mode)` function, there is a problem with window jumping after converting to a relative path.
```vim
  if (mode == "preview" or mode == "preview_no_picker") and view.View.float.enable then
    -- ignore "WinLeave" autocmd on preview
    -- because the registered "WinLeave"
    -- will kill the floating window immediately
    set_current_win_no_autocmd(target_winid, { "WinLeave", "BufEnter" })
  else
    set_current_win_no_autocmd(target_winid, { "BufEnter" })
  end

  pcall(vim.api.nvim_cmd, command, { output = false })
  lib.set_target_win()
```

Can perform simple steps to reproduce this problem:
```vim
:NvimTreeOpen
" jump to a window to open file
:lcd ..
" jump to NvimTree window and select a file to open
```
This will open an incorrect path.

I just make a PR to fix it.

alex-courtis:
Thanks @epheien 

I did expect several issues like these. We will fix them.

paulodiovani:
I was looking for this. Thanks.

I'll be using from now on and report if I find any issue.

alex-courtis:
This is looking good... over a month and no issues.

Are you happy to finish the experiment and make this the one and only way to open files @epheien  ?

wrldspawn:
[...]

PR #2850, opened 2024-07-24
fix: change dir with explorer nodes unchanged

linepi:
# Situation
Every time I change the path in the explorer, the expanded folders in a certain directory will be reset. It's annoyning.
Here are some demonstration GIF images.

api: `api.tree.change_root_to_parent`
From path `~/tmp/t` to `~/tmp`

api: `api.tree.change_root_to_node`
Enter folder `~/tmp/t`

# Object
Change dir with explorer nodes unchanged.

### DIFF
--- a/lua/nvim-tree/actions/root/change-dir.lua
+++ b/lua/nvim-tree/actions/root/change-dir.lua
@@ -85,7 +85,11 @@ M.force_dirchange = add_profiling_to(function(foldername, should_open_view)
     if should_change_dir() then
       cd(M.options.global, foldername)
     end
-    core.init(foldername)
+    if M.options.explorer_node_unchanged then
+      core.change_root(foldername)
+    else
+      core.init(foldername)
+    end
   end
 
   if should_open_view then
--- a/lua/nvim-tree/core.lua
+++ b/lua/nvim-tree/core.lua
@@ -3,6 +3,8 @@ local explorer = require "nvim-tree.explorer"
 local live_filter = require "nvim-tree.live-filter"
 local view = require "nvim-tree.view"
 local log = require "nvim-tree.log"
+local Iterator = require "nvim-tree.iterators.node-iterator"
+local utils = require "nvim-tree.utils"
 
 local M = {}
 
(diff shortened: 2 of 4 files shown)


lua/nvim-tree/explorer/init.lua
```diff
@@ -45,6 +46,58 @@ function Explorer.new(path)
...
+  end
+  return explorer
+end
+
+---@param child_path string
+---@param oldexplorer Explorer
+---@return Explorer
+function Explorer.new_from_parent(child_path, oldexplorer)
+  ---@class Explorer
+  local explorer = setmetatable({
```
alex-courtis, line 75: We should use the constructor here: `Explorer.new` or similar.

alex-courtis (changes requested):
Resource leaks aside, this does pass the initial testing.

As this is a change in default behaviour, this should be an experiment as per https://github.com/nvim-tree/nvim-tree.lua/issues/2819

Not going to lie: I'm not comfortable with `new_from_parent` - there's a lot of magic in there.

Alternative proposal:
* create a temporary graph with open node state
* keep existing explore/node creation unchanged
* apply the graph to the new explorer

lua/nvim-tree/actions/root/change-dir.lua
```diff
@@ -85,7 +85,7 @@ M.force_dirchange = add_profiling_to(function(foldername, should_open_view)
     if should_change_dir() then
       cd(M.options.global, foldername)
     end
-    core.init(foldername)
+    core.change_root(foldername)
```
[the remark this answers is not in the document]
linepi: There should be no leaks now, please review.

linepi:
> As this is a change in default behaviour, this should be an experiment as per #2819
I have now created an option for this action, and I believe this will not affect the default behavior, and thus will not impact existing users.

alex-courtis:
> > As this is a change in default behaviour, this should be an experiment as per #2819
> 
> I have now created an option for this action, and I believe this will not affect the default behavior, and thus will not impact existing users.

We don't need an option; you've identified and fixed a bug and the new behaviour is correct.

We just have to be careful as this is core functionality.

lua/nvim-tree/core.lua
```diff
@@ -25,6 +27,49 @@ function M.init(foldername)
...
+  end
+  local root_parent_cwd = vim.fn.fnamemodify(utils.path_remove_trailing(TreeExplorer.absolute_path), ":h")
+  if root_parent_cwd == path then
+    local newTreeExplorer = explorer.Explorer.new(path)
+    if newTreeExplorer == nil then
+      return
+    end
+    for _, node in ipairs(newTreeExplorer.nodes) do
+      if node.absolute_path == TreeExplorer.absolute_path then
+        node.nodes = TreeExplorer.nodes
```
alex-courtis, line 43: These nodes are being destroyed when `TreeExplorer:destroy()` is executed.

Test case:
create.sh
```sh
#!/bin/sh

mkdir -p a/1/A
mkdir -p a/2/A
mkdir -p a/2/B
mkdir -p b/1/A
mkdir -p b/2/B

touch foo
touch a/1/foo
touch a/1/A/bar
touch a/2/A/baz
touch a/2/B/foo
touch b/1/A/bar
touch b/2/B/baz
```

Open from directory `a` and expand all `E`.

From another terminal or cmd mode `touch a/1/A/zero` and see it appear.

`-` to change directory up.

`rm a/1/A/zero` does not update the tree until manual refresh `R`.

This is because the watcher has been destroyed.

alex-courtis (changes requested):
This is good. It's a "graft" that puts a branch from the old tree in the new, which is the most efficient method.

We need to cut the branch from the old tree so that it is not destroyed with the old tree.

Suggestion: find and cut off the branch first, then destroy the old tree.

lua/nvim-tree/core.lua
linepi: Oh, I overlooked that fact. Let me make some changes. Thank you for providing the test case.

alex-courtis:
This looks great; I'm limited by time however I will get to it.

alex-courtis:
Closing as this PR is stale. Please reopen if you wish to continue this work.

PR #2872, opened 2024-08-17
refactor(#2830): multi instance marks

### LINKED ISSUE #2830: Multi Instance: Refactor: nvim-tree.marks
* #2255 

`nvim-tree.marks` is a stateful singleton. Make it an instantiable class owned by `Explorer`.

All usages must go through the `TreeExplorer` singleton.

alex-courtis:
fixes #2830

Follow the explorer member with back references pattern.


alex-courtis:
These aren't stateful and don't benefit from being members.

Move them into the Marks class instead, simplifying as they can directly access the data.

PR #2893, opened 2024-09-01
fix(#2794): sshfs compatibility

### LINKED ISSUE #2794: `nvim-tree` allows the creation of files over `sshfs`, but they can't be accessed
### Description

I'm using `nosduco/remote-sshfs.nvim` to work remotely over sshfs.

`nvim-tree` doesn't show anything from the mounted folder on the server, however, you can perform edit operations from `nvim-tree` directly, such as creating a new file (see image, I basically pressed `a` on top of the folder in nvim-tree panel).

I checked on server and the file was indeed created.

Thanks

### Neovim version

```text
NVIM v0.10.0
Build type: Release
LuaJIT 2.1.1716656478


"nvim-tree.lua": { "branch": "master", "commit": "517e4fbb9ef3c0986da7047f44b4b91a2400f93c" }
```

macOS Sonoma 14.5 (23F79)

### Windows variant

_No response_

### nvim-tree version

{ "branch": "master", "commit": "517e4fbb9ef3c0986da7047f44b4b91a2400f93c" }

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

local function load_plugins()
  require("packer").startup {
    {
    "wbthomason/packer.nvim",
    "nvim-tree/nvim-tree.lua",
    'nosduco/remote-sshfs.nvim',
    "nvim-telescope/telescope.nvim", branch = "0.1.x",
    "nvim-lua/plenary.nvim",
    },

    config = {
      package_root = package_root,
    },
  }
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
    require("nvim-tree").setup {}
    require('remote-sshfs').setup {
      connections = {
        ssh_configs = { -- which ssh configs to parse for hosts list
          vim.fn.expand "$HOME" .. "/.ssh/config",
          "/etc/ssh/ssh_config",
        },
        sshfs_args = { -- arguments to pass to the sshfs command
          "-o reconnect",
          "-o auto_cache",
          "-o Ciphers=aes128-ctr",
          "-o ConnectTimeout=5",
          "-C",
          "-o cache_timeout=60",
          "-o cache=yes",
        },
      },
      mounts = {
        base_dir = vim.fn.expand "$HOME" .. "/.sshfs/", -- base directory for mount points
        unmount_on_exit = true, -- run sshfs as foreground, will unmount on vim exit
      },
      handlers = {
        on_connect = {
          change_dir = true, -- when connected change vim working directory to mount point
        },
        on_disconnect = {
          clean_mount_folders = false, -- remove mount point folder on disconnect/unmount
        },
        on_edit = {}, -- not yet implemented
      },
      ui = {
        select_prompts = false, -- not yet implemented
        confirm = {
          connect = true, -- prompt y/n when host is selected to connect to
          change_dir = false, -- prompt y/n to change working directory on connection (only applicable if handlers.on_connect.change_dir is enabled)
        },
      },
      log = {
        enable = false, -- enable logging
        truncate = false, -- truncate logs
        types = { -- enabled log types
          all = false,
          util = false,
          handler = false,
          sshfs = false,
        },
      },
    }
    require('telescope').setup {}

end

-- UNCOMMENT this block for diagnostics issues, substituting pattern and cmd as appropriate.
-- Requires diagnostics.enable = true in setup.
--[[
vim.api.nvim_create_autocmd("FileType", {
  pattern = "lua",
  callback = function()
    vim.lsp.start { cmd = { "lua-language-server" } }
  end,
})
]]
```

### Steps to reproduce

```
1. nvim -nu /tmp/nvt-min.lua
2. :RemoteSSHFSConnect (pick server; have it listed in your ~/.ssh/config)
3. :NvimTreeOpen
4. expand the mounted folder (should be ~ on your server)
```

### Expected behavior

List all files, directories, normal navigation as one would have locally.

### Actual behavior

image

mxple:
Added type fallback to make nvim-tree work with sshfs.

fixes #2794

### DIFF
--- a/lua/nvim-tree/explorer/explore.lua
+++ b/lua/nvim-tree/explorer/explore.lua
@@ -43,7 +43,7 @@ local function populate_children(handle, cwd, node, git_status, parent)
       local stat = vim.loop.fs_stat(abs)
 
       -- Type must come from fs_stat and not fs_scandir_next to maintain sshfs compatibility
-      local type = (vim.loop.fs_stat(abs) or {}).type
+      local type = stat and stat.type or nil
 
       local filter_reason = parent.filters:should_filter_as_reason(abs, stat, filter_status)
       if filter_reason == FILTER_REASON.none and not nodes_by_path[abs] then
--- a/lua/nvim-tree/explorer/explore.lua
+++ b/lua/nvim-tree/explorer/explore.lua
@@ -35,6 +35,7 @@ local function populate_children(handle, cwd, node, git_status, parent)
     end
 
     local abs = utils.path_join { cwd, name }
+    t = t or (vim.loop.fs_stat(abs) or {}).type
 
     if Watcher.is_fs_event_capable(abs) then
       local profile = log.profile_start("explore populate_children %s", abs)
--- a/lua/nvim-tree/explorer/explore.lua
+++ b/lua/nvim-tree/explorer/explore.lua
@@ -29,27 +29,30 @@ local function populate_children(handle, cwd, node, git_status, parent)
   })
 
   while true do
-    local name, t = vim.loop.fs_scandir_next(handle)
+    local name, _ = vim.loop.fs_scandir_next(handle)
     if not name then
       break
     end
 
     local abs = utils.path_join { cwd, name }
-    t = t or (vim.loop.fs_stat(abs) or {}).type
 
     if Watcher.is_fs_event_capable(abs) then
       local profile = log.profile_start("explore populate_children %s", abs)
 
       ---@type uv.fs_stat.result|nil
       local stat = vim.loop.fs_stat(abs)
+
+      -- Type must come from fs_stat and not fs_scandir_next to maintain sshfs compatibility
+      local type = (vim.loop.fs_stat(abs) or {}).type
+
       local filter_reason = parent.filters:should_filter_as_reason(abs, stat, filter_status)
       if filter_reason == FILTER_REASON.none and not nodes_by_path[abs] then
         local child = nil
-        if t == "directory" and vim.loop.fs_access(abs, "R") then
+        if type == "directory" and vim.loop.fs_access(abs, "R") then
           child = builders.folder(node, abs, name, stat)
-        elseif t == "file" then
+        elseif type == "file" then
           child = builders.file(node, abs, name, stat)
-        elseif t == "link" then
+        elseif type == "link" then
           local link = builders.link(node, abs, name, stat)
           if link.link_to ~= nil then
             child = link
--- a/lua/nvim-tree/actions/fs/remove-file.lua
+++ b/lua/nvim-tree/actions/fs/remove-file.lua
@@ -56,13 +56,18 @@ local function remove_dir(cwd)
   end
 
   while true do
-    local name, t = vim.loop.fs_scandir_next(handle)
+    local name, _ = vim.loop.fs_scandir_next(handle)
     if not name then
       break
     end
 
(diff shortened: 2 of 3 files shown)


lua/nvim-tree/explorer/explore.lua
```diff
@@ -35,6 +35,7 @@ local function populate_children(handle, cwd, node, git_status, parent)
     end
 
     local abs = utils.path_join { cwd, name }
+    t = t or (vim.loop.fs_stat(abs) or {}).type
```
alex-courtis, line 38: This is an extra (not expensive as usually short circuited) call to fs_stat.

Could we put this after line 44?

alex-courtis (changes requested):
Great! Nice and simple.

Please also leave a comment about the use case that this test addresses and why.

lua/nvim-tree/explorer/explore.lua
mxple: Moved after line 44, please review.

lua/nvim-tree/explorer/explore.lua
```diff
@@ -41,14 +41,18 @@ local function populate_children(handle, cwd, node, git_status, parent)
 
       ---@type uv.fs_stat.result|nil
       local stat = vim.loop.fs_stat(abs)
+
+      -- Type must come from fs_stat and not fs_scandir_next to maintain sshfs compatibility
+      local type = (vim.loop.fs_stat(abs) or {}).type
```
alex-courtis, line 46:
```suggestion
      local type = stat and stat.type or nil
```

It looks like we're still calling fs_stat twice, which is a big performance hit.

I might be misunderstanding, but it looks like type should be nil on an fs_stat failure.

alex-courtis (changes requested):

lua/nvim-tree/explorer/explore.lua
mxple: You're right, that's an oversight by me.

alex-courtis:
This is working really nicely. The new files/directories are being picked up as soon as the OS sees them.

`t` is nil for sshfs but `type` is correct.
```lua
      -- Type must come from fs_stat and not fs_scandir_next to maintain sshfs compatibility
      local type = stat and stat.type or nil
      log.line("dev", "populate_children %.06s\t%.06s\t%s", vim.inspect(t), vim.inspect(type), abs)
```

Adding a directory does work, however you need to close/open the containing folder.

alex-courtis:
We need to go further: there are other places that need this fix.

Reload is the most obvious case which will resolve the add case above:
https://github.com/mxple/nvim-tree.lua/blob/ea55ef12036897fdc4476b115a395d2a34965c82/lua/nvim-tree/explorer/init.lua#L129

remove-file looks to be a similar case.

Please look for all cases of broken functionality or use of `fs_scandir_next` that we can resolve with this fix.

alex-courtis (changes requested):
I spoke too soon when claiming this was simple...

mxple:
Good catch on the create/delete/reload functionality. I did not think to test that. I added the same type checking for reloading and deletion and in my testing, remote folders seem to behave identical to local folders. Please confirm and let me know if any further changes are needed.

alex-courtis (approved):
Works nicely, many thanks for your contribution.

As you're working with sshfs I'd be most grateful for future PRs if you discover similar issues in the future.

Issue #2902, opened 2024-09-10
Memory leak and hanging when using the Live Filter feature


opsidjflksdf:
### Description

Hello - I'm encountering a memory leak that starts eating into RAM as soon as I have used the live filter feature (default shortcut f). The neovim process RAM usage starts going up and it keeps going. Before I figured what caused it, I caught it at 30 GB of RAM, after a day or so.
It also causes that particular nvim instance to become unresponsive after not too long (minutes).
This is on Windows 10. neovim usually takes about 3MB. As soon as I have used the live filter in nvim tree, the process goes to something like 12MB, and slowly rising from there. Even after you have closed/cleared the filter. Very easy to reproduce. Would be great if someone could look into it. I can provide more testing and details.
Thanks!

### Neovim version

```text
0.10.0
```

### Operating system and version

Windows 10

### nvim-tree version

1.6.1

### Clean room replication

```lua
Please see above
```

### Steps to reproduce

Please see above

### Expected behavior

Please see above

### Actual behavior

Please see above

alex-courtis:
Caveat: nvim-tree team do not have access to nor expertise with windows.

I attemtped a replication:
* open the linux repository (it's large)
* navigate to `Documentation`
* `E`
* `f` yml
* `F`

Please provide a clean room configuration and specific instruction on how to replicate this issue.

Issue #2906, opened 2024-09-13
Arithmetic on a nil value error when opening file in child git repo with NvimTree Open


ianhomer:
### Description

With NvimTree open, when I open a file (with Telescope oldfiles) in a child project directory of the root directory and this child project directory is a git repository then I get the following error.

```txt
Error executing vim.schedule lua callback: ...r/start/nvim-tree.lua/lua/nvim-tree/explorer/explore.lua:69: attempt to perform
 arithmetic on a nil value
stack traceback:
        ...r/start/nvim-tree.lua/lua/nvim-tree/explorer/explore.lua:69: in function 'populate_children'
        ...r/start/nvim-tree.lua/lua/nvim-tree/explorer/explore.lua:92: in function 'explore'
        ...cker/start/nvim-tree.lua/lua/nvim-tree/explorer/init.lua:267: in function '_load'
        ...cker/start/nvim-tree.lua/lua/nvim-tree/explorer/init.lua:73: in function 'expand'
        ...vim-tree.lua/lua/nvim-tree/actions/finders/find-file.lua:66: in function '_apply_fn_on_node'
        .../nvim-tree.lua/lua/nvim-tree/iterators/node-iterator.lua:63: in function 'iter'
        .../nvim-tree.lua/lua/nvim-tree/iterators/node-iterator.lua:66: in function 'iterate'
        ...vim-tree.lua/lua/nvim-tree/actions/finders/find-file.lua:76: in function 'fn'
        ...t/nvim-tree.lua/lua/nvim-tree/actions/tree/find-file.lua:64: in function 'fn'
        ...n/site/pack/packer/start/nvim-tree.lua/lua/nvim-tree.lua:280: in function 'callback'
        .../pack/packer/start/nvim-tree.lua/lua/nvim-tree/utils.lua:455: in function <.../pack/packer/start/nvim-tree.lua/lua
/nvim-tree/utils.lua:454>
```

[...]

PR #2856, opened 2024-07-28
feat(#2225): add renderer.hidden_display to show a summary of hidden files below the tree

evertonse:
This PR introduces the rendering of hidden count files per folder, a feature that closes https://github.com/nvim-tree/nvim-tree.lua/issues/2225

Option ``renderer.hidden_display``, allows users to customize how hidden files count information are displayed per folder.
It can be either a function or  a string. It defaults to "none".
- "none": Doesn't inform anything about hidden files.
- "simple": Shows how many hidden files are in a folder.
- "all": Shows how many files are hidden and the number of hidden files per reason why they're hidden.

#### Video for Ilustration
The video below show that we can trace the ``reason`` for exclusion. Also, shows that it works with ``live filter`` and takes into account ``indent marks`` and ``indent width``. 

https://github.com/user-attachments/assets/c5e7812e-ec95-4298-b82b-a94221cde96e

If a function is provided to ``renderer.hidden_display``, it receives a table ``hidden_count`` where keys are reasons and values are the count of hidden files for that reason.

```lua
hidden_count = {
  bookmark = <num>,
  buf = <num>,
  custom = <num>,
  dotfile = <num>,
  git = <num>,
  live_filter = <num>,
}
```
If an error occurs inside this function, we issue a warning:

https://github.com/user-attachments/assets/118a3029-630e-4258-9c93-19e7fe51ac3c

Also exposes the highlight "NvimTreeHiddenDisplay".

## Minimal Setup 
Setup using custom function
```lua
require('nvim-tree').setup {
  renderer = {
    -- hidden_display = 'all',
    hidden_display = function(hidden_count)
      local total_count = 0
      for reason, count in pairs(hidden_count) do
        total_count = total_count + count
      end

      if total_count > 0 then
        return '> ' .. tostring(total_count) .. ' le hidden'
      end
      return nil
    end,
  },
}
```

PR #2907, opened 2024-09-13
fix(#2906): resource leak on populate children

ianhomer:
FIx for https://github.com/nvim-tree/nvim-tree.lua/issues/2906

It looks like there are some situations where the explorer tries to populate a node when it is already populated. This leads to a situation in the code where the filter reason is none, however the node has already been loaded. This in turn leads to an attempt to record the hidden stats for the reason. Since the reason however is none and the `hidden_stats `object has understandably not been initialised to capture stats for reason none, the following exception is thrown

```txt
Error executing vim.schedule lua callback: ...r/start/nvim-tree.lua/lua/nvim-tree/explorer/explore.lua:69: attempt to perform
 arithmetic on a nil value
stack traceback:
        ...r/start/nvim-tree.lua/lua/nvim-tree/explorer/explore.lua:69: in function 'populate_children'
```

The fix changes the logic so that no attempt to record hidden stats are made and the node is not reloaded. I'm assuming here that the node has been loaded OK if the path is in the `nodes_by_path` object. I welcome any feedback on whether that is a good assumption.

### DIFF
--- a/lua/nvim-tree/explorer/explore.lua
+++ b/lua/nvim-tree/explorer/explore.lua
@@ -46,22 +46,24 @@ local function populate_children(handle, cwd, node, git_status, parent)
       local type = stat and stat.type or nil
 
       local filter_reason = parent.filters:should_filter_as_reason(abs, stat, filter_status)
-      if filter_reason == FILTER_REASON.none and not nodes_by_path[abs] then
-        local child = nil
-        if type == "directory" and vim.loop.fs_access(abs, "R") then
-          child = builders.folder(node, abs, name, stat)
-        elseif type == "file" then
-          child = builders.file(node, abs, name, stat)
-        elseif type == "link" then
-          local link = builders.link(node, abs, name, stat)
-          if link.link_to ~= nil then
-            child = link
+      if filter_reason == FILTER_REASON.none then
+        if not nodes_by_path[abs] then
+          local child = nil
+          if type == "directory" and vim.loop.fs_access(abs, "R") then
+            child = builders.folder(node, abs, name, stat)
+          elseif type == "file" then
+            child = builders.file(node, abs, name, stat)
+          elseif type == "link" then
+            local link = builders.link(node, abs, name, stat)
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Love your work @ianhomer 
FYI as a trusted member of the nvim-tree community you can push branches etc. to the main repo.

We're back to the pre-filter logic - not populating when nodes_by_path contains: https://github.com/nvim-tree/nvim-tree.lua/pull/2856/files#diff-1b05da4f503ff1df76cfeb7ef723f260d51acd287b0486f9de0ed61b2eeaa042L21

alex-courtis:
I'm buggered if I can replicate the telescope example, however this looks correct and might resolve #2902

ianhomer:
> I'm buggered if I can replicate the telescope example, however this looks correct and might resolve #2902

It took me a while to isolate the steps to trip this.  I do have a set of repositories (that I can't share) and it would happen every time I did the steps to reproduce. I ended up deleting stuff to find the minimal set to get the error, but may be I isolated steps for my env only.  It feels like a race condition and perhaps it's also dependent on hardware performance. I just tested the steps to reproduce again and I do still get the error on the commit before this fix merged.

I've also verified that the error doesn't happen for me on the latest release (after this fix merged), so looks good to me.

alex-courtis:
> It feels like a race condition and perhaps it's also dependent on hardware performance. I just tested the steps to reproduce again and I do still get the error on the commit before this fix merged.

Ah... you might be referring to this shameful hack:

https://github.com/nvim-tree/nvim-tree.lua/blob/cd9c6db77f7c8b348dfa14d805ea28a69e5e718f/lua/nvim-tree/explorer/init.lua#L447

> I've also verified that the error doesn't happen for me on the latest release (after this fix merged), so looks good to me.

Sounds like we've fixed the symptom at least. Let's monitor.

alex-courtis:
I'm hoping we can get rid of that ineffective lock during  https://github.com/orgs/nvim-tree/projects/1

PR #2927, opened 2024-09-23
fix: invalid explorer on open

TinusgragLin:
The return value of `core.get_explorer()` could be changed by the `core.init(..)` between the binding (L266) and its later uses (L281 onwards): 
https://github.com/nvim-tree/nvim-tree.lua/blob/8405ecfbd6bb08a94ffc9c68fef211eea56e8a3b/lua/nvim-tree/lib.lua#L266-L284


alex-courtis:
Thanks for the quick fix, I'll try to review// merge tonight

Issue #2961, opened 2024-10-19
[On windows] Root of nvim-tree changed unexpectedly and buffer not listed by telescope


ljie-PI:
### Description

With this change (https://github.com/nvim-tree/nvim-tree.lua/commit/45a93d99794fff3064141d5b3a50db98ce352697#diff-eeb21711fceb0abba1450bd1adcbecc4fd90cf8ce4d5d7a8233f886c1bebd41fR375), I found it doesn't work well on Windows for 2 cases:
1. When I open a file from nvim-tree, the root of nvim-tree window changed to the directory of that file
2. The buffer of this file can not be listed when I use telescope command `<cmd>lua require('telescope.builtin').buffers(require('telescope.themes').get_dropdown{previewer = false})<cr>`

### Neovim version

```text
NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1713484068
```

### Operating system and version

Windows 11

### Windows variant

Command Prompt

### nvim-tree version

2a268f6

### Clean room replication

```lua
I'm on windows so don't have /tmp/nvt-min.lua.

Pasting my config here:

local status_ok, nvim_tree = pcall(require, "nvim-tree")
if not status_ok then
  return
end

local api_status_ok, nvim_tree_api = pcall(require, "nvim-tree.api")
if not api_status_ok then
  return
end

local function customized_on_attach(bufnr)
  local function opts(desc)
    return { desc = "nvim-tree: " .. desc, buffer = bufnr, noremap = true, silent = true, nowait = true }
  end
  -- default mappings
  nvim_tree_api.config.mappings.default_on_attach(bufnr)
  -- custom mappings
  vim.keymap.set('n', 'c', nvim_tree_api.tree.change_root_to_node, opts('CD'))
  vim.keymap.set('n', 'l', nvim_tree_api.node.open.edit, opts('Open'))
  vim.keymap.set('n', '<CR>', nvim_tree_api.node.open.edit, opts('Open'))
  vim.keymap.set('n', 'o', nvim_tree_api.node.open.edit, opts('Open'))
  vim.keymap.set('n', 'h', nvim_tree_api.node.navigate.parent_close, opts('Close Directory'))
  vim.keymap.set('n', 'v', nvim_tree_api.node.open.vertical, opts('Open: Vertical Split'))
end

local keymap = vim.api.nvim_set_keymap
keymap("n", "<leader>e", ":NvimTreeToggle<CR>", { noremap = true, silent = true })

[...]

Issue #2862, opened 2024-08-03
Wrong filepath opened when selecting file


LandonHarter:
### Description

I am unable to open a file and read it because nvim-tree, when selecting a file, will open a nonexistent file. The route it should be opening should be `\app\(pages)\(app)\document\[id]\page.tsx` but instead opens `\app(pages)(app)\document\[id]\page.tsx`. I don't seem to have this issue with other plugins such as telescope, where I can use the file finder and it will give me a preview of the content (I will provide an image below). My guess is when it tries to put a backslash before a parenthesis, it registers it as just a parenthesis (kind of how in regex you have to put a backslash before some characters).

# Incorrect Path
![image](https://github.com/user-attachments/assets/c0657986-6ae2-405c-9b18-99dbebd00de8)

# Telescope Displaying Content Correctly
![image](https://github.com/user-attachments/assets/454ec5d0-4827-4bf0-9313-b8301901314c)

### Neovim version

```text
NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1713484068
```

### Operating system and version

Windows 11

### nvim-tree version

v1.5.0 (latest)

### Clean room replication

```lua
I am a bit confused about this step, but you can find my neovim setup here:
https://github.com/LandonHarter/neovim/master
```

### Steps to reproduce

1. Create a directory that starts and ends with parenthesis (maybe try nesting a few together like mine in the photo)
2. Create a file in one of the new directories
3. Try to open the file with nvim-tree

### Expected behavior

Opens the file

### Actual behavior

Tries to open a nonexistent file leading to me being unable to edit the file

alex-courtis:
When opening on windows, special characters will be escaped: https://github.com/nvim-tree/nvim-tree.lua/blob/8704b6f7fc0bdbc4e0a754fa1e1d3fd3c9bf4969/lua/nvim-tree/utils.lua#L263

It looks like there's some handling of parenthesis in there.

No nvim-tree team members have access to or experience with windows.

I'm going to ask you to attempt a fix as we have no way to test or verify it.

As a sanity check first: [please enumerate your feature flags](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development#os-feature-flags) to ensure that windows is being correctly detected.

PR #2903, opened 2024-09-11
fix(#2862): windows path replaces backslashes with forward slashes

dogeggz:
### Changes:
- Replaces `\` with `/` in file paths for Windows.
- Applies the updated path handling to functions that use the filename (e.g. open, open in tab, split, etc.).

### Note:
- Verified that all filename-related tasks work fine on my Windows machine.
- Not sure if this change will affect anything else though.

PR #2962, opened 2024-10-19
fix(#2961): windows: escape brackets and parentheses when opening file

ljie-PI:
Description
With this change (https://github.com/nvim-tree/nvim-tree.lua/commit/45a93d99794fff3064141d5b3a50db98ce352697#diff-eeb21711fceb0abba1450bd1adcbecc4fd90cf8ce4d5d7a8233f886c1bebd41fR375), I found it doesn't work well on Windows for 2 cases:

1. When I open a file from nvim-tree, the root of nvim-tree window changed to the directory of that file
2. The buffer of this file can not be listed when I use telescope command `<cmd>lua require('telescope.builtin').buffers(require('telescope.themes').get_dropdown{previewer = false})<cr>
Neovim version`

This PR is to revert the change and fix the issue #2961 and #2862

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -331,9 +331,9 @@ local function open_in_new_window(filename, mode)
 
   local fname
   if M.relative_path then
-    fname = vim.fn.fnameescape(utils.path_relative(filename, vim.fn.getcwd()))
+    fname = utils.escape_special_chars(vim.fn.fnameescape(utils.path_relative(filename, vim.fn.getcwd())))
   else
-    fname = vim.fn.fnameescape(filename)
+    fname = utils.escape_special_chars(vim.fn.fnameescape(filename))
   end
 
   local command
@@ -370,33 +370,27 @@ end
 ---@param mode string
 ---@param filename string
 function M.fn(mode, filename)
-  local fname
-  if utils.is_windows then
-    fname = filename
-  else
-    fname = utils.escape_special_chars(filename)
-  end
   if type(mode) ~= "string" then
     mode = ""
   end
 
   if mode == "tabnew" then
-    return open_file_in_tab(fname)
+    return open_file_in_tab(filename)
   end
 
   if mode == "drop" then
-    return drop(fname)
+    return drop(filename)
   end
 
   if mode == "tab_drop" then
-    return tab_drop(fname)
+    return tab_drop(filename)
   end
 
   if mode == "edit_in_place" then
-    return edit_in_current_buf(fname)
+    return edit_in_current_buf(filename)
   end
 
-  local buf_loaded = is_already_loaded(fname)
+  local buf_loaded = is_already_loaded(filename)
 
   local found_win = utils.get_win_buf_from_path(filename)
   if found_win and (mode == "preview" or mode == "preview_no_picker") then
@@ -404,7 +398,7 @@ function M.fn(mode, filename)
   end
 
   if not found_win then
-    open_in_new_window(fname, mode)
+    open_in_new_window(filename, mode)
   else
     vim.api.nvim_set_current_win(found_win)
     vim.bo.bufhidden = ""
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -329,14 +329,12 @@ local function open_in_new_window(filename, mode)
     set_current_win_no_autocmd(target_winid, { "BufEnter" })
   end
 
-  print("filename: " .. filename)
   local fname
   if M.relative_path then
     fname = utils.escape_special_chars(vim.fn.fnameescape(utils.path_relative(filename, vim.fn.getcwd())))
   else
     fname = utils.escape_special_chars(vim.fn.fnameescape(filename))
   end
-  print("fname: " .. fname)
 
   local command
   if create_new_window then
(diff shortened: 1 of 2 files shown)


ljie-PI:
link the issue: https://github.com/nvim-tree/nvim-tree.lua/issues/2961

lua/nvim-tree/utils.lua
```diff
@@ -59,6 +59,17 @@ function M.path_basename(path)
   return path:sub(i + 1, #path)
 end
 
+--- Check if there are parentheses before brackets, it causes problems for windows.
+--- Refer to issue #2862 and #2961 for more details.
+local function has_parentheses_and_brackets(path)
```
alex-courtis, line 64: Nice work, that's a strange one.

lua/nvim-tree/utils.lua
```diff
@@ -272,14 +288,22 @@ function M.canonical_path(path)
...
+
 --- Escapes special characters in string if windows else returns unmodified string.
 ---@param path string
 ---@return string|nil
 function M.escape_special_chars(path)
   if path == nil then
     return path
   end
-  return M.is_windows and path:gsub("\\", "/") or path
+  return M.is_windows and escape_special_char_for_windows(path) or path
```
alex-courtis, line 306: Thank you, this one definitely should be behind the windows feature flag.

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -370,36 +370,35 @@ end
 ---@param mode string
 ---@param filename string
 function M.fn(mode, filename)
-  local fname = utils.escape_special_chars(filename)
```
alex-courtis, line 373: This is a change in behaviour that will affect linux and macos users.

Please put this behind the windows feature flag, to ensure that it only applies to windows users.

Something like
```
local fname
if utils.is_windows then
  fname = filename
else
  fname = utils.escape_special_chars(filename)
end
```

alex-courtis (changes requested):
Many thanks for this fix! This all looks reasonable and I can see how you have tested comprehensively.

The `is_windows` feature flag [only applies to powershell](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development#os-feature-flags), hence you do not need to test WSL.

The nvim-tree developers don't have access to nor expertise with windows, if there is an issue we expect you to resolve it.

- [ ] All windows fixes must go behind the feature flags `utils.is_windows` etc. The existing codepaths for non-windows must remain unchanged.

lua/nvim-tree/actions/node/open-file.lua
ljie-PI: Hi @alex-courtis , change of this line is caused by the revert of previous fix. I think there wasn't such line at the begining.

lua/nvim-tree/utils.lua
```diff
@@ -68,13 +79,18 @@ function M.path_relative(path, relative_to)
...
 
-  local _, r = path:find(M.path_add_trailing(relative_to), 1, true)
-  local p = path
+  local norm_path = path
+  if has_parentheses_and_brackets(path) then
+    norm_path = path:gsub("/", "\\")
+  end
+
+  local _, r = norm_path:find(M.path_add_trailing(relative_to), 1, true)
+  local p = norm_path
```
alex-courtis, line 88: This path is executed for all OS, which may cause issues for such paths. Please execute only when `is_windows`

This injected error is definitely thrown on linux, showing that the code path is executed:
```lua
--- Check if there are parentheses before brackets, it causes problems for windows.
--- Refer to issue #2862 and #2961 for more details.
local function has_parentheses_and_brackets(path)
  error("has_parentheses_and_brackets")
```
```
E5108: Error executing lua: ...d/site/pack/packer/start/ljie-PI/lua/nvim-tree/utils.lua:65: has_parentheses_and_brackets called
stack traceback:
        [C]: in function 'error'
        ...d/site/pack/packer/start/ljie-PI/lua/nvim-tree/utils.lua:65: in function 'has_parentheses_and_brackets'
        ...d/site/pack/packer/start/ljie-PI/lua/nvim-tree/utils.lua:84: in function 'path_relative'
        .../packer/start/ljie-PI/lua/nvim-tree/explorer/filters.lua:161: in function 'custom'
        .../packer/start/ljie-PI/lua/nvim-tree/explorer/filters.lua:247: in function 'should_filter_as_reason'
```

alex-courtis (changes requested):
Looking good, just the one windows feature flag check needed.

Alternative idea:
I was able to cleanly `git revert 45a93d99794fff3064141d5b3a50db98ce352697` on master
You could instead do that on a new branch and apply your fixes on top.
That would result in cleaner history and a simpler PR.
Your choice.

lua/nvim-tree/utils.lua
ljie-PI: Added flag on line 84, so that norm_path will keep unchanged for non-windows OS.

lua/nvim-tree/actions/node/open-file.lua
ljie-PI: Yes, I reverted all changes of #2903 . 

In #2903 `util.escape_special_chats` is called here and impact all file names below, which are not correct. (my tests failed)

Now and before #2903  `utils.escape_special_chars` is called in the `open_in_new_window` function (line 334 and 336), so I think I'm just following the logic before #2903 . And the is_windows flag is already in the `utils.escape_special_chars` function, so not added outside `utils.escape_special_chars`

alex-courtis:
Thank you

I'll get to this review on the weekend.

alex-courtis (approved):

PR #2966, opened 2024-10-25
chore: fix lib prompt doc for neovim nightly

alex-courtis:


lua/nvim-tree/lib.lua
```diff
@@ -108,7 +108,7 @@ end
 ---@param items_short string[]
 ---@param items_long string[]
 ---@param kind string|nil
----@param callback fun(item_short: string)
+---@param callback fun(item_short: string|nil)
```
alex-courtis, line 111: Callers always check nil

Issue #2969, opened 2024-10-26
After a rename, the node loses selection


opsidjflksdf:
### Description

After you rename a (selected) node (file or folder) such that it changes its position in the current sorting order, the selection stays with the same number index within the tree (thus a different file/folder ending up being selected), instead of moving together with the node that was renamed.

In rename-file.lua, function fn(), there is code to update the selection:
`find_file(utils.path_remove_trailing(new_file_path))`
but the new_file_path variable only contains the filename or part of it (depending on the rename option chosen),  rather than the entire path, even if relative.
rename() itself is called with
`M.rename(node, prepend .. new_file_path .. append)`
I tested and doing something like the following instead (calling find_file() with the same destination path that rename() was called with), fixes the issue (selection properly moves with the renamed node):
```
local new_full_file_path = prepend .. new_file_path .. append
M.rename(node, new_full_file_path )
[...]
find_file(utils.path_remove_trailing(new_full_file_path))
```
Could this be implemented? I wanted to run it by you rather than create a pull request in case there might be other considerations here. Thanks!

### Neovim version

```text
NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info
```

### Operating system and version

Windows 10

### Windows variant

Native Windows

### nvim-tree version

1.7.1 and latest master both have the issue

### Clean room replication

```lua
nvim-tree with any file hierarchy
```

### Steps to reproduce

In the nvim-tree window, set the selection on a node - file or folder, and then rename it with shortcut e or r. Selection doesn't move with the renamed item.

### Expected behavior

Selection should move with the renamed item.

### Actual behavior

Selection doesn't move with the renamed item.

alex-courtis:
Not able to replicate on linux/alacritty.

![20241027_093230](https://github.com/user-attachments/assets/d8910d60-dc26-433e-b490-d6d3e0e652c0)

![20241027_093243](https://github.com/user-attachments/assets/8704289c-2f48-4a42-816c-a8c39827a2c9)

It sounds likely a windows issue, due to the path difficulties in that OS.

nvim-tree team have no access to nor expertise with windows. Please do create the PR.

Please ensure changes are behind the windows feature flags `utils.is_wsl` or `utils.windows`. See [wiki: OS Feature Flags](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development#os-feature-flags)

opsidjflksdf:
[...]

PR #2974, opened 2024-10-27
fix(#2969): After a rename, the node loses selection

opsidjflksdf:
Use the full path, as used in rename(), for the subsequent call to find_file(), in order to correctly re-select the renamed item.
Fix for 
https://github.com/nvim-tree/nvim-tree.lua/issues/2969


alex-courtis (approved):
Works beautifully, many thanks for your contribution!

Issue #2990, opened 2024-11-05
Diagnostic Indicators Not Updating


alex-courtis:
### Description

Diagnostic signs are eventually showing correctly however they take some time.

Regression since 82ab19e

Can I please leave this with you @des-b ?

### Neovim version

```text
: ; nvim --version
NVIM v0.10.2
Build type: RelWithDebInfo
LuaJIT 2.1.1727870382
```

### Operating system and version

Linux 6.11.5-arch1-1

### nvim-tree version

82ab19e

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvt-min/site]])
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing nvim-tree and dependencies.")
  vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
vim.cmd([[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]])
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
_G.setup = function()
  require("nvim-tree").setup({
    diagnostics = {
      enable = true,
    }
  })
end

-- UNCOMMENT this block for diagnostics issues, substituting pattern and cmd as appropriate.
-- Requires diagnostics.enable = true in setup.
vim.api.nvim_create_autocmd("FileType", {
  pattern = "lua",
  callback = function()
    vim.lsp.start {
      name = "my-luals",
      cmd = { "lua-language-server" },
      root_dir = vim.loop.cwd(),
    }
[...]

PR #2980, opened 2024-11-01
fix(#2954): resolve occasional tree flashing on diagnostics, set tree buffer options in deterministic order

des-b:
fixes #2954

PR #2998, opened 2024-11-10
fix(#2990): Do not check if buffer is buflisted in diagnostics.update()

des-b:
This ensures that LSP diagnostics of files which are not manually opened by users are rendered by nvim-tree diagnostic indicators.

However when users attach an LSP to nvim-tree this will bring back flashing as attempted to fix in #2980. Fixing this should probably done by checking data passed via diagnostic events (DiagnosticChanged and CocDiagnosticChanged).

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -169,10 +169,7 @@ function M.update()
     log.profile_end(profile)
 
     local bufnr = view.get_bufnr()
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Works nicely, thanks for the fix @des-b 

The whole view business will be resolved / rewritten during https://github.com/orgs/nvim-tree/projects/1

If you have any spare time your assistance would be gratefully appreciated!

PR #3006, opened 2024-11-17
fix(api): correct argument types in `wrap_node` and `wrap_node_or_nil`

devxpain:
The `wrap_node` and `wrap_node_or_nil` functions now correctly accept `Node?` to handle nil values, resolving a warning about incorrect argument counts in `api.tree.change_root_to_node()`.


alex-courtis (approved):
Many thanks for the fix @devxpain 

API is a bit of a mess right now, but it will be properly tidied and annotated during the long running project [Multiple Instances](https://github.com/orgs/nvim-tree/projects/1)

Issue #2262, opened 2023-06-10
Previous window is not selected unless :NvimTreeFocus is used


zhimsel:
### Description

From https://github.com/nvim-tree/nvim-tree.lua/issues/1985#issuecomment-1582896807

> Is there really no way to open to the window that was previously active before selecting the nvim-tree window (regardless on navigation method)?
> 
> I have the same problem as @paulrouget, however I often use the mouse to start interacting with nvim-tree (to open a file, for instance). If I don't use :NvimTreeFocus, then it opens the file in the last window to have previously used :NvimTreeFocus (or presumably have been created by nvim-tree).
> 
> Would it be possible to use the lua equivalent of [:winnr(#)](https://neovim.io/doc/user/builtin.html#winnr()) (if it exists)?
> 
> Would it be possible to track the previous window by an event that fires every time you leave a window? Like maybe the WinLeave [autocmd event](https://neovim.io/doc/user/autocmd.html#autocmd-events)? We could add another WinLeave autocmd similar to [the one defined here](https://github.com/nvim-tree/nvim-tree.lua/blob/f5d970d4506f385b29534252d8c15a782fa53034/lua/nvim-tree.lua#L320-L327), but using a negated pattern to ignore unwanted buffers like NvimTree_*.

---

## TL;DR

This plugin can only track the "previous active window" when it's focused via `:NvimTreeFocus`. If the nvim-tree is focused by any other method, it will still refer to the window where `:NvimTreeFocus` was lat called, not the _actual_ last window.

### Neovim version

```text
NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux 6.2.12-zen1-1-zen, MacOS 13.4

### nvim-tree version

Latest master (034511714bacfadc5008e49f73fcef67e5613840 as of testing)

### Minimal config

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
[...]

PR #3008, opened 2024-11-19
feat: add option to set last active window as target for file actions

devxpain:
- Add new option `set_last_win_as_target` to `opts.actions`
- Update documentation to reflect new option
- Set default value of `set_last_win_as_target` to `true`

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -570,6 +570,7 @@ Following is the default configuration. See |nvim-tree-opts| for details. >lua
       },
       actions = {
         use_system_clipboard = true,
+        set_last_win_as_target = true,
         change_dir = {
           enable = true,
           global = false,
@@ -1432,6 +1433,14 @@ Useful when path is not in `.gitignore` or git integration is disabled.
 A boolean value that toggle the use of system clipboard when copy/paste
 function are invoked. When enabled, copied text will be stored in registers
 '+' (system), otherwise, it will be stored in '1' and '"'.
+  Type: `boolean`, Default: `true`
+
+*nvim-tree.actions.set_last_win_as_target*
+A boolean value that toggles the behavior of setting the last active window as
+the target window for `nvim-tree`. When enabled, the last active window will be
+set as the target for file actions (like opening files), ensuring better window
+management. When disabled, the default behavior applies.
+
   Type: `boolean`, Default: `true`
 
 *nvim-tree.actions.change_dir*
@@ -2887,6 +2896,7 @@ highlight group is not, hard linking as follows: >
 |nvim-tree.actions.open_file.window_picker.picker|
 |nvim-tree.actions.remove_file.close_window|
 |nvim-tree.actions.use_system_clipboard|
+|nvim-tree.actions.set_last_win_as_target|
 |nvim-tree.auto_reload_on_write|
 |nvim-tree.diagnostics.debounce_delay|
 |nvim-tree.diagnostics.enable|
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -570,7 +570,7 @@ Following is the default configuration. See |nvim-tree-opts| for details. >lua
       },
       actions = {
         use_system_clipboard = true,
-        set_last_win_as_target = true,
+        set_last_win_as_target = false,
         change_dir = {
           enable = true,
           global = false,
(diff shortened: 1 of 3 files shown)


alex-courtis:
Does `api.node.open.no_window_picker()` achieve the functionality you are after?

The `actions.open_file.window_picker` apply even when the `window_picker.enable` is false, as does the above API. This needs to be properly documented, see similar https://github.com/nvim-tree/nvim-tree.lua/discussions/3005

doc/nvim-tree-lua.txt
```diff
@@ -570,6 +570,7 @@ Following is the default configuration. See |nvim-tree-opts| for details. >lua
       },
       actions = {
         use_system_clipboard = true,
+        set_last_win_as_target = true,
```
alex-courtis, line 573: This would be a change to existing behaviour; we'll need to default to false.

lua/nvim-tree.lua
```diff
@@ -232,6 +232,17 @@ local function setup_autocommands(opts)
       end,
     })
   end
+
+  if opts.actions.set_last_win_as_target then
+    create_nvim_tree_autocmd("BufLeave", {
```
alex-courtis, line 237: Please avoid autocommands; this should be done in `open-file.lua`

lua/nvim-tree.lua
```diff
@@ -232,6 +232,17 @@ local function setup_autocommands(opts)
...
+      callback = function()
+        local buf_name = vim.api.nvim_buf_get_name(0)
+        if not string.find(buf_name, "NvimTree_") then
```
alex-courtis, line 240: `utils.is_nvim_tree_buf` should be used.

lua/nvim-tree.lua
```diff
@@ -232,6 +232,17 @@ local function setup_autocommands(opts)
       end,
     })
   end
+
+  if opts.actions.set_last_win_as_target then
+    create_nvim_tree_autocmd("BufLeave", {
+      callback = function()
+        local buf_name = vim.api.nvim_buf_get_name(0)
```
alex-courtis, line 239: bufnr / bufname should come from the `ev` argument passed. See `:help nvim_create_autocmd()`

alex-courtis:
Looking at similar issue #2262 `vim.fn.winnr("#")` may be a more reliable approach.

devxpain:
I’ve updated the code as requested and reviewed the behavior of api.node.open.no_window_picker(). I didn’t encounter any issues.

Looking at similar issue #2262, I see vim.fn.winnr("#") mentioned as a more reliable approach. I’m not sure how it applies to this pull request. Could you clarify or provide an example?

Let me know if there’s anything else that needs adjustment!

alex-courtis:
> Looking at similar issue #2262, I see vim.fn.winnr("#") mentioned as a more reliable approach. I’m not sure how it applies to this pull request. Could you clarify or provide an example?

The idea is to use the actual previous window, according to nvim, rather than the one tracked via `lib.target_winid`.

It could be something like:
```lua
local function get_target_winid(mode)
  local target_winid
  if not M.window_picker.enable or mode == "edit_no_picker" or mode == "preview_no_picker" then
    target_winnr = vim.fn.winnr("#")
```

We would likely need to check that against `usable_win_ids` and fall back to first.

Looking through other issues it seems that this is not the best approach: https://github.com/nvim-tree/nvim-tree.lua/issues/2262#issuecomment-1585963649

Let's stick with tracking.

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -426,6 +426,15 @@ function M.setup(opts)
     opts.actions.open_file.window_picker.chars = tostring(opts.actions.open_file.window_picker.chars):upper()
   end
   M.window_picker = opts.actions.open_file.window_picker
+  if opts.actions.set_last_win_as_target then
+    vim.api.nvim_create_autocmd("BufLeave", {
+      callback = function()
+        if utils.is_nvim_tree_buf then
```
alex-courtis, line 432: This always evaluates to true, as it's not actually calling the function.

alex-courtis:
I'm sorry, this is not a change we should make:
* window selection code is fragile enough already
* bypasses or invalidates other window selection options
* will have undesirable effects or possibly break in some cases when using multiple tabs; this will be resolved in https://github.com/nvim-tree/nvim-tree.lua/issues/2255

I think the best option for you is to write a `actions.open_file.window_picker.picker` function that exactly meets your use case.
You could perform your own window tracking and return your desired choice from that function.
You can even invoke a window picker in situations you choose e.g. `s1n7ax/nvim-window-picker`

PR #3010, opened 2024-11-23
fix(#3009): nvim < 0.10 apply view options locally

### LINKED ISSUE #3009: nvim-tree buffer options applied to entire window
### Description

I have set settings like `relativenumber`, `number` and `colorcolumn` in my vimrc which is sourced in my `init.lua`. After updating nvim-tree, I found that these settings were getting overridden by the plugin. I was able to trace the behaviour change to https://github.com/nvim-tree/nvim-tree.lua/pull/2968 which introduced the following lines
https://github.com/nvim-tree/nvim-tree.lua/blob/1f3ffd6af145af2a4930a61c50f763264922c3fe/lua/nvim-tree/view.lua#L167-L169

Before this change, the lines were
https://github.com/nvim-tree/nvim-tree.lua/blob/d41b4ca013ed89e41b9c0ecbdae5f1633e42f7fa/lua/nvim-tree/view.lua#L150-L152

The reason for the behaviour change seems to be that `vim.opt_local` was applying the settings only to the current buffer while `vim.api.nvim_win_set_option` is applying the settings to the entire window.

From my understanding, the benefit of using `vim.api.nvim_win_set_option` is to support earlier versions of nvim. I'm not familiar with lua or the nvim API, so I don't know the exact fix if the code can't be reverted to use `vim.opt_local`.

### Neovim version

```text
NVIM v0.9.5
Build type: Release
LuaJIT 2.1.1700206165
```

Debian

### Windows variant

_No response_

### nvim-tree version

v1.7

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1
vim.opt.colorcolumn = "80"

local function load_plugins()
    require("packer").startup({
        {
            "wbthomason/packer.nvim",
            { "nvim-tree/nvim-tree.lua", tags = "v1.6.1" },
        },
        config = {
            package_root = package_root,
        },
    })
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
    require("nvim-tree").setup()
end
```

### Steps to reproduce

1. touch ./abc.txt
2. nvim -nu /tmp/nvt-min.lua .
3. :NvimTreeToggle
4. open abc.txt

### Expected behavior

the `colorcolumn` at col 80 should show up. It does so up till `nvim-tree` v1.6.1

### Actual behavior

You will see that the `colorcolumn` is not visible.

alex-courtis:
fixes #3009

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -164,8 +164,10 @@ local function set_window_options_and_buffer()
     local eventignore = vim.api.nvim_get_option("eventignore") ---@diagnostic disable-line: deprecated
     vim.api.nvim_set_option("eventignore", "all") ---@diagnostic disable-line: deprecated
 
+    -- #3009 vim.api.nvim_win_set_option does not set local scope without explicit winid.
+    -- Revert to opt_local instead of propagating it through for just the 0.10 path.
     for k, v in pairs(M.View.winopts) do
-      vim.api.nvim_win_set_option(0, k, v) ---@diagnostic disable-line: deprecated
+      vim.opt_local[k] = v
     end
 
     vim.api.nvim_set_option("eventignore", eventignore) ---@diagnostic disable-line: deprecated


arjan-bal (approved):
LGTM! Verified that this fixes the issue locally.

Issue #3018, opened 2024-12-06
Vim(normal):Can't re-enter normal mode from terminal mode


shy-robin:
### Description

When I focus on nvim tree panel, I open vim-floaterm and switch floaterm, it occurs error:

```
Error executing vim.schedule lua callback: ...n/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:201: Error executing lua: vim/_editor.lua:0: nvim_exec2(): Vim(normal):Can't re-enter normal mode from terminal mode
stack traceback:
	[C]: in function 'nvim_exec2'
	vim/_editor.lua: in function 'cmd'
	...n/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:206: in function <...n/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:201>
	[C]: in function 'nvim_buf_call'
	...n/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:201: in function <...n/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:200>
stack traceback:
	[C]: in function 'nvim_buf_call'
	...n/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:201: in function <...n/.local/share/nvim/lazy/nvim-tree.lua/lua/nvim-tree.lua:200>
```

https://github.com/user-attachments/assets/84cf59df-203d-489a-9917-367580ae572a

I have read the source code of nvim-tree.lua and I found the reason. When I switch floaterm, the cursor will focus on nvim-tree panel quickly and the mode now is the terminal mode. Because I config `centralize_selection = true`, it will use `zz` to centralize selection but now is terminal mode, cannot use `zz` command and it causes error.

The relevant source code is as follows:

```lua
  if opts.view.centralize_selection then
    create_nvim_tree_autocmd("BufEnter", {
      pattern = "NvimTree_*",
      callback = function()
        vim.schedule(function()
          vim.api.nvim_buf_call(0, function()
            vim.cmd([[norm! zz]]) -- -> causes error in terminal mode
          end)
        end)
      end,
    })
  end
```

So I think we should first determine whether the mode is terminal mode or not, like this:

```lua
  if opts.view.centralize_selection then
    create_nvim_tree_autocmd("BufEnter", {
      pattern = "NvimTree_*",
      callback = function()
        vim.schedule(function()
[...]

PR #3019, opened 2024-12-06
fix(#3018): error when focusing nvim-tree when in terminal mode

shy-robin:
fix: https://github.com/nvim-tree/nvim-tree.lua/issues/3018


alex-courtis (approved):

alex-courtis:
I think I replicated the issue using `<C-\><C-n>` and it seems to be fixed.

PR #3020, opened 2024-12-07
fix: view.width functions may return strings

ljie-PI:
In the function `get_size()`, if the parameter `size` is a function which return string like "20%", the function returns the string instead of a number. It fails when `vim.api.nvim_win_set_width(winnr, new_size)` to set the width.

The `get_size()` funtion can parse such strings, so invoke the function again to fix this.


alex-courtis:
Works for:

```lua
    width = {
      max = function()
        return "40%"
      end,
      min = function()
        return "30%"
      end,
      padding = function()
        return "1%"
      end,
    },
```

alex-courtis (approved):
Many thanks @ljie-PI !

Updated doc myself for quick turnaround.

Issue #3015, opened 2024-12-03
Do not ignore `right_aligned` icons when computing max window width


simonmandlik:
**Is your feature request related to a problem? Please describe.**
I like the nvim-tree window to take as much width as needed as long as it doesn't cover over `60%` of the screen. For this reason, I'm using

```lua
view.width = {
   min = -1,
   max=60%
}
```

This works very nice, but when used together with `right_align`ed icon extmarks, the longest paths are not always shown in full, even if there is plenty of width in the window. This is likely because the max width computation doesn't take these extmarks into account.

It would be great to have the width computation take extmarks into account.

I guess the modification could be done here?
https://github.com/nvim-tree/nvim-tree.lua/blob/ca7c4c33cac2ad66ec69d45e465379716ef0cc97/lua/nvim-tree/view.lua#L302-L335

alex-courtis:
> I guess the modification could be done here?

That looks correct, a pull request would be most gratefully appreciated, see [CONTRIBUTING.md](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md)

> `max=60%`

Should this be `max="60%"`

simonmandlik:
> a pull request would be most gratefully appreciated

Here https://github.com/nvim-tree/nvim-tree.lua/pull/3022

> Should this be `max="60%"`

Sure, sorry, made a typo here :)

PR #3022, opened 2024-12-08
fix(#3015): dynamic width no longer truncates on right_align icons

simonmandlik:
Fixes https://github.com/nvim-tree/nvim-tree.lua/issues/3015

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -321,8 +321,19 @@ local function grow()
     max_width = get_width(M.View.max_width) - padding
   end
 
-  for _, l in pairs(lines) do
+  local ns_id = vim.api.nvim_get_namespaces()["NvimTreeExtmarks"]
+  for line_nr, l in pairs(lines) do
     local count = vim.fn.strchars(l)
+    -- also add space for right-aligned icons
+    local extmarks = vim.api.nvim_buf_get_extmarks(M.get_bufnr(), ns_id, { line_nr, 0 }, { line_nr, -1 }, { details = true })
+    for _, extmark in ipairs(extmarks) do
+      local virt_texts = extmark[4].virt_text
+      if virt_texts then
+        for _, virt_text in ipairs(virt_texts) do
+          count = count + vim.fn.strchars(virt_text[1])
+        end
+      end
+    end
     if resizing_width < count then
       resizing_width = count
     end


alex-courtis:
Thanks for the fix, however I'm having trouble replicating and validating it.

Please provide a [clean room configuration](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development#clean-room) and instructions on how I can demonstrate the fix with it.

simonmandlik:
Instructions:

1. run `nvim -nu minimal.lua`
2. Edit the longest file in any git repo. Here, I use `nvim-tree` directly so I change something in `.release-please-manifest.json` so that it get a right-aligned icon
3. This is the bug

image

Even though the max width is set to 100%, the whole filename is not shown due to the icon covering a part of it.

With this PR, it looks like this

image

`minimal.lua`:
```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      -- "~/Downloads/nvim-tree.lua",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup({
    view = {
      width = {
        min = -1,
        max = "100%",
        padding = 0
      }
    },
    renderer = {
      icons = {
        git_placement = "right_align"
      }
    }
  })
end
```

simonmandlik:
Addressed the CI issues

alex-courtis:
This change does address the issue for width whilst focused. Unfortunately the width when changing is a regression.

View is quite fragile. It is targeted for a refactor during https://github.com/orgs/nvim-tree/projects/1 however that will not address this issue.

A full solution is desirable, let's see what we can come up with.

lua/nvim-tree/view.lua
```diff
@@ -321,8 +321,19 @@ local function grow()
     max_width = get_width(M.View.max_width) - padding
   end
 
-  for _, l in pairs(lines) do
+  local ns_id = vim.api.nvim_get_namespaces()["NvimTreeExtmarks"]
+  for line_nr, l in pairs(lines) do
     local count = vim.fn.strchars(l)
+    -- also add space for right-aligned icons
+    local extmarks = vim.api.nvim_buf_get_extmarks(0, ns_id, { line_nr, 0 }, { line_nr, -1 }, { details = true })
```
alex-courtis, line 328: This looks like it might be the issue: we're getting extmarks for the current buffer rather than the tree buffer.

simonmandlik: Thank you, fixed!

alex-courtis (approved):
Works nicely for a variety of width cases.
Many thanks for resolving this one!

PR #3042, opened 2025-01-02
fix(#3041): use vim.diagnostic.get for updating diagnostics

### LINKED ISSUE #3041: Diagnostic indicators are based only on messages from the "DiagnosticChanged" event, not from the entire buffer
### Description

Diagnostic indicators are based only on messages from the "DiagnosticChanged" event, not from the entire buffer

I use nvim for Go (GoLang) development. I set up the lsp and linter using https://github.com/mfussenegger/nvim-lint.

I created my own autocmd: when saving the buffer (BufWritePost event) using the plugin https://github.com/mfussenegger/nvim-lint the linter is started, and its messages are added to diagnostics for the current buffer with the warning level.

In my case, it works like this:
1) I save the current buffer
2) LSP adds diagnostics
3) The linter is triggered
4) The linter adds diagnostics

When updating diagnostics, nvim-tree processes only those diagnoses that are contained in the "DiagnosticChanged" event. This is a problem for me because the "DiagnosticChanged" event occurs two times: in step 2 and in step 4.

If the diagnostics added by the linter will not contain diagnostic with the error level, and in the nvim-tree settings I will specify:

```
severity = {
  min = vim.diagnostic.severity.ERROR,
  max = vim.diagnostic.severity.ERROR,
}
```

then the diagnostics will not be displayed at all for the file in the nvim-tree.

I think it's more correct to analyze the entire diagnostics list for the buffer, which can be obtained using the `vim.diagnostic.get` function.

### Neovim version

```text
NVIM v0.10.2
Build type: RelWithDebInfo
LuaJIT 2.1.1731601260
```

Linux 5.15.173-1-MANJARO

### Windows variant

_No response_

### nvim-tree version

master

### Clean room replication

```lua
Not provided
```

### Steps to reproduce

Not provided

### Expected behavior

_No response_

### Actual behavior

_No response_

Yu-Leo:
Fixes #3041

### DIFF
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -128,7 +128,6 @@ function M.update_lsp(ev)
 
   local profile_event = log.profile_start("DiagnosticChanged event")
 
-  ---@type vim.Diagnostic[]
   local diagnostics = vim.diagnostic.get(ev.buf)
 
   -- use the buffer from the event, as ev.data.diagnostics will be empty on resolved diagnostics
--- a/lua/nvim-tree/diagnostics.lua
+++ b/lua/nvim-tree/diagnostics.lua
@@ -129,7 +129,7 @@ function M.update_lsp(ev)
   local profile_event = log.profile_start("DiagnosticChanged event")
 
   ---@type vim.Diagnostic[]
-  local diagnostics = ev.data.diagnostics
+  local diagnostics = vim.diagnostic.get(ev.buf)
 
   -- use the buffer from the event, as ev.data.diagnostics will be empty on resolved diagnostics
   local bufname = uniformize_path(vim.api.nvim_buf_get_name(ev.buf))


alex-courtis (approved):
Many thanks for deep diving.

I did wonder why there were two instances of `Diagnostics[]` however this looks far more reliable.

In the interests of speed I'll remove the now unnecessary @type

lua/nvim-tree/diagnostics.lua
```diff
@@ -128,8 +128,7 @@ function M.update_lsp(ev)
 
   local profile_event = log.profile_start("DiagnosticChanged event")
 
-  ---@type vim.Diagnostic[]
-  local diagnostics = ev.data.diagnostics
+  local diagnostics = vim.diagnostic.get(ev.buf)
```
alex-courtis, line 131: Looks safe for 0.10

PR #3044, opened 2025-01-06
fix: hijack directory "BufEnter", "BufNewFile" events are nested

phanen:
I have some autocmds on `DirChanged` to record my directory history.

Currently, with `hijack_directories` enabled, `gf` or `:e /some/dir` will jump to directory without triggering the `DirChanged` events.


alex-courtis (approved):
Good catch.

This has some potential for unforseen breakages or user issues. Please watch https://github.com/nvim-tree/nvim-tree.lua/issues in case a user reports an issue so that we can fix it fast.

Issue #2896, opened 2024-09-07
Documentation: API luals Annotations


alex-courtis:
Add luals annotations for all API functions, incorporating the existing named param classes.

alex-courtis:
No longer part of multi-instance.

alex-courtis:
Removed opts luadoc.

alex-courtis:
Splitting.

alex-courtis:
?

alex-courtis:
Consider generation of help from API, possibly via another tool.

Issue #2832, opened 2024-07-13
Multi Instance: Refactor: API Accept winid


alex-courtis:
Many API families such as `api.marks` operate globally.

Add options to all API that are window specific to take a `winid`.

```
      • {winid}          (number)         apply to the specified winid, otherwise first currently visible
```

This will be functional, however the fallback will be current behaviour.

PR #3054, opened 2025-01-26
feat(#1984): add quit_on_open and focus opts to various api.node.open functions

### LINKED ISSUE #1984: Open a file and keep tree open, but another hotkey for open file and close the tree afterwards
### Discussed in https://github.com/nvim-tree/nvim-tree.lua/discussions/1980

<sup>Originally posted by **mangelozzi** February  9, 2023</sup>
I found there is an option: `nvim-tree.actions.open_file.quit_on_open`, but the problem with this it always closes the tree afterwards.
I wish to have `o` perform as default (keep the tree open after opening the file), but have `<CR>` open the buffer and close the tree.

Currently these seems to be the only options (tried them all):
```
`<CR>`            edit                open a file or folder; root will cd to the above directory
`o`
`<2-LeftMouse>`
`<C-e>`           edit_in_place       edit the file in place, effectively replacing the tree explorer
`O`               edit_no_picker      same as (edit) with no window picker
```
`edit_in_place`  is similar but then I have 2 windows after opening the file, and have to go to the other window and close it.

Looking for something like this:
```
require("nvim-tree").setup({
    view = {
        mappings = {
            list = {
                { key = "<CR>", action = "quit_on_open" },
            },
        },
    },    
})
````</div>

### PR DESCRIPTION
**Pull request by GCrispino**
fixes #1984

As discussed on #1984 , this should implement the ability of specifying the following two options in the `api.node.open.edit` method:

- `quit_on_open`: if set to true, it closes the tree when file related to the node is opened;
- `focus`: if set to ~~true~~ false, opens the file but keep the focus on the tree. Similar to what's done in the preview feature, but actually creating a new buffer for each file.

For now I've added a draft commit that supports the first option and then just checks it and closes the tree view in the edit API method.
Not sure if that's the best option, but at first I decided to not forward it to the action level to not mix it with the `quit_on_open` global option that can be passed in to the action, and also to keep the changes restricted to the API code only.

Thoughts?

### DIFF
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -239,12 +239,14 @@ local function edit(mode, node, edit_opts)
 
   edit_opts = edit_opts or {}
 
-  if edit_opts.quit_on_open then
+  local mode_unsupported_quit_on_open = mode == "drop" or mode == "tab_drop" or mode == "edit_in_place"
+  if not mode_unsupported_quit_on_open and edit_opts.quit_on_open then
     view.close(cur_tabpage)
   end
 
+  local mode_unsupported_focus = mode == "drop" or mode == "tab_drop" or mode == "edit_in_place"
   local focus = edit_opts.focus == nil or edit_opts.focus == true
-  if not focus then
+  if not mode_unsupported_focus and not focus then
     -- if mode == "tabnew" a new tab will be opened and we need to focus back to the previous tab
     if mode == "tabnew" then
       vim.cmd(":tabprev")
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -233,12 +233,14 @@ Api.fs.copy.relative_path = wrap_node(wrap_explorer_member("clipboard", "copy_pa
 local function edit(mode, node, edit_opts)
   local file_link = node:as(FileLinkNode)
   local path = file_link and file_link.link_to or node.absolute_path
+  local cur_tabpage = vim.api.nvim_get_current_tabpage()
+
   actions.node.open_file.fn(mode, path)
 
   edit_opts = edit_opts or {}
 
   if edit_opts.quit_on_open then
-    view.close()
+    view.close(cur_tabpage)
   end
 
   local focus = edit_opts.focus == nil or edit_opts.focus == true
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -244,6 +244,7 @@ end
 ---@return fun(node: Node, edit_opts: NodeEditOpts?)
 local function open_or_expand_or_dir_up(mode, toggle_group)
   ---@param node Node
+  ---@param edit_opts NodeEditOpts?
   return function(node, edit_opts)
     local root = node:as(RootNode)
     local dir = node:as(DirectoryNode)
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -243,6 +243,10 @@ local function edit(mode, node, edit_opts)
 
   local focus = edit_opts.focus == nil or edit_opts.focus == true
   if not focus then
+    -- if mode == "tabnew" a new tab will be opened and we need to focus back to the previous tab
+    if mode == "tabnew" then
+      vim.cmd(":tabprev")
+    end
     view.focus()
   end
 end
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -225,6 +225,7 @@ Api.fs.copy.relative_path = wrap_node(wrap_explorer_member("clipboard", "copy_pa
 ---
 ---@class NodeEditOpts
 ---@field quit_on_open boolean|nil default false
+---@field focus boolean|nil default true
 
 ---@param mode string
 ---@param node Node
@@ -234,8 +235,15 @@ local function edit(mode, node, edit_opts)
   local path = file_link and file_link.link_to or node.absolute_path
   actions.node.open_file.fn(mode, path)
 
-  if edit_opts and edit_opts.quit_on_open then
+  edit_opts = edit_opts or {}
+
+  if edit_opts.quit_on_open then
     view.close()
+  end
+
+  local focus = edit_opts.focus == nil or edit_opts.focus == true
+  if not focus then
+    view.focus()
   end
 end
 
--- a/lua/nvim-tree/api.lua
+++ b/lua/nvim-tree/api.lua
@@ -222,21 +222,29 @@ Api.fs.copy.absolute_path = wrap_node(wrap_explorer_member("clipboard", "copy_ab
 Api.fs.copy.filename = wrap_node(wrap_explorer_member("clipboard", "copy_filename"))
 Api.fs.copy.basename = wrap_node(wrap_explorer_member("clipboard", "copy_basename"))
 Api.fs.copy.relative_path = wrap_node(wrap_explorer_member("clipboard", "copy_path"))
+---
+---@class NodeEditOpts
+---@field quit_on_open boolean|nil default false
 
 ---@param mode string
 ---@param node Node
-local function edit(mode, node)
+---@param edit_opts NodeEditOpts?
+local function edit(mode, node, edit_opts)
   local file_link = node:as(FileLinkNode)
   local path = file_link and file_link.link_to or node.absolute_path
   actions.node.open_file.fn(mode, path)
+
+  if edit_opts and edit_opts.quit_on_open then
+    view.close()
+  end
 end
 
 ---@param mode string
 ---@param toggle_group boolean?
----@return fun(node: Node)
+---@return fun(node: Node, edit_opts: NodeEditOpts?)
 local function open_or_expand_or_dir_up(mode, toggle_group)
   ---@param node Node
-  return function(node)
+  return function(node, edit_opts)
     local root = node:as(RootNode)
     local dir = node:as(DirectoryNode)
 
@@ -245,7 +253,7 @@ local function open_or_expand_or_dir_up(mode, toggle_group)
     elseif dir then
       dir:expand_or_collapse(toggle_group)
     elseif not toggle_group then
-      edit(mode, node)
+      edit(mode, node, edit_opts)
     end
   end
 end
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -2007,10 +2007,18 @@ fs.print_clipboard()                      *nvim-tree-api.fs.print_clipboard()*
 Parameters: ~
   • {node} (Node|nil) file or folder
 
-node.open.edit({node})                        *nvim-tree-api.node.open.edit()*
+node.open.edit({node}, {opts})                        *nvim-tree-api.node.open.edit()*
     File:   open as per |nvim-tree.actions.open_file|
     Folder: expand or collapse
     Root:   change directory up
+
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
 
                                *nvim-tree-api.node.open.replace_tree_buffer()*
 node.open.replace_tree_buffer({node})
@@ -2018,31 +2026,79 @@ node.open.replace_tree_buffer({node})
     nvim-tree window.
 
                                   *nvim-tree-api.node.open.no_window_picker()*
-node.open.no_window_picker({node})
+node.open.no_window_picker({node}, {opts})
     |nvim-tree-api.node.edit()|, window picker will never be used as per
     |nvim-tree.actions.open_file.window_picker.enable| `false`
 
-node.open.vertical({node})                *nvim-tree-api.node.open.vertical()*
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
+
+node.open.vertical({node}, {opts})                *nvim-tree-api.node.open.vertical()*
     |nvim-tree-api.node.edit()|, file will be opened in a new vertical split.
 
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
+
                                 *nvim-tree-api.node.open.vertical_no_picker()*
-node.open.vertical_no_picker({node})
+node.open.vertical_no_picker({node}, {opts})
     |nvim-tree-api.node.vertical()|, window picker will never be used as per
     |nvim-tree.actions.open_file.window_picker.enable| `false`
 
-node.open.horizontal({node})            *nvim-tree-api.node.open.horizontal()*
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
+
+node.open.horizontal({node}, {opts})            *nvim-tree-api.node.open.horizontal()*
     |nvim-tree-api.node.edit()|, file will be opened in a new horizontal split.
 
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
+
                               *nvim-tree-api.node.open.horizontal_no_picker()*
-node.open.horizontal_no_picker({node})
+node.open.horizontal_no_picker({node}, {opts})
     |nvim-tree-api.node.horizontal()|, window picker will never be used as per
     |nvim-tree.actions.open_file.window_picker.enable| `false`
 
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
+
                                 *nvim-tree-api.node.open.toggle_group_empty()*
-node.open.toggle_group_empty({node})
+node.open.toggle_group_empty({node}, {opts})
     Toggle |nvim-tree.renderer.group_empty| for a specific folder.
     Does nothing on files.
     Needs |nvim-tree.renderer.group_empty| set.
+
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
 
 node.open.drop({node})                        *nvim-tree-api.node.open.drop()*
     Switch to window with selected file if it exists.
@@ -2053,8 +2109,16 @@ node.open.drop({node})                        *nvim-tree-api.node.open.drop()*
     Folder: expand or collapse
     Root:   change directory up
 
-node.open.tab({node})                          *nvim-tree-api.node.open.tab()*
+node.open.tab({node}, {opts})                          *nvim-tree-api.node.open.tab()*
     |nvim-tree-api.node.edit()|, file will be opened in a new tab.
+
+    Parameters: ~
+      • {node} (Node|nil) file or folder
+      • {opts} (table) optional parameters
+
+    Options: ~
+      • {quit_on_open} (boolean) quits the tree when opening the file
+      • {focus} (boolean) keep focus in the tree when opening the file
 
                                           *nvim-tree-api.node.open.tab_drop()*
 node.open.tab_drop({node})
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: lua/nvim-tree/api.lua [hunk 1]
```diff
@@ -222,21 +222,29 @@ Api.fs.copy.absolute_path = wrap_node(wrap_explorer_member("clipboard", "copy_ab
...
+    view.close()
+  end
 end
 
 ---@param mode string
 ---@param toggle_group boolean?
----@return fun(node: Node)
+---@return fun(node: Node, edit_opts: NodeEditOpts?)
 local function open_or_expand_or_dir_up(mode, toggle_group)
   ---@param node Node
```
alex-courtis, line 246: Missing param, CI got upset: https://github.com/nvim-tree/nvim-tree.lua/actions/runs/12975816599/job/36187330634?pr=3054

Needs
```lua
  ---@param edit_opts NodeEditOpts?
```

I recommend setting up [wiki: Lua Language Server](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development#lua-language-server), which will show a warning about these sorts of things.

alex-courtis (commented):
Works beautifully:
```lua
  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.edit(nil, { quit_on_open = true })
  end, opts("Open Quit"))
```
Handles directories, root etc. as expected.

> For now I've added a draft commit that supports the first option and then just checks it and closes the tree view in the edit API method.
> Not sure if that's the best option, but at first I decided to not forward it to the action level to not mix it with the quit_on_open global option that can be passed in to the action, and also to keep the changes restricted to the API code only.

I like that approach, it's really clear what's going on. The `open_file` is far too complex and needs to be split up into file opening and post-actions; this is the first step.

alex-courtis:
If it helps, you're a member of the project and can now push branches directly, rather than your fork.

I've added "fixes #1984" to the description: that issue will be automatically closed once this is merged.

gegoune:
@GCrispino Hi! 👋 quick bit: instead of adding WIP to subject you can open draft PR which is a way clearer way to communicate. 

Congrats on your first PR here. :)

GCrispino:
@alex-courtis 

> Works beautifully:
> 
> ```lua
>   vim.keymap.set("n", "<S-CR>", function()
>     api.node.open.edit(nil, { quit_on_open = true })
>   end, opts("Open Quit"))
> ```
> 
> Handles directories, root etc. as expected.
> 
> > For now I've added a draft commit that supports the first option and then just checks it and closes the tree view in the edit API method.
> > Not sure if that's the best option, but at first I decided to not forward it to the action level to not mix it with the quit_on_open global option that can be passed in to the action, and also to keep the changes restricted to the API code only.
> 
> I like that approach, it's really clear what's going on. The `open_file` is far too complex and needs to be split up into file opening and post-actions; this is the first step.

Great! Will continue following this then 👍🏼 


> If it helps, you're a member of the project and can now push branches directly, rather than your fork.
> 
> I've added "fixes #1984" to the description: that issue will be automatically closed once this is merged.

Oh yeah, totally forgot about this 😅 . Definitely easier to just push a new branch into the project, will do this on the next one, thanks for the reminder 👍🏼 

----

@gegoune 

> @GCrispino Hi! 👋 quick bit: instead of adding WIP to subject you can open draft PR which is a way clearer way to communicate.

Oh yeah sorry wasn't aware GitHub had draft PRs - set this one to be draft for now and removed the "WIP" prefix, thanks for the recommendation 👍🏼 

> Congrats on your first PR here. :)

Thank you!

lua/nvim-tree/api.lua
GCrispino: Ok will do, thanks :)

GCrispino:
I believe this is now done, so removing the draft status.

Something I'd like to note is that, since other methods on api.node (such as `api.node.open.vertical` and `api.node.open.horizontal`) wrap the `edit` function that is modified in this PR, this means that these methods will implicitly gain support for the two new options added here.

Would this be a problem, or if this isn't, would we need to add documentation anywhere for this?

alex-courtis:
> this means that these methods will implicitly gain support for the two new options added here.

Wow... they do indeed. We'll need to test them all...

> would we need to add documentation anywhere for this?

Proposal:

add `{opts}` param to each `nvim-tree-api.node.open.` function in 6.3.
add a note something like this to cover them all

```
==============================================================================
 6.3 API NODE                                             *nvim-tree-api.node*

Parameters for all `nvim-tree-api.node` functions: ~
  • {node} (Node|nil) file or folder

Parameters for all `nvim-tree-api.node.open` functions: ~

Options: ~
    • {force} (boolean) delete even if buffer is modified, default false
    • {quit_on_open} (boolean) ...
```

alex-courtis:
Testing OK, one failure with `.tab`

I didn't have the energy to test all open variants...

Can I please leave the `drop` functions with you?

```lua
function M.on_attach(bufnr)
  local function opts(desc)
    return { desc = "nvim-tree: " .. desc, buffer = bufnr, noremap = true, silent = true, nowait = true }
  end

  api.config.mappings.default_on_attach(bufnr)

  vim.keymap.set("n", "?", api.tree.toggle_help, opts("Help"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.edit(nil, {})
  end, opts("Edit"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.edit(nil, { quit_on_open = true, focus = false, })
  end, opts("Edit Quit No Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.edit(nil, { quit_on_open = true, focus = true, })
  end, opts("Edit Quit Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.edit(nil, { quit_on_open = false, focus = false, })
  end, opts("Edit No Focus"))


  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.vertical(nil, {})
  end, opts("Vert"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.vertical(nil, { quit_on_open = true, focus = false, })
  end, opts("Vert Quit No Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.vertical(nil, { quit_on_open = true, focus = true, })
  end, opts("Vert Quit Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.vertical(nil, { quit_on_open = false, focus = false, })
  end, opts("Vert No Focus"))


  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.horizontal_no_picker(nil, {})
  end, opts("Horiz No Picker"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.horizontal_no_picker(nil, { quit_on_open = true, focus = false, })
  end, opts("Horiz No Picker Quit No Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.horizontal_no_picker(nil, { quit_on_open = true, focus = true, })
  end, opts("Horiz No Picker Quit Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.horizontal_no_picker(nil, { quit_on_open = false, focus = false, })
  end, opts("Horiz No Picker No Focus"))


  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.replace_tree_buffer(nil, {})
  end, opts("Replace Tree Buffer"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.replace_tree_buffer(nil, { quit_on_open = true, focus = false, })
  end, opts("Replace Tree Buffer Quit No Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.replace_tree_buffer(nil, { quit_on_open = true, focus = true, })
  end, opts("Replace Tree Buffer Quit Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.replace_tree_buffer(nil, { quit_on_open = false, focus = false, })
  end, opts("Replace Tree Buffer No Focus"))


  -- these do work, even thought they do not make much sense
  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.preview(nil, {})
  end, opts("Preview"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.preview(nil, { quit_on_open = true, focus = false, })
  end, opts("Preview Quit No Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.preview(nil, { quit_on_open = true, focus = true, })
  end, opts("Preview Quit Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.preview(nil, { quit_on_open = false, focus = false, })
  end, opts("Preview No Focus"))


  -- FAIL
  -- tree is not closed, but focus is correct
  -- skip options for this one?
  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.tab(nil, {})
  end, opts("Tab"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.tab(nil, { quit_on_open = true, focus = false, })
  end, opts("Tab Quit No Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.tab(nil, { quit_on_open = true, focus = true, })
  end, opts("Tab Quit Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.tab(nil, { quit_on_open = false, focus = false, })
  end, opts("Tab No Focus"))


  -- these do work, even thought they do not make much sense
  -- this function really should not be inside open
  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.toggle_group_empty(nil, {})
  end, opts("Toggle Group Empty"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.toggle_group_empty(nil, { quit_on_open = true, focus = false, })
  end, opts("Toggle Group Empty Quit No Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.toggle_group_empty(nil, { quit_on_open = true, focus = true, })
  end, opts("Toggle Group Empty Quit Focus"))

  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.toggle_group_empty(nil, { quit_on_open = false, focus = false, })
  end, opts("Toggle Group Empty No Focus"))



end
```

GCrispino:
> Proposal:
> 
> add `{opts}` param to each `nvim-tree-api.node.open.` function in 6.3. add a note something like this to cover them all
> 
> ```
> ==============================================================================
>  6.3 API NODE                                             *nvim-tree-api.node*
> 
>     Parameters for all `nvim-tree-api.node` functions: ~
>       • {node} (Node|nil) file or folder
> 
>     Optional {opts} for all `nvim-tree-api.node.open` functions: ~
>       • {force} (boolean) delete even if buffer is modified, default false
>       • {quit_on_open} (boolean) ...
> ```

I'm fine with this proposal 👍🏼 


> Testing OK, one failure with `.tab`
> 
> I didn't have the energy to test all open variants...
> 
> Can I please leave the `drop` functions with you?

Yeah sure, I just didn't have the time to give this another look.
I just have two questions regarding this.

1. On your `on_attach` function referenced on your latest message for testing the changes in this PR, you have multiple mappings for `<S-CR>`
1.1. do you comment all of them and leave the one you're trying to test uncommented, or do you have something special to make all of them work?
2. Has the addition of an automated testing tool/framework been discussed for nvim-tree? I'm a newbie in the nvim world so I don't really know how this is handled generally, but it seems like that would extremely helpful in cases like this one

alex-courtis:
> * 1.1. do you comment all of them and leave the one you're trying to test uncommented, or do you have something special to make all of them work?

Sorry, yes, I should have mentioned. Really hacky.

> * Has the addition of an automated testing tool/framework been discussed for nvim-tree? I'm a newbie in the nvim world so I don't really know how this is handled generally, but it seems like that would extremely helpful in cases like this one

Yes. It is very much needed, we just lack the time/people to select a framework and start building tests.

There's a refactoring task for `NodeIterator` and I'm honestly too scared to do it without tests.

alex-courtis:
> I just pushed some commits with the fix for the `tab` functions.
> 
> I still need to test the `drop` functions though

`tab` works as expected, as does `tab_drop`.

`drop` is not for these cases
```lua
  -- new file is opened, however the previous one remains in the window
  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.drop(nil, { quit_on_open = true, focus = false, })
  end, opts("Drop No Focus"))

  -- new file is opened, however the previous one remains in the window
  vim.keymap.set("n", "<S-CR>", function()
    api.node.open.drop(nil, { quit_on_open = true, focus = true, })
  end, opts("Drop Quit Focus"))
```

```sh
echo foo > foo
echo bar > bar
```

* open nvim
* open bar
* navigate to foo
* `api.node.open.drop(nil, { quit_on_open = true, focus = false, })`
* foo buffer is opened, bar is not dropped and remains in the window

Repeat for `api.node.open.drop(nil, { quit_on_open = true, focus = true, })`

lua/nvim-tree/api.lua
```diff
@@ -222,21 +222,44 @@ Api.fs.copy.absolute_path = wrap_node(wrap_explorer_member("clipboard", "copy_ab
...
+
+  if edit_opts.quit_on_open then
+    view.close(cur_tabpage)
+  end
+
+  local focus = edit_opts.focus == nil or edit_opts.focus == true
+  if not focus then
+    -- if mode == "tabnew" a new tab will be opened and we need to focus back to the previous tab
+    if mode == "tabnew" then
+      vim.cmd(":tabprev")
```
alex-courtis, line 250:
```suggestion
      vim.cmd.tabprev()
```

When neovim gets around to exposing these as proper API we will be ready.

alex-courtis:
CI is currently failing, see https://github.com/nvim-tree/nvim-tree.lua/pull/3060#issuecomment-2628628783

alex-courtis (changes requested):
Looking good!

You accidentally picked an issue that hits one of the most annoying, fragile parts of this codebase `node.open_file.fn` :)

- [ ] drop https://github.com/nvim-tree/nvim-tree.lua/pull/3054#issuecomment-2628669497
- [ ] add help https://github.com/nvim-tree/nvim-tree.lua/pull/3054#issuecomment-2620494062

GCrispino:
Hey @alex-courtis , apologies for the delay here, but I have been busy with other things in the past few weeks and haven't had the opportunity to test this further until today.

So I did a few tests and couldn't come up with a fix for these scenarios with the `api.node.open.drop`, but I have some small findings, perhaps you can have a clue on what's going on here:

**About the `focus` functionality on drop:**

It seems like when dropping the new nvim-tree window changes or something like that, and a different window is focused instead.

For example, when following the steps below with the `foo` and `bar` files created [here](https://github.com/nvim-tree/nvim-tree.lua/pull/3054#issuecomment-2628669497):
1. open nvim
2. open bar
3. run `:NvimTreeOpen`
4. run `:lua vim.print(vim.api.nvim_get_current_win())` - in my case it outputted 1002
5. run `:drop foo`
  - `foo` is opened 
6. navigate back to tree and run `:lua vim.print(vim.api.nvim_get_current_win()) again` - in my case it outputted **1003**

Even when testing with nvim-tree without my modified code, I could reproduce this.

This might be part of my lack of knowledge of vim's inner APIs, but I would expect that the second call to `vim.api.nvim_get_current_win()` yields 1002 again - or is it expected that the window gets modified?


**About the `close` functionality on drop:**

The nvim-tree window is closed indeed, but for some reason it doesn't open the new file (in your example, `foo`).

I don't know why that's the case, but I could accidentally find out that when putting an `vim.fn.input()` statement before [this line](https://github.com/nvim-tree/nvim-tree.lua/pull/3054/files#diff-69640a8794ab2834e7b05cd80b8c0516749f6e893813a81b4e71117f346e809fR243), just to pause execution and try understanding a bit more how this is working, after pressing enter for continuing execution, the `foo` file was opened as expected 🤔 .

Do you have any idea of why this happened? 😅

alex-courtis:
> Hey @alex-courtis , apologies for the delay here, but I have been busy with other things in the past few weeks and haven't had the opportunity to test this further until today.

There is never any pressure or rush; this is a community project with members all having limited availability. My window is also limited - usually Sun/Mon.

> So I did a few tests and couldn't come up with a fix for these scenarios with the `api.node.open.drop`, but I have some small findings, perhaps you can have a clue on what's going on here:
> 
> **About the `focus` functionality on drop:**
> 
> It seems like when dropping the new nvim-tree window changes or something like that, and a different window is focused instead.
> 
> For example, when following the steps below with the `foo` and `bar` files created [here](https://github.com/nvim-tree/nvim-tree.lua/pull/3054#issuecomment-2628669497):
> 
> 1. open nvim
> 2. open bar
> 3. run `:NvimTreeOpen`
> 4. run `:lua vim.print(vim.api.nvim_get_current_win())` - in my case it outputted 1002
> 5. run `:drop foo`
> 
> * `foo` is opened
> 
> 6. navigate back to tree and run `:lua vim.print(vim.api.nvim_get_current_win()) again` - in my case it outputted **1003**
> 
> Even when testing with nvim-tree without my modified code, I could reproduce this.
> 
> This might be part of my lack of knowledge of vim's inner APIs, but I would expect that the second call to `vim.api.nvim_get_current_win()` yields 1002 again - or is it expected that the window gets modified?
> 
> **About the `close` functionality on drop:**
> 
> The nvim-tree window is closed indeed, but for some reason it doesn't open the new file (in your example, `foo`).
> 
> I don't know why that's the case, but I could accidentally find out that when putting an `vim.fn.input()` statement before [this line](https://github.com/nvim-tree/nvim-tree.lua/pull/3054/files#diff-69640a8794ab2834e7b05cd80b8c0516749f6e893813a81b4e71117f346e809fR243), just to pause execution and try understanding a bit more how this is working, after pressing enter for continuing execution, the `foo` file was opened as expected 🤔 .

I like that trick... I'll definitely be using it in the future.

> Do you have any idea of why this happened? 😅

We're in difficult territory here; it's very hard to predict or react to window focus changes that nvim itself is making. Honestly, any workarounds usually end up being buggy and thus we avoid them. Events are sometimes not even sent, when debugging them.

## Proposal: limit the scope.

We know which actions are safe and work, and some of the actions like preview do not make any sense.

Limit the scope of this option to known good and desirable cases only.

Proposed actions to remove:
* `.*drop.*`
* `replace_tree_buffer`
* `preview`
* `toggle_group_empty`

Tab is the only special case, but it's rock solid.

We'd need to document the new option for each function in `nvim-tree-api.node`

Don't worry about `api.lua` complexity; it's a complete mess and will be overhauled #2896 #2832

GCrispino:
> I like that trick... I'll definitely be using it in the future.

That was mostly myself trying to compensate the lack of a proper debugging setup... do you have one for developing nvim plugins? I would love to have that set up on my machine.

> We're in difficult territory here; it's very hard to predict or react to window focus changes that nvim itself is making. Honestly, any workarounds usually end up being buggy and thus we avoid them. Events are sometimes not even sent, when debugging them.

got it :pray:


> ## Proposal: limit the scope.
> We know which actions are safe and work, and some of the actions like preview do not make any sense.
> 
> Limit the scope of this option to known good and desirable cases only.
> 
> Proposed actions to remove:
> 
> * `.*drop.*`
> * `replace_tree_buffer`
> * `preview`
> * `toggle_group_empty`
> 
> Tab is the only special case, but it's rock solid.
> 
> We'd need to document the new option for each function in `nvim-tree-api.node`
> 
> Don't worry about `api.lua` complexity; it's a complete mess and will be overhauled #2896 #2832

Ok that makes sense - most of the actions would probably not be used in this context anyway.

But something wasn't clear to me: are you proposing to completely removing these actions?
Or are you suggesting that we don't try to make the new `quit_on_open` and `focus` options compatible with these actions?

alex-courtis:
> Ok that makes sense - most of the actions would probably not be used in this context anyway.
> 
> But something wasn't clear to me: are you proposing to completely removing these actions? Or are you suggesting that we don't try to make the new `quit_on_open` and `focus` options compatible with these actions?

Don't try. We're adding unnecessary code paths that may indeed break with window focus etc.

The "good" ones are fine - we know there are no unexpected focus changes and we are in complete control of focus.

> That was mostly myself trying to compensate the lack of a proper debugging setup... do you have one for developing nvim plugins? I would love to have that set up on my machine.

I just add `"dev"` [log lines](https://github.com/nvim-tree/nvim-tree.lua), however this now gives me a means to actually pause and investigate.

When I'm desparate I'll debug nvim and tree events, something like
```lua
vim.api.nvim_create_autocmd("DirChanged", {
  callback = function(data)
    print(vim.inspect(data))
  end,
})

-- tree
api.events.subscribe(api.events.Event.NodeRenamed, function(data)
  log.line("dev", "Event.NodeRenamed %s %s", data.old_name, data.new_name)
end)
```

GCrispino:
Thanks for the tips on debugging :) 

> Don't try. We're adding unnecessary code paths that may indeed break with window focus etc.
> 
> The "good" ones are fine - we know there are no unexpected focus changes and we are in complete control of focus.

Ok that makes sense, was just trying to make sure I got it :)

Should we then throw an error/warning or something to not let the user use these new options for the problematic actions?

If you agree, I drafted that up in a separate branch. Can apply the commit in this one if you think it makes sense.
Here it is:

https://github.com/GCrispino/nvim-tree.lua/commit/1b51026854fdf661f61b82a3c73af64d123cd23a#diff-69640a8794ab2834e7b05cd80b8c0516749f6e893813a81b4e71117f346e809fR276-R310

alex-courtis:
> Should we then throw an error/warning or something to not let the user use these new options for the problematic actions?

Sorry... I was vague...

We should not make the new options available at all.

`edit(` should only execute the `if not focus then` block when it's a known good action e.g. edit, vertical etc.

It looks like we can test `mode`

alex-courtis:
As these options only apply to specific actions, we'll need to explicitly define the option for each e.g.

```help
node.open.vertical({node}, {opts})        *nvim-tree-api.node.open.vertical()*
    |nvim-tree-api.node.edit()|, file will be opened in a new vertical split.

    Parameters: ~
      • {node} (Node|nil) file or folder

    Options: ~
      • {force} (boolean) delete even if buffer is modified, default false
      • {quit_on_open} (boolean) TODO
```

GCrispino:
> > Should we then throw an error/warning or something to not let the user use these new options for the problematic actions?
> 
> Sorry... I was vague...
> 
> We should just ignore the new options available for the bad actions.
> 
> `edit(` should only execute the `if not focus then` block when it's a known good action e.g. edit, vertical etc.
> 
> It looks like we can test `mode`

Ok that makes sense to me.
I can do as you said and test if the value of `mode` corresponds to one of the api methods you listed in [this comment](https://github.com/nvim-tree/nvim-tree.lua/pull/3054#issuecomment-2660717546).

I just have one question about the api methods and their corresponding modes - these methods usually have modes corresponding to their names (e.g. `Api.node.open.edit` uses `edit` mode and `Api.node.open.drop` uses `drop`), however `Api.node.open.replace_tree_buffer` uses `edit_in_place`.
My question then is that if we can safely check on the `edit` function for mode `edit_in_place` for identifying if the current method being called is `Api.node.open.replace_tree_buffer`.

If we can, I believe I can implement this check just as you mentioned, otherwise I might have to do it outside of the scope of the `edit` function (maybe a little bit like I did in the draft code available in the branch I linked in my last comment). Does this make sense to you?

alex-courtis:
Honestly, whatever is simplest and easiest for you to implement. New
wrapper, existing wrapper, new function etc.

This API file will be overhauled and all those common functions unrolled or
combined with the edit action.

We've accepted that API.lua will get worse before it gets better :)

alex-courtis (approved):
Works beautifully, nice and clear. Many thanks for dealing with this most problematic one @GCrispino !

GCrispino:
@alex-courtis I was meaning to give this one last test before asking you to give it a final review, but thanks for jumping in and approving and merging it 😅 🙏🏼  !

alex-courtis:
> @alex-courtis I was meaning to give this one last test before asking you to give it a final review, but thanks for jumping in and approving and merging it 😅 🙏🏼 !

Please accept my apologies, that was rude.

In the future I'll always ensure that you're ready and satisfied before merging.

PR #3063, opened 2025-02-01
chore: POC plenary test

alex-courtis:
Plenary test is used by
telescope
lualine
packer

It allows static mocking of nvim functions.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -10,10 +10,12 @@ export DIR_PLENARY
 
 if [ "${#}" -eq 1 ]; then
 	TEST_NAME="${1}"
-else
+elif [ -z "${TEST_NAME}" ]; then
 	TEST_NAME="tests"
 fi
 export TEST_NAME
+
+echo "testing: ${TEST_NAME}"
 
 nvim --headless \
 	--clean \
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -2,15 +2,18 @@
 
 set -e
 
-# TODO add ability for user to specify a test file or directory
-
-# TODO clone plenary
-
 DIR_REPO="$(git rev-parse --show-toplevel)"
 export DIR_REPO
 
 DIR_PLENARY="${DIR_REPO}/plenary.nvim"
 export DIR_PLENARY
+
+if [ "${#}" -eq 1 ]; then
+	TEST_NAME="${1}"
+else
+	TEST_NAME="tests"
+fi
+export TEST_NAME
 
 nvim --headless \
 	--clean \
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -5,12 +5,15 @@ set -e
 # TODO add ability for user to specify a test file or directory
 
 REPO_DIR="$(git rev-parse --show-toplevel)"
+export REPO_DIR
+
 PLENARY_DIR="${REPO_DIR}/plenary.nvim"
+export PLENARY_DIR
 
 nvim --headless \
 	--clean \
 	--noplugin \
-	-c "set runtimepath+=${PLENARY_DIR}" \
+	-u "tests/minimal_init.lua" \
 	-c "lua require('plenary.test_harness').test_directory('tests/', { minimal_init = './tests/minimal_init.lua' })" \
 	-c "qa!"
 
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+set -e
+
+# TODO add ability for user to specify a test file or directory
+
+REPO_DIR="$(git rev-parse --show-toplevel)"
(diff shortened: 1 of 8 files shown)


gegoune:
I can't recall where and when it was but I have seen projects moving away from plenary tests. Might be worth exploring what top plugins in the ecosystem use/do.

tests/unit/utils_spec.lua
```diff
@@ -0,0 +1,37 @@
+---@type Luassert
+local assert = require("luassert")
+local stub = require("luassert.stub")
+
+local utils = require("nvim-tree.utils")
+
+describe("utils.path_add_trailing", function()
+  before_each(function()
+  end)
```
GCrispino, line 9: Is this intended?

scripts/test.sh
GCrispino, line 1: Should we add a command in Makefile (e.g. `make test`) that runs this?

GCrispino (commented):
I don't have a lot of experience in the neovim development space, so I don't think I can weigh in too much here, sorry :/ .

I guess my only question would be setup wise. I checked the branch out on MacOS and ran the script in `./scripts/test.sh`, but got an error - I believe I need to install plenary or something like that?

GCrispino:
> I can't recall where and when it was but I have seen projects moving away from plenary tests. Might be worth exploring what top plugins in the ecosystem use/do.

I know very little about this but I watched [this video](https://www.youtube.com/watch?v=zwEZJIXYnwI) on the subject, which uses plenary and mentions [mini.test](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-test.md).

tests/unit/utils_spec.lua
alex-courtis: Whoops... didn't need that one...

scripts/test.sh
alex-courtis: Yup. It can then run in CI.

alex-courtis:
> I guess my only question would be setup wise. I checked the branch out on MacOS and ran the script in `./scripts/test.sh`, but got an error - I believe I need to install plenary or something like that?

Whoops... forgot to add that step - check out plenary at ./plenary.nvim

Early days; many thanks for taking a look!

alex-courtis:
> I know very little about this but I watched [this video](https://www.youtube.com/watch?v=zwEZJIXYnwI) on the subject, which uses plenary and mentions [mini.test](https://github.com/echasnovski/mini.nvim/blob/main/readmes/mini-test.md).

The mini plugins are rather good, looks promising. I have homework to do.

alex-courtis:
Tidied and added missing bits.

It looks like plenary is a [subset of busted](https://github.com/nvim-lua/plenary.nvim/blob/master/README.md#plenarytest_harness), which isn't well liked any more.

Let's see how mini goes...

alex-courtis:
I gave mini.test a go:
* optional busted (plenary) compatibility
* nice style of laying out tests via sets
* high level of control over child vim process
* all the mini plugins have very comprehensive tests

Thoughts:
* sets are nice, but way too many strings; could probably use table keys directly
* asserts are pretty basic
* no plugins but mini uses it

Dealbreaker:
* doesn't provide mocks/stubs https://github.com/echasnovski/mini.test/blob/main/doc/mini-test.txt#L37

My Conclusion:
* Plenary works nicely

Same utils test in mini; didn't setup the harnesses, which are very close to plenary anyway.
```lua
local utils = require("nvim-tree.utils")

local new_set = MiniTest.new_set
local eq = MiniTest.expect.equality

-- Create (but not start) child Neovim object
local child = MiniTest.new_child_neovim()

local T = new_set({
  hooks = {
    post_once = child.stop,
  },
})

T["utils.path_add_trailing()"] = new_set()

T["utils.path_add_trailing()"]["trailing added"] = function()
  eq("foo/", utils.path_add_trailing("foo"))
end

T["utils.path_add_trailing()"]["trailing already present"] = function()
  eq("foo/", utils.path_add_trailing("foo/"))
end

T["utils.canonical_path()"] = new_set()

T["utils.canonical_path()"]["not windows"] = function()
  eq("c:\\foo\\bar", utils.canonical_path("c:\\foo\\bar"))
end

T["utils.canonical_path()"]["is windows"] = function()
  -- missing the ability to mock/stub
  -- https://github.com/echasnovski/mini.test/blob/main/doc/mini-test.txt#L37

  -- vim.fn.has.on_call_with("win32unix").returns(1)
  eq("c:\\foo\\bar", utils.canonical_path("C:\\foo\\bar"))
end

return T
```

alex-courtis:
Successful POC, closing.

alex-courtis:
Plenary is being [retired 2026-06-30](https://github.com/nvim-lua/plenary.nvim?tab=readme-ov-file#important-notice) 

A new framework will need to be found.

PR #3066, opened 2025-02-17
fix: window picker: hide fillchars: stl and stlnc

simonmandlik:
This addresses a minor issue that I have with the window picker — if `stl` and/or `stlnc` `fillchars` are set, they are not hidden when the picker is shown:

This example uses the replication from [here](https://github.com/nvim-tree/nvim-tree.lua/wiki/Troubleshooting#clean-room-replication) together with `:set fillchars+=stl:*,stlnc:*`:

image

The selection characters are hard to spot and hidden. This PR addresses this problem by temporarily disabling these characters when the window picker is displayed.


alex-courtis (approved):
Many thanks for the fix!

I've added you to this repository so that you may push branches and run CI.

PR #3096, opened 2025-04-01
fix: avoid deprecated usage of `vim.validate`

carlosala:
`vim.validate` in its spec form is deprecated.
This fixes it.

### DIFF
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -349,7 +349,11 @@ end
 ---@param dst_pos string value pos
 ---@param remove boolean
 function M.move_missing_val(src, src_path, src_pos, dst, dst_path, dst_pos, remove)
-  local ok, err = pcall(vim.validate, {
+  local ok, err = pcall(function(arr)
+    for key, value in pairs(arr) do
+      vim.validate(key, value[1], value[2])
+    end
+  end, {
     src = { src, "table" },
     src_path = { src_path, "string" },
     src_pos = { src_pos, "string" },


alex-courtis (changes requested):
Many thanks for finding and fixing this. We would need to provide a backwards compatible shim for the minimum supported nvim version 0.9.

However, I reckon we should just remove this validate completely:
* This one and only one use was from a time when we were heavily experimenting with legacy options and we didn't have LSP / type checking / CI.
* We have proper luadoc type annotations now

carlosala:
Sounds good. Yeah, I didn't know it wasn't available on older versions 👍🏻

angryluck:
What about adding an nvim-version check? I.e. something like this

```lua
  local safe_validate = function(arr)
    if vim.fn.has("nvim-0.9") == 1 then
      for key, args in pairs(arr) do
        vim.validate(key, unpack(args))
      end
    else
      vim.validate(arr)
    end
  end
  local ok, err = pcall(safe_validate, {
    src = { src, "table" },
    -- rest
```

alex-courtis:
> What about adding an nvim-version check? I.e. something like this
> 
> ```lua
>   local safe_validate = function(arr)
>     if vim.fn.has("nvim-0.9") == 1 then
>       for key, args in pairs(arr) do
>         vim.validate(key, unpack(args))
>       end
>     else
>       vim.validate(arr)
>     end
>   end
>   local ok, err = pcall(safe_validate, {
>     src = { src, "table" },
>     -- rest
> ```

Sorry, I mean just delete this whole check block that's not needed:

https://github.com/nvim-tree/nvim-tree.lua/pull/3096/files#diff-7c68f0fc800988f242ac61f7db0c2167124815cb6b65f1d01a99df9a4bb609f1R352-R368

carlosala:
Hey! Closing then, since it can be removed. Thanks for the review!

PR #3107, opened 2025-04-19
fix: reliably dispatch exactly one TreeOpen and TreeClose events

devansh08:
Fixes `TreeOpen` and `TreeClose` event dispatch, per this discussion: https://github.com/nvim-tree/nvim-tree.lua/pull/3105#issuecomment-2807238355. Post this, both these events should only be dispatched exactly once in every case.

Just to note, for `TreeClose`, I found out while implementing that autocmd on `WinClosed` does not match with the file name. Instead it matches with the window ID, which we would not know (I think?).
So instead, I had to write it as a condition in the callback to check the filetype of the underlying buffer to be `NvimTree`. Do let me know if anything can be improved here.

Any other thoughts on this ? @alex-courtis 

Thanks!

### DIFF
--- a/lua/nvim-tree.lua
+++ b/lua/nvim-tree.lua
@@ -236,6 +236,17 @@ local function setup_autocommands(opts)
       end,
     })
   end
+
+  -- Handles event dispatch when tree is closed by `:q`
+  create_nvim_tree_autocmd("WinClosed", {
+    pattern = "*",
+    ---@param ev vim.api.keyset.create_autocmd.callback_args
+    callback = function(ev)
+      if vim.api.nvim_get_option_value("filetype", { buf = ev.buf }) == "NvimTree" then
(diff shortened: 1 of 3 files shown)


alex-courtis:
> So instead, I had to write it as a condition in the callback to check the filetype of the underlying buffer to be `NvimTree`. Do let me know if anything can be improved here.

Fantastic! That is very reliable.

alex-courtis (approved):
Works beautifully, with one and only one events sent.

Close, other windows present:
- [x] :q
- [x] :NvimTreeClose

Open:
- [x] `vim.keymap.set("n", "<leader>c", function() api.tree.open({ current_window = true }) end,  { noremap = true })`
- [x] `vim.keymap.set("n", "<leader>n", function() api.tree.open({ current_window = false }) end, { noremap = true })`
- [x] :NvimTreeOpen
- [x] :NvimTreeToggle

PR #2864, opened 2024-08-04
refactor(#2826): multi instance nvim-tree.view

raaymax:
WIP: still not working

fixes https://github.com/nvim-tree/nvim-tree.lua/issues/2826

PR #3109, opened 2025-04-20
refactor(#2826): multi instance view

### LINKED ISSUE #2826: Multi Instance: Refactor: nvim-tree.view
* #2255 

`nvim-tree.view` is a stateful singleton. Make it an instantiable class owned by `Explorer`.

All usages must go through the `TreeExplorer` singleton.

This will be more difficult than other such refactors, as configuration and state are conflated and both mutated. See #2811

alex-courtis:
fixes #2826 

View does not mutate options, hence no need to deep clone anything.

Preserved the tab states bits; they still function as there is still one and only one explorer.

Testing methodology:
1. Add an `error("$callsite")` at each call to a View method
2. Test to ensure error is raised
3. Remove error
4. Repeat test to ensure correctness


alex-courtis:
Apologies @raaymax I abandoned trying to rebase #2864 , I applied the changes by hand. Notable:
* Did not do the deep clone of the options member; they are not mutated by view nor callers
* Used classic class structure

alex-courtis:
- [ ] Bug: tree does not refresh and becomes unusable following cd

alex-courtis:
This is working, however it's not useful due to the extensive use of the globals.

Alternative approach: extract the global functions into a separate global module, to be addressed in the next milestone.

alex-courtis:
Needs a re-think and alternative approach.

alex-courtis:
Split instance Window and global View.

Dogfooding.

alex-courtis:
Splitting out view was a step backwards.

Revert that and continue in #3153

Issue #3119, opened 2025-05-02
Customizable filter composition logic


gracepetryk:
When using the git_clean and no_buffer filters together, I think it would be more useful to show files that are either dirty or open in a buffer. The current behavior is to only show files that are both dirty and open in a buffer. 

**Can this functionality be implemented utilising API?**
Not currently, the filter logic is not exposed in the public API.  

**Describe the solution you'd like**
Exposing the various filter functions to the public api in a way that they could be composed in the custom filter would be my preferred solution. Something like this would be nice:
```lua
require('nvim-tree').setup({
  filters = {
    no_buffer = false,
    git_clean = false,
    custom = function (path)
      local filter_api = require('nvim-tree.api.filter')

      -- return true only when *both* git_clean and no_buffer would filter the path. These
      -- functions should not check the enabled/disabled state of the function so that the
      -- custom filter can have precedence while still using their result.
      return filter_api.git_clean(path) and filter_api.no_buffer(path)
    end
  }
})

```

**Describe alternatives you've considered**
I've implemented the behavior I want in my config with a couple janky runtime patches to `nvim-tree`'s internal filter API:

```lua
local filters = require('nvim-tree.explorer.filters')
local enum = require('nvim-tree.enum')

local should_filter = filters.should_filter
local should_filter_as_reason = filters.should_filter_as_reason

filters.should_filter = function (self, path, fs_stat, status)
  if self.state.no_buffer and not self:buf(path, status.bufinfo) then
    return false
  end

  if self.state.git_clean and not self:git(path, status.project) then
    return false
  end

  return should_filter(self, path, fs_stat, status)
end

filters.should_filter_as_reason = function (self, path, fs_stat, status)
  if not should_filter(self, path, fs_stat, status) then
    return enum.FILTER_REASON.none
  end

  return should_filter_as_reason(self, path, fs_stat, status)
end
```

**Additional context**

<img width="1152" alt="Image" src="https://github.com/user-attachments/assets/5226213a-78b3-4d94-bede-d78989cea066" />

alex-courtis:
That's a fantastic idea, something we should have done earlier.

I like the API - we can use that to express the built in filters, allowing users to follow them as examples.

A pull request would be most gratefully appreciated, see [CONTRIBUTING.md](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md)

I'm very happy to work with you, iterating on a solution.

[...]

PR #3121, opened 2025-05-06
feat(#3119): Expose filter functions to the api

gracepetryk:
Addresses #3119. Filter methods have all been normalized to accept a single `path` variable so that they may be easily called during a call to `custom`. Results are cached on the path for each render loop to avoid computing filters more than once if the filter is enabled and also called in `custom`.

### DIFF
--- a/lua/nvim-tree/actions/finders/search-node.lua
+++ b/lua/nvim-tree/actions/finders/search-node.lua
@@ -23,7 +23,7 @@ local function search(search_dir, input_path)
   local function iter(dir)
     local realpath, path, name, stat, handle, _
 
-    local filter_status = explorer.filters:prepare()
+    explorer.filters:prepare()
 
     handle, _ = vim.loop.fs_scandir(dir)
     if not handle then
@@ -46,7 +46,7 @@ local function search(search_dir, input_path)
         break
       end
 
-      if not explorer.filters:should_filter(path, stat, filter_status) then
+      if not explorer.filters:should_filter(path) then
         if string.find(path, "/" .. input_path .. "$") then
           return path
         end
--- a/lua/nvim-tree/explorer/filters.lua
+++ b/lua/nvim-tree/explorer/filters.lua
@@ -4,14 +4,23 @@ local FILTER_REASON = require("nvim-tree.enum").FILTER_REASON
 local Class = require("nvim-tree.classic")
 
 ---@alias FilterType "custom" | "dotfiles" | "git_ignored" | "git_clean" | "no_buffer" | "no_bookmark"
+---@alias GitFilterType "git_clean" | "git_ignored"
+
+---@class FilterStatus
+---@field project GitProject | nil
+---@field bufinfo table
+---@field bookmarks table
 
 ---@class (exact) Filters: Class
 ---@field enabled boolean
 ---@field state table<FilterType, boolean>
+---@field api table<string, fun(self: Filters, path: string): boolean|nil>
 ---@field private explorer Explorer
 ---@field private exclude_list string[] filters.exclude
 ---@field private ignore_list table<string, boolean> filters.custom string table
 ---@field private custom_function (fun(absolute_path: string): boolean)|nil filters.custom function
+---@field protected status FilterStatus
+---@field private filter_cache table<string, FILTER_REASON>
 local Filters = Class:extend()
 
 ---@class Filters
@@ -38,6 +47,8 @@ function Filters:new(args)
     no_bookmark = self.explorer.opts.filters.no_bookmark,
   }
 
+  self.filter_cache = {}
+
   local custom_filter  = self.explorer.opts.filters.custom
   if type(custom_filter) == "function" then
     self.custom_function = custom_filter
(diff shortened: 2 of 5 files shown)


alex-courtis:
Nice work! I'll get to a proper review / test in the next couple of days.

alex-courtis:
Apologies for the delay, my time is limited right now.

This deserves considerable review and test time which I hope to have in the next few days.

lua/nvim-tree/actions/finders/search-node.lua
```diff
@@ -23,7 +23,7 @@ local function search(search_dir, input_path)
   local function iter(dir)
     local realpath, path, name, stat, handle, _
 
-    local filter_status = explorer.filters:prepare()
+    explorer.filters:prepare()
```
alex-courtis, line 26: Great clean up! All the state is in filters now.

lua/nvim-tree/explorer/filters.lua
```diff
@@ -4,14 +4,23 @@ local FILTER_REASON = require("nvim-tree.enum").FILTER_REASON
...
 ---@class (exact) Filters: Class
 ---@field enabled boolean
 ---@field state table<FilterType, boolean>
+---@field api table<string, fun(self: Filters, path: string): boolean|nil>
 ---@field private explorer Explorer
 ---@field private exclude_list string[] filters.exclude
 ---@field private ignore_list table<string, boolean> filters.custom string table
 ---@field private custom_function (fun(absolute_path: string): boolean)|nil filters.custom function
+---@field protected status FilterStatus
+---@field private filter_cache table<string, FILTER_REASON>
```
alex-courtis, line 23:
```suggestion
---@field private filter_cache table<string, FILTER_REASON>
---@field private custom      fun(self: Filters, path: string): boolean
---@field private dotfile     fun(self: Filters, path: string): boolean
---@field private git_ignored fun(self: Filters, path: string): boolean
---@field private git_clean   fun(self: Filters, path: string): boolean
---@field private buf         fun(self: Filters, path: string): boolean
---@field private bookmark    fun(self: Filters, path: string): boolean
```

This is annoying, but it will keep the language server from complaining about inject-field

You could even wrap them in the constructor, although I'm not sure how clean that would look.

alex-courtis:
It's working nicely with a variety of composed filters and tests on `path`, with `toggle_custom_filter` working as expected.

Unfortunately this one's not updating until we manually refresh:
```lua
  filters = {
    enable = true,
    git_ignored = false,
    dotfiles = false,
    git_clean = false,
    no_buffer = false,
    no_bookmark = false,
    custom = function(path)
      return path:match(".*Makefile") and not api.filters.no_buffer(path)
    end,
    exclude = {},
```

`no_buffer` requires the `Buf:filter_buffer_` event in `Explorer` to apply the filter, and it's only fired when `no_buffer = true`.

Possible solution: fire the event when the user has a custom function. That's not the best for performance, happy for any other ideas.

alex-courtis:
This wasn't you, but we need to change the description of the default mapping Hidden->Custom
```lua
  vim.keymap.set("n", "U",              api.tree.toggle_custom_filter,      opts("Toggle Filter: Hidden"))
```

lua/nvim-tree/explorer/filters.lua
```diff
@@ -50,6 +61,35 @@ function Filters:new(args)
...
+---@private
+---@param fn fun(self: Filters, path: string): boolean
+---@param reason FILTER_REASON
+---@return fun(self: Filters, path: string): boolean
+local function cache_wrapper(fn, reason)
+
+  ---@param self Filters
+  ---@param path string
+  ---@return FILTER_REASON
+  ---@diagnostic disable:invisible
```
alex-courtis, line 75: I've had no success in resolving it. I'll keep at it...

lua/nvim-tree/explorer/filters.lua
alex-courtis, line 23:
```suggestion
---@field private filter_cache table<string, FILTER_REASON>
```
---@field private filter_cache table<FILTER_REASON, table<string, boolean>>

alex-courtis:
I think we can go one step further with the custom function: pass a node instead of a path, allowing the user to see the type, stat, children etc.

All of the filters could take a node, for consistency and ease of use, as well as aiding any future builtins.

There is extra complexity: we need to supply the user with a sanitised node rather than the actual, as per `Decorator`. See their cached creation:
https://github.com/gracepetryk/nvim-tree.lua/blob/ea5097a1e2702b4827cb7380e7fa0bd6da87699c/lua/nvim-tree/renderer/builder.lua#L236-L238

alex-courtis (changes requested):
This is looking fantastic, very straight-forward for the user.

Please
- [ ] resolve annotations suggestions
- [ ] update default map description: https://github.com/nvim-tree/nvim-tree.lua/pull/3121#issuecomment-2885681866
- [ ] consider using `Node` over `path`: https://github.com/nvim-tree/nvim-tree.lua/pull/3121#issuecomment-2885731834
- [ ] propagate buffer events: https://github.com/nvim-tree/nvim-tree.lua/pull/3121#issuecomment-2885680479

Apologies for the volume of comments; this is a significant change.

alex-courtis:
Any progress on this one @gracepetryk ?

Would you like me to complete this and merge it?

No pressure...

gracepetryk:
Hi @alex-courtis, I haven't had the time to work on this  I thought it would. If you'd like to complete + merge it go ahead :)

PR #3125, opened 2025-05-11
fix(#3124): fix icon padding for "right_align" placements, notably for dotfiles

### LINKED ISSUE #3124: Git icons for dotfiles have no padding if right aligned
### Description

If a file starts with a `.` (dot), all padding will be missing on the git icons if `renderer.icons.git_placement` is set to `"right_align"`.

The issue has existed since I started using right aligned git icons in November so I don't think it's a regression.

### Neovim version

```text
NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1713773202
```

Linux 6.12.10-zen1

### Windows variant

_No response_

### nvim-tree version

e7d1b7da

### Clean room replication

```lua
return {
  "nvim-tree/nvim-tree.lua",
  opts = {
    renderer = {
      icons = {
        padding = "  ",
        glyphs = {
          git = {
            unstaged = "",
            staged = "",
            unmerged = "",
            renamed = "",
            untracked = "",
            deleted = "",
            ignored = "",
          },
        },
      },
    }
  }
}
```

### Steps to reproduce

1. Create a file with a dot in front of the name and one without
2. run `git init`
3. open nvim and look at the git icons in nvim-tree

`mkdir foo && cd foo && touch bar .baz && git init && nvim`

### Expected behavior

The git icons should have their padding preserved:

Image

Image

### Actual behavior

The git icons do not have their padding preserved:

Image

Image

rossw01:
fixes #3124 

Bug is caused in `builder.lua`:
```lua
for _, d in ipairs(self.decorators) do
  add_to_end(rights, d:icons_right_align(not d:is(UserDecorator) and node or api_node))
end
```

In the case of tracked dotfiles, `d:icons_right_align(not d:is(UserDecorator) and node or api_node)` returns `{}` from the HiddenDecorator, but inside `add_to_end(t1, t2)` we only check `if not t2`. Not sure if this was intentional, I didn't notice any unexpected behaviour while testing but there may be edge cases I hadn't thought of.

### DIFF
--- a/lua/nvim-tree/renderer/builder.lua
+++ b/lua/nvim-tree/renderer/builder.lua
@@ -135,7 +135,7 @@ end
 function Builder:format_line(indent_markers, arrows, icon, name, node)
   local added_len = 0
   local function add_to_end(t1, t2)
-    if not t2 then
+    if not t2 or vim.tbl_isempty(t2) then
       return
     end
     for _, v in ipairs(t2) do


alex-courtis:
> Not sure if this was intentional, I didn't notice any unexpected behaviour while testing but there may be edge cases I hadn't thought of.

That was not intentional; it was a bad copy-paste when creating the decorator classes. The same issue occurs for modified and this fix resolves it as well.

From the comment it looks like the `:icons` method originally tested `self.explorer.opts.renderer.icons.show.git` and similar, however that test was moved to the constructor. There's no need to change that as this fix covers it and will assist people with user decorators.

alex-courtis (approved):

Issue #2746, opened 2024-04-11
Float window for full filenames has wrong bg


simonmandlik:
### Description

Setting `full_name = true`, there is an inconsistent color of the background of the floating window, depending on `culopt` settting in neovim.

### Neovim version

```text
NVIM v0.10.0-dev-2274+g1d5f86f69-Homebrew
Build type: Release
LuaJIT 2.1.1703358377
```

### Operating system and version

macOS 14.4.1

### nvim-tree version

81eb8d519233c105f30dc0a278607e62b20502fd

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvt-min/site]]
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup {
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  }
end
if vim.fn.isdirectory(install_path) == 0 then
  print "Installing nvim-tree and dependencies."
  vim.fn.system { "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path }
end
load_plugins()
require("packer").sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua setup()]]
vim.opt.termguicolors = true
vim.opt.cursorline = true

-- MODIFY NVIM-TREE SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
vim.opt.culopt = "number"
_G.setup = function()
    require("nvim-tree").setup {
        view = { cursorline = true },
        renderer = { full_name = true, }
    }
end
```

### Steps to reproduce

1. nvim -nu /tmp/nvt-min.lua
2. :NvimOpenTree

### Expected behavior

I would expect to either show no cursorline at all (there are no numbers in nvim tree), or the floating window bg to be consistent with cursorline bg for shorter items.

### Actual behavior

[...]

PR #3128, opened 2025-05-16
fix(#2746): background and right aligned icons in floating windows

simonmandlik:
Closes https://github.com/nvim-tree/nvim-tree.lua/issues/2746

Also I added functionality for copying over right-aligned icons to the float.

Before:

Screenshot 2025-05-16 at 8 48 57

After:

Screenshot 2025-05-16 at 8 47 58

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -302,21 +302,6 @@ function M.open(options)
   log.profile_end(profile)
 end
 
----@param extmarks table
----@return number
-function M.extmarks_length(extmarks)
-  local length = 0
-  for _, extmark in ipairs(extmarks) do
-    local details = extmark[4]
-    if details and details.virt_text then
-      for _, text in ipairs(details.virt_text) do
-        length = length + vim.fn.strchars(text[1])
-      end
-    end
-  end
-  return length
-end
-
 local function grow()
   local starts_at = M.is_root_folder_visible(require("nvim-tree.core").get_cwd()) and 1 or 0
   local lines = vim.api.nvim_buf_get_lines(M.get_bufnr(), starts_at, -1, false)
@@ -344,7 +329,7 @@ local function grow()
     local count = vim.fn.strchars(l)
     -- also add space for right-aligned icons
     local extmarks = vim.api.nvim_buf_get_extmarks(M.get_bufnr(), ns_id, { line_nr, 0 }, { line_nr, -1 }, { details = true })
-    count = count + M.extmarks_length(extmarks)
+    count = count + utils.extmarks_length(extmarks)
     if resizing_width < count then
       resizing_width = count
     end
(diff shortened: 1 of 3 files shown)


lua/nvim-tree/view.lua
```diff
@@ -302,6 +302,21 @@ function M.open(options)
   log.profile_end(profile)
 end
 
+---@param extmarks table
```
alex-courtis, line 305:
```suggestion
---@param vim.api.keyset.get_extmark_item[] as per vim.api.nvim_buf_get_extmarks
```

lua/nvim-tree/view.lua
```diff
@@ -302,6 +302,21 @@ function M.open(options)
...
+---@return number
+function M.extmarks_length(extmarks)
```
alex-courtis, line 307: Thanks for abstracting this.

Small request: move this to `utils.lua`. View is being converted into a stateful explorer member as part of the multi-instance view initiative: https://github.com/nvim-tree/nvim-tree.lua/pull/3109

alex-courtis:
> On second thought, copying over right aligned icons is not a very good idea—they are nearly always visible anyway and as implemented in this PR, they might be copied in a different order than in the tree window.

Good call. It's non-trivial and doesn't really add much value.

alex-courtis:
> There is also one more improvement in the condition when to show the full name in float. It now takes into account also right aligned icons.

Apologies, I think I might not be understanding; it seems to be adding whitespace to the float, for the right icons:

```lua
-- vim.opt.culopt = "number"
require("nvim-tree").setup({
  view = {
    width = 18,
  },
  renderer = {
    full_name = true,
    icons = {
      modified_placement = "right_align",
      hidden_placement = "right_align",
      diagnostics_placement = "right_align",
      bookmarks_placement = "right_align",
    },
  },
})
```

No icon:


Icon:

alex-courtis (changes requested):
Great work as always @simonmandlik 

please:

- [x] move `extmarks_length` to util
- [ ] fix the annotation nit

alex-courtis:
Apologies, my time is limited, I'll look at this next weekend.

simonmandlik:
> > There is also one more improvement in the condition when to show the full name in float. It now takes into account also right aligned icons.
> 
> Apologies, I think I might not be understanding; it seems to be adding whitespace to the float, for the right icons:
> 
> ```lua
> -- vim.opt.culopt = "number"
> require("nvim-tree").setup({
>   view = {
>     width = 18,
>   },
>   renderer = {
>     full_name = true,
>     icons = {
>       git_placement = "right_align",
>       modified_placement = "right_align",
>       hidden_placement = "right_align",
>       diagnostics_placement = "right_align",
>       bookmarks_placement = "right_align",
>     },
>   },
> })
> ```
> 
> No icon: 
> 
> Icon: 

Thanks for pointing that out, there was a bug, which now should be fixed. This PR, besides fixing https://github.com/nvim-tree/nvim-tree.lua/issues/2746, also fixes this behavior:

`master`:
image

image

PR:
image

image

alex-courtis (approved):
Many thanks for the thorough fix!

PR #3129, opened 2025-05-16
fix: "Invalid buffer id" on closing nvim-tree window

chrishrb:
Hello there,

I got the following error if I try to close the [mcphub.nvim](https://github.com/ravitemer/mcphub.nvim) window with q. This fixes this error:

image


alex-courtis (approved):
Many thanks for the contribution!

Issue #3132, opened 2025-05-16
Collapse directories under current highlighted directory


itsramiel:
**Can this functionality be implemented utilising API?**
nvim-tree exposes extensive API (see `:h nvim-tree-api`). Can it be used to achieve your goal? Is there a missing API that would make it possible?
The api does not cover it as mentioned [here](https://github.com/nvim-tree/nvim-tree.lua/discussions/3126#discussioncomment-13165481)

**Is your feature request related to a problem? Please describe.**
Sometimes I want to unclutter my workspace by closing all open directories and focus on the current one

**Describe the solution you'd like**
A mapping that would do the above

**Describe alternatives you've considered**
None

**Additional context**
[This discussion](https://github.com/nvim-tree/nvim-tree.lua/discussions/3126)

alex-courtis:
Changing the signature for consistency:

```
tree.collapse_all({node}, {opts})          *nvim-tree-api.tree.collapse_all()*
    Collapse the tree.

    Parameters: ~
      • {node} (Node|nil) collapse only children of this directory
      • {opts} (table) optional parameters

    Options: ~
      • {keep_buffers} (boolean?) do not collapse nodes with open buffers, default false
```

PR #3133, opened 2025-05-16
feat(#3132): add api.node.expand and api.node.collapse

itsramiel:
## Summary
Support passing `node` to `collapse_all` to allow collapsing all dirs under a given node instead of always root node.

Addresses #3132 by providing an api


itsramiel:
@alex-courtis Please review and lmk if you have any comments. Then I'll update the api docs

alex-courtis:
I saw the issue before seeing this PR.

Please use an `opts` parameter as per draft https://github.com/nvim-tree/nvim-tree.lua/issues/3132#issuecomment-2888688187

lua/nvim-tree/actions/tree/modifiers/collapse-all.lua
```diff
@@ -23,21 +23,30 @@ local function buf_match()
   end
 end
 
+---@param node Node
 ---@param keep_buffers boolean
-function M.fn(keep_buffers)
+function M.fn(node, keep_buffers)
```
alex-courtis, line 28: As we are changing the signature of this API, we're going to need to retain backwards compatibility, in the case of the user calling it like `collapse_all(true)`

See https://github.com/nvim-tree/nvim-tree.lua/blob/1ae1c33ce1f4f600dbb9dd8a15b95fbd215358e3/lua/nvim-tree/actions/tree/toggle.lua#L7-L29 for a complex example.

alex-courtis (changes requested):
Many thanks for the fast PR. Please:

- [ ] handle legacy api calls  https://github.com/nvim-tree/nvim-tree.lua/pull/3133#discussion_r2094297404
- [ ] use opts - my apologies, I advised you incorrectly https://github.com/nvim-tree/nvim-tree.lua/issues/3132#issuecomment-2888688187

alex-courtis:
Test cases:

```lua
  vim.keymap.set("n", "nk", function()
    local node = api.tree.get_node_under_cursor()
    api.tree.collapse_all(node, true)
  end, opts("Collapse Node Keep"))

  vim.keymap.set("n", "nn", function()
    local node = api.tree.get_node_under_cursor()
    api.tree.collapse_all(node, false)
  end, opts("Collapse Node No Keep"))

  vim.keymap.set("n", "ak", function()
    api.tree.collapse_all(nil, true)
  end, opts("Collapse All Keep"))

  vim.keymap.set("n", "an", function()
    api.tree.collapse_all(nil, false)
  end, opts("Collapse All No Keep"))

  -- collapse-all.lua:43: attempt to index local 'node' (a boolean value)
  vim.keymap.set("n", "lk", function()
    api.tree.collapse_all(true)
  end, opts("Collapse All Legacy Keep"))

  -- collapse-all.lua:43: attempt to index local 'node' (a boolean value)
  vim.keymap.set("n", "ln", function()
    api.tree.collapse_all(true)
  end, opts("Collapse All Legacy No Keep"))
```

alex-courtis:
Apologies, my time is limited, I'll look at this next weekend.

itsramiel:
Hi @alex-courtis 

I am fine with taking a step back and rethinking how to implement that. Take your time.

On a side note. I appreciate how much effort you are putting to not create breaking changes and maintain backwards compat, but doesnt it sometime make sense to take the hard decision and introduce a breaking version with a major version? I mean if expand_all assumes the current node is the node to expand, doesnt it make sense to do the same with `collapse_all` to maintain consistency?

alex-courtis:
> On a side note. I appreciate how much effort you are putting to not create breaking changes and maintain backwards compat, but doesnt it sometime make sense to take the hard decision and introduce a breaking version with a major version?

That time will inevitably come, however it would be best to wait until we have a planned large feature / fix that must break API.

> I mean if expand_all assumes the current node is the node to expand, doesnt it make sense to do the same with `collapse_all` to maintain consistency?

Exactly! Consistency is the hard part, and users have come to expect it.

# Proposal after sleeping in it

## For nodes

### `api.node.collapse`

- Move the new collapse functionality to a new API function `api.node.collapse`
- Revert existing `api.tree.collapse_all`
- When `api.node.collapse` is passed a nil node, operate on the `node_at_cursor`
  - if that's a directory, collapse it
  - if it's a file, collapse the parent directory, like `api.fs.create`

### `api.node.expand`

This is outside the scope of this change, however bonus points if you feel like implementing this.

- Create new API function `api.node.expand`
- It has the same functionality as `api.tree.expand_all` except
  - When it's passed a nil node and a directory is not under the cursor it will expand the parent directory, as per `api.node.collapse`

## For tree

### `api.tree.collapse_all`

This is consistent with other functions and doesn't need a change.

### `api.tree.expand_all`

Using node under cursor or root is inconsistent, however any change would break users. If they want more control they can use the new `api.node.expand`.

## Default Mappings

- Change `C` description from `Collapse` to `Collapse All`
- Add `A` `Collapse`
- Add `X` `Expand`

alex-courtis:
Apologies for messing you around on this one.

Please tell me what you think of the proposal.

I think the minimum we need to do here is move the changes to `api.node.collapse`. Anything else is nice to have.

alex-courtis:
Any progress on this one @itsramiel ?

Would you like me to complete this and merge it?

No pressure...

itsramiel:
@alex-courtis I am busy this period so I cant get to it. Feel free to take it if you have the time

PR #3118, opened 2025-04-28
fix(#3117): allow changing filename's casing

ktnlvr:
Closes (#3117). Seems a little crude, but I don't think there is an easier way.

Issue #3117, opened 2025-04-27
Renaming a file to the same name with different casing on a WSL


ktnlvr:
### Description

When renaming a file that is on windows into its version with a different casing the NvimTree raises an error and refuses the rename. For instance, renaming `A.txt` -> `a.txt` is impossible, unless done via another unoccupied name `A.txt` -> `tmp.txt` -> `a.txt`.

### Neovim version

```text
NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1713484068
```

### Operating system and version

Debian 12, Linux 5.15.167.4-microsoft-standard-WSL2

### Windows variant

WSL

### nvim-tree version

5bea2b3

### Clean room replication

```lua
No additional configuration required.
```

### Steps to reproduce

1. Open WSL.
2. Go to a directory that belongs to windows (usually `/mnt/c/Users/<USERNAME>`).
3. Ensure that you have read/write permissions for the directory
4. Create a file with an all-capital name (e.g. `A.txt`)
5. Open it in neovim using `nvim -nu /tmp/nvt-min.lua`. 
6. Open the sidebar using `:NvimTreeOpen`.
7. Locate and navigate to the created file (e.g. `A.txt`).
8. Press `r` and rename it to the lowercase version of its name (e.g. `A.txt` -> `a.txt`).
9. Observe `[NvimTree] Cannot rename /mnt/c/Users/<USERNAME>/A.txt -> /mnt/c/Users/<USERNAME>/a.txt: file already exists`.

### Expected behavior

Step 9 should leave the file renamed and output like `[NvimTree] /mnt/c/Users/<USERNAME>/A.txt -> /mnt/c/Users/<USERNAME>/a.txt`.

### Actual behavior

Step 9, error, no file renamed.

alex-courtis:
Honestly, I don't think this is achievable with windows mapping a case insensitive with a case sensitive file system.

After many attempts we did finally get this working on macos #2814, with just one file system.

The nvim-tree team doesn't have expertise with or access to windows however you're welcome to [contribute a fix.](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md)

Please ensure that all new or modified codepaths are behind the appropriate [windows feature flag.](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development#os-feature-flags)

PR #3135, opened 2025-05-24
fix(#3117): windows: change file/dir case

ktnlvr:
Like #3118 but without the broken commit history. Works for CMD/Powershell.


alex-courtis (approved):

PR #3136, opened 2025-05-25
fix(#3134): setting one glyph to "" no longer disables others

### LINKED ISSUE #3134: Feature request: "blank" icon glyphs turn off all icons of that type
**Is this a question?**
No

No

I am using `renderer.icons.glyphs.git` on a daily basis, but the "ignore" icon seems superficial and distracting especially in repos with a lot of ignored paths.

The solution is to set `renderer.icons.glyphs.git` to whitespace, e.g. `" "`, however, when git icons are right-aligned and the longest name in nvim-tree window is ignored, this adds unnecessary whitespace to the right:

Image

Setting the icon glyph to an empty string `""` doesn't solve this, because then the whole set of icons is disabled due to this condition:

https://github.com/nvim-tree/nvim-tree.lua/blob/6709463b2d18e77f7a946027917aa00d4aaed6f4/lua/nvim-tree/renderer/decorator/init.lua#L115-L119

If the point of this condition is to only save compute, would it make sense to remove it or somehow improve the logic such that it is possible to achieve the above?

simonmandlik:
fixes #3134

### DIFF
--- a/lua/nvim-tree/renderer/decorator/init.lua
+++ b/lua/nvim-tree/renderer/decorator/init.lua
@@ -112,9 +112,8 @@ function Decorator:define_sign(icon)
       vim.fn.sign_undefine(name)
     end
 
-    -- don't use sign if not defined
+    -- don't render sign if empty
     if #icon.str < 1 then
-      self.icon_placement = "none"
       return
     end
 


alex-courtis (approved):
Many thanks for the great fix!

alex-courtis:
Tested OK in a sampling of situations.

Definitely a bug.

PR #3145, opened 2025-06-14
fix: window picker ignore hidden window

phanen:
Hidden win should never be used as target.

## repro
Use neovim nightly
```lua
for name, url in pairs {
  nvt = 'https://github.com/nvim-tree/nvim-tree.lua',
} do
  local install_path = vim.fn.fnamemodify('nvim_issue/' .. name, ':p')
    vim.fn.system { 'git', 'clone', '--depth=1', url, install_path }
  end
  vim.opt.runtimepath:append(install_path)
end
require('nvim-tree').setup {}
require('vim._extui').enable({})
```

```sh
nvim --clean -u nvim-tree-hide.lua +"%bd!" +NvimTreeOpen
```

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -39,7 +39,7 @@ local function usable_win_ids()
     end
 
     local win_config = vim.api.nvim_win_get_config(id)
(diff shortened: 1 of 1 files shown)


alex-courtis (approved):
Many thanks for the quick fix @phanen !

PR #3175, opened 2025-08-04
perf(#3171): use vim.system() instead of vim.fn.system() to execute git toplevel

przepompownia:


przepompownia:
This PR does not resolve #3171 - it only prevents cursor flickering in the case described there. Please notice unchanged win32 part - I cannot test it.

przepompownia:
Pure Nvim example:

```lua
local timer = assert(vim.uv.new_timer())
timer:start(0, 600, function ()
  vim.schedule(function ()
    vim.fn.system({'true'})
  end)
end)
```

alex-courtis:
Alternatively, I'm open to bumping the minimum neovim version to 0.11. Do you reckon it's time @gegoune ? We'd be at n-1 released nvim released version.

It would involve:
- [ ] updating the minimum version check at startup
- [ ] removing backwards compatibility shims for 0.9
- [ ] updating readme/help/any other doc

That would be best done in a separate PR first.

alex-courtis:
> Please notice unchanged win32 part - I cannot test it.

I think that should be OK, as vim's abstraction layer should handle it.

Can you confirm that `out` is the same before and after.

przepompownia:
- https://github.com/neovim/neovim/milestones - on the Nvim side I do not see 0.10.x, 
- maintaining compatibility with the previous stable version of Nvim seems to be a common practice in the plugins world
- OS distributions can keep very old Nvim version, e.g. 0.7.2 for Debian stable https://packages.debian.org/search?keywords=neovim. Users who want a newer version can download it from the release page.

alex-courtis:
> * https://github.com/neovim/neovim/milestones - on the Nvim side I do not see 0.10.x,

0.11 is current stable
0.10 was released 2024-05-16
0.9 on 2023-04-07
0.8 on 2022-09-30

We moved to minimum version 0.9 on 2024-06-09

alex-courtis (approved):
Works on both codepaths, many thanks for the fix @przepompownia

gegoune:
> Alternatively, I'm open to bumping the minimum neovim version to 0.11. Do you reckon it's time @gegoune ? We'd be at n-1 released nvim released version.

> That would be best done in a separate PR first.

n-1 should be enough in cases where it makes it easier to maintain the code. If no breaking changes happened along the way we can extend it a bit, but more as a happy accident rather than our policy.

alex-courtis:
> > 
> 
> n-1 should be enough in cases where it makes it easier to maintain the code. If no breaking changes happened along the way we can extend it a bit, but more as a happy accident rather than our policy.

Cool, I guess we should document n-1 at the same time.

PR #3190, opened 2025-08-17
fix: get diagnostics icons from config (#1826)

### LINKED ISSUE #1826: Add option to use diagnostic signs
nvim-tree exposes extensive API (see `:h nvim-tree-api`). Can it be used to achieve your goal? Is there a missing API that would make it possible?
Given stable status of nvim-tree it's preferred to add new API than new functionality.

No

```lua
  use({
    'nvim-tree/nvim-tree.lua',
    requires = 'nvim-tree/nvim-web-devicons',
    config = function()
      require('nvim-tree').setup({
        diagnostics = {
          enable = true,
          -- This is the additional option I'm proposing
          use_diagnostic_signs = true,
          on_dirs = true,
        },
      })
    end,
  })
```

```lua
  use({
    'nvim-tree/nvim-tree.lua',
    requires = 'nvim-tree/nvim-web-devicons',
    config = function()
      require('nvim-tree').setup({
        diagnostics = {
          enable = true,
          icons = {
            error = vim.fn.sign_getdefined('DiagnosticSignError')[1].text,
            hint = vim.fn.sign_getdefined('DiagnosticSignHint')[1].text,
            info = vim.fn.sign_getdefined('DiagnosticSignInfo')[1].text,
            warning = vim.fn.sign_getdefined('DiagnosticSignWarn')[1].text,
          },
        },
      })
    end,
  })

```

**Additional context**
I think this is probably the common scenario. Most folks want to share the diagnostic signs across plugins related to LSP. I'd be happy to open a PR for this if it is determined to be valuable.

igorlfs:
Closes #1826 

Based on https://github.com/nvim-lualine/lualine.nvim/pull/1091, which never got merged


lua/nvim-tree.lua
```diff
@@ -303,7 +314,7 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
     special_files = { "Cargo.toml", "Makefile", "README.md", "readme.md" },
     hidden_display = "none",
     symlink_destination = true,
-    decorators = { "Git", "Open", "Hidden", "Modified", "Bookmark", "Diagnostics", "Copied", "Cut", },
+    decorators = { "Git", "Open", "Hidden", "Modified", "Bookmark", "Diagnostics", "Copied", "Cut" },
```
alex-courtis: Please don't change the formatting of unchanged lines; this make git history a bit hard to read.

lua/nvim-tree.lua
```diff
@@ -415,10 +426,10 @@ local DEFAULT_OPTS = { -- BEGIN_DEFAULT_OPTS
...
     },
     icons = {
-      hint = "",
-      info = "",
-      warning = "",
-      error = "",
+      hint = get_diagnostics_icons(vim.diagnostic.severity.HINT) or "",
+      info = get_diagnostics_icons(vim.diagnostic.severity.INFO) or "",
+      warning = get_diagnostics_icons(vim.diagnostic.severity.WARN) or "",
+      error = get_diagnostics_icons(vim.diagnostic.severity.ERROR) or "",
```
alex-courtis: When the user copies and uses this default config this fails, as `get_diagnostics_icons` is local to this file.

No matter, we will work around this another way.

alex-courtis (changes requested):
Thanks for looking into this one @igorlfs - this will be most useful.

The timing isn't working, however. At nvim-tree startup, diagnostics may not be configured.

## Alternative Implementation

Retrieve the signs dynamically in `DiagnosticsDecorator`. This will ensure that we pick up the configured signs, and allow the user to change them at runtime.

igorlfs:
> Retrieve the signs dynamically in DiagnosticsDecorator. This will ensure that we pick up the configured signs, and allow the user to change them at runtime.

Just pushed a commit following this implementation. LMK if that's not what you meant.

lua/nvim-tree/renderer/decorator/diagnostics.lua
```diff
@@ -3,6 +3,17 @@ local diagnostics = require("nvim-tree.diagnostics")
...
 
+---@param severity vim.diagnostic.Severity
+---@return string?
+local function get_diagnostics_icons(severity)
+  local config = vim.diagnostic.config() or {}
+  local signs = config.signs or {}
+  if type(signs) == "function" then
+    signs = signs(0, 0)
+  end
+  return (type(signs) == "table" and signs.text and signs.text[severity]) or nil
```
alex-courtis: This is safe and is returning correctly with the following test cases:

```lua
vim.diagnostic.config({
  signs = function(namespace, bufnr)
    return {
      text = {
        "e", "w", "i", "h",
      }
    }
  end,
})
```

```lua
vim.diagnostic.config({
  signs = {
    text = {
      "e", "w", "i", "h",
    }
  }
})
```

lua/nvim-tree/renderer/decorator/diagnostics.lua
```diff
@@ -51,7 +62,7 @@ function DiagnosticsDecorator:new(args)
     self.diag_icons = {}
     for name, sev in pairs(ICON_KEYS) do
       self.diag_icons[sev] = {
-        str = self.explorer.opts.diagnostics.icons[name],
+        str = get_diagnostics_icons(sev) or self.explorer.opts.diagnostics.icons[name],
```
alex-courtis: Please make this more efficient and call `get_diagnostics_icons` once in this constructor. It could return a single table with all severities populated.

We don't have to do exactly that; we just don't want to call `vim.diagnostic.config()` and extract all the icons multiple times as it is expensive.

lua/nvim-tree/renderer/decorator/diagnostics.lua
```diff
@@ -3,6 +3,17 @@ local diagnostics = require("nvim-tree.diagnostics")
 local Decorator = require("nvim-tree.renderer.decorator")
 local DirectoryNode = require("nvim-tree.node.directory")
 
+---@param severity vim.diagnostic.Severity
+---@return string?
+local function get_diagnostics_icons(severity)
```
alex-courtis: Please make this function a private member.

lua/nvim-tree/renderer/decorator/diagnostics.lua
igorlfs: Sorry, should I add a test for that or something?

lua/nvim-tree/renderer/decorator/diagnostics.lua
igorlfs: I inlined it.

alex-courtis:
Fantastic, thank you so much for going the extra mile.

I'll get to a review this weekend.

lua/nvim-tree/renderer/decorator/diagnostics.lua
alex-courtis: No need, we don't actually have any testing.

That's more for my reference so that I can test again.

alex-courtis (approved):
Works beautifully, many thanks for your contribution!

Fixing the trivial CI issues in the name of speed.

PR #3197, opened 2025-09-02
fix(#3196): utils.read_file may hold onto fd

### LINKED ISSUE #3196: utils.read_file may hold onto fd
### Description

`require('nvim-tree.utils').read_file` does not close `fd` if `stat` is falsy, potentially leaving hanging file descriptors.

### Neovim version

```text
NVIM v0.12.0-dev-582e071-dirty
Build type: Release
LuaJIT 2.1.1731601260
```

Linux 6.6.87.2-microsoft-standard-WSL2

### Windows variant

WSL

### nvim-tree version

master

### Clean room replication

```lua
vim.pack.add({ 'https://github.com/nvim-tree/nvim-tree.lua' }, { load = true })
```

### Steps to reproduce

1. `nvim`
2. Do anything that makes `nvim-tree.lua` call `require('nvim-tree.utils').read_file`
3. On the off chance that it fails to `fstat(2)`, it leaves a hanging file descriptor

### Expected behavior

`require('nvim-tree.utils').read_file` calls `vim.loop.fs_close(fd)` prior to returning, if `stat` is falsy.

### Actual behavior

`require('nvim-tree.utils').read_file` does not call `vim.loop.fs_close(fd)` prior to returning, when `stat` is falsy.

sepruko:
fixes #3196

### DIFF
--- a/lua/nvim-tree/utils.lua
+++ b/lua/nvim-tree/utils.lua
@@ -26,6 +26,7 @@ function M.read_file(path)
   end
   local stat = vim.loop.fs_fstat(fd)
   if not stat then
+    vim.loop.fs_close(fd)
     return ""
   end
   local data = vim.loop.fs_read(fd, stat.size, 0)


alex-courtis:
It looks like that function is unused. It was added for gitignore support https://github.com/nvim-tree/nvim-tree.lua/commit/eadead6072fe019ae45be33377528b33cb7a3da2 which has been rewritten several times over the years. We use git itself to track ignored now, rather than reading the `.gitignore` file.

Have you got a use case that uses this function? Maybe a recipe or extension plugin?

I reckon we should just delete this function.

sepruko:
> It looks like that function is unused.

I'm not overly familiar with the codebase, I was just perusing GitHub for the status quo on reading from a file and noticed this bug. I could've sworn I did a `ripgrep` on the codebase first and found several references, but maybe I was out of it and mindlessly searched for a different pattern.

> Have you got a use case that uses this function? Maybe a recipe or extension plugin?

No, and anything of the sort I feel can just roll their own similarly written function.

> I reckon we should just delete this function.

Probably, the entire file it's in seems to be using the deprecated `vim.loop` which will need to be thought about (with regards to minimum-supported Neovim version), too.

PR #3205, opened 2025-10-06
fix: prevent NvimTree to be alternate buffer when tab open

phanen:
`<c-t>` on file, then sometimes I press `<c-^>` and mistakenly goto `NvimTree` buffer.

### DIFF
--- a/lua/nvim-tree/actions/node/open-file.lua
+++ b/lua/nvim-tree/actions/node/open-file.lua
@@ -200,7 +200,7 @@ local function open_file_in_tab(filename)
   end
   vim.cmd.tabnew()
   vim.bo.bufhidden = "wipe"
-  -- HACK: prevent NvimTree to be alternate buffer
+  -- Following vim.fn.tabnew the # buffer may be set to the tree buffer. There is no way to clear the # buffer via vim.fn.setreg as it requires a valid buffer. Clear # by setting it to a new temporary scratch buffer.
   if utils.is_nvim_tree_buf(vim.fn.bufnr("#")) then
     local tmpbuf = vim.api.nvim_create_buf(false, true)
     vim.fn.setreg("#", tmpbuf)


alex-courtis:
Tested OK for base below and a few variants:

1. :NvimTreeOpen
2. \<navigate to file>
3. \<C-t>
4. \<C-^> or :e# or :b#

```
:ls!
  1  a   "[No Name]"                    line 1
  2u a-  "NvimTree_1"                   line 0
  3 %a   ".clang-format"                line 1
```

No `#` on nvim tree buffer :heavy_check_mark:

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -198,7 +198,15 @@ local function open_file_in_tab(filename)
   if M.relative_path then
     filename = utils.path_relative(filename, vim.fn.getcwd())
   end
-  vim.cmd("tabe " .. vim.fn.fnameescape(filename))
+  vim.cmd.tabnew()
+  vim.bo.bufhidden = "wipe"
+  -- HACK: prevent NvimTree to be alternate buffer
```
alex-courtis: I don't think this is a hack. There's no other option.

```suggestion
  -- Following vim.fn.tabnew the # buffer may be set to the tree buffer. There is no way to clear the # buffer via vim.fn.setreg as it requires a valid buffer. Clear # by setting it to a new temporary scratch buffer.
```

lua/nvim-tree/actions/node/open-file.lua
```diff
@@ -198,7 +198,15 @@ local function open_file_in_tab(filename)
   if M.relative_path then
     filename = utils.path_relative(filename, vim.fn.getcwd())
   end
-  vim.cmd("tabe " .. vim.fn.fnameescape(filename))
+  vim.cmd.tabnew()
+  vim.bo.bufhidden = "wipe"
```
alex-courtis, line 202: Is this needed? Should it go into the new if block so as not to affect existing functionality?

phanen: no, iirc when :tabnew then :edit an existing buffer, the scratch buffer will be set as the alternate file, this ensure we always delete the new scratch buffer

alex-courtis: I see, thank you. The assumption that the current buffer is nvim-tree sounds reasonable.

It's working for the permutations of this, executed from outside the buffer so it sounds like we're all good.

```lua
api.node.open.tab(api.tree.get_node_under_cursor(), { focus = true, quit_on_open = true })
```

alex-courtis (approved):
Really appreciate all your work @phanen

PR #3210, opened 2025-11-05
perf(commands): defer module loading

przepompownia:


przepompownia:
I [use](https://github.com/przepompownia/nvim-arctgx/blob/master/lua/arctgx/pluginConfigs/nvim-tree.lua) my custom, simple Lua module lazy [loader](https://github.com/przepompownia/nvim-arctgx/blob/master/lua/arctgx/lazy.lua). In that way nvim-tree isn't setup and no module is loaded before calling function defined in keymap.

It has a drawback: `NvimTree*` commands definitions are not loaded at that point too. 

Rarely, but still, I would like to start using nvim-tree with such command.

I know that direct call to `require('nvim-tree.commands').setup()` can be not recommended way, but (with this change) it allows to load command definitions without loading modules not used yet.

That `setup()` logic could exist in `plugin/nvim-tree.lua` instead.

lua/nvim-tree/commands.lua
```diff
@@ -111,7 +108,7 @@ local CMDS = {
       bar = true,
     },
     command = function(c)
-      view.resize(c.args)
+      require("nvim-tree.view").resize(c.args)
```
alex-courtis, line 111: Unrelated, but we could probably change this to `api.tree.resize`.
All commands would then use public API.

lua/nvim-tree/commands.lua
```diff
@@ -12,7 +9,7 @@ local CMDS = {
       complete = "dir",
     },
     command = function(c)
-      api.tree.open({ path = c.args })
+      require("nvim-tree.api").tree.open({ path = c.args })
```
alex-courtis, line 12: Initial thought was to use a common function like
```lua
local function api()
  require("nvim-tree.api")
end
```
however that would just obscure things.
The inline require make it really clear what's going on.

alex-courtis (commented):
Normally I'd give dire warnings about lazy loading, predictable ordering and not using API, however you know exactly what you're doing :)

This looks good/safe and won't change anything.

> That `setup()` logic could exist in `plugin/nvim-tree.lua` instead. 

I'm not quite sure what you mean... would that allow you to avoid requiring commands? Would that not load all the modules?

Either way, sounds good. Do you want to just push it?

przepompownia:
> Normally I'd give dire warnings about lazy loading, predictable ordering and not using API, however you know exactly what you're doing :)
> 
> This looks good/safe and won't change anything.
> 
> > That `setup()` logic could exist in `plugin/nvim-tree.lua` instead.
> 
> I'm not quite sure what you mean... would that allow you to avoid requiring commands? Would that not load all the modules?
> 
> Either way, sounds good. Do you want to just push it?

~OK, I can test permissions at the git level: here I have no permissions to merge.~ Regardless if it works, I can then create a PR with commands definitions moved to `plugin` directory. 

Ah, you probably meant pushing plugin approach here :grinning: 

This change seems to be safe to merge separately. Using `plugin/` may require discussion on some details, so let me propose it separately.

Issue #3213, opened 2025-11-11
Feature request: take root folder into account when computing width in `View.grow`


simonmandlik:
Would it be possible to take width of root folder (first line) into account when computing width of an nvim tree window?

Root folder might be quite long, but one can easily limit the width with `view.width.max`.

Thank you

alex-courtis:
That sounds like a fantastic idea! It would need to be an option (default off) so as not to break the existing user experience.

As usual, pull requests are most gratefully appreciated, see
* [CONTRIBUTING.md](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md)
* [Wiki: Development](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development)

PR #3214, opened 2025-11-15
feat(#3213): add view.width.lines_excluded option

simonmandlik:
Closes https://github.com/nvim-tree/nvim-tree.lua/issues/3213

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -822,6 +822,15 @@ longest line.
       Type: `string | number | fun(): number|string`
       Default: `-1`
 
+    *nvim-tree.view.width.lines_excluded*
+    Exclude these lines when computing width.
+    Supported values: `"root".
+      Type: `table`
+      Default:
+        `{`
+          `"root"`
+        `}`
+
     *nvim-tree.view.width.padding*
     Extra padding to the right.
       Type: `number | fun(): number|string`
@@ -3323,6 +3332,7 @@ highlight group is not, hard linking as follows: >
 |nvim-tree.view.side|
 |nvim-tree.view.signcolumn|
 |nvim-tree.view.width|
+|nvim-tree.view.width.lines_excluded|
 |nvim-tree.view.width.max|
 |nvim-tree.view.width.min|
 |nvim-tree.view.width.padding|
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -12,6 +12,9 @@ local M = {}
 
 local DEFAULT_MIN_WIDTH = 30
 local DEFAULT_MAX_WIDTH = -1
+local DEFAULT_LINES_EXCLUDED = {
+  "root",
+}
 local DEFAULT_PADDING = 1
 
 M.View = {
@@ -303,7 +306,7 @@ function M.open(options)
 end
 
 local function grow()
-  local starts_at = M.is_root_folder_visible(require("nvim-tree.core").get_cwd()) and 1 or 0
+  local starts_at = (M.is_root_folder_visible(require("nvim-tree.core").get_cwd()) and M.View.root_excluded) and 1 or 0
   local lines = vim.api.nvim_buf_get_lines(M.get_bufnr(), starts_at, -1, false)
   -- number of columns of right-padding to indicate end of path
   local padding = get_size(M.View.padding)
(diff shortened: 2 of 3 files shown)


doc/nvim-tree-lua.txt
```diff
@@ -822,6 +822,10 @@ longest line.
       Type: `string | number | fun(): number|string`
       Default: `-1`
 
+    *nvim-tree.view.width.include_root*
```
gegoune: Can you think of anything else that could be added in the future? I am asking because we could have an enum `nvim-tree.view.width.include` instead of multiple `include_*`.

simonmandlik: Not really. As things stand, with `include_root = true` all lines in the buffer will be considered.

We could start excluding something, e.g. use only either folders/files to compute the width, but it is becoming very niche and not overly useful IMHO

simonmandlik:
Added small improvements to the `grow()` algorithm including potential early exit

alex-courtis:
Apologies, my time is very limited right now; I will likely review this change on the weekend.

lua/nvim-tree/view.lua
```diff
@@ -314,31 +315,25 @@ local function grow()
...
+  local final_width = M.View.initial_width
+  local max_width = math.huge
+  if M.View.max_width ~= -1 then
+    max_width = get_width(M.View.max_width)
   end
 
   local ns_id = vim.api.nvim_get_namespaces()["NvimTreeExtmarks"]
   for line_nr, l in pairs(lines) do
-    local count = vim.fn.strchars(l)
+    local line_width = vim.fn.strchars(l)
```
alex-courtis, line 328: Thank you for clarifying the variable names.

lua/nvim-tree/view.lua
```diff
@@ -314,31 +315,25 @@ local function grow()
...
     -- also add space for right-aligned icons
     local extmarks = vim.api.nvim_buf_get_extmarks(M.get_bufnr(), ns_id, { line_nr, 0 }, { line_nr, -1 }, { details = true })
-    count = count + utils.extmarks_length(extmarks)
-    if resizing_width < count then
-      resizing_width = count
-    end
-    if M.View.adaptive_size and max_width >= 0 and resizing_width >= max_width then
-      resizing_width = max_width
+    line_width = line_width + utils.extmarks_length(extmarks) + padding
+    final_width = math.max(final_width, line_width)
```
alex-courtis, line 332: Could we replace `final_width` with `M.View.initial_width`, to make it a bit clearer, as final_width is not mutated?

```suggestion
    local final_width = math.max(M.View.initial_width, line_width)
```

alex-courtis:
Testing in `/home/alex/src/linux/arch/alpha/include` which has children with short names:

Defaults OK:

```lua
    -- width = { },
```

Root should grow. FAIL: it doesn't.

```lua
    width = {
      include_root = true,
    },
```

Root explicitly should not grow. FAIL: it actually does.
It looks like `include_root` option might be inverted. I'm inverting the values going forward.

```lua
    width = {
      include_root = false,
    },
```

Long root constrained to max, OK:

```lua
    width = {
      include_root = false,
      max = 20,
    },
```

Respects min, OK:

```lua
    width = {
      include_root = false,
      min = 50,
    },
```

Respects padding, OK:

```lua
    width = {
      include_root = false,
      padding = 5,
    },
```

Respects empty table defaults: OK after changing `DEFAULT_INCLUDE_ROOT` to `true`

```lua
    width = { },
```

alex-courtis (changes requested):
Apologies for the late review. Looking good, please:

- [ ] simplify this https://github.com/nvim-tree/nvim-tree.lua/pull/3214/files#r2572631362
- [x] resolve the inverted logic for `include_root` https://github.com/nvim-tree/nvim-tree.lua/pull/3214#issuecomment-3590680230
- [x] refactor `include_root` option https://github.com/nvim-tree/nvim-tree.lua/pull/3214/files/3d234325d1ed5411a41260ff92718b4f41540717#r2530138318

doc/nvim-tree-lua.txt
alex-courtis: I think this is the safest option. We've been burned in the past when it comes to options like these, when unexpected use cases arise. Refactoring options is a pain for devs and for the users.

An enum would be best, with values in `ACCEPTED_STRINGS`. That will be clearest when we finally get to annotating the options in api meta.

lua/nvim-tree/view.lua
simonmandlik: Sorry but I don't understand. `final_width` here is used to store the width that the window is resized to. It is assigned before the for loop over all lines and is used to resize the window afterwards. Why declare it local to the for loop?

If we replace `math.max(final_width, line_width)` with `math.max(M.View.initial_width, line_width)`, this would break the algorithm, the window would be resized to either initial width or width of the last line, all other lines would be ignored.

alex-courtis: > Sorry but I don't understand. `final_width` here is used to store the width that the window is resized to. It is assigned before the for loop over all lines and is used to resize the window afterwards. Why declare it local to the for loop?
> 
> If we replace `math.max(final_width, line_width)` with `math.max(M.View.initial_width, line_width)`, this would break the algorithm, the window would be resized to either initial width or width of the last line, all other lines would be ignored.

Please accept my sincere apologies for wasting your time; I did not see that `final_width` was updated inside the loop.

alex-courtis:
Defaults OK:

```lua
    -- width = { },
```

Root should grow. OK.

```lua
    width = {
      lines_excluded = "none",
    },
```

Root explicitly should not grow. OK.

```lua
    width = {
      lines_excluded = "root",
    },
```

Long root constrained to max, OK:

```lua
    width = {
      lines_excluded = "none",
      max = 20,
    },
```

Respects min, OK:

```lua
    width = {
      lines_excluded = "none",
      min = 50,
    },
```

Respects padding, OK:

```lua
    width = {
      lines_excluded = "none",
      padding = 5,
    },
```

Respects empty table defaults with root excluded. OK

```lua
    width = { },
```

alex-courtis:
Having `lines_excluded` as as a single enum value isn't working, as it won't allow us to add more exclusions in the future e.g. symlinks.

It can be changed to a table of enums e.g.
```lua
    width = {
      lines_excluded = {
        "root",
        "symlink",
      },
  },
```
For now it can just contain `"root"`.

After some frustrating hacking, I've added enum validation to `validate_options`

Do you think we can implement this?

[ACCEPTED_ENUMS.patch](https://github.com/user-attachments/files/23837441/ACCEPTED_ENUMS.patch)

simonmandlik:
@alex-courtis done

alex-courtis:
Defaults OK:

```lua
    -- width = { },
```

Root should grow. OK.

```lua
    width = {
      lines_excluded = { },
    },
```

Root explicitly should not grow. OK.

```lua
    width = {
      lines_excluded = {
        "root",
      }
    },
```

Long root constrained to max, OK:

```lua
    width = {
      max = 20,
      lines_excluded = { }
    },
```

Respects min, OK:

```lua
    width = {
      lines_excluded = { },
      min = 50,
    },
```

Respects padding, OK:

```lua
    width = {
      lines_excluded = "none",
      padding = 5,
    },
```

Respects empty table defaults with root excluded. OK

```lua
    width = { },
```

Checks type. OK.

```lua
    width = {
      lines_excluded = "foo",
    },
```

Checks value. OK.

```lua
    width = {
      lines_excluded = {
        "foo",
      },
    },
```

Issue #3226, opened 2025-12-16
failed to open nvim-tree when temp file exists


wsdjeg:
```
Found a swap file by the name "~\AppData\Local\nvim-data\swap\D%%wsdjeg%dashboard-nvim%NvimTree_1.swp"
             dated: �the�nam� "~�App 12�\Lo�al\ 07 22:19:32 2025
         file name: D:\wsdjeg\dashboard-nvim\NvimTree_1
          modified: no
         user name: 0   host name: WSDJEG-PC
        process ID: 18204
While opening file "NvimTree_1"
      CANNOT BE FOUND
(1) Another program may be editing the same file.  If this is the case,
    be careful not to end up with two different instances of the same
    file when making changes.  Quit, or continue with caution.
(2) An edit session for this file crashed.
    If this is the case, use ":recover" or "nvim -r NvimTree_1"
    to recover the changes (see ":help recovery").
    If you did this already, delete the swap file "C:\Users\wsdjeg\AppData\Local\nvim-data\swap\D%%wsdjeg%dashboard-nvim%NvimTree_1.swp"
    to avoid this message.

E5108: Error executing lua: vim/_editor.lua:447: nvim_exec2(), line 1: Vim:Error executing Lua callback: vim/_editor.lua:447: nvim_exec2()[1]..nvim_exec2() called at nv
im_exec2():1, line 1: Vim:Error executing Lua callback: ...undle_dir\nvim-tree\nvim-tree.lua/lua/nvim-tree/view.lua:104: Vim:E325: ATTENTION
stack traceback:
        [C]: in function 'nvim_buf_set_name'
        ...undle_dir\nvim-tree\nvim-tree.lua/lua/nvim-tree/view.lua:104: in function 'create_buffer'
        ...undle_dir\nvim-tree\nvim-tree.lua/lua/nvim-tree/view.lua:292: in function 'open'
        D:\bundle_dir\nvim-tree\nvim-tree.lua/lua/nvim-tree/lib.lua:34: in function 'open_view_and_draw'
        D:\bundle_dir\nvim-tree\nvim-tree.lua/lua/nvim-tree/lib.lua:129: in function 'open'
        ...e\nvim-tree.lua/lua/nvim-tree/actions/tree/find-file.lua:52: in function 'find_file'
[...]

PR #3227, opened 2025-12-16
fix(#3226): set &swapfile=false before setting tree buffer name, avoiding any potential collisions with a swapfile

wsdjeg:
close https://github.com/nvim-tree/nvim-tree.lua/issues/3226


wsdjeg:
Thanks.

PR #3238, opened 2026-01-08
feat: add default <Del> mapping for api.fs.remove

ivankra:
\<Del\> is a common mapping to delete file/folder in many tree explorers such as vim's netrw or vscode. But in nvim-tree it fails with `E21: Cannot make changes, 'modifiable' is off`. Let's add it to the default key mappings as an alternative to "d".


alex-courtis:
We don't normally change default key mappings, however I reckon this should be OK. @gegoune ?

gegoune:
@alex-courtis I won't be using it (being on MacBook), `d` is more vim-idiomatic anyway. I am fine with that change, shouldn't break for many users, if any at all.

ivankra:
I think \<Del\> is a more universal mapping, less surprising to new users, especially anyone coming from vim's builtin netrw which has \<Del\>/D for delete, d=mkdir. And since there was no mapping for \<Del\> - it gives an error, it shouldn't really break anyone, unless doing some weird stuff with nvimtree buffers.

gegoune:
> Nice one, didn't realise macbook didn't have a delete...
> 

It does, `fn+backspace` is `delete`, just so much easier to type `d`.

ivankra:
Added `<Del>` to `doc/nvim-tree-lua.txt`

alex-courtis (changes requested):
It looks like we're missing an addition in help: https://github.com/nvim-tree/nvim-tree.lua/actions/runs/20954090775/job/60215519930?pr=3238

You should run `make help-update`: https://github.com/nvim-tree/nvim-tree.lua/blob/000ca6bcddd11ec0bf782e3c9794838f4c6d7384/CONTRIBUTING.md#adding-new-actions

ivankra:
Oh, I see, there were two places in the help file. I ran `make help-check`.

PR #3239, opened 2026-01-08
fix: incorrect window width when right_align icons present

simonmandlik:
This PR fixes a bug introduced in https://github.com/nvim-tree/nvim-tree.lua/pull/3214

In `grow()` when `M.View.root_excluded` is `true` lines starts from the second line, but later in the for loop we index starting from 1 always.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -325,7 +325,8 @@ local function grow()
   end
 
   local ns_id = vim.api.nvim_get_namespaces()["NvimTreeExtmarks"]
-  for line_nr, l in pairs(lines) do
+  for i, l in pairs(lines) do
+    local line_nr = starts_at + i - 1
     local line_width = vim.fn.strchars(l)
     -- also add space for right-aligned icons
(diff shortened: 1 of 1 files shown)


alex-courtis:
Thanks for looking after this one @simonmandlik 

This looks right, however I cannot identify the bug. Can you please decscribe how to reproduce it so that I can test?

simonmandlik:
There are several ingredients needed:

- use right-aligned icons
- use the root for width computation
- have an icon shown on the widest item

Before the fix:

image

With the fix:

image


```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
    },
    config = {
      package_root = package_root,
    },
  })
end
end
load_plugins()
require("packer").sync()
vim.opt.termguicolors = true
vim.opt.cursorline = true

_G.setup = function()
  require("nvim-tree").setup({
    view = {
      width = {
        min = 0,
        padding = 0,
        lines_excluded = {},
      }
    },
    renderer = {
      icons = {
      }
    }
  })
end
```

PR #3252, opened 2026-01-24
refactor: use Node objects in Filters:bookmark for improved type safety

xiantang:
This pull request updates the bookmark filtering logic in the `Filters:bookmark` function to work with a new data structure for bookmarks. Instead of mapping bookmark paths to a filetype or string, bookmarks now map to `Node` objects, allowing for more robust and type-safe operations.

**Refactor bookmark filtering to use Node objects:**

* The `bookmarks` parameter now expects a table mapping bookmark paths to `Node` objects instead of strings or filetypes, enabling more structured handling of bookmarks.
* The logic for checking bookmark types and parent-child relationships has been updated to use the `Node` interface, specifically using the `as(DirectoryNode)` method for type checking and path comparisons.

### DIFF
--- a/lua/nvim-tree/explorer/filters.lua
+++ b/lua/nvim-tree/explorer/filters.lua
@@ -127,7 +127,7 @@ function Filters:bookmark(path, path_type, bookmarks)
   if not self.state.no_bookmark then
     return false
   end
-
+  -- if bookmark is empty, we should see a empty filetree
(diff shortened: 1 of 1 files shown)


xiantang:
@alex-courtis, please take a review first, I spent a bit of time understanding the Lua type check, and there is no logic for FileNode, so i only add for dir

Issue #2668, opened 2024-02-07
Full And Sanitized Node Documentation


stefgodin:
**Is this a question?**
No

**Can this functionality be implemented utilizing API?**
Not a functionality per say.

**Is your feature request related to a problem? Please describe.**
Many functions of the API retrieves nodes such as `get_node_under_cursor()`.
There doesn't seem to be any documentation about the structure of a node except in the [sort API section](https://github.com/nvim-tree/nvim-tree.lua/blob/8cbb1db8e90b62fc56f379992e622e9f919792ce/doc/nvim-tree-lua.txt#L679). But there are some missing fields in the doc.

**Describe the solution you'd like**
It would be nice to have a small section of the documentation dedicated to the content of a node. This would make it officially part of the API.

**Describe alternatives you've considered**
Check the [node builder source code](https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/explorer/node-builders.lua) but that does not prevent invisible breaking changes to the node API.

**Additional context**
This issue is raised after a [short post on reddit](https://www.reddit.com/r/neovim/comments/1akjoc3/what_can_i_do_with_a_nvimtree_node/).

I don't know if this qualifies as a feature request, feel free to move it around.

nvim-tree is an awesome plugin and you did a really good job.

alex-courtis:
We do have https://github.com/nvim-tree/nvim-tree.lua/blob/6a99f5af78b5dd531d49ec0a5158dd0f9a7c992b/lua/nvim-tree/node.lua to document all the contents of various nodes, however that's not super informative.

Proposal: add more descriptions and include it in help.

alex-courtis:
@stefgodin _meta api documentation has been added for all the `nvim_tree.api.Node` types, as they are exposed via API.

https://github.com/nvim-tree/nvim-tree.lua/blob/master/lua/nvim-tree/_meta/api.lua

Does that satisfy your requirements?

More work needs to be done to add them to `api.lua` and help, as mentioned in #2832

stefgodin:
@alex-courtis  Yes! Thank you. Feel free to close the issue.

alex-courtis:
Partially addressed by #3244 

Outstanding: `nvim_tree.api.Node` subclasses, with `:as` etc. documented
Requires: conversion of the classes to classic

Most API and callbacks use real nodes, only `get_nodes` and decorators use `nvim_tree.api.*Node`. Decision:
[...]

PR #3268, opened 2026-02-15
feat(#2994): add visual mode operations: copy, cut, delete, trash, toggle bookmark

### LINKED ISSUE #2994: Visual Selection
### Discussed in https://github.com/nvim-tree/nvim-tree.lua/discussions/2993

<sup>Originally posted by **l00sed** November  9, 2024</sup>
First of all, thank you so much to the open source community, the plugin author, and plugin contributors. Nvim Tree is a fantastic plugin and part of my everyday workflow. 🌳

### Some Background, Visual Selection

The vim-centric way to delete multiple lines of text is to first `Shift-v`— visually select a full line of text. Next, with the visual line selection, I could navigate (`j` or `k`) to expand the selection area down or up, respectively. See below:

https://github.com/user-attachments/assets/d33580ef-8841-4d20-b7db-c440973861ea

I would love to mimic this behavior for selecting multiple files in Nvim Tree, using `<S-v>` to visually select files or folders, then pressing `c` (for example, to copy) or `d` (to delete)— or running any action against that selection of files.

Performing the selection this way, using the standard Vim visual selection, would feel like a seamless extension of normal buffer behavior.

### Example Scenario

I want to delete several markdown files from a folder, `blog`. The files are in consecutive order, so I can `<S-v>` to visually select the first file to delete (`car-hacking.mdx`) and then press `j` several times to expand the selection downward. Pressing `d` deletes the selected files.

https://github.com/user-attachments/assets/a26a7290-b97f-4633-8fc7-6dd6db329059

Does anyone have an existing recipe that makes this behavior possible? Perhaps using the existing buffer marking (`m`) behavior as a way to integrate the key-bindings somehow?

I suspect there are other people interested in this as well— or those who may have already found a solution. Funny enough, I found a [stack overflow](https://stackoverflow.com/questions/5745506/vim-modifiable-is-off) issue related to NERD Tree which was likely asked in the hope of achieving the same visual-selection behavior in this other Neovim tree-based file manager.

**Other relevant discussions:**
- [batch delete consecutive files (discussion#2677)](https://github.com/nvim-tree/nvim-tree.lua/discussions/2677)</div>

v3ceban:
## Add visual selection operations

Closes #2994
See also: #2993

### Summary

Adds visual selection support to nvim-tree. Users can visually select a range of nodes (charwise, linewise, or blockwise) and operate on them with a single keypress, matching standard Vim visual mode behavior.

Supported operations:
- **Toggle bookmark** (`m`)
- **Copy** (`c`)
- **Cut** (`x`)
- **Delete** (`d` and `Del`) -- single confirmation prompt for the entire selection
- **Trash** (`D`) -- single confirmation prompt for the entire selection

### Design

Existing API functions (`api.fs.copy.node`, `api.fs.cut`, `api.fs.remove`, `api.fs.trash`, `api.marks.toggle`) are now **mode-dependent** -- no new API functions are added. When called in visual mode they operate on the selected range; in normal mode they behave exactly as before.

Default keymaps use `{ "n", "x" }` mode tables so the same key works in both modes. The `g?` help window shows a mode column (e.g. `n`, `nx`) for each mapping, with mode characters sorted for deterministic display.

- **Visual range detection** uses `vim.fn.line("v")` / `vim.fn.line(".")` while still in visual mode, avoiding `<Esc>` + `'<`/`'>` marks. All visual and select modes are supported.
- **`Explorer:get_nodes_in_range()`** reuses the existing `get_nodes_by_line()` infrastructure.
- **Node class check** for single-vs-array dispatch: `type(node_or_nodes) == "table" and node_or_nodes.is and node_or_nodes:is(Node)` ensures correct handling when passed a table that is not a class instance.
- **Descendant filtering** (`utils.filter_descendant_nodes`): when a directory and its children are both selected, only the directory is operated on. Walks the full `.parent` chain to handle `group_empty` grouped nodes and collapsed directories.
- **Single prompt** for delete/trash rather than per-node confirmation. Respects `ui.confirm.default_yes` config; prompt construction extracted to `utils.confirm_prompt()`.
- **Batch conflict resolution** for paste: non-conflicting items are pasted immediately; conflicting items are resolved in a single prompt with rename-suffix, overwrite-all, or skip-all options.
- **`Clipboard:finish_paste()`** calls `reload_explorer()` to refresh the tree after paste. This ensures the tree reflects filesystem changes regardless of watcher state, and clears clipboard highlights even when all conflicts are skipped.
- Visual mode is exited synchronously (`nvim_feedkeys` with `"nx"` flags) before any operation executes.

### Files Changed

- `utils.lua` -- `filter_descendant_nodes()`, `confirm_prompt()`, `is_visual_mode()`, `exit_visual_mode()`, `get_visual_nodes()`
- `api/impl/post.lua` -- `wrap_node_or_visual()` dispatcher for mode-dependent API functions
- `explorer/init.lua` -- `Explorer:get_nodes_in_range(start_line, end_line)`
- `actions/fs/remove-file.lua` -- `remove_one()` / `remove_many()` split, `Node` class check dispatch
- `actions/fs/trash.lua` -- `trash_one()` / `trash_many()` split, `Node` class check dispatch
- `actions/fs/clipboard.lua` -- `bulk_clipboard()`, `resolve_conflicts()`, `finish_paste()`, `Node` class check dispatch in `copy()` / `cut()`
- `marks/init.lua` -- `toggle_one()` extracted, `toggle()` handles single node or array with `Node` class check
- `keymap.lua` -- `{ "n", "x" }` mode tables for visual-capable keymaps
- `help.lua` -- mode column display, mode merging and sorting for duplicate lhs+desc entries
- `scripts/help-defaults.sh` -- parse both `"n"` and `{ "n", "x" }` mode forms for help doc generation
- `_meta/api/fs.lua`, `_meta/api/marks.lua` -- updated docs with "In visual mode..." notes
- `api.lua` -- mode-dependent behavior note
- `doc/nvim-tree-lua.txt` -- regenerated help with mode column


alex-courtis:
I see many Copilot comments on this PR.

Was AI used to generate this code?

v3ceban:
> I see many Copilot comments on this PR.

Yeah, it's just the review thingy. Didn't even know it'd work in someone else's repo.

> Was AI used to generate this code?

A bit. Initially, to find the right API functions in the nvim-tree codebase for the bindings that I left in #2994, and then for the automated review, obviously.

alex-courtis:
You can fix help generation by adding other mode handling to the script:

```patch
diff --git a/scripts/help-defaults.sh b/scripts/help-defaults.sh
index 11ffefe..609072f 100755
--- a/scripts/help-defaults.sh
+++ b/scripts/help-defaults.sh
@@ -45,7 +45,7 @@ sed -i -e "/${begin}/,/${end}/{ /${begin}/{p; r /tmp/ON_ATTACH_DEFAULT.lua

 # help human
 echo > /tmp/ON_ATTACH_DEFAULT.help
-sed -E "s/^ *vim.keymap.set\(\"n\", \"(.*)\",.*api(.*),.*opts\(\"(.*)\".*$/'\`\1\`' '\3' '|nvim_tree.api\2()|'/g
+sed -E "s/^ *vim.keymap.set\(\".\", \"(.*)\",.*api(.*),.*opts\(\"(.*)\".*$/'\`\1\`' '\3' '|nvim_tree.api\2()|'/g
 " /tmp/ON_ATTACH_DEFAULT.lua | while read -r line
 do
    eval "printf '%-17.17s %-26.26s %s\n' ${line}" >> /tmp/ON_ATTACH_DEFAULT.help
```

You can then commit the updated help as per https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md#help-documentation

alex-courtis:
> Yeah, it's just the review thingy. Didn't even know it'd work in someone else's repo.

Please don't use automated reviews in the future.

I'll get to a review this week.

Uanela:
This such a great feat, I was already looking forward to start working on this, I was just tired of jump from tree to oil and so on for some simple tasks.

alex-courtis:
I haven't had a chance to review the code, however cursory initial testing looks good. This is fantastic as it opens the door for future visual (and other) mode operations.

Looking at the new API like `nvim_tree.api.fs.copy.visual()`: do we actually need it?

Rather than adding new API, we could make the existing API context dependent: when `vim.api.nvim_get_mode().mode == "v"` or `V` or maybe `x` we do the visual copy otherwise we do "normal" node copy.

alex-courtis:
Disregard the CI that has not run; new test matricies have not yet filtered through.

v3ceban:
> I haven't had a chance to review the code, however cursory initial testing looks good. This is fantastic as it opens the door for future visual (and other) mode operations.

Thanks! And no worries - I’ve been quite busy myself - might not have a chance to jump on this again till tomorrow

> Rather than adding new API, we could make the existing API context dependent: when `vim.api.nvim_get_mode().mode == "v"` or `V` or maybe `x` we do the visual copy otherwise we do "normal" node copy.

Yeah, this sounds good to me. No reason to duplicate the apis if existing ones are ok to extend. I was a bit cautious to not break any direct api usage accidentally with the new mode, but the risk is probably low here.

What should we do with mapping and help menu though? I assume having a separate entry for visual mode won’t make sense then.

I can extend the description of existing maps to mention visual mode operations.

Or, perhaps, we could even skip mentioning it in the description at all? At least to me this felt like a “natural” thing to try when I first attempted to delete multiple files. Mentioning it for every supported mapping may just add redundancy.

---

sry for typos - sent from my phone

alex-courtis:
> Yeah, this sounds good to me. No reason to duplicate the apis if existing ones are ok to extend. I was a bit cautious to not break any direct api usage accidentally with the new mode, but the risk is probably low here.

Your caution is most gratefully appreciated.

For the mapping itself, we can set multiple modes (table) e.g.
```lua
  vim.keymap.set({"n","v"}, "c",              api.fs.copy.node,                   opts("Copy"))
```

> What should we do with mapping and help menu though? I assume having a separate entry for visual mode won’t make sense then.
 
We could be lazy with the help screen and just add a mode column in the middle, something like `nx` or `nv` or `nV`

`nvim-tree-quickstart-help` could mirror this, as the first two columns are the same.

> I can extend the description of existing maps to mention visual mode operations.
> 
> Or, perhaps, we could even skip mentioning it in the description at all? At least to me this felt like a “natural” thing to try when I first attempted to delete multiple files. Mentioning it for every supported mapping may just add redundancy.

That should be OK.

vim help already has `nvim-tree-mappings-default` and the users will see it in the help screen and `nvim-tree-quickstart-help`

We can add a catch all note to `nvim-tree-api` in the "Generally, functions accepting..." section to describe mode dependent behaviour of API that has modal mappings.

v3ceban:
> We could be lazy with the help screen and just add a mode column in the middle, something like `nx` or `nv` or `nV`
> 
> `nvim-tree-quickstart-help` could mirror this, as the first two columns are the same.
> 
> vim help already has `nvim-tree-mappings-default` and the users will see it in the help screen and `nvim-tree-quickstart-help`
> 
> We can add a catch all note to `nvim-tree-api` in the "Generally, functions accepting..." section to describe mode dependent behaviour of API that has modal mappings.

Ok, this sounds good to me! I’ll get to it whenever I have a chance (most likely in the next ~24 hours)

alex-courtis:
Cheers. No hurry at all.

v3ceban:
@alex-courtis Thanks for the patience with this! It's ready for review.

Kaiser-Yang:
Wonderful PR! Hope it can be merged.

lua/nvim-tree/api.lua
```diff
@@ -24,7 +24,9 @@
 ---local api = require("nvim-tree.api")
 ---api.tree.reload()
 ---```
----Generally, functions accepting a [nvim_tree.api.Node] as their first argument will use the node under the cursor when that argument is not present or nil. e.g. the following are functionally identical:
+---Generally, functions accepting a [nvim_tree.api.Node] as their first argument will use the node under the cursor when that argument is not present or nil. Some functions are mode-dependent: when invoked in visual mode they will operate on all nodes in the visual selection instead of a single node. See |nvim-tree-mappings-default| for which mappings support visual mode.
```
alex-courtis, line 27: Many thanks for the detail.

lua/nvim-tree/_meta/api/fs.lua
```diff
@@ -28,6 +28,7 @@ function nvim_tree.api.fs.copy.filename(node) end
 
 ---
 ---Copy to the nvim-tree clipboard.
+---In visual mode, copies all nodes in the visual selection.
```
alex-courtis, line 31: Adding the explicit notes is most gratefully appreciated.

lua/nvim-tree/api/impl/post.lua
```diff
@@ -172,18 +256,18 @@ function M.hydrate(api)
   api.tree.winid = view.winid
 
   api.fs.create = wrap_node_or_nil(actions.fs.create_file.fn)
-  api.fs.remove = wrap_node(actions.fs.remove_file.fn)
-  api.fs.trash = wrap_node(actions.fs.trash.fn)
+  api.fs.remove = wrap_node_or_visual_bulk(actions.fs.remove_file.fn, "marks", "bulk_delete_nodes")
+  api.fs.trash = wrap_node_or_visual_bulk(actions.fs.trash.fn, "marks", "bulk_trash_nodes")
```
alex-courtis: The original single functions called the actions.fs "domain", wheras the new bulk versions are calling the marks domain, which results in difficult to read and maintain code.

The bulk operations should reside in the actions.fs domain, not marks.

See https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2839038833

lua/nvim-tree/marks/init.lua
alex-courtis, line 1: There are significant refactors and changes in this class that do not appear necessary which results in unclear git history and a non-atomic commit.

- The prompt has been changed to respect `default_yes`. Whilst I appreciate the initiative, it is unrelated to this functionality and should be reverted. We can make that change in a future pull request.
- `bulk_delete_nodes` and trash have been added. They do not appear related to marks, which already has the bulk operations which do not require any change. They appear to be used only by API.fs 
- Many portions have been refactored or rewritten, such as extracting the `bulk_remove` and `bulk_trash` methods. These will be removed, however for future reference, the general OO principal of dynamic dispatch should be used instead of proxying a method with a, quite frankly, riduculous number of arguments. This is not readable or maintainable code.

In fact, it does not appear that any changes are required to this class: the existing `Marks:toggle` is successfully called repeatedly via `wrap_node_or_visual`, which is the only additional marks functionality.

scripts/help-defaults.sh
```diff
@@ -44,11 +44,20 @@ sed -i -e "/${begin}/,/${end}/{ /${begin}/{p; r /tmp/ON_ATTACH_DEFAULT.lua
...
-" /tmp/ON_ATTACH_DEFAULT.lua | while read -r line
+sed -E '
+  s/^ *vim\.keymap\.set\(\{([^}]+)\}, *"([^"]+)",.*api(.*),.*opts\("([^"]*)".*/\1 \2 \3 \4/
+  t reformat
+  s/^ *vim\.keymap\.set\("(.)", *"([^"]+)",.*api(.*),.*opts\("([^"]*)".*/\1 \2 \3 \4/
+  t reformat
+  d
+  :reformat
+  s/"//g
+  s/, //g
```
alex-courtis, line 58: Many thanks for splitting this sed program and making it readable.

v3ceban:
@alex-courtis Thank you for a detailed and in-depth review. This is a great learning opportunity for me and I sincerely appreciate your feedback! I believe I addressed all your concerns - please do let me know if anything is still not up to standards, is unrelated, or missing.

lua/nvim-tree/api/impl/post.lua
```diff
@@ -86,6 +86,76 @@ local function wrap_explorer_member(explorer_member, member_method)
   end
 end
 
+---Check if the current mode is visual (v, V, or CTRL-V).
+---@return boolean
+local function is_visual_mode()
+  local mode = vim.api.nvim_get_mode().mode
+  return mode == "v" or mode == "V" or mode == "\22" -- \22 is CTRL-V
```
alex-courtis: lualine agrees with this mode detection.

However, we should add the various select modes as well - some users like to use the mouse.

See https://github.com/nvim-lualine/lualine.nvim/blob/47f91c416daef12db467145e16bed5bbfe00add8/lua/lualine/utils/mode.lua

lua/nvim-tree/actions/fs/trash.lua
```diff
@@ -119,6 +119,46 @@ function M.fn(node)
...
+
+    if M.config.ui.confirm.default_yes then
+      prompt_input = prompt_select .. " Y/n: "
+      items_short = { "", "n" }
+      items_long = { "Yes", "No" }
+    else
+      prompt_input = prompt_select .. " y/N: "
+      items_short = { "", "y" }
+      items_long = { "No", "Yes" }
+    end
```
alex-courtis: These lines are identical to those in `fn()`. Please extract a function. The signature can look something like:

```lua
---@class TrashPrompt
---@field prompt_input string
---@field items_short string[]
---@field items_long string[]

---@param prompt_select string
---@return TrashPrompt
local function trash_prompt(prompt_select)
```

We normally want to avoid making changes to unrelated code, however having duplicated in code overrides this.

lua/nvim-tree/actions/fs/remove-file.lua
```diff
@@ -158,6 +158,50 @@ function M.fn(node)
...
+
+    if M.config.ui.confirm.default_yes then
+      prompt_input = prompt_select .. " Y/n: "
+      items_short = { "", "n" }
+      items_long = { "Yes", "No" }
+    else
+      prompt_input = prompt_select .. " y/N: "
+      items_short = { "", "y" }
+      items_long = { "No", "Yes" }
+    end
```
alex-courtis: As per https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862129812

lua/nvim-tree/actions/fs/remove-file.lua
```diff
@@ -158,6 +158,50 @@ function M.fn(node)
...
+  if #nodes == 0 then
+    return
+  end
+
+  local function execute()
+    for i = #nodes, 1, -1 do
+      if nodes[i].name ~= ".." then
+        M.remove(nodes[i])
+      end
+    end
```
alex-courtis: Good work removing the root node from remove and trash operations.

However, please
- use `node:is(RootNode)` instead of comparing the name
- use a simple list iterator `ipairs` instead of an integer incrementing loop. That is the lua way.

Same applies for trash.

lua/nvim-tree/api/impl/post.lua
```diff
@@ -86,6 +86,76 @@ local function wrap_explorer_member(explorer_member, member_method)
...
+          for _, n in ipairs(nodes) do
+            fn(n, ...)
+          end
+        end
+      end
+    else
+      node = node or wrap_explorer("get_node_at_cursor")()
+      if node then
+        return fn(node, ...)
+      end
```
alex-courtis: This duplicates `wrap_node`. Please call that.

lua/nvim-tree/api/impl/post.lua
```diff
@@ -86,6 +86,76 @@ local function wrap_explorer_member(explorer_member, member_method)
...
+---@param opts? WrapNodeOrVisualOpts
+---@return fun(node: Node?, ...): any
+local function wrap_node_or_visual(fn, opts)
+  opts = opts or {}
+  return function(node, ...)
+    if is_visual_mode() then
+      local nodes = get_visual_nodes()
+      if nodes then
+        if opts.filter_descendants ~= false then
+          nodes = utils.filter_descendant_nodes(nodes)
```
alex-courtis: Good work putting `filter_descendant_nodes` in utils.

Please move similar to utils:
- is_visual_mode
- exit_visual_mode
- get_visual_nodes

lua/nvim-tree/api/impl/post.lua
```diff
@@ -173,18 +243,18 @@ function M.hydrate(api)
   api.tree.winid = view.winid
 
   api.fs.create = wrap_node_or_nil(actions.fs.create_file.fn)
-  api.fs.remove = wrap_node(actions.fs.remove_file.fn)
-  api.fs.trash = wrap_node(actions.fs.trash.fn)
+  api.fs.remove = wrap_node_or_visual(actions.fs.remove_file.fn, { visual_fn = actions.fs.remove_file.visual_fn })
+  api.fs.trash = wrap_node_or_visual(actions.fs.trash.fn, { visual_fn = actions.fs.trash.visual_fn })
```
alex-courtis: These are very difficult to read and do not match the style of the surrounding code i.e. one api interface mapped to one function/method.

Let's refactor this:

Single entry point for remove:
```lua
---@param node Node
local function remove_one_node(node)
  ---
end

---@param nodes Node[]
local function remove_many_nodes(nodes)
  ---
end

---@param node_or_nodes Node|Node[] single node when called in normal mode, multiple nodes in visual mode
function M.fn(node_or_nodes)
  if node_or_nodes:is(Node) then
    remove_one_node(node_or_nodes)
  else
    remove_many_nodes(node_or_nodes)
  end
```

We then map the api to this single entry point:
```lua
  api.fs.remove = wrap_node_or_visual(actions.fs.remove_file.fn)
```

Yes, this goes against what I said earlier about proxies, however this is the lesser of two evils.

We now have the problem of how to choose whether to filter descendents. Plan:

- The implementation chooses to do the filtering itself.
- wrap_node_or_visual always passes all visual nodes.
- remove/trash etc. call filter_descendant_nodes themselves.

This moves implementation details from the api to the implementation itself.

lua/nvim-tree/api/impl/post.lua
```diff
@@ -247,7 +317,7 @@ function M.hydrate(api)
 
   api.marks.get = wrap_node(wrap_explorer_member("marks", "get"))
   api.marks.list = wrap_explorer_member("marks", "list")
-  api.marks.toggle = wrap_node(wrap_explorer_member("marks", "toggle"))
+  api.marks.toggle = wrap_node_or_visual(wrap_explorer_member("marks", "toggle"), { filter_descendants = false })
```
alex-courtis: As per https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862232144

lua/nvim-tree/api/impl/post.lua
```diff
@@ -173,18 +243,18 @@ function M.hydrate(api)
...
-  api.fs.trash = wrap_node(actions.fs.trash.fn)
+  api.fs.remove = wrap_node_or_visual(actions.fs.remove_file.fn, { visual_fn = actions.fs.remove_file.visual_fn })
+  api.fs.trash = wrap_node_or_visual(actions.fs.trash.fn, { visual_fn = actions.fs.trash.visual_fn })
   api.fs.rename_node = wrap_node(actions.fs.rename_file.fn(":t"))
   api.fs.rename = wrap_node(actions.fs.rename_file.fn(":t"))
   api.fs.rename_sub = wrap_node(actions.fs.rename_file.fn(":p:h"))
   api.fs.rename_basename = wrap_node(actions.fs.rename_file.fn(":t:r"))
   api.fs.rename_full = wrap_node(actions.fs.rename_file.fn(":p"))
-  api.fs.cut = wrap_node(wrap_explorer_member("clipboard", "cut"))
+  api.fs.cut = wrap_node_or_visual(wrap_explorer_member("clipboard", "cut"))
```
alex-courtis, line 202: As per https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862232144

Passing a table of `Node` or a single `Node` to cut and copy has the added benefit of allowing us to resolve the [multiple prompts](https://github.com/nvim-tree/nvim-tree.lua/pull/3268#issuecomment-3970492899)

alex-courtis (changes requested):
This review is partially complete. It is very time consuming and I am limited in time.

Following have not yet been reviewed:
- help window
- help doc
- `exit_visual_mode`
- `get_visual_nodes`
- `Explorer:get_nodes_in_range`
- `filter_descendant_nodes`

Please:

- [ ] [add select modes](https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862081893)
- [ ] [move utils into place](https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862196921)
- [ ] [extract common prompt code](https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862129812)
- [ ] [detect root based on class and user a regular lua iterator](https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862144877)
- [ ] [simplify post.lua with 1:1 api to implementation mappings](https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2862232144)
- [ ] following above, deal with multiple copy message spam

alex-courtis:
We are correctly receiving copy/cut messages, however a message is received for every node, which spams the user with prompts.

That is unacceptable UX and users have (correctly) raised issues about similar in the past.

One message is all we can send. It'll have to be a summary like:
`[NvimTree] 5 nodes added to clipboard.`



`c`

v3ceban:
> We are correctly receiving copy/cut messages, however a message is received for every node, which spams the user with prompts.
> 
> That is unacceptable UX and users have (correctly) raised issues about similar in the past.
> 
> One message is all we can send. It'll have to be a summary like: `[NvimTree] 5 nodes added to clipboard.`
> 
> 
> `c`
> 

I wasn't able to replicate the exact issue you described, but it led me to discover a related problem I hadn't tested — pasting files that already exist at the destination triggered a separate prompt for each file, which is the same kind of prompt spam we wanted to avoid.

To fix this, I updated the `Clipboard` class to handle paste conflicts in batch. When conflicts are detected, a single prompt offers three options: rename with a suffix (r), overwrite all (y), or skip all (n). If the chosen suffix still conflicts, it re-prompts. I realize this increases the scope of the PR, but I think it's necessary — without it, pasting a visual selection with conflicts would produce the same per-file prompt spam.

Demo: https://asciinema.org/a/ErimWquwpSyKlpE1

v3ceban:
> This review is partially complete. It is very time consuming and I am limited in time.

No worries, thank you for taking the time to review this in the first place. Appreciate all the feedback. I believe I did address your comments - let me know if anything is still missing, incorrect, or requires a refactor.

alex-courtis:
Thanks for the updates,  I'll get to a review shortly.

lua/nvim-tree/api/impl/post.lua
```diff
@@ -86,6 +86,25 @@ local function wrap_explorer_member(explorer_member, member_method)
   end
 end
 
+---Wrap a function to be mode-dependent: in visual mode, pass all nodes in the
+---visual range; in normal mode, pass a single node. The implementation decides
+---how to handle each case.
```
alex-courtis, line 91: Good description.

lua/nvim-tree/utils.lua
```diff
@@ -483,4 +483,82 @@ function M.enumerate_options(opts, was_set)
...
+  return filtered
+end
+
+---Build confirmation prompt strings based on default_yes config.
+---@param prompt_select string
+---@param default_yes boolean
+---@return string prompt_input
+---@return string[] items_short
+---@return string[] items_long
+function M.confirm_prompt(prompt_select, default_yes)
```
alex-courtis, line 509: Very nice, this can be used for future functionality.

lua/nvim-tree/actions/fs/remove-file.lua
```diff
@@ -165,6 +157,51 @@ function M.fn(node)
...
+      end
+    end)
+  else
+    execute()
+  end
+end
+
+---@param node_or_nodes Node|Node[]
+function M.fn(node_or_nodes)
+  if type(node_or_nodes) == "table" and node_or_nodes.is then
```
alex-courtis: Good work checking for a table - my comment completely forgot to do that.

This tests for the presence of the `is` method: that node_or_nodes is a class. This is necessary as we may be passed a table that is not a class. We do still need to check that it is an instance of the node class - a third condition: `node_or_nodes:is(Node)` see `:help nvim-tree-class`

Same applies to trash, clipboard and marks.

Further api sanitization is well outside the scope of this issue and will be addressed by #2668

lua/nvim-tree/marks/init.lua
```diff
@@ -100,9 +100,9 @@ function Marks:clear()
   self.explorer.renderer:draw()
 end
 
----@public
+---@private
```
alex-courtis, line 104: Good work changing the visibility.

lua/nvim-tree/utils.lua
```diff
@@ -483,4 +483,82 @@ function M.enumerate_options(opts, was_set)
...
+    end
+  end
+  local filtered = {}
+  for i, node in ipairs(nodes) do
+    if not dominated[i] then
+      table.insert(filtered, node)
+    end
+  end
+  return filtered
+end
```
alex-courtis, line 501: Using path mapping is working for most cases, however we need to account for the edge cases:
* windows: wsl and powershell with its variable silly C:/ etc.
* directory and file symlinks

Instead of matching the paths, we can directly query the node hierarchy. Something like this overly verbose sample:
```lua
    log.line("dev", "nodes = %s", node.absolute_path)
  end

  ---Filter function: is a node a descendent of nodes?
  ---@param node Node
  ---@return boolean true when node.parent is not present in nodes
  local function is_descendent_of_nodes(node)
    return not vim.tbl_contains(nodes, node.parent)
  end

  local filtered2 = vim.tbl_filter(is_descendent_of_nodes, nodes)

  for _, node in ipairs(filtered2) do
    log.line("dev", "filtered2 = %s", node.absolute_path)
  end
```

loops or filter functions or inline filters? There is no preference - it's a matter of personal taste, similar to java.

lua/nvim-tree/api/impl/post.lua
```diff
@@ -86,6 +86,25 @@ local function wrap_explorer_member(explorer_member, member_method)
...
+---@param fn fun(node_or_nodes: Node|Node[], ...): any
+---@return fun(node: Node?, ...): any
+local function wrap_node_or_visual(fn)
+  local wrapped = wrap_node(fn)
+  return function(node, ...)
+    if utils.is_visual_mode() then
+      local nodes = utils.get_visual_nodes()
+      if nodes then
```
alex-courtis, line 99: Good safety check.

lua/nvim-tree/explorer/init.lua
```diff
@@ -642,6 +642,22 @@ function Explorer:find_node(fn)
...
+  local nodes_by_line = self:get_nodes_by_line(core.get_nodes_starting_line())
+  local nodes = {}
+  for line = start_line, end_line do
+    local node = nodes_by_line[line]
+    if node and node.absolute_path then
+      table.insert(nodes, node)
+    end
+  end
+  return nodes
+end
```
alex-courtis, line 659: Tested OK for all 3 cases including `{root_folder_label}` false

scripts/help-defaults.sh
```diff
@@ -44,11 +44,20 @@ sed -i -e "/${begin}/,/${end}/{ /${begin}/{p; r /tmp/ON_ATTACH_DEFAULT.lua
...
+' /tmp/ON_ATTACH_DEFAULT.lua | while read -r mode lhs apipath desc
 do
-	eval "printf '%-17.17s %-26.26s %s\n' ${line}" >> /tmp/ON_ATTACH_DEFAULT.help
+  printf ' %-17.17s %-4.4s %-26.26s %s\n' "\`${lhs}\`" "${mode}" "${desc}" "|nvim_tree.api${apipath}()|" >> /tmp/ON_ATTACH_DEFAULT.help
```
alex-courtis, line 61: That new loop is far safer / more readable than the eval.

scripts/help-defaults.sh
```diff
@@ -44,11 +44,20 @@ sed -i -e "/${begin}/,/${end}/{ /${begin}/{p; r /tmp/ON_ATTACH_DEFAULT.lua
            }; /${end}/p; d; }" "${WIP}"
 
 # help human
+# extract mode, lhs, api, desc; handle both "n" and {"n", "x"} mode forms
 echo > /tmp/ON_ATTACH_DEFAULT.help
-sed -E "s/^ *vim.keymap.set\(\"n\", \"(.*)\",.*api(.*),.*opts\(\"(.*)\".*$/'\`\1\`' '\3' '|nvim_tree.api\2()|'/g
-" /tmp/ON_ATTACH_DEFAULT.lua | while read -r line
+sed -E '
+  s/^ *vim\.keymap\.set\(\{([^}]+)\}, *"([^"]+)",.*api(.*),.*opts\("([^"]*)".*/\1 \2 \3 \4/
```
alex-courtis, line 51: Works for edge cases and variations e.g.
`  vim.keymap.set({ "n", "x", "i", }, "d",              api.fs.remove,                      opts("Delete"))`

`  vim.keymap.set({ "n", "x", }, "d",              api.fs.remove,                      opts("Delete"))`

lua/nvim-tree/help.lua
alex-courtis, line 1: This is working well, however there is a nondeterministic ordering of the mode, which looks a bit odd as the `n`s are not lined up.

```
┌─────────────────────────────────────────────┐                         
│nvim-tree mappings                    exit: q│
│                       sort by description: s│
│ <2-LeftMouse>  n  Open                      │
│ <2-RightMouse> n  CD                        │
│ <C-]>          n  CD                        │
│ <C-E>          n  Open: In Place            │
│ <C-K>          n  Info                      │
│ <C-R>          n  Rename: Omit Filename     │
│ <C-T>          n  Open: New Tab             │
│ <C-V>          n  Open: Vertical Split      │
│ <C-X>          n  Open: Horizontal Split    │
│ <BS>           n  Close Directory           │
│ <CR>           n  Open                      │
│ <Del>          xn Delete                    │
│ <Tab>          n  Open Preview              │
│ .              n  Run Command               │
│ -              n  Up                        │
│ >              n  Next Sibling              │
│ <              n  Previous Sibling          │
│ ?              n  Help                      │
│ B              n  Toggle Filter: No Buffer  │
│ C              n  Toggle Filter: Git Clean  │
│ D              xn Trash                     │
│ E              n  Expand All                │
│ F              n  Live Filter: Clear        │
│ H              n  Toggle Filter: Dotfiles   │
│ I              n  Toggle Filter: Git Ignored│
│ J              n  Last Sibling              │
│ K              n  First Sibling             │
│ L              n  Toggle Group Empty        │
```

This is overly critical and a nitpik. It's completely up to you whether you'd like to address it. More for your general knowledge.

It looks like the modes returned from `vim.api.nvim_buf_get_keymap` might not be deterministically ordered, despite our passing an ordered list.

We can sort and merge a string list like that using table.sort and .concat, which might make the merging a bit simpler.

Demo, put this in a file e.g. `/tmp/foo.lua` and `:so` it:
```lua
local modes = {
  "x",
  "n",
}

print("unsorted table = " .. vim.inspect(modes))

local unsorted = table.concat(modes, "")

print("unsorted string = " .. unsorted)

table.sort(modes)

print("sorted table = " .. vim.inspect(modes))

local sorted = table.concat(modes, "")

print("sorted string = " .. sorted)
```

lua/nvim-tree/help.lua
```diff
@@ -139,10 +152,10 @@ local function compute(map)
   }
 
   -- mappings, left padded 1
-  local fmt = string.format(" %%-%ds %%-%ds", max_lhs, max_desc)
+  local fmt = string.format(" %%-%ds %%-%ds %%-%ds", max_lhs, max_mode, max_desc)
```
alex-courtis, line 165: Max is working correctly when there are mappings like `{ "v", "o", "s", "n", }`

alex-courtis:
> To fix this, I updated the `Clipboard` class to handle paste conflicts in batch. When conflicts are detected, a single prompt offers three options: rename with a suffix (r), overwrite all (y), or skip all (n). If the chosen suffix still conflicts, it re-prompts. I realize this increases the scope of the PR, but I think it's necessary — without it, pasting a visual selection with conflicts would produce the same per-file prompt spam.

That's a good call. Users will be doing more bulk pastes now that visual mode is available and we should harden it, with the skip option quite likely to be used. The rename suffix is a better option than rename.

alex-courtis (changes requested):
Looking great, only a few small changes required.

# Please

- [ ] [Check for node class](https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2881338413) 
- [x] [Filter by parent](https://github.com/nvim-tree/nvim-tree.lua/pull/3268/changes#r2881415245)

Preparing for release:
- You'll be "on call" for this change and will be mentioned in an issue/discussion
- If there is a breaking issue, we prefer to push a fix fast
  - If that's not safe or achievable, we will have to roll back and redo the change with the fix

v3ceban:
@alex-courtis Thanks again for the review! I’ll take a more detailed look and work on addressing the issues and edge cases you raised shortly.

I’m okay with being mentioned in the issues. Will do my best to respond and create PRs with fixes in a timely manner :)

lua/nvim-tree/utils.lua
v3ceban: Replaced path-string prefix matching with `node.parent` hierarchy traversal, as suggested. It still walks the full parent chain (not just one level) to correctly handle `group_empty` grouped nodes and collapsed directories where parents are not in the visual selection.

v3ceban:
@alex-courtis Ready for re-review (no rush, I know you're busy!)

Changes since last review:
  - Added `node_or_nodes:is(Node)` class check in all 4 dispatch sites
  - `filter_descendant_nodes` walks full `.parent` chain instead of path-prefix matching
  - Sorted mode characters in help display (thx for the nit, I completely missed that!)
  - `finish_paste` uses `reload_explorer()` to avoid stale tree after paste (missed this bug initially, it was creating a partially updated tree with pasted files added but not removed visually)
  - Updated PR description to reflect current state

alex-courtis:
Thanks for the updates, I'll get to a review shortly.

lua/nvim-tree/utils.lua
alex-courtis: Great work identifying and dealing with grouped nodes.

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -196,23 +148,119 @@ function Clipboard:clear_clipboard()
...
+    self:bulk_clipboard(utils.filter_descendant_nodes(node_or_nodes), self.data.copy, self.data.cut, "cut to")
+  end
+end
+
+---Clear clipboard for action and reload to reflect filesystem changes from paste.
+---@private
+---@param action ClipboardAction
+function Clipboard:finish_paste(action)
+  self.data[action] = {}
+  self.explorer:reload_explorer()
```
alex-courtis, line 208: Good catch. That's the safest approach.

alex-courtis:
This will be released as 1.16.0 once it's settled.

v3ceban:
@alex-courtis Thanks again for your time and all the reviews! Happy to see this in master!

PR #3261, opened 2026-02-11
fix(#3178): handle Windows paths in ignore_dirs and add .zig-cache to defaults

Uanela:
This PR fixes a critical memory leak issue on Windows where `nvim-tree.lua` spawns infinite filesystem watchers when the `.zig-cache/tmp` folder creates temporary files during a build process. The watcher locks these files, preventing Zig from deleting them and causing memory to grow by 10-20 MB/second.

## Changes

1. **Added `.zig-cache` to default `ignore_dirs`** - This directory contains build artifacts not under version control, similar to existing defaults like `node_modules` and `build`.

2. **Fixed Windows path handling in `ignore_dirs`** - The forward slash prefix (e.g., `/node_modules`) used in default ignore patterns wasn't matching Windows paths (e.g., `M:\path\to\node_modules`). The fix converts forward slashes to backslashes on Windows to ensure proper pattern matching.

## Technical Details

The existing `utils.escape_special_characters` utility was not suitable for this fix as it only handles parentheses and curly braces in paths, not the fundamental path separator difference between Unix (`/`) and Windows (`\`).

The solution implements OS-specific path normalization in `is_folder_ignored()` by converting forward slashes to backslashes on Windows systems before pattern matching with `vim.fn.match()`.

## Testing

Tested on Windows 10 with Zig build process. The `.zig-cache` directory is now properly ignored, and existing defaults like `node_modules` now correctly match on Windows paths.

Closes #3178

PR #3275, opened 2026-02-23
chore: build scripts use sed -E for cross-OS compatibility

alex-courtis:
Many thanks to @Uanela  for finding this one: https://github.com/nvim-tree/nvim-tree.lua/pull/3261#issuecomment-3932089099

I'd be really grateful for your testing with the "standard" macos sed.

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1231,9 +1231,11 @@ Config: renderer                                   *nvim-tree-config-renderer*
 
     Specify {decorators} is a list e.g. `{ "Git", MyDecorator, "Cut" }`
 
-    {root_folder_label} has 3 forms:
+    {root_folder_label}          *nvim_tree.config.renderer.root_folder_label*
+
+    Controls the root folder name and visibility:
     • `string`: |filename-modifiers| format string, default `":~:s?$?/..?"`
-    • `boolean`: `true` to disable
+    • `false`: to disable
     • `fun(root_cwd: string): string`: return a literal string from root's
       absolute path e.g. >lua
         my_root_folder_label = function(path)
@@ -1266,8 +1268,8 @@ Config: renderer                                   *nvim-tree-config-renderer*
       • {full_name}?               (`boolean`, default: `false`) Display nodes
                                    whose name length is wider than the width
                                    of nvim-tree window in floating window.
-      • {root_folder_label}?       (`string|boolean|(fun(root_cwd: string): string)`)
-                                   (default: `":~:s?$?/..?"`)
+      • {root_folder_label}?       (`nvim_tree.config.renderer.root_folder_label`, default: `":~:s?$?/..?"`)
+                                   |nvim_tree.config.renderer.root_folder_label|
       • {indent_width}?            (`integer`, default: `2`) Number of spaces
                                    for each tree nesting level. Minimum 1.
       • {hidden_display}?          (`nvim_tree.config.renderer.hidden_display`, default: `none`)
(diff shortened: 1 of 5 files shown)


v3ceban:
Oh, I had a lot of "fun" with this as well in the past! [GNU](https://www.gnu.org/software/sed/manual/sed.html) and [BSD](https://man.freebsd.org/cgi/man.cgi?sed(1)) (default macOS) sed are slightly different, particularly around the `-i` flag (GNU: `sed -i`, BSD: `sed -i ''`) and how the `r` command behaves inside brace groups and with multiple `-e` expressions. The person [here](https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed) summarized the differences well. Below are a couple of workarounds that I found with a simple example from the `help-defaults.sh`:

---

Current usage:
```bash
sed -i -e "/${inject}/r /tmp/DEFAULT_OPTS.6.lua" -e "/${inject}/d" "${WIP}"
```

Workarounds:

1. Avoid using `-i` flag entirely and do a two-stage pipe to first filter, then delete, then replace:
```bash
# NOTE: POSIX `r` command reads the filename until end of line,
# so a newline is required after the filename to terminate it.
# GNU sed treats the end of the -e string as a line boundary, so it works, but technically it's not POSIX compliant.
sed "/${inject}/r /tmp/DEFAULT_OPTS.6.lua
" "${WIP}" | sed "/${inject}/d" > "${WIP}.tmp" && mv "${WIP}.tmp" "${WIP}"
```
2. Maintain two variants of the command and use the correct one for the given OS:
```bash
if [ "$(uname -s)" = "Darwin" ]; then
  sed -i '' "/${inject}/r /tmp/DEFAULT_OPTS.6.lua
  " "${WIP}"
  sed -i '' "/${inject}/d" "${WIP}"
else
  sed -i -e "/${inject}/r /tmp/DEFAULT_OPTS.6.lua" -e "/${inject}/d" "${WIP}"
fi
```
3. Replace sed with **POSIX** [awk](https://www.man7.org/linux/man-pages/man1/awk.1p.html), which should behave the same on both systems:
```bash
awk "/${inject}/{while((getline line < \"/tmp/DEFAULT_OPTS.6.lua\") > 0) print line; next} {print}" "${WIP}" > "${WIP}.tmp" && mv "${WIP}.tmp" "${WIP}"
```
4. Require macOS users to install [gsed](https://formulae.brew.sh/formula/gnu-sed) and run the scripts with it on PATH:
```bash
PATH="$HOMEBREW_PREFIX/opt/gnu-sed/libexec/gnubin:$PATH" make help-update
 ```
 or conditionally call `gsed` instead of `sed` based on user's OS (similar to the second workaround).

Uanela:
@alex-courtis I will test this one shortly

alex-courtis:
> Below are a couple of workarounds that I found with a simple example from the `help-defaults.sh`:

Nice one @v3ceban , many solutions there.

1 is clear and simple. Using -i is not really necessary and is fragile and confusing.

3 is pragmatic, simple, compatible and there is precedent with doc-comments.sh

2 and 4 aren't very friendly and create friction.

Let's see how it tests.

Uanela:
So far it LGTM. maybe if this one comes back we can try some more detailed (aka kind complex) solution. Sorry for such big delay.

alex-courtis:
> So far it LGTM. maybe if this one comes back we can try some more detailed (aka kind complex) solution. Sorry for such big delay.

Cheers for the rebase.

Merging this one, yes, we should simplify at some point. #3268 first.

Issue #3281, opened 2026-02-27
view.width.max function returning -1 is not respected


alex-courtis:
### Description

When `view.width.max` is set to a function that returns `-1` the view does not resize to fit the content.

```
: ; git bisect good
776a5cdfac948b490e06f1d1d22c4cb986e40699 is the first bad commit
commit 776a5cdfac948b490e06f1d1d22c4cb986e40699
Author: Šimon Mandlík <simon.mandlik@gmail.com>
Date:   Sun Nov 30 22:29:34 2025 +0100

    feat(#3213): add view.width.lines_excluded option (#3214)

    feat(#3213): add `view.width.lines_excluded` option

 doc/nvim-tree-lua.txt  | 10 ++++++++++
 lua/nvim-tree.lua      | 67 +++++++++++++++++++++++++++++++++++++++++++------------------------
 lua/nvim-tree/view.lua | 33 ++++++++++++++++-----------------
 3 files changed, 69 insertions(+), 41 deletions(-)
```

I'd be most grateful if you could look into this one @simonmandlik 

### Neovim version

```text
NVIM v0.11.6
Build type: RelWithDebInfo
LuaJIT 2.1.1767980792
Run "nvim -V1 -v" for more info
```

### Operating system and version

Linux 6.18.9-arch1-2

### nvim-tree version

ae16aab5fd1792b97758622b9fb53408ec804fd8

### Clean room replication

```lua
vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1

vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvt-min/site]])
local package_root = "/tmp/nvt-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
  require("packer").startup({
    {
      "wbthomason/packer.nvim",
      "nvim-tree/nvim-tree.lua",
      "nvim-tree/nvim-web-devicons",
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. "/plugin/packer_compiled.lua",
      display = { non_interactive = true },
    },
  })
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing nvim-tree and dependencies.")
[...]

PR #3283, opened 2026-02-27
fix(#3281): fix a bug when a view width of -1 is returned from a function

simonmandlik:
Fixes https://github.com/nvim-tree/nvim-tree.lua/issues/3281


alex-courtis (approved):
Really appreciate the quick fix.

Issue test cases pass.

Issue #3292, opened 2026-03-20
Memory leak when file watcher enabled on Windows


andriysavin:
### Description

I'm observing enourmous memory consumption by neovim caused by nvim-tree plugin when using with large directory hiearchies like C: drive as root. My findings show that the cause of that is file system watcher. I know that the plugin is documented (aka by-design) to perform poorly in such scenarios, but I have some more thoughts on this.
1. I tried to do memory profiling of neovim in VS, and found out that most allocations occur in calls to luv_fs_event_cb which apparently puts lua callback request in a queue, and the queue is growing continuously.
2. Browsing nvim-tree source code and monitoring logs, I found out that nvim-tree registers FS watcher for every child directory and file _separately_ when a directory is expanded. I saw the comment that recursive events monitoring is not working for some reason (though AFAIK on windows FS watcher API works ok with hierarchies). What I think this leads to is that with huge amount of callback regisrations (I counted almost 40 000 (!) in the log for just expanding c:\Users\MyUserName\) neovim api has to poll for events (via ReadDirectoryChangesW system call) for each file system item separately. Results return happens asynchronously, so callback requests are queued in parallel from multiple threads, which increases chances of queue growing faster than it can be processed.
3. As mentioned in 2, the amount of watcher registrations does not correlate with amount of expanded direcories and visible children. In the log I see many records like "Watcher:create 'C:\Users\MyUserName\AppData\Local\Microsoft\VSApplicationInsights\vstelf<some guid>'", while I just expanded "c:\Users\MyUserName\". In nvim-tree code I see that [children exploration](https://github.com/nvim-tree/nvim-tree.lua/blob/dfeeb12aaae37f23285101cf875416c29052b22a/lua/nvim-tree/explorer/init.lua#L438) goes just one level, but in some cases it [goes deeper recursively](https://github.com/nvim-tree/nvim-tree.lua/blob/dfeeb12aaae37f23285101cf875416c29052b22a/lua/nvim-tree/explorer/init.lua#L455).  
I don't quite understand why, by I believe this is the reason the number of watched items is so huge.
[...]

Issue #3178, opened 2025-08-08
Memory leak caused by watcher on .zig-cache/tmp files


caboyd:
### Description

nvim-tree.lua spawns infinite watchers when the .zig-cache/tmp folder creates a temporary file during a build process. nvim-tree.lua locks the files and prevents zig from deleting it.

I assume this a windows only issue because windows locks files when they are open.

nvim-tree.log spits out this line 1933 times per second and nvim.exe grows in memory about 10-20 megabytes per second.

```
[2025-08-07 17:54:56] [watcher] event_cb 'M:\ComputerScience\GitLocal\c-engine\.zig-cache\tmp\2c5da8ef08bfc59e' 'M:\ComputerScience\GitLocal\c-engine\.zig-cache\tmp\2c5da8ef08bfc59e'
[2025-08-07 17:54:56] [watcher] node event scheduled refresh explorer:watch:M:\ComputerScience\GitLocal\c-engine\.zig-cache\tmp\2c5da8ef08bfc59e
```

### Solution
add ```/.zig-cache``` to default ignore_dirs for watcher

https://github.com/nvim-tree/nvim-tree.lua/blob/dd2364d6802f7f57a98acb8b545ed484c6697626/lua/nvim-tree.lua#L417-L424

### Neovim version

```text
NVIM v0.11.3
Build type: Release
LuaJIT 2.1.1741730670
```

### Operating system and version

Windows 10

### Windows variant

Powershell

### nvim-tree version

nvim-tree-v1.13.0

### Clean room replication

```lua
uneccesary
```

### Steps to reproduce

1. make sure ./zig-cache is not filtered or ignored
2. nvim
3. zig build

alex-courtis:
That's quite reasonable, as there is little value in watching those files which are not under version control.

This is the same use case as the others.

I'd be most grateful for your raising a Pull Request to add this:
* [CONTRIBUTING.md](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md)
* [Wiki: Development](https://github.com/nvim-tree/nvim-tree.lua/wiki/Development)

caboyd:
It appears on windows it doesn't resolve the forward slash into the correct directory so I was required to remove them. 

``` lua
     filesystem_watchers = {
        enable = true,
        debounce_delay = 50,
        ignore_dirs = {
          ".ccls-cache",
          "build",
          "node_modules",
          "target",
          ".zig-cache",
[...]

PR #3298, opened 2026-03-24
feat(#3292): add nvim_tree.config.filesystem_watchers.whitelist_dirs - strongly recommended for windows

alex-courtis:
Adds a watcher whitelist.

@andriysavin I'd be most grateful if you thoroughly tested the fix branch 

```sh
cd /path/to/nvim-tree.lua
git pull
git checkout fix/3292
```

When you're finished testing:
```sh
git checkout master
```

### DIFF
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1701,9 +1701,9 @@ Config: filesystem_watchers             *nvim-tree-config-filesystem-watchers*
       • Backslashes must be escaped e.g. `"my-proj/\\.build$"`
       • Literal backslashes must be double escaped to avoid |/magic| e.g.
         `"C:\\\\src\\\\my-proj"`
-    • OR a function that is passed an absolute path and returns:
-      • {ignore_dirs}: `true` to disable
-      • {whitelist_dirs}: `true` to enable
+    • OR a `fun(path: string): boolean` that is passed an absolute path:
+      • {ignore_dirs}: return `true` to disable
+      • {whitelist_dirs}: return `true` to enable
 
     {whitelist_dirs} is STRONGLY recommended for windows:
     |nvim-tree-os-specific|.
--- a/lua/nvim-tree/_meta/config/filesystem_watchers.lua
+++ b/lua/nvim-tree/_meta/config/filesystem_watchers.lua
@@ -11,9 +11,9 @@ error("Cannot require a meta file")
 ---- A table of |regular-expression| that will be passed to `vim.fn.match`
 ---   - Backslashes must be escaped e.g. `"my-proj/\\.build$"`
 ---   - Literal backslashes must be double escaped to avoid |/magic| e.g. `"C:\\\\src\\\\my-proj"`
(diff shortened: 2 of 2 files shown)


andriysavin:
Thanks, I'm looking. Have a silly question: are white-listed directories (as well as ignored directories) supposed to be plain paths or patterns? I'm asking because matching doesn't seem to work (everything is ignored regardless of whitelist). I see you're using `vim.fn.match(p, dir)`, and I believe it expects regex as second parameter. For instance, `:lua=vim.fn.match("E:\\Code", "E:\\Code")` returns -1, and this is exactly how whitelisted directories are matched.

alex-courtis:
> Thanks, I'm looking. Have a silly question: are white-listed directories (as well as ignored directories) supposed to be plain paths or patterns? I'm asking because matching doesn't seem to work (everything is ignored regardless of whitelist). I see you're using `vim.fn.match(p, dir)`, and I believe it expects regex as second parameter. For instance, `:lua=vim.fn.match("E:\\Code", "E:\\Code")` returns -1, and this is exactly how whitelisted directories are matched.

They're patterns, as are `{ignore_dirs}`. I wasn't really clear in the help, I'll polish that a bit.

We need double escapes for backslashes: https://github.com/nvim-tree/nvim-tree.lua/issues/3178#issuecomment-3862970723

This case is a problem:
```lua
local expr = 'E:\\Code';

local pat = 'E:\\\\Code';

local ret = vim.fn.match(expr, pat)

print(expr .. " " .. pat .. " match ret=" .. ret)

local pat2 = 'E:\\Code';

local ret2 = vim.fn.match(expr, pat2)

print(expr .. " " .. pat2 .. " NO match ret2=" .. ret2)
```

```
E:\Code E:\\Code match ret=0
E:\Code E:\Code NO match ret2=-1
```

alex-courtis:
I've rewritten the help: please review for clarity: https://github.com/nvim-tree/nvim-tree.lua/pull/3298/changes#diff-b5b2d1abcb86ead5a2319f7e065a7545b834a407cc9f420590d7acebb285fca9

andriysavin:
Thanks, with double escaping the path works well. 

I've been using the whitelist throughout the day and it does the job! I don't see any unusual memory allocation when browsing non-whitelisted locations. This doesn't fix watcher digging whitelisted dirs, but at least in those dirs not that many events happen to cause memory growth. 

And updated docs look great, thanks again!

alex-courtis:
> Thanks, with double escaping the path works well.
> 
> I've been using the whitelist throughout the day and it does the job! I don't see any unusual memory allocation when browsing non-whitelisted locations. This doesn't fix watcher digging whitelisted dirs, but at least in those dirs not that many events happen to cause memory growth.
> 
> And updated docs look great, thanks again!

Thanks for the thorough testing, merging.

You can add more {ignore_dirs} that will be still obeyed inside whitelisted. I've had to add a few over the years.

andriysavin:
Great, thanks for your work!

PR #3302, opened 2026-03-28
perf(#3257): remove setup for renderer, log, appearance

### LINKED ISSUE #3257: Performance: Remove all setup functions
Part of #3231 : Performance: Load Only Necessary Modules On setup and requiring API

This MUST be broken into multiple PRs: logical groups such as git or individual modules. There is a high blast radius and fast revert must be available.

Module setup should not be invoked at nvim-tree setup time, as they require the world which is expensive.

They should use the config module created in #3256

The setup function should be completely removed. In the case of any config being mutated such as `system-open` setup, it should be moved to the config module, executed at setup time.

Scope at c07ce43

`rg -l "function.*\.setup\(opts\)" lua`

- [x] actions/fs/*
- [x] actions/node/*
- [x] actions/root/*
- [x] actions/tree/*
- [x] buffers.lua
- [x] diagnostics.lua
- [x] explorer/watch.lua
- [x] git/*
- [x] help.lua
- [x] keymap.lua
- [x] lib.lua
- [x] notify.lua
- [x] renderer/components/*
- [x] view.lua
- [x] watcher.lua

alex-courtis:
fixes #3257

### DIFF
--- a/lua/nvim-tree/actions/fs/rename-file.lua
+++ b/lua/nvim-tree/actions/fs/rename-file.lua
@@ -192,8 +192,4 @@ function M.rename_full(node)
   prompt_to_rename(node, ":p")
 end
 
-function M.setup(opts)
-  config.g.filesystem_watchers = opts.filesystem_watchers
-end
-
 return M


alex-courtis:
superseded by #3302

PR #3313, opened 2026-04-06
fix: eliminate flicker when opening tree window

0ax1:
Suppress events during the open+reposition to prevent intermediate redraws.

### DIFF
--- a/lua/nvim-tree/view.lua
+++ b/lua/nvim-tree/view.lua
@@ -121,8 +121,14 @@ local function open_window()
   if config.g.view.float.enable then
     vim.api.nvim_open_win(0, true, open_win_config())
   else
-    vim.api.nvim_command("vsp")
+    local ei = vim.o.eventignore
+    vim.o.eventignore = "all"
+    vim.api.nvim_open_win(M.get_bufnr(), true, {
+      split = config.g.view.side or "left",
+      width = view_state.get_width(),
+    })
     M.reposition_window()
+    vim.o.eventignore = ei
   end
   setup_tabpage(vim.api.nvim_get_current_tabpage())
   set_window_options_and_buffer()


lua/nvim-tree/view.lua
```diff
@@ -121,8 +121,14 @@ local function open_window()
   if config.g.view.float.enable then
     vim.api.nvim_open_win(0, true, open_win_config())
   else
-    vim.api.nvim_command("vsp")
+    local ei = vim.o.eventignore
+    vim.o.eventignore = "all"
+    vim.api.nvim_open_win(M.get_bufnr(), true, {
```
alex-courtis, line 126: What's the reason for this change from `vsp`?

This code is very fagile and has to handle a lot of edge cases.

alex-courtis (changes requested):

Please can you provide a test case that can demonstrate the problem. I can then validate the fix.

lua/nvim-tree/view.lua
gegoune: LLM might have not 'known' about it though…

0ax1: `vim.api.nvim_command("vsp")` opens a vsplit showing the current buffer in the new split first and then swapping it after in my understanding. What I'm trying to do here is ignore displaying intermediate steps via `local ei = vim.o.eventignore`, set the correct buffer from the start `vim.api.nvim_open_win(M.get_bufnr()`, position the window like before `M.reposition_window()`, and then finally make the changes visible `vim.o.eventignore = ei`. 

Right now I see a glitch each time I toggle nvim-tree to visible, as in I can see that vsp initially splits the buffer has the content of the current buffer, before it repositions and shows the content of the file tree.

So the repro really only is toggle nvim-tree, not using a float window.

0ax1: @gegoune Yeah I've been iterating with on this with Claude, and might be totally possible that some subtleties have been missed. I never really went deep on nvim plugins before but this seemed reasonable as a first iteration. 

@alex-courtis Maybe it's possible to keep `vsp` if we ignore intermediate events `local ei = vim.o.eventignore`?

0ax1:
> Many thanks for your contribution!
> 
> Please can you provide a test case that can demonstrate the problem. I can then validate the fix.

Thanks for having a look! I mentioned a flow in the other comment. Let me know if you meant an automated test.

lua/nvim-tree/view.lua
alex-courtis: It looks like `set_window_options_and_buffer` is already temporarily setting eventignore via `vim.api.nvim_set_option`. Using eventignore is very dangerous and we use it only when we absolutely have to.

I can't see the flicker however I do understand how it occurs. If I comment out `set_window_options_and_buffer` it does indeed show the current buffer in the tree window.

> @alex-courtis Maybe it's possible to keep vsp if we ignore intermediate events local ei = vim.o.eventignore?

I did try wrapping vsp with the eventignore, however that had no effect - the current buffer is still displayed.

alex-courtis (changes requested):
I'm really reluctant to
- make this `nvim_open_win` change to open a new window with the tree buffer, as there is already a lot of code to switch the buffer in, that would need to be remediated
- add an extremely risky `eventignore`, which I'm not convinced is actually helping

This is a very fragile area of code and has broken many times in history, hence we don't make changes unless absolutely necessary.

How about a less risky compromise:
1. Use `:vnew` so that the window opens with a new empty buffer
2. Make the new buffer a `help scratch-buffer` with `bufhidden=wipe`

I'm open to using lua API for 2, however changing 1 from vimscript is just too risky.

0ax1:
Seems like a quite an intricate part of the logic in nvim-tree. Happy to keep this as is to not accidentally disrupt other ppls flow.

PR #3334, opened 2026-06-11
feat(#1549): copy files between multiple vim instances

### LINKED ISSUE #1549: copied files working in multiple vim instances
i use use neovim instance for each of projects of mine.  sometimes i want to copy file from one to another and its really hard to do with nvim-tree clipboard working only current instance

Uanela:
Closes #1549 

## Summary

This PR introduces the feature to allow copying files between multiple vim instances plus some code small refactors to improve the codebase.

### Changes

-  core changes where made into clipboard by reusing `copy_absolute_path` and `paste` by passing `opts` with `use_protocol`.
- keymaps added `bgy` - copy using protocol and `bgp` paste using protocol (prefered `bgy`+`bgp` instead of `bgy`+`gp` from #1549 because of consistency and I tell from own xp while testing)
-  Added customizable protocol label through `actions.clipboard.protocol`
-  Made `copy_basename` and `copy_absolute_path` visual capable, reason almost ZERO for me, but I think is something users would expect to exist, mainly now that we've many visual operations

## What Can Be Improved (My thoughts)

1 - Maybe rename `bgy` to `bgc` to be consistent with `c` our current copy key
2 - Maybe add `bgx` to allow cutting the files (which will only add a slight change into the copied text), and if we want to allow cutting should it be on copying to clipboard time or decide when pasting, e.g:
- `bgx` and then `bgp`, basically I decided to `cut` on copying to clipboard time
- `bgc` and then `bgp` (copy) or `bgxp` (or whatever) to mean paste
- or apply both.
I am leaning more toward first option to match what we do in our OS already we decide cut/copy at the time we are selecting the file(s).

BTW this was a wild and exciting PR for me, the conversation under #1549 and the changes regarding visual copy/cut/delete from @v3ceban helped a lot.


Uanela:
I am here thinking about the protocol thing, I think maybe it should be removed so that it can allow me to easily copy file from my OS explorer to nvim basically I copy the path the `bgp` and done.

I we gotta build extensions for OS.

Uanela:
Then if we remove the protocol we have no need to a new command like `bgy` just do `gy` in `n` or `x` and get the list of file paths, the `gp` to paste those.

alex-courtis:
> * `bgx` and then `bgp`, basically I decided to `cut` on copying to clipboard time

That sounds like a reasonable solution that would avoid a lot of complexity. The user will be explicitly cutting and will be aware of this behaviour.

Looking great, I'll get to a review shortly.

alex-courtis:
> I am here thinking about the protocol thing, I think maybe it should be removed so that it can allow me to easily copy file from my OS explorer to nvim basically I copy the path the `bgp` and done.

That would be very useful.

I've had a very quick look at the XDG spec for extended clipboard handling but found nothing. Experimentation is needed.

Uanela:
Should I remove the protocol for a while?

alex-courtis:
> Should I remove the protocol for a while?

If it works now, leave it. We can change it later if we want, as we don't need to worry about compatibility.

Uanela:
It does works fine and without it would work as well even with lesser code.

If we leave the protocol won't be able to simply copy a path from wherever
and then `gp` (gp over bgy because the copy would simply be what we use the
`gy`).

But if it is fine for you that's ok. the job is done.

Would only be left on the table whether add `bgx` (with protocol) or a
`gpx|gx` (without protocol would mean paste while cutting).

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -304,12 +320,42 @@ function Clipboard:resolve_conflicts(conflict, destination, action, action_fn)
...
+    return
+  end
+
+  local prefix = self.protocol .. ": "
+  if content:sub(1, #prefix) ~= prefix then
+    return
+  end
+
+  local nodes = {}
+  local absolute_paths = vim.split(content:sub(#prefix + 1, #content), ", ")
```
alex-courtis, line 338: Working nicely, mixed file and dir copied properly with a visual `bgy` and `bgp`

`nvim-tree: /home/alex/src/linux/xxx/, /home/alex/src/linux/aa`

lua/nvim-tree/config.lua
```diff
@@ -232,6 +232,9 @@ M.d = { -- config-default-start
   },
   actions = {
     use_system_clipboard = true,
+    clipboard = {
+      protocol = "nvim-tree"
+    },
```
alex-courtis, line 237: This doesn't seem to add much value, however you've addressed it in #3336

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -176,10 +182,20 @@ function Clipboard:bulk_clipboard(nodes, from, to, verb)
   self.explorer.renderer:draw()
 end
 
+---@private
+---@param node_or_nodes Node|Node[]
+---@return boolean
+function Clipboard:is_nodes_array(node_or_nodes)
```
alex-courtis, line 188: Thank you for abstracting.

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -304,12 +320,42 @@ function Clipboard:resolve_conflicts(conflict, destination, action, action_fn)
...
+      node_args.name = vim.fn.fnamemodify(absolute_path:sub(1, -2), ":t")
+      table.insert(nodes, DirectoryNode(node_args))
+    else
+      table.insert(nodes, FileNode(node_args))
+    end
+  end
+  return nodes
+end
+
 ---Paste cut or copy with batch conflict resolution.
```
alex-courtis, line 352: When pasting a file that doesn't exist e.g. `nvim-tree: /home/alex/src/linux/foo/`, the error is:

```
[NvimTree]
Could not copy /home/alex/src/linux/foo/
 - ENOENT: no such file or directory: /home/alex/src/linux/foo/
```

We should check that the file/dir exists before pasting, so that we can give a meaningful message like
```Could not paste /home/alex/src/linux/foo/ ```

File: lua/nvim-tree/node/init.lua [hunk 5]
```diff
@@ -159,4 +159,14 @@ function Node:expand(expand_opts)
   end
 end
 
+function Node:get_basename()
```
alex-courtis, line 162: Thank you for abstracting. We could simplify by adding a root implementation e.g.

```lua
function RootNode:get_basename()
  return vim.fn.fnamemodify(self.explorer.absolute_path, ":t:r")
end
```

then simplifying the base node implementation.

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -318,7 +364,7 @@ function Clipboard:do_paste(node, action, action_fn)
       node = dir:last_group_node()
     end
   end
-  local clip = self.data[action]
+  local clip = opts and opts.use_protocol and self:get_nodes_from_reg() or self.data[action]
```
alex-courtis, line 367: ~~This falls back correctly when~~
~~`actions.use_system_clipboard = false,` and `&clipboard` unset~~

~~However, to be safe we should disable protocol when we do not have a system clipboard.~~

Edit: it is fine, the extra complexity is not worth it

alex-courtis (changes requested):
This is looking great. Small changes for this which also look to apply to #3336 

- [ ] split Node:get_basename
- [ ] nicer error when pasting file that doesn't exist

Briefly looking at #3336 it seems a lot simpler and more natural than this change. Review incoming shortly.

Uanela:
> This is looking great. Small changes for this which also look to apply to #3336
> 
> * [ ]  split Node:get_basename
> * [ ]  nicer error when pasting file that doesn't exist
> 
> Briefly looking at #3336 it seems a lot simpler and more natural than this change. Review incoming shortly.

Exactly and it makes it much easier for the user to interact with e.g: copying absolute path from another places and copy/cut to the project inside nvim.

lua/nvim-tree/actions/fs/clipboard.lua
Uanela: The `clipboard` config is out anyway.

lua/nvim-tree/actions/fs/clipboard.lua
Uanela: Really missed this one

Uanela:
I will wait for the review of #3336 so that we can decide which one gets merged.

Uanela:
This one is dead in favor of #3336

Issue #1992, opened 2023-02-12
nvim-tree no longer works with rmagatti/auto-session following "Open At Startup" option removal.


Cudochi:
### Description

I updated my nvim-tree plugin, and saw that the startup options had all been marked as deprecated.
I never had a problem with them, in fact, it's what made me migrate from NERDTree at the time : 
The ability to save and restore session with rmagatti/auto-session while always keeping a file tree on the side.

### Neovim version

```text
NVIM v0.8.0-dev-1081-g907fc8ac3
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-10 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/runner/work/neovim/neovim/build/cmake.config -I/home/runner/work/neovim/neovim/src -I/home/runner/work/neovim/neovim/.deps/usr/include -I/usr/include -I/home/runner/work/neovim/neovim/build/src/nvim/auto -I/home/runner/work/neovim/neovim/build/include
Compilé par runner@fv-az202-333

Features: +acl +iconv +tui
See ":help feature-compile"

         fichier vimrc système : "$VIM/sysinit.vim"
               $VIM par défaut : "/share/nvim"

Run :checkhealth for more info
```

### Operating system and version

Ubuntu 22.04.1 LTS

### nvim-tree version

36e29c3

### Minimal config

```lua
Nvim-tree config :

require'nvim-tree'.setup { -- BEGIN_DEFAULT_OPTS
  sync_root_with_cwd = true,
  respect_buf_cwd = true,
  update_focused_file = {
    enable = true,
    update_root = true
  },
[...]

PR #3335, opened 2026-06-12
feat: experiment: restore nvim-tree on session load, see https://github.com/nvim-tree/nvim-tree.lua/issues/3343

igorlfs:
Hello,

I'm on a quest to make all of my plugins (vim) "[session](https://neovim.io/doc/user/usr_21/#_sessions) friendly". For nvim-tree this sounds pretty straightforward, so I went ahead and implemented it myself.

**Problem**: nvim-tree leaves stale empty buffers after restoring a session

**Solution**: detect these buffers on `SessionLoadPost` and reopen them.

### DIFF
--- a/lua/nvim-tree/autocmd.lua
+++ b/lua/nvim-tree/autocmd.lua
@@ -17,6 +17,23 @@ function M.global()
     end,
   })
 
+  if vim.fn.has("nvim-0.13") == 1 and config.g.experimental.session_restore_nvim then
+    vim.api.nvim_create_autocmd("SessionWritePre", { ---@diagnostic disable-line: param-type-mismatch
+      group = augroup_id,
+      callback = function()
+        require("nvim-tree.session").save()
+      end,
+    })
+
+    vim.api.nvim_create_autocmd("SessionLoadPost", {
+      group = augroup_id,
+      callback = function()
+        require("nvim-tree.session").restore()
+      end,
+    })
+  end
+
+
   if config.g.tab.sync.open then
     vim.api.nvim_create_autocmd("TabEnter", {
       group = augroup_id,
--- a/lua/nvim-tree/autocmd.lua
+++ b/lua/nvim-tree/autocmd.lua
@@ -27,6 +27,9 @@ function M.global()
 
     vim.api.nvim_create_autocmd("SessionLoadPost", {
       group = augroup_id,
+      -- Workaround: this event fires multiple times (report upstream),
+      -- failure notifications could show up many times.
+      once = true,
       callback = function()
         require("nvim-tree.session").restore()
       end,
--- a/lua/nvim-tree/autocmd.lua
+++ b/lua/nvim-tree/autocmd.lua
@@ -21,7 +21,9 @@ function M.global()
     vim.api.nvim_create_autocmd("SessionWritePre", { ---@diagnostic disable-line: param-type-mismatch
       group = augroup_id,
       callback = function()
-        require("nvim-tree.session").save()
+        -- We must schedule to wait until `vim.v.this_session` is properly updated
+        -- This prevents overwriting the data if a new session is created with a new `:mksession`
+        vim.schedule(require("nvim-tree.session").save)
       end,
     })
 
--- a/lua/nvim-tree/session.lua
+++ b/lua/nvim-tree/session.lua
@@ -1,5 +1,9 @@
 local M = {}
 
+local function load_state()
+  return vim.json.decode(vim.g.NvimTreeState or "{}") or {}
+end
+
 function M.save()
   local cwd = require("nvim-tree.core").get_cwd()
   vim.g.NvimTreeState = vim.json.encode({ cwd = cwd })
@@ -11,7 +15,10 @@ function M.restore()
   ---@type table<integer,boolean>
   local tabs = {}
 
-  local session_state = vim.json.decode(vim.g.NvimTreeState or "{}") or {}
+  local ok, session_state = pcall(load_state)
+  if not ok then
+    require("nvim-tree.notify").warn(string.format("Failed to restore cwd: %s", session_state))
+  end
   local path = session_state and session_state.cwd or nil
 
   -- Save tabs with leftover nvim-tree buffers
--- a/lua/nvim-tree/session.lua
+++ b/lua/nvim-tree/session.lua
@@ -1,12 +1,59 @@
+local notify = require("nvim-tree.notify")
+
 local M = {}
 
-local function load_state()
-  return vim.json.decode(vim.g.NvimTreeState or "{}") or {}
+local storepath = vim.fn.stdpath("data") .. "/nvim-tree-session.json"
+
+local function decode_session_file()
+  local file = io.open(storepath, "r")
+
+  if file then
+    local content = file:read("*all")
+    file:close()
+    if content and content ~= "" then
+      return vim.json.decode(content)
+    end
+  end
+  -- Notifying on failure could be noisy, as it'd happen if the file simply did not exist
+end
+
+local function write_session_file(session_file_content)
+  local file, errmsg = io.open(storepath, "w")
+
+  if file then
+    file:write(vim.json.encode(session_file_content))
+    file:close()
+  else
+    notify.warn(string.format("Invalid session file (%s): %s", storepath, errmsg))
+  end
 end
 
 function M.save()
-  local cwd = require("nvim-tree.core").get_cwd()
-  vim.g.NvimTreeState = vim.json.encode({ cwd = cwd })
+  -- We either always save the session data, which ends up populating the storepath with empty data
+  -- Or we gate by the tree's visibility, which could create "out of sync" scenario:
+  -- if the tree is initially open and the session is saved, it won't be overridden later,
+  -- if the tree is no longer visible. However, when restoring, this is harmless.
+  if require("nvim-tree.api").tree.is_visible({ any_tabpage = true }) then
+    local ok, session_file_content = pcall(decode_session_file)
+
+    if not ok then
+      notify.warn(string.format("Failed to decode session file (%s): %s", storepath, session_file_content))
+      return
+    end
+
+    -- might be nil at this point
+    session_file_content = session_file_content or {}
+
+    local cwd = require("nvim-tree.core").get_cwd()
+
+    session_file_content[vim.v.this_session] = { cwd = cwd }
+
+    local ok2, err = pcall(write_session_file, session_file_content)
+
+    if not ok2 then
+      notify.warn(string.format("Failed to write session file (%s): %s", storepath, err))
+    end
+  end
 end
 
 function M.restore()
@@ -15,11 +62,14 @@ function M.restore()
   ---@type table<integer,boolean>
   local tabs = {}
 
-  local ok, session_state = pcall(load_state)
+  local ok, session_file_content = pcall(decode_session_file)
+
+  -- Failing to restore cwd is not fatal, we can continue
   if not ok then
-    require("nvim-tree.notify").warn(string.format("Failed to restore cwd: %s", session_state))
+    notify.warn(string.format("Failed to decode session file (%s): %s", storepath, session_file_content))
   end
-  local path = session_state and session_state.cwd or nil
+
+  local path = session_file_content and session_file_content[vim.v.this_session] and session_file_content[vim.v.this_session].cwd or nil
 
   -- Save tabs with leftover nvim-tree buffers
   for _, tab in ipairs(vim.api.nvim_list_tabpages()) do
--- a/lua/nvim-tree/session.lua
+++ b/lua/nvim-tree/session.lua
@@ -1,0 +1,46 @@
+local M = {}
+
+function M.save()
+  local cwd = require("nvim-tree.core").get_cwd()
+  vim.g.NvimTreeState = vim.json.encode({ cwd = cwd })
+end
+
+function M.restore()
+  local api = require("nvim-tree.api").tree
+
+  ---@type table<integer,boolean>
+  local tabs = {}
+
+  local session_state = vim.json.decode(vim.g.NvimTreeState or "{}") or {}
+  local path = session_state and session_state.cwd or nil
+
+  -- Save tabs with leftover nvim-tree buffers
+  for _, tab in ipairs(vim.api.nvim_list_tabpages()) do
+    for _, win in ipairs(vim.api.nvim_tabpage_list_wins(tab)) do
+      local buf = vim.api.nvim_win_get_buf(win)
+      if api.is_tree_buf(buf) then
+        tabs[tab] = true
+      end
+    end
+  end
+
+  -- The new window may inherit options from the current one if we don't schedule
+  vim.schedule(function()
+    api.close_in_all_tabs()
+
+    for tab, _ in pairs(tabs) do
+      if vim.api.nvim_tabpage_is_valid(tab) then
+        local win = vim.api.nvim_tabpage_get_win(tab)
+
+        -- Invoke `open` from each tab's current window to prevent having to switch tabs
+        if vim.api.nvim_win_is_valid(win) then
+          vim.api.nvim_win_call(win, function()
+            api.open({ path = path })
+          end)
+        end
+      end
+    end
+  end)
+end
+
+return M
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1911,6 +1911,13 @@ Config: experimental                           *nvim-tree-config-experimental*
     In the event of a problem please disable the experiment and raise an
     issue.
 
+    Fields: ~
+      • {session_restore_nvim}?  (`boolean`, default: `false`) Restore
+                                 nvim-tree buffers when restoring vim sessions
+                                 (requires 0.13+). To restore the working
+                                 directory, ensure that |'sessionoptions'|
+                                 includes "globals"
+
 ==============================================================================
 Config: log                                             *nvim-tree-config-log*
 
@@ -2216,6 +2223,7 @@ Following is the default configuration, see |nvim_tree.config| for details. >lua
         persist = false,
       },
       experimental = {
+        session_restore_nvim = false
       },
       log = {
         enable = false,
--- a/doc/nvim-tree-lua.txt
+++ b/doc/nvim-tree-lua.txt
@@ -1914,9 +1914,7 @@ Config: experimental                           *nvim-tree-config-experimental*
     Fields: ~
       • {session_restore_nvim}?  (`boolean`, default: `false`) Restore
                                  nvim-tree buffers when restoring vim sessions
-                                 (requires 0.13+). To restore the working
-                                 directory, ensure that |'sessionoptions'|
-                                 includes "globals"
+                                 (requires 0.13+).
 
 ==============================================================================
 Config: log                                             *nvim-tree-config-log*
--- a/lua/nvim-tree/_meta/config/experimental.lua
+++ b/lua/nvim-tree/_meta/config/experimental.lua
@@ -10,7 +10,6 @@ error("Cannot require a meta file")
 ---@class nvim_tree.config.experimental
 ---
 ---Restore nvim-tree buffers when restoring vim sessions (requires 0.13+).
----To restore the working directory, ensure that |'sessionoptions'| includes "globals"
 ---(default: `false`)
 ---@field session_restore_nvim? boolean
 --Example below for future reference:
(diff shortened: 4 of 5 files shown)


lua/nvim-tree/autocmd.lua
```diff
@@ -17,6 +17,41 @@ function M.global()
     end,
   })
 
+  vim.api.nvim_create_autocmd("SessionLoadPost", {
```
igorlfs: I went with this location since it was the standard, but it might a good idea have this as an eager autocmd. Otherwise, any means of "lazy loading" may end up calling `setup` too late (sessions could be loaded relatively early, as in, part of the initilization with `nvim -S Session.vim`).

lua/nvim-tree/autocmd.lua
```diff
@@ -17,6 +17,41 @@ function M.global()
...
+        api.close_in_all_tabs()
+
+        for tab, _ in pairs(tabs) do
+          if vim.api.nvim_tabpage_is_valid(tab) then
+            local win = vim.api.nvim_tabpage_get_win(tab)
+
+            if vim.api.nvim_win_is_valid(win) then
+              vim.api.nvim_win_call(win, function()
+                api.open()
+              end)
```
igorlfs: There's an edge case with this approach: if `nvim-tree` was the only buf from a tabpage, it will not be restored. I think that's fine, as it's sort of unlikely. Could add a comment as a reference, or otherwise look into it. The vim api is not very tab friendly, but I guess it should be possible to hook an unholy `:[N]tabnew`...

igorlfs:
Am I missing something or is the CI failure unrelated?

alex-courtis:
We've had a lot of problems with session integration in the past e.g. #1992, with nodeterministic issues surrounding plogin load and event order etc.

The outcome was that we elected not to support sessions, using a [recipe](https://github.com/nvim-tree/nvim-tree.lua/wiki/Recipes#workaround-when-using-rmagattiauto-session) instead.

First question: is an error thrown on startup, or does the session simply fail to correctly load?

Options:
- Publish your solution as a recipe
- Package your solution as an [extension plugin](https://github.com/nvim-tree/nvim-tree.lua/wiki/Integrations-And-Extension-Plugins)

Are the above possible using API?

alex-courtis:
> Am I missing something or is the CI failure unrelated?

Apologies, an upstream Nvim change broke the build, fixed on master #3337

igorlfs:
> We've had a lot of problems with session integration in the past

Thanks, I should have looked this up.

> The outcome was that we elected not to support sessions

That's unfortunate to hear. If this decision is final, feel free to close this PR. Else, since the discussion in the other threads is lengthy, could you propose some "up to date" requirements, with, potentially, some compromises? For instance, [this comment](https://github.com/nvim-tree/nvim-tree.lua/issues/1992#issuecomment-1455504628) suggests that folks using session plugins ([rmagatti/auto-session](https://github.com/rmagatti/auto-session) in particular) should "be on their own".

The comment also mentions "persisting something about the buffer" (which is also mentioned in [this comment](https://github.com/nvim-tree/nvim-tree.lua/issues/1992#issuecomment-1441215381)). Regarding that, I have an idea which is already employed by some plugins (ref: [kulala.nvim](https://github.com/mistweaverco/kulala.nvim/pull/937), [nvim-dap-view](https://github.com/igorlfs/nvim-dap-view/commit/b8eb02427993e454f7da658e0bcf3d1110423494)): **storing the state inside a global variable**. Global variables are restored when `'sessionoptions'` includes `globals` (the conditions are slightly stricter and there are other caveats, but I digress).

This approach could be leveraged to restore additional information, besides plainly reopening the buffers. At a most basic level, it could be used to properly restore the `cwd`. But maybe it could be used to fully restore the tree's state (although I think this would be more complex, since files may have been deleted or whatever).

If we are willing to use features only available on nightly, we can also take advantage of the ([recently merged](https://github.com/neovim/neovim/pull/39688)) `SessionWritePre`, a new event that could also be helpful with the "stateful restore".

A rough outline of what I have in mind for an MVP is:

- Leave "session plugin users on their own"
- When a `:mksession` is issued, use `SessionWritePre` to update the global variable that stores nvim-tree's state
- Setup an **eager** autocmd on `SessionLoadPost` to restore the state. What I mean by "eager" is that it should be "reachable" from `plugin/nvim-tree.lua` (potentially "inlined"). Given that session plugins are not interfering with their own `SessionLoadPost` hooks _and_ that the user isn't using any shenanigans to lazy-load the plugin, this should be much less "race prone".

Of course, that's a really "happy path" plan. If you are uncomfortable with the 'globals' approach, I was thinking about raising a discussion upstream, regarding the "best practices for plugins handling sessions" (aka `lua-plugin-sessions`). If the other ideas are also "far fetched", I'm open to alternatives. And if we can't settle down on the outline, I'm happy with closing and keeping this as a module on my own config. I could publish it as a recipe as well (current implementation would is easily "copy-pasteable").

> Are the above possible using API?

Current implementation is not relying on any internals.

---

> First question: is an error thrown on startup, or does the session simply fail to correctly load?

Do you mean **if** an error is thrown on startup, or something else? Raising an error on `VimEnter` does not prevent the session from loading. Buffers are restored as usual. `SessionLoadPre` is also fine. Haven't checked other events (I'm not sure about the event's load order, if you have any specific hints, I can dig into them). I could try a more "destructive" error (e.g., manipulating windows, etc) if needed.

alex-courtis:
> > We've had a lot of problems with session integration in the past
> 
> Thanks, I should have looked this up.
> 
> > The outcome was that we elected not to support sessions
> 
> That's unfortunate to hear. If this decision is final, feel free to close this PR. Else, since the discussion in the other threads is lengthy, could you propose some "up to date" requirements, with, potentially, some compromises? For instance, [this comment](https://github.com/nvim-tree/nvim-tree.lua/issues/1992#issuecomment-1455504628) suggests that folks using session plugins ([rmagatti/auto-session](https://github.com/rmagatti/auto-session) in particular) should "be on their own".

The problems were all around reliability and robustness - many instances of errors thrown on startup. A solution that is very defensive and absolutely does not throw exceptions is acceptable.

> 
> The comment also mentions "persisting something about the buffer" (which is also mentioned in [this comment](https://github.com/nvim-tree/nvim-tree.lua/issues/1992#issuecomment-1441215381)). Regarding that, I have an idea which is already employed by some plugins (ref: [kulala.nvim](https://github.com/mistweaverco/kulala.nvim/pull/937), [nvim-dap-view](https://github.com/igorlfs/nvim-dap-view/commit/b8eb02427993e454f7da658e0bcf3d1110423494)): **storing the state inside a global variable**. Global variables are restored when `'sessionoptions'` includes `globals` (the conditions are slightly stricter and there are other caveats, but I digress).
> 
> This approach could be leveraged to restore additional information, besides plainly reopening the buffers. At a most basic level, it could be used to properly restore the `cwd`. But maybe it could be used to fully restore the tree's state (although I think this would be more complex, since files may have been deleted or whatever).

Setting cwd is quite achievable: `nvim_tree.api.tree.open()` will set nvim-tree cwd to Nvim pwd. The desired cwd can be explicitly passed via `{path}`

Full state could be very problematic, due to the great number of configuration options and live state changes (e.g. no-buffer filter) that we have no means of serialising or persisting.

A `nvim_tree.api.tree.find_file()` might be possible however there may be unforseen problems.

> 
> If we are willing to use features only available on nightly, we can also take advantage of the ([recently merged](https://github.com/neovim/neovim/pull/39688)) `SessionWritePre`, a new event that could also be helpful with the "stateful restore".

That changes everything; an official and time predictable mechanism. Yes, new Nvim features can be used via a [version check](https://github.com/nvim-tree/nvim-tree.lua/blob/master/CONTRIBUTING.md#backwards-compatibility) to enable only when available, see `Explorer`

> 
> A rough outline of what I have in mind for an MVP is:
> 
> * Leave "session plugin users on their own"
> * When a `:mksession` is issued, use `SessionWritePre` to update the global variable that stores nvim-tree's state
> * Setup an **eager** autocmd on `SessionLoadPost` to restore the state. What I mean by "eager" is that it should be "reachable" from `plugin/nvim-tree.lua` (potentially "inlined"). Given that session plugins are not interfering with their own `SessionLoadPost` hooks _and_ that the user isn't using any shenanigans to lazy-load the plugin, this should be much less "race prone".
> 
> Of course, that's a really "happy path" plan. If you are uncomfortable with the 'globals' approach, I was thinking about raising a discussion upstream, regarding the "best practices for plugins handling sessions" (aka `lua-plugin-sessions`). If the other ideas are also "far fetched", I'm open to alternatives. And if we can't settle down on the outline, I'm happy with closing and keeping this as a module on my own config. I could publish it as a recipe as well (current implementation would is easily "copy-pasteable").

A predictable `SessionWritePre` makes a globals approach desirable. The code would be simple on read and write, taking all the guess-work away, thus making it quite robust.

Yes to a global, however please to raise also start a discussion as there may be other unconsidered options or the opportunity for further upstream changes.

Please try this MVP:
- global write
- global read
- open with cwd
  - multiple tabs is up to you at this stage

> 
> > Are the above possible using API?
> 
> Current implementation is not relying on any internals.
> 
> > First question: is an error thrown on startup, or does the session simply fail to correctly load?
> 
> Do you mean **if** an error is thrown on startup, or something else? Raising an error on `VimEnter` does not prevent the session from loading. Buffers are restored as usual. `SessionLoadPre` is also fine. Haven't checked other events (I'm not sure about the event's load order, if you have any specific hints, I can dig into them). I could try a more "destructive" error (e.g., manipulating windows, etc) if needed.

No error right now is good, I was quite concerned. With defensive programming we can keep it that way.

alex-courtis:
There is an alternative to globals: a json state file.

There is precedent with bookmark persistence: `:help nvim_tree.config.bookmarks`

igorlfs:
Latest push now uses the "globals" approach to save and restore the `cwd`.

> start a discussion as there may be other unconsidered options or the opportunity for further upstream changes.

[Here it is](https://github.com/neovim/neovim/issues/40336)!

> There is an alternative to globals: a json state file.

I can look into that, given the precedent. But we can also wait for some feedback on the upstream discussion, whatever you prefer.

---- 

CI is failing because the event does exist on stable 🤔 
Thoughts on that?

alex-courtis:
> Latest push now uses the "globals" approach to save and restore the `cwd`.
> 
> > start a discussion as there may be other unconsidered options or the opportunity for further upstream changes.
> 
> [Here it is](https://github.com/neovim/neovim/issues/40336)!
> 
> > There is an alternative to globals: a json state file.
> 
> I can look into that, given the precedent. But we can also wait for some feedback on the upstream discussion, whatever you prefer.

That sounds sensible. The global is a good placeholder until we learn more.

We can always refactor / change this at a later date as it is not API.

> 
> CI is failing because the event does exist on stable 🤔 Thoughts on that?

That is annoying. It looks like we'll have to use a suppression for the new path instead of the usual old path:
```lua
    vim.api.nvim_create_autocmd("SessionWritePre", { ---@diagnostic disable-line: param-type-mismatch
```

Those suppressions are regularly cleaned up, when we increase the minimum version of Nvim.

lua/nvim-tree/autocmd.lua
alex-courtis: That's fine. We are very aware of how insufficient the tab api is and how much trouble we've had with it.

Let's stick to the happy path cases to keep it simple and reliable.

Yes please: lots of comments in here.

igorlfs:
CI is happy now, should I squash?

alex-courtis (changes requested):
Let's run this as an experiment: we can make changes as needed and the user expectations will be correctly set.

- [x] create an experimental config `experimental.session_restore_nvim_0.13`
- [x] add doc to `lua/nvim-tree/_meta/config/experimental.lua`
- [x] gate the autocommands behind the experiment flag
- [x] create an issue similar to #2819 to describe the new functionality and give users a place to raise issues and discuss - I'll pin it

alex-courtis:
> CI is happy now, should I squash?

Up to you. The PR will be sqash merged anyway.

alex-courtis:
https://github.com/neovim/neovim/issues/40336#issuecomment-4757303962

Uppercase globals seem to be a step in the right direction.

Hopefully more guidance follows.

igorlfs:
Checklist done 🫡  (I can't edit your comment 😭)

Edit: I named the option `session_restore_nvim` because the doc-gen did not like having to deal with a period inside an option name. Hopefully the name is fine, else a rename is trivial.

lua/nvim-tree/autocmd.lua
igorlfs: > Yes please: lots of comments in here.

Is the current amount enough?

alex-courtis (changes requested):
Looking good so far...

Apologies for suggesting a `.` in the name.
 
I'm not able to achieve session restore for cwd, I think I might be doing something wrong.

Testing with nightly 0.13 3b8c19ea46

```
VIMRUNTIME=/home/alex/src/neovim/runtime /home/alex/src/neovim/build/bin/nvim ...
```

Test case:

Added at the end of pre:
```lua
print("SessionWritePre set vim.g.NvimTreeState", vim.g.NvimTreeState)
```

start of post:
```lua
print("SessionLoadPost read vim.g.NvimTreeState", vim.g.NvimTreeState)
```

Set `experimental.session_restore_nvim`

- `:mksession`
  `SessionWritePre set vim.g.NvimTreeState {"cwd":"/home/alex/src/nvim-tree/r/3335"}`
- `:echo g:NvimTreeState`
  `{"cwd":"/home/alex/src/nvim-tree/r/3335"}`
- exit
- start with `-S Session.vim`
```
SessionLoadPost read vim.g.NvimTreeState nil
SessionLoadPost path nil
SessionLoadPost read vim.g.NvimTreeState nil
SessionLoadPost path nil
```
- `:echo g:NvimTreeState`
  `E121: Undefined variable: g:NvimTreeState`

Multiple `SessionLoadPost` events on startup is strange - is that expected?

igorlfs:
> I think I might be doing something wrong

I suspect so. `sessionoptions` must include "globals", which is a non-default flag. Added a note in the documentation.

> Multiple SessionLoadPost events on startup is strange - is that expected?

I think that's odd on its own, but unrelated.

alex-courtis:
> > I think I might be doing something wrong
> 
> I suspect so. `sessionoptions` must include "globals", which is a non-default flag. Added a note in the documentation.

Got it. After setting it, I am able to restore the session for multiple tabs.

Given the need for that option, and [Justin's cautious blessing of using stdpath data](https://github.com/neovim/neovim/issues/40336#issuecomment-4881931791), I think it would be best to move to a file.

With a file, it would "just work" out of the box. Following experiment success, this would be a no option / always enabled feature. Nvim-tree state persistence a reasonable expectation for session users and they will be the only ones hitting this codepath.

> 
> > Multiple SessionLoadPost events on startup is strange - is that expected?
> 
> I think that's odd on its own, but unrelated.

`SessionLoadPost` actions are idempotent, so I reckon it should be OK.

alex-courtis (changes requested):
Very close now, this is greatly exceeding my expectations.

Please:
- [x] move to `session.lua`
- [x] handle json parse errors
- [x] [persist in a file](https://github.com/nvim-tree/nvim-tree.lua/pull/3335#issuecomment-4898862086), say, `nvim-tree-session.json`

Remember that this is experimental and not API, hence we can change things like persistence in the future, in a breaking manner.

# Move to new module

Apologise, I should have asked for this earlier. This is functionality that is first class and will likely be enhanced in the future. We also want to keep `autocmds.lua` as minimal as possible.

The callbacks could be something like:
```lua
    callback = function()
      require("nvim-tree.session").restore()
    end,
```
Why require inline inside the callback? Requires at startup time should be minimal, for fastest possible startup.

# Invalid json handling

We do need some error handling. Test case:
- create a session and exit
- modify the state in Session.vim to something invalid `let NvimTreeState = "{\"cwd\":\"/home/alex/src/nvim-tree/r/3335/foo bar/two three\"} foo bar"`
Error message 4 times:
```
Error in command line..script /home/alex/src/nvim-tree/r/3335/Session.vim[125]..SessionLoadPost Autocommands for "*":
Lua callback: ...packer/start/nvim-tree.lua.dev/lua/nvim-tree/autocmd.lua:40: Expected the end but found invalid token at character 61
stack traceback:
        [C]: in function 'decode'
        ...packer/start/nvim-tree.lua.dev/lua/nvim-tree/autocmd.lua:40: in function <...packer/start/nvim-tree.lua.dev/lua/nvim-tree/autocmd.lua:32>
```

Bookmark failures do present any json issues with some context and stop after the first failure.
```
[NvimTree] Failed to load bookmarks: ...ker/start/nvim-tree.lua.dev/lua/nvim-tree/marks/init.lua:50: Expected comma or array end but found invalid token at character 4
```
(oversight: bookmark fails should also be printing the path of the bookmark file)

alex-courtis:
RE nightly check failure, I'll fix that on master. It might be time to remove nightly from all PRs.

Edit: see https://github.com/nvim-tree/nvim-tree.lua/pull/3345

igorlfs:
> I think it would be best to move to a file.

Done in afad15129139ffd61432f54909b274533af0f8bc, although, I must say, this added quite a bit of complexity.

> `SessionLoadPost` actions are idempotent, so I reckon it should be OK.

This ended up causing trouble once I introduced error notifications: sometimes up to 7 notifications would pop up! With UI2 that's not a big deal, since it "accumulates" repeated messages and just shows a counter. But, 7 notifications is odd nonetheless.

lua/nvim-tree/session.lua
```diff
@@ -0,0 +1,103 @@
...
+      notify.warn(string.format("Failed to decode session file (%s): %s", storepath, session_file_content))
+      return
+    end
+
+    -- might be nil at this point
+    session_file_content = session_file_content or {}
+
+    local cwd = require("nvim-tree.core").get_cwd()
+
+    session_file_content[vim.v.this_session] = { cwd = cwd }
```
alex-courtis, line 49: I like using the session path as a key - minimise any ambiguities.

Test: Multiple sessions in one file are correctly loaded and the correct session is updated.

lua/nvim-tree/autocmd.lua
```diff
@@ -17,6 +17,28 @@ function M.global()
...
+        -- This prevents overwriting the data if a new session is created with a new `:mksession`
+        vim.schedule(require("nvim-tree.session").save)
+      end,
+    })
+
+    vim.api.nvim_create_autocmd("SessionLoadPost", {
+      group = augroup_id,
+      -- Workaround: this event fires multiple times (report upstream),
+      -- failure notifications could show up many times.
+      once = true,
```
alex-courtis, line 34: Seems to work nicely; I only get messages once and sessions loaded correctly.

If we run into problems, say, execution on the first event is too early, we could use the debouncer instead, for delayed one-shot e.g.

```lua
    utils.debounce("DiagnosticChanged redraw", config.g.diagnostics.debounce_delay, function()
      local profile_redraw = log.profile_start("DiagnosticChanged redraw")

      local explorer = core.get_explorer()
      if explorer then
        explorer.renderer:draw()
      end

      log.profile_end(profile_redraw)
    end)
```

alex-courtis (approved):
This is just fantastic. You've taken me from no-sessions-supported to first class session functionailty.

Storing at mksession is the key: we know exactly what the state is and persist it. Previous attempts have always been pure guesswork.

# Test Cases

OK Write / Read / Update Multiple Sessions In One File

## Write

### Permission Bad

`chmod 444 ~/.local/share/nvim/nvim-tree-session.json`

`mksession`

`[NvimTree] Invalid session file (/home/alex/.local/share/nvim/nvim-tree-session.json): /home/alex/.local/share/nvim/nvim-tree-session.json: Permission denied`

### Bad Contents

`echo 'foo bar baz' > ~/.local/share/nvim/nvim-tree-session.json`

`mksession`

`[NvimTree] Failed to decode session file (/home/alex/.local/share/nvim/nvim-tree-session.json): Expected value but found invalid token at character 1`

Content preserved so that user may recover.

### Write Fail

Artificial test: add `file:close()` before `file:write` 

`mksession`

`[NvimTree] Failed to write session file (/home/alex/.local/share/nvim/nvim-tree-session.json): ...packer/start/nvim-tree.lua.dev/lua/nvim-tree/session.lua:25: attempt to use a closed file`

## Read

### Bad Contents

`echo 'foo bar baz' > ~/.local/share/nvim/nvim-tree-session.json`

`[NvimTree] Failed to decode session file (/home/alex/.local/share/nvim/nvim-tree-session.json): Expected value but found invalid token at character 1`

Content preserved so that user may recover.

## Tabs

Did not test leftover nvim-tree buffers

alex-courtis:
Hope you don't mind, I merged master first.

alex-courtis:
Issue is pinned and mentioned in the commit message.

I'm watching the issue and asked people to tag you if there are any problems.

igorlfs:
Thanks for the patience, I'm glad we got here!

PR #3336, opened 2026-06-14
feat(#1549): p will paste files from another nvim-tree instance, gp will cut/paste

Uanela:
Hello @alex-courtis this is simply a variation of what we've been discussing at #3334 I just wanted to show it in practice and is btw the thing that I've been dogfooding, I am thinking in even adding option to download a file directly when an url is on the register (maybe for my own locally if this is too nichy).

## Changes

1 - `gy` - remains as it is, and now supports visual operation which copies a list of comma separated absolute paths
2 - `gp` - this is used to paste while copying
3 - `gx` - this is used to paste while cutting.

https://github.com/user-attachments/assets/e35223ad-1fe2-4af9-b0e0-fae08b7c7944


Uanela:
Added `gp` to automatically download when find a `^http`, currently is just local on my machine.

image

image

lua/nvim-tree/node/init.lua
```diff
@@ -159,4 +159,14 @@ function Node:expand(expand_opts)
   end
 end
 
+function Node:get_basename()
+  if self.name == ".." then
```
alex-courtis: On further reflection, changing the `".."` makes me nervous. Yes it's legacy, however it is present in many places in `Clipboard` hence we shouldn't change it when building this feature.

Let's keep it localised - please move this method to `Clipboard` so that it doesn't spread to other functionality.

alex-courtis (changes requested):
This is looking great, only minor changes required.

Please:
- [x] handle spaces etc. in register as per `gp` test case, moving to a [newline delimited format as per "standard"](https://github.com/nvim-tree/nvim-tree.lua/pull/3336#issuecomment-4725104964) with newlines between files, but not after the last file
- [x] [move `Node:get_basename` to `Clipboard`](https://github.com/nvim-tree/nvim-tree.lua/pull/3336/changes#r3424844102)

Test cases:

# gy

- [x] copy one `/home/alex/src/linux/.mailmap`
- [x] visual copy `/home/alex/src/linux/.pylintrc,/home/alex/src/linux/.rustfmt.toml`
- [x] copy root `/home/alex/src/linux/`
- [ ] ~~visual copy root and other nodes~~
  root is not copied, however that's OK. This is existing behaviour and arguably a feature, as copying root and other files makes no sense.

# ge

- [x] copy one `.mailmap`
- [x] visual copy `.pylintrc,.rustfmt`
- [x] copy root `linux`
- [ ] ~~visual copy root and other nodes, ok, see above~~

# gp

- [x] paste one `/home/alex/src/linux/MAINTAINERS`
- [x] paste many `/home/alex/src/linux/COPYING,/home/alex/src/linux/CREDITS`
- [ ] FAIL - paste with spaces `    /home/alex/src/linux/CREDITS,   /home/alex/src/linux/Kbuild,   /home/alex/src/linux/Kconfig  `
  [NvimTree] Could not copy     /home/alex/src/linux/CREDITS - ENOENT
- [x] paste one inexistent `/home/alex/src/linux/xxx`
  [NvimTree] Could not copy /home/alex/src/linux/xxx - ENOENT
- [x] paste many, one inexistent `/home/alex/src/linux/Kbuild,/home/alex/src/linux/xxx,/home/alex/src/linux/MAINTAINERS`
  [NvimTree] Could not copy /home/alex/src/linux/xxx - ENOENT
  others copied correctly
- [x] paste mixed nodes and directories `/home/alex/src/linux/zzz/,/home/alex/src/linux/aaa`

# gx

- [x] Test cases as per `gp`
- [x] Multiple source removed when in same instance `/home/alex/src/linux/Kconfig,/home/alex/src/linux/MAINTAINERS`
- [x] Multiple source removed in other instance in the same directory `/home/alex/src/linux/.clang-format,/home/alex/src/linux/.clippy.toml`

alex-courtis:
> I am thinking in even adding option to download a file directly when an url is on the register (maybe for my own locally if this is too nichy).

That would be great, I have some use cases for it. Out of scope, future PR please :)

alex-courtis:
Wild and crazy idea after reading comments on previous PR:

## Remove `gp`: `p` can handle all the cases

- If there are files copied or cut in the current instance, paste them as per original `Clipboard:paste()`
- Else if the system/register clipboard contains files, paste them as per `use_register = true`

## `Clipboard:copy()` and `:cut()` copy to system clipboard

These methods should also work as per `gy`, copying absolute paths to the system/register clipboard

This is something I've tried to do many times - `c` a file in one instance and `p` into another, which of course fails.

## `gx`

This is a niche case, however would need to be retained: we don't have any information in the clipboard to identify cut vs paste hence it must be explicit.

alex-courtis:
I finally installed the gnome file manager nautilus and inspected the clipboard mime types when copying multiple files:

```sh
: ; wl-paste --list-types
x-special/gnome-copied-files
text/plain;charset=utf-8
text/uri-list
application/vnd.portal.filetransfer
application/vnd.portal.files


: ; wl-paste --type text/plain\;charset=utf-8
/home/alex/src/linux/.gitattributes
/home/alex/src/linux/.gitignore
/home/alex/src/linux/.mailmap


: ; wl-paste --type text/uri-list
file:///home/alex/src/linux/.gitattributes
file:///home/alex/src/linux/.gitignore
file:///home/alex/src/linux/.mailmap


: ; wl-paste --type application/vnd.portal.filetransfer
10924863569610452011%


: ; wl-paste --type application/vnd.portal.files
11848482484090875390%


: ; wl-paste --type x-special/gnome-copied-files
copy
file:///home/alex/src/linux/.gitattributes
file:///home/alex/src/linux/.gitignore
file:///home/alex/src/linux/.mailmap%

: ;
```

**Today I think we should at least match the `text/plain` standard - use newline delimited files instead of comma delimited. Note that there is no newline at the end of the last file. Spaces are not escaped.**

`text/uri-list` is interesting and something that we could do in the future, however I'm not sure how easy or possible it is to set the mime type is within Nvim.

`x-special/gnome-copied-files` is completely gnome specific, however it inspires the possibly of something like `x-special/nvim-tree-copied-files`, which we could use to specify the copy/cut operation. Maybe not so useful.

For reference, this is what a `gy` results in:
```sh
: ; wl-paste --list-types
text/plain
text/plain;charset=utf-8
TEXT
STRING
UTF8_STRING

: ; wl-paste --type UTF8_STRING
/home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig

: ; wl-paste --type TEXT
/home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig

: ; wl-paste --type STRING
/home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig

: ; wl-paste --type text/plain
/home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig

: ; wl-paste --type text/plain\;charset=utf-8
/home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig

: ;
```

KDE dolpin operates similarly except that it does not copy `text/plain`

Uanela:
> Wild and crazy idea after reading comments on previous PR:
> 
> ## Remove `gp`: `p` can handle all the cases
> 
> - If there are files copied or cut in the current instance, paste them as per original `Clipboard:paste()`
> - Else if the system/register clipboard contains files, paste them as per `use_register = true`
> 
> ## `Clipboard:copy()` and `:cut()` copy to system clipboard
> 
> These methods should also work as per `gy`, copying absolute paths to the system/register clipboard
> 
> This is something I've tried to do many times - `c` a file in one instance and `p` into another, which of course fails.
> 
> ## `gx`
> 
> This is a niche case, however would need to be retained: we don't have any information in the clipboard to identify cut vs paste hence it must be explicit.

Makes total sense removing `gp` and making `c` and `x` take text to register.

The only thing is that `x` would still mean nothing, only when `gx` it would cut. Or maybe let `gp` to mean paste while cutting bc to some extent `gx` sounds like cut paths bc of `gy`, or even `xp` or something else.

Uanela:
> I finally installed the gnome file manager nautilus and inspected the clipboard mime types when copying multiple files:
> 
> ```sh
> : ; wl-paste --list-types
> x-special/gnome-copied-files
> text/plain;charset=utf-8
> text/uri-list
> application/vnd.portal.filetransfer
> application/vnd.portal.files
> 
> 
> : ; wl-paste --type text/plain\;charset=utf-8
> /home/alex/src/linux/.gitattributes
> /home/alex/src/linux/.gitignore
> /home/alex/src/linux/.mailmap
> 
> 
> : ; wl-paste --type text/uri-list
> file:///home/alex/src/linux/.gitattributes
> file:///home/alex/src/linux/.gitignore
> file:///home/alex/src/linux/.mailmap
> 
> 
> : ; wl-paste --type application/vnd.portal.filetransfer
> 10924863569610452011%
> 
> 
> : ; wl-paste --type application/vnd.portal.files
> 11848482484090875390%
> 
> 
> : ; wl-paste --type x-special/gnome-copied-files
> copy
> file:///home/alex/src/linux/.gitattributes
> file:///home/alex/src/linux/.gitignore
> file:///home/alex/src/linux/.mailmap%
> 
> : ;
> ```
> 
> **Today I think we should at least match the `text/plain` standard - use newline delimited files instead of comma delimited. Note that there is no newline at the end of the last file. Spaces are not escaped.**
> 
> `text/uri-list` is interesting and something that we could do in the future, however I'm not sure how easy or possible it is to set the mime type is within Nvim.
> 
> `x-special/gnome-copied-files` is completely gnome specific, however it inspires the possibly of something like `x-special/nvim-tree-copied-files`, which we could use to specify the copy/cut operation. Maybe not so useful.
> 
> For reference, this is what a `gy` results in:
> ```sh
> : ; wl-paste --list-types
> text/plain
> text/plain;charset=utf-8
> TEXT
> STRING
> UTF8_STRING
> 
> : ; wl-paste --type UTF8_STRING
> /home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig
> 
> : ; wl-paste --type TEXT
> /home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig
> 
> : ; wl-paste --type STRING
> /home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig
> 
> : ; wl-paste --type text/plain
> /home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig
> 
> : ; wl-paste --type text/plain\;charset=utf-8
> /home/alex/src/linux/Kbuild,/home/alex/src/linux/Kconfig
> 
> : ;
> ```
> 
> KDE dolpin operates similarly except that it does not copy `text/plain`

That's great.

The unique problem usinf this approach is that it will resamble protocol and hence we won't be able to paste from anywhere else by simply grabbing the absolute path and vice-versa maybe.

This would suite better if we kept `bgy` to use protocol and keep `gy` as it is (only add the newline support), because if I visually `gy` I expect a list of absolute paths not nvim-tree specific thing.

alex-courtis:
> The only thing is that `x` would still mean nothing, only when `gx` it would cut. Or maybe let `gp` to mean paste while cutting bc to some extent `gx` sounds like cut paths bc of `gy`, or even `xp` or something else.

Yeah that's a tricky one, there's no "natural" option. I'm easy for any mapping, however we can't do `xp` as `x` is taken.

alex-courtis:
> This would suite better if we kept `bgy` to use protocol and keep `gy` as it is (only add the newline support), because if I visually `gy` I expect a list of absolute paths not nvim-tree specific thing.

Yes, let's just go with newlines for now. This branch is working.

We can add a protocol in the future :)
Remember that we can "break" this in the future as we're defining communication between nvim-tree instances which is not API.

Uanela:
> > This would suite better if we kept `bgy` to use protocol and keep `gy` as it is (only add the newline support), because if I visually `gy` I expect a list of absolute paths not nvim-tree specific thing.
> 
> Yes, let's just go with newlines for now. This branch is working.
> 
> We can add a protocol in the future :)
> Remember that we can "break" this in the future as we're defining communication between nvim-tree instances which is not API.

The protocol will break OS level operation, which I think is one of the biggest win for this.

And yes protocol would make this nvim specific but I think it would be cooler to make it work anywhere.

## Changes To Be Made

- Add register operation on `c` and `x`
- Add multline support
- Remove `gx` in favor of `gp` to mean paste while cutting. This avoids ambiguity as `p` will operate locally and between instances
- Make `p` decide whether use the clipboard or the register

Together with those mentioned earlier.

Uanela:
All changes made and all test passing.

## Concern
- I `c` on instance B (it also triggers register, easy to warn on next update) and then go to  instance A and `c` again (erases B thing in register and adds it's content) then I go back to instance B and do `p` it will paste what is in clipboard because I didn't remove it basically I will need first to clear the clipboard (there ain't no keymaps only api) and then `p` to paste from instance A. Is kind a GOTCHA moment but maybe is to wrap head around.

## Additional changes made, mostly refactor
- Unified all node attribute copying function to `clipboard.copy_node_attribute(n, attribute)` so that all 4 could easily be handled with node or nodes
- Made all copy.basename|filename|path|absolute_path work on visual mode

alex-courtis:
> And yes protocol would make this nvim specific but I think it would be cooler to make it work anywhere.

We could do both eventually - continue to put the files in the `text/plain` and also put into `x-special/nvim-tree-copied-files` and `text/uri-list` for full interoparability with everything.

alex-courtis:
> * I `c` on instance B (it also triggers register, easy to warn on next update) and then go to  instance A and `c` again (erases B thing in register and adds it's content) then I go back to instance B and do `p` it will paste what is in clipboard because I didn't remove it basically I will need first to clear the clipboard (there ain't no keymaps only api) and then `p` to paste from instance A. Is kind a GOTCHA moment but maybe is to wrap head around.

I don't think there's much we can do about that, which is kind of an edge case. It does work as described for the base cases, and multiple copies (especially cuts) like this will likely "break" most applications.

lua/nvim-tree/_meta/api/fs.lua
```diff
@@ -66,6 +66,18 @@ function nvim_tree.api.fs.cut(node) end
 ---@param node? nvim_tree.api.Node
 function nvim_tree.api.fs.paste(node) end
 
+---
+---Paste nodes from the system register as files to nvim-tree while copying.
+---
+---@param node_or_nodes? nvim_tree.api.Node | nvim_tree.api.Node[]
+function nvim_tree.api.fs.paste_from_register_copying(node_or_nodes) end
```
alex-courtis: Do we still need these two? They don't appear to be used. `p` does the job now :)

lua/nvim-tree/_meta/config/actions.lua
```diff
@@ -24,8 +24,6 @@ error("Cannot require a meta file")
 ---[nvim_tree.config.actions.remove_file]
 ---@field remove_file? nvim_tree.config.actions.remove_file
 
-
-
```
alex-courtis, line 28: I'm guessing the style tooling fixed this.

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -413,46 +453,73 @@ function Clipboard:print_clipboard()
...
+---@param opts? { notify?: boolean }
+function Clipboard:copy_node_attribute(node_or_nodes, attribute, opts)
+  opts = opts and opts or {}
+  local content
+  local node_attribute_getters = {
+    basename = function(n) return self:get_node_basename(n) end,
+    filename = function(n) return self:get_node_filename(n) end,
+    relative_path = function(n) return self:get_node_relative_path(n) end,
+    absolute_path = function(n) return self:get_node_absolute_path(n) end,
+  }
```
alex-courtis: I really appreciate the care and effort you've made with this refactor, however:
- it obfuscates the flow of control: it's no longer easy to "LSP References" `get_node_relative_path` to map it directly to API/impl
- it's doesn't appear necessary to achieve the goal of copy between instances

Let's limit the scope of this change to the minimum required to achieve the functionality: atomic changes.
- It ensures that we can look at git history and identify what was the new functionality and what was a refactor.
- In the case of a problem it allows us to identify whether the problem was the functionality or the refactor.
- Allows reverting/reapplying of individual changes in the event of an issue, instead of having to revert the larger change

I'm always open to refactors and polishing, however please raise separate PRs for them.

alex-courtis (commented):
This is looking fantastic. It feels just right to copy between.

Existing test cases look good, and I can successfully paste from gnome now!!!

Please:
- [x] [remove new API if not needed](https://github.com/nvim-tree/nvim-tree.lua/pull/3336/changes#r3445130550)
- [x] [revert `copy_node_attributes` refactor](https://github.com/nvim-tree/nvim-tree.lua/pull/3336/changes#r3445325613)
- [x] resolve multiple copy test case below

New test cases:

# Needs Fixing - Multiple Copy

Directory with files a b c d

## Multiple Single Copies

Normal copy a

a marked in green

Clip OK:
```
/home/alex/src/linux/foo/a
```

Normal copy b

a and b marked in green

Clip missing a:
```
/home/alex/src/linux/foo/b
```

## Multiple Visual Copy Toggling

Visual Copy a and b

a and b marked in green

Clip OK:
```
/home/alex/src/linux/foo/a
/home/alex/src/linux/foo/b
```

Visual Copy b and c

a and c marked in green

Clip Incorrect:
```
/home/alex/src/linux/foo/b
/home/alex/src/linux/foo/c
```

In both cases it looks like we need to put the final contents of `clipboard.data` into the clipboard, not the visual operation nodes.

lua/nvim-tree/actions/fs/clipboard.lua
Uanela: Thanks for the alert @alex-courtis, should I open a dedicated refactor PR? (I think know from the points you brought).

I just tried to automate to because I was making all of them to support visual mode. Should I simply drop the visual support also, or maybe another PR right away?

Uanela:
All changes done Mr @alex-courtis 

- Revered `copy_node_attribute`
- Removed unneeded APIs
- Fixed the multiple copy by passing exactly what is at clipboard.data to the register.

The last fix was kind odd, even the way it works because on your 2nd example, if I `b` and `c` I would expect those to be green (kind of) but it removes `b` and adds `c` which also makes sense, but was a gotcha to me, but I simply mirrored this behavior.

alex-courtis:
Thinking about the commit message - how to advertise this new feature. What do you think about

feat(#1549): p will paste files from another nvim-tree instance, gp will cut/paste

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -176,28 +177,40 @@ function Clipboard:bulk_clipboard(nodes, from, to, verb)
   self.explorer.renderer:draw()
 end
 
+---@private
+---@param node_or_nodes Node|Node[]
+---@return boolean
+function Clipboard:is_nodes_array(node_or_nodes)
```
alex-courtis, line 183: I was pretty hard on you about the `node_attribute_getters` refactor.

This refactor is good as it is necessary to avoid (already) duplicated code for the new feature.

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -304,6 +317,30 @@ function Clipboard:resolve_conflicts(conflict, destination, action, action_fn)
...
+  end
+
+  local nodes = {}
+  local absolute_paths = vim.split(content:sub(1, #content), "\n")
+
+  for _, absolute_path in ipairs(absolute_paths) do
+    local node_args = { absolute_path = absolute_path, name = vim.fn.fnamemodify(absolute_path, ":t"), explorer = self.explorer }
+    if absolute_path:sub(-1) == "/" then
+      node_args.name = vim.fn.fnamemodify(absolute_path:sub(1, -2), ":t")
+      table.insert(nodes, DirectoryNode(node_args))
```
alex-courtis: Creating (dummy) nodes is a smart way to work with the existing clipboard code to avoid massive changes.

We do need to destroy `Node:destroy()` these nodes when we are done, to release resources e.g. watchers.

They are mixed in with current instance nodes in the `clip` and `conflict` tables in `Clipboard:do_paste` so we cannot just destroy all of those nodes.

Suggestion: keep the dummy nodes from `self:get_nodes_from_reg()` in another table. Destroy each node in the table at:
- early exit around `if not stats and err_name ~= "ENOENT" then`
- the end of `Clipboard:do_paste`

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -437,18 +475,20 @@ function Clipboard:copy_filename(node)
   end
 end
 
+---@private
```
alex-courtis, line 496: Thank you.

alex-courtis (changes requested):
This is just great, really clean and minimalistic - `4 files changed, 107 insertions(+), 34 deletions(-)` - great a major feature.

# General Test Cases

- [x] multiple `c` or `x` - working exactly as per current behaviour
- [x] `gp` working nicely for copied and cut files
- [x] `gy` working as expected

```sh
mkdir -p foo foo/a ; touch foo/1 foo/2 foo/3 foo/a/1 bar baz
```

- [x] some files existing OK
copy bar into foo, then copy bar and baz into foo
`Overwrite /home/alex/src/linux/foo/bar ? R(ename)/y/n:`
baz copied OK

- [x] spaces at end of one file OK
```
/home/alex/src/linux/bar  
/home/alex/src/linux/baz
```
`[NvimTree] Could not copy /home/alex/src/linux/bar   - ENOENT: no such file or directory: /home/alex
/src/linux/bar  `

# Needs Work

- [x] [Destroy Dummy Nodes](https://github.com/nvim-tree/nvim-tree.lua/pull/3336/changes#r3463898971)

## Doc, Suggestions Below

- [x] add API move

Please add `api.fs.move` to `lua/nvim-tree/_meta/api/fs.lua`

- [x] help
```help
==============================================================================
Introduction                                          *nvim-tree-introduction*

Features

    - Automatic updates
    - File type icons
    - Git integration
    - Diagnostics integration: LSP and COC
    - (Live) filtering
    - Rename, delete, create
    - Cut, copy, paste between instances
    - Highly customisable
```

- [x] api paste doc
```lua
---
---Paste files from:
---- current nvim-tree clipboard (if present)
---- another nvim-tree instance via system clipboard
---
---If {node} is a file it will pasted in the parent directory.
---
---@param node? nvim_tree.api.Node
function nvim_tree.api.fs.paste(node) end
```

## Very much optional, as it is an extreme edge case

- [ ] prompts to overwrite directory on bad clipboard

Manually create clipboard:
```
/home/alex/src/linux/bar

/home/alex/src/linux/baz
```

Navigate to `foo` directory and `p`

`Overwrite /home/alex/src/linux/foo/ ? R(ename)/y/n:`

Select n and both files are copied as expected.

Occurs when there are any blank lines, including first or last line.

The last line case is the critical case: copying file paths using vim via, say, `yy` or visual line `y` will add the newline.

nvim-tree `gy` and visual `gy` are OK as they don't add the trailing newline

# Prior Test Cases

## `p`

Using `vim.o.clipboard = "unnamedplus"`

- [x] paste one
- [x] paste many
- [x] paste one inexistent `/home/alex/src/linux/xxx`
  [NvimTree] Could not copy /home/alex/src/linux/xxx - ENOENT
- [x] paste many, one inexistent
```
/home/alex/src/linux/COPYING
/home/alex/src/linux/xxx
/home/alex/src/linux/CREDITS
```
`[NvimTree] Could not copy /home/alex/src/linux/xxx - ENOENT: no such file or directory: /home/alex/s
rc/linux/xxx`
  others copied correctly
- [x] paste mixed files and directories
```
/home/alex/src/linux/foo/a/
/home/alex/src/linux/foo/1
```

## `gp`

Tests as per `p` OK

lua/nvim-tree/actions/fs/clipboard.lua
Uanela: This was exactly the purpose of the refactor, that's why I did it. maybe we can do it on the future.

lua/nvim-tree/actions/fs/clipboard.lua
Uanela: I called destroy to everything on the clip variable itself and everything looks fine.

Uanela:
All solved @alex-courtis , the only thing now is if there will be a problem calling `destroy` at all `clip` nodes variable under `do_paste` because it seems to be harmless because we are changing location of the nodes and they will be redrawn btw, and when there is conflicts the scenario is the same.

I think this one is ready for a merge.

alex-courtis:
> All solved @alex-courtis , the only thing now is if there will be a problem calling `destroy` at all `clip` nodes variable under `do_paste` because it seems to be harmless because we are changing location of the nodes and they will be redrawn btw, and when there is conflicts the scenario is the same.
> 
> I think this one is ready for a merge.

The `clip` nodes are being destroyed, evidenced via `log.types.watcher`

That's working correctly for the cut/paste case, but is a problem for copy/paste

## OK Cut/Paste

```sh
rm -rf 3336 ; mkdir -p 3336/ 3336/zzz/ 3336/yyy/a ; touch 3336/yyy/1 3336/yyy/2 3336/yyy/3 3336/yyy/a/1 ; cd 3336
```
- open nvim and tree in 3336
- open yyy and zzz
- cut yyy
- paste in zzz

```
[2026-06-26 08:39:23] [watcher] Watcher:destroy '/home/alex/src/nvim-tree/r/3336/yyy'
[2026-06-26 08:39:23] [watcher] Event:destroy '/home/alex/src/nvim-tree/r/3336/yyy'
[2026-06-26 08:39:23] [watcher] Watcher:destroy '/home/alex/src/nvim-tree/r/3336/yyy/a'
[2026-06-26 08:39:23] [watcher] Event:destroy '/home/alex/src/nvim-tree/r/3336/yyy/a'
``` 

Destroying the original `yyy` is expected

## Needs Work: Copy/Paste

Copy paste is destroying the source directories:

Same test as above except copy yyy

```
[2026-06-26 08:42:38] [watcher] Watcher:destroy '/home/alex/src/nvim-tree/r/3336/yyy'
[2026-06-26 08:42:38] [watcher] Event:destroy '/home/alex/src/nvim-tree/r/3336/yyy'
[2026-06-26 08:42:38] [watcher] Watcher:destroy '/home/alex/src/nvim-tree/r/3336/yyy/a'
[2026-06-26 08:42:38] [watcher] Event:destroy '/home/alex/src/nvim-tree/r/3336/yyy/a'
```

This is a problem as the watcher and node for `3336/yyy` no longer function. 

- `touch 3336/yyy/4`

`3336/yyy/4` is not shown. 

## Needs Work: Paste Failure

Nodes are not destroyed if the paste fails.

Cause an artificial failure at line 368
```lua
if true then
```

Same test as copy/paste except copy from another instance

The watcher for the dummy nodes is created but never destroyed
```
[2026-06-26 09:03:47] [watcher] Watcher:create '/home/alex/src/nvim-tree/r/3336/yyy/' nil
[2026-06-26 09:03:47] [watcher] Event:create '/home/alex/src/nvim-tree/r/3336/yyy/'
[2026-06-26 09:03:47] [watcher] Event:start '/home/alex/src/nvim-tree/r/3336/yyy/'
```

## Two Changes Required

I'm happy to create a patch if you'd like. Some people like that, some people do not, hence I ask.

1. Destroy only the dummy nodes from `Clipboard:get_nodes_from_reg()`
  - create another table containing just the nodes from above
  - destroy just that table
2. Also destroy on stat failure, before the early return ~371

alex-courtis (changes requested):
We are so very close. Looking forward to using this one.

- [x] https://github.com/nvim-tree/nvim-tree.lua/pull/3336#issuecomment-4804854782

Uanela:
Done @alex-courtis thanks for your careful review on this one up until now it's been a pleasure working with you.

alex-courtis (approved):
Retested [scenario](https://github.com/nvim-tree/nvim-tree.lua/pull/3336#issuecomment-4804854782) with permutations of cut/paste/gp and all working nicely.

I did also notice that the clipboard is cleared on `[NvimTree] ... removed from clipboard.`. That's a nice touch, adding a bit of safety.

Many thanks, this is a great addition to nvim-tree!

alex-courtis:
[released :1st_place_medal: ](https://github.com/nvim-tree/nvim-tree.lua/releases)

Uanela:
Found out something very weird for macos explorer (finder) regarding this feature:

1 - When a path has spaces it puts it on single quotes
image

2 - When you copy multiple absolute paths, it instead joins the space separated
image
image

In order to make it work seamlessly as we designed this one feature, I need to add a custom quick actions (shortcut) to get the files paths correctly, should we add this macOs behavior natively? or simply create a discussion/issue and add the shortcut tricky as the recommended way?

@alex-courtis @gegoune 

I just added @gegoune so that we can have a mac user to feel my pain, hahahahhh.

alex-courtis:
> In order to make it work seamlessly as we designed this one feature, I need to add a custom quick actions (shortcut) to get the files paths correctly, should we add this macOs behavior natively? or simply create a discussion/issue and add the shortcut tricky as the recommended way?

If that's the finder behaviour, it's the "macos default" for the main clipboard and we should integrate with that.

Proposal: change behaviour based on the OS:
- default is current newline separated
- on macos use the above space/quote scheme

Remember we can change the formats at any time. It's not API and we don't need to maintain backwards compatibility between, say, two different nvim-tree versions running at the same time.

PR #3339, opened 2026-06-18
feat: add feature to download from url when `p` from register

Uanela:
Depends on #3336

## Summary

This PR simply takes #3336 one step further and allows downloading files directly to nvim when it finds URL on the register on `p`.

- There is no edge case that I've just forgotten what was is it and not quite sure if it is real a problem (may I'll update if I remember it)
- It only works when the text starts with http I don't if should support other protocols. btw passing an url that doens't start with http will be treated as a folder because my folder/file can be github.com/user/some-blog.png which is also a valid link on http context.


lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -176,28 +177,40 @@ function Clipboard:bulk_clipboard(nodes, from, to, verb)
   self.explorer.renderer:draw()
 end
 
+---@private
+---@param node_or_nodes Node|Node[]
+---@return boolean
+function Clipboard:is_nodes_array(node_or_nodes)
+  if type(node_or_nodes) == "table" and node_or_nodes.is and node_or_nodes:is(Node) then
```
gegoune: How about avoiding conditional altogether
```suggestion
  return type(node_or_nodes) == "table" and node_or_nodes.is and node_or_nodes:is(Node)
```

gegoune (commented):
I wasn't following recent discussions too closely, got one question: why do we want nvim-tree to handle fetching files over http?

Uanela:
> I wasn't following recent discussions too closely, got one question: why do we want nvim-tree to handle fetching files over http?

I read this through my phones notification bar and it sounded like `whey do we need cars to fly`, rsrsrsrs.

This PR as more to do with #3336 and is basically a one tiny addition to allow download files, WHY? On #3336 I added a feature to allow `c` and `p` between different nvim-tree instances this discussion was undergoing at #1549, as the feature simply copies the absolute path(s) and pastes into another instance and nvim-tree deals with it and make copy/cut an horizon of possibilites were opened:

1 - Copy/Cut files from anywhere on the OS as far as you got the absolute path, I can go to finder/explorer simply grab the absolute path and go paste into nvim-tree
2 - Most of the time when download something to a project (image, audio, video, etc) you download it to your OS explorer then copy to your IDE (and on vim we basically can't), then I thought why do I need to download first and copy to my project why not simply download directly to my project then this PR feature came along just grab the link and `p` on nvim-tree and it will download it.

alex-courtis:
Hope you don't mind, I changed the target branch for reviewability. It will change to master after merge.

lua/nvim-tree/actions/fs/clipboard.lua
```diff
@@ -388,7 +385,18 @@ function Clipboard:do_paste(node, action, action_fn)
 
   -- Paste non-conflicting items immediately
   for _, item in ipairs(no_conflict) do
-    do_paste_one(item.node.absolute_path, item.dest, action, action_fn)
+    local absolute_path = item.node.absolute_path
+    if absolute_path:sub(1, #"http") == "http" then
+      notify.info("Downloading " .. absolute_path .. " to " .. item.dest .. "...")
+      local result = vim.fn.system({ "curl", "-sL", absolute_path, "-o", item.dest })
```
alex-courtis, line 405: As much as it pains me to talk about netrw, it appears that it handle almost all protocols:

e.g.
```
nvim --clean
:Nread https://raw.githubusercontent.com/nvim-tree/nvim-tree.lua/refs/heads/master/README.md
```

Perhaps we could invoke it to do all the work.

alex-courtis:
There is also https://github.com/miversen33/netman.nvim which we could integrate with to provide the file transfers.

Uanela:
I will take a while to get hands on this @alex-courtis. but I will work on this based your feedback.

PR #3350, opened 2026-07-15
style(#3223): change prompt of create file to "Create "

### LINKED ISSUE #3223: [Suggestion] Change "Create file" to "Create"
This is very subtle and not really important, but I think it would make more sense if the prompt for creating a new file / folder used `Create ...` instead of `Create file ...`, since the target is not always a file.
```
Create file /path/to/a/folder/
```
reads a little unnatural, as a folder is not a file. Using `Create ...` would be more inclusive and accurate.

Data5tream:
Changes the file creation prompt from "Create file" to "Create ", as suggested in the issue #3223 

resolves #3223


alex-courtis (approved):

PR #3351, opened 2026-08-03
fix: null deref in case git.config is empty

eyalz800:
image


alex-courtis (approved):
Many thanks for your fix!

PR #3353, opened 2026-08-26
fix: change_dir with global = false does not change the working directory

joshw1013:
## Issue
With ```actions.change_dir.global = false```, changing the directory manually through nvim-tree fails to change to the working directory for the user. This is caused by Explorer:cd, running :lcd, which is window-scoped and applies to the current window, which is nvim-tree itself. The editing window's cwd is unchanged, so :! and pickers still resolve against the old root.

## Fix
My code does lcd on the target window as well, under the case that the current window is nvim-tree.


alex-courtis:
> ## Issue
> With `actions.change_dir.global = false`, changing the directory manually through nvim-tree fails to change to the working directory for the user. This is caused by Explorer:cd, running :lcd, which is window-scoped and applies to the current window, which is nvim-tree itself. The editing window's cwd is unchanged, so :! and pickers still resolve against the old root.
> 
> ## Fix
> My code does lcd on the target window as well, under the case that the current window is nvim-tree.

Thank you for creating this pull request.

> This is caused by Explorer:cd, running :lcd, which is window-scoped and applies to the current window, which is nvim-tree itself. The editing window's cwd is unchanged, so :! and pickers still resolve against the old root.

That appears to be working as documented for `global = false`. Confirming: are you talking about the nvim-tree window or another window with a file buffer?

I am clearly not understanding the nature of the problem. Please provide a detailed test case to demonstrate the problem and a fix e.g.
- open nvim-tree in directory x
- open file y
- focus nvim-tree
- <c-]> on dir z

joshw1013:
Sure I can explain in more detail. I think the best way is to just show you the video of what is going on from my config. The first one is the current nvim-tree code, and the second is the one with my fix. 

You can see that in the first one:
* I open nvim-tree
* use `<C-]>` to change the directory
* quit nvim-tree
* I open nvim-tree back up again
* The directory change did not stick

If you try to reproduce it and don't see that, it's possible that without these settings 
```
sync_root_with_cwd = true
respect_buf_cwd = true
```
it won't reset nvim-tree's root, but you should still be able to see that `:pwd` gives the wrong directory.

The reason this is going on, is that when I type `<C-]>`, nvim-tree window is currently open, meaning the `lcd` gets applied to the nvim-tree window, not to the target window, which is the one that opened up nvim-tree, and the one that we care about. 

Also, here's my config if it helps you at all:
https://github.com/joshw1013/nvim/blob/main/lua/josh/plugins/nvim-tree.lua

Let me know if it makes sense now / if you need anything else

https://github.com/user-attachments/assets/27c501fe-5328-473c-aa14-53d5723ff337


https://github.com/user-attachments/assets/188cc33f-b0c8-4611-8962-26ddddba2b56

alex-courtis:
Many thanks for the detailed reproduction.

I understand the problem (directory change does not persist after quitting tree) and the solution (lcd in the buffer window).

Unfortunately I don't think we can make this change:
- Current expected behaviour will change, disrupting users that don't expect it
  - The cwd of the buffer window will change
  - This will break users (myself included) that use, say, lsp diagnostics expecting the project root
- The tree is ephemeral: it is destroyed when you close it and has no "memory"

Is there any other solution you can think of to achieve your goal? I am guessing that `change_dir.global` is not a solution that meets your workflow.

gegoune:
As a generic approach we could emit events, allowing users to do what and when they want, keeping nvim-tree itself lean.

alex-courtis:
> As a generic approach we could emit events, allowing users to do what and when they want, keeping nvim-tree itself lean.

We could definitely add a `TreeChangeDir` event, maybe with a param `scope: One of "global" or "local"` as per `vim.api.keyset.option`

